Issue #26, opened 2019-10-30
[Feature] Build 'Resources' section


kwajiehao:
Build the user interface to modify `Resources` on Isomer

kwajiehao:
Fixed in #42

PR #42, opened 2019-11-07
Add basic resource room functionality

prestonlimlianjie:
### Overview
This PR adds the resource room functionality to the frontend (https://github.com/isomerpages/isomercms-frontend/issues/26). 

### High latency in the resource room because there are too many API calls

The image below is the InVision design for the resource room. 

Screenshot 2019-11-11 at 12 39 37 PM

In order to display the `title`, `post-type`, and `date`, the frontend would need to retrieve the data in the frontmatter of each resource page.

However, this means the backend would have had to call many GitHub APIs:
- 1 API call to list directory in the `resources` folder, and then
- 1 API call for each directory to list the names of all files in the directory, and then
- 1 API call for each file to retrieve the front matter.

If there were on average `m` directories and `n` files in each directory, the frontend would have to call `mn+1` GitHub APIs, which could take a really long time. 

### Reducing API calls by caching `title`, `post-type`, and `date` in the resource page filename

Instead, we can reduce the number of GitHub API calls to `m+1` by caching the fields we needed to display in the file names of the resource pages.

And so, the frontend only calls:
- 1 API call to list directory in the `resources` folder (to get the resource categories), and then
- 1 API call for each directory to list the names of all files in the directory (resource category).

And so the resource pages file names follow the format `{date}-{post-type}-{title}.md`, where:
- `date` is a string with the format `YYYY-MM-DD`,
- `post-type` is a string with two types: `post` or `download`,
- `title` is a case-sensitive string that can contain the following characters `a` to `z`, `A` to `Z`,  `0` to `9`, and hyphen.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -10,7 +10,7 @@
       content="Web site created using create-react-app"
     />
     <link rel="apple-touch-icon" href="logo192.png" />
-    <!-- <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous"> -->
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
     <!--
       manifest.json provides metadata used when your web app is installed on a
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
--- a/public/index.html
+++ b/public/index.html
@@ -10,7 +10,7 @@
       content="Web site created using create-react-app"
     />
     <link rel="apple-touch-icon" href="logo192.png" />
-    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
+    <!-- <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous"> -->
     <!--
       manifest.json provides metadata used when your web app is installed on a
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
--- a/src/components/ResourceCard.jsx
+++ b/src/components/ResourceCard.jsx
@@ -52,7 +52,7 @@ const ResourceCardModal = ({
   </div>
 )
 
-export default class TemplateResourceCard extends Component {
+export default class ResourceCard extends Component {
   constructor(props) {
     super(props);
     this.state = {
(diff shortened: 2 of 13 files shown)


src/utils.js
```diff
@@ -57,3 +58,75 @@ export function changeFileName(event, context) {
...
+}
+
+// Takes in a resource file name and prettifies it.
+// =================
+// Each fileName comes in the format of `{date}-{type}-{title}.md`
+// A sample fileName is 2019-08-23-post-CEO-made-a-speech.md
+// {date} is YYYY-MM-DD, e.g. 2019-08-23
+// {type} is either `post` or `download`
+// {title} is a string containing [a-z,A-Z,0-9] and all whitespaces are replaced by hyphens
+export function prettifyResourceFileName(fileName) {
```
kwajiehao: This only works if existing files are named as such. We get erroneous data because current files do not have `type` included in their file names, and as a result, `title` is also missing the first word in the title

kwajiehao: This only works if existing files are named as such. We get erroneous data because current files do not have type included in their file names, and as a result, title is also missing the first word in the title

src/components/ResourceCard.jsx
```diff
@@ -0,0 +1,310 @@
...
+              resourceCategories={resourceCategories}
+              changeHandler={this.changeHandler}
+              saveHandler={this.saveHandler}
+              deleteHandler={this.deleteHandler}
+              permalinkFileUrlToggle={this.permalinkFileUrlToggle}
+              isNewPost={isNewPost}
+            />
+          )
+          : null}
+        <Link to={`/sites/${siteName}/resources/${category}/pages/${fileName}`}>Edit</Link>
```
kwajiehao: The backend endpoint does not work, so when you click into `Edit` it does not return the content

src/components/ResourceCard.jsx
```diff
@@ -0,0 +1,310 @@
...
+            value={slugifyResourceCategory(resourceCategory.dirName)}
+            label={prettifyResourceCategory(resourceCategory.dirName)}
+          />
+        ))
+      }
+    </select>
+    <p>Title</p>
+    <input value={dequoteString(title)} id="title" onChange={changeHandler} />
+    <p>Date</p>
+    <input value={date} id="date" onChange={changeHandler} />
```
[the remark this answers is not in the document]
prestonlimlianjie: Agreed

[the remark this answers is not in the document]
prestonlimlianjie: We'll also have to implement a frontend validator for date

src/components/ResourceCard.jsx
prestonlimlianjie: Deleted

src/components/ResourceCard.jsx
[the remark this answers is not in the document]
kwajiehao: Logged in #45

src/components/ResourceCard.jsx
```diff
@@ -0,0 +1,310 @@
...
+    <button type="button" onClick={permalinkFileUrlToggle}>{permalink ? 'Switch to download' : 'Switch to post'}</button>
```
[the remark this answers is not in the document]
kwajiehao: Logged in #46. We should consider in the first place whether we want users to be able to touch the permalinks at all - we can set the permalinks ourselves by slugifying the title. What do you think?

In any case, this is not urgent, so let's put it in the parking lot.

src/utils.js
kwajiehao: Actionable: remember to rename a repository's resources files before we do any sort of testing on it.

kwajiehao: Loggedi n #47

public/index.html
```diff
@@ -10,7 +10,7 @@
       content="Web site created using create-react-app"
     />
     <link rel="apple-touch-icon" href="logo192.png" />
-    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
+    <!-- <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous"> -->
```
kwajiehao: This breaks our Edit pages side-by-side functionality

kwajiehao: Re-added bootstrap in acac0ac

PR #61, opened 2019-11-20
feat: make header dynamic

prestonlimlianjie:
This PR allows us to pass in the following props: `title`, `backButtonText`, and `backButtonUrl` from each page so that we can make the header dynamic.

### DIFF
--- a/src/layouts/EditPage.jsx
+++ b/src/layouts/EditPage.jsx
@@ -96,14 +96,13 @@ export default class EditPage extends Component {
     const { editorValue } = this.state;
     return (
       <>
-        <Header />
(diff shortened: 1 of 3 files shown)


taufiqmmhd:
ag lgtm

src/layouts/EditPage.jsx
```diff
@@ -96,14 +96,13 @@ export default class EditPage extends Component {
     const { editorValue } = this.state;
     return (
       <>
-        <Header />
+        <Header
```
kwajiehao: We should add this for `EditCollectionPage` as well

PR #66, opened 2019-11-20
[Feat] Implement media page

kwajiehao:
This PR accomplishes the following objectives:
- Create a `Media` option in the `Sidebar` which contains links to `Images` and `Files`
- Remove `Images` and `Files` from the `Sidebar`
- Implement `Pages`-like styling for both the `Images` and `Files` pages

### DIFF
--- a/src/layouts/Media.jsx
+++ b/src/layouts/Media.jsx
@@ -1,0 +1,94 @@
+import React, { Component } from 'react';
+import { Link } from 'react-router-dom';
+import axios from 'axios';
+import Bluebird from 'bluebird';
+import PropTypes from 'prop-types';
+import Header from '../components/Header';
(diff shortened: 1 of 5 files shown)


src/layouts/Media.jsx
```diff
@@ -0,0 +1,92 @@
...
+          numFiles: objectCounter(documentResp.data.documents),
+        });
+      });
+    } catch (err) {
+      console.log(err);
+    }
+  }
+
+  render() {
+    const { numImages, numFiles } = this.state;
```
prestonlimlianjie: `numImages` and `numFiles` aren't used

kwajiehao: I'll get Yixin to style it so we can include it

PR #71, opened 2019-11-21
style: zoom out the display for EditPage and EditCollectionPage

kwajiehao:
This PR tidies up the CSS for EditPage and EditCollectionPage so that it looks more presentable and less zoomed in.


prestonlimlianjie:
@artylope what do you think about the preview pane on the right? I think it might be a little too zoomed in still.

Screenshot 2019-11-21 at 12 15 10 PM

Issue #68, opened 2019-11-20
[Bug] PageSettingModal save button does not work


taufiqmmhd:
![image](https://user-images.githubusercontent.com/7150533/69240860-5f6c5c80-0bd8-11ea-8785-47b7f999ed75.png)

Clicking on the save button deletes the file. There might be a weird condition where the old file gets deleted and the `POST` request to create the file doesn't get sent.

Issue #39, opened 2019-11-07
Missing `key` props for list items


kwajiehao:
Throughout the CMS, we get this `Warning: Each child in a list should have a unique "key" prop.` messages because we have failed to define `key` props in our list elements. This could lead to unpredictable errors if/when we manipulate our interface's elements.

PR #72, opened 2019-11-21
fix: refresh form only after all API calls finish

taufiqmmhd:
This PR is a general fix before our code freeze on 2019/11/21 which addresses the following issues:
- #39: Provide React `key` prop for list items
- #68: Fix the save button for `PageSettingsModals` by ensuring that the form only refreshes after all API calls are done.
- Fix breadcrumbs in `EditPage` and `EditCollectionPage`


prestonlimlianjie:
This branch’s history is quite strange

PR #73, opened 2019-11-21
feat: improve loading ux

prestonlimlianjie:
This PR improves the UX for page loading in Pages and Resources. 

At present, some page elements  that the user should not be able to interact with are displayed while the page is still loading. This PR hides these page elements and only displays them when the page is fully loaded.

### DIFF
--- a/src/layouts/Resources.jsx
+++ b/src/layouts/Resources.jsx
@@ -259,7 +259,7 @@ export default class Resources extends Component {
                     <>
                       { resourcePages
                         ? (
-                          <ResourcePages 
+                          <ResourcePages
                             resourcePages={resourcePages}
                             siteName={siteName}
                             settingsToggle={this.settingsToggle}
@@ -296,10 +296,10 @@ ResourcePages.propTypes = {
       siteName: PropTypes.string.isRequired,
       category: PropTypes.string.isRequired,
       fileName: PropTypes.string.isRequired,
-    })
-  ),
+    }),
+  ).isRequired,
   siteName: PropTypes.string.isRequired,
-}
+};
 
 ResourceCard.propTypes = {
   fileName: PropTypes.string.isRequired,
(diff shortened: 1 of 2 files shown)


src/layouts/Resources.jsx
```diff
@@ -58,7 +82,7 @@ export default class Resources extends Component {
     super(props);
     this.state = {
       resourceCategories: [],
-      resourcePages: [],
+      resourcePages: undefined,
```
kwajiehao: Would it be better to initialize `resourcePages` as `[]` for consistency? Since in our other files we don't initialize with `undefined`. Then, instead of using `resourcePages.length > 0` we can test for whether the array is empty with lodash `_.isEmpty`

src/layouts/Resources.jsx
```diff
@@ -233,23 +257,15 @@ export default class Resources extends Component {
                   )
                   : (
                     <>
-                      {/* Display resource cards */}
-                      {resourcePages.length > 0
+                      { resourcePages
```
kwajiehao: Test for `!_.isEmpty(resourcePages)` if we initialize with empty array

src/layouts/Resources.jsx
```diff
@@ -41,6 +41,30 @@ const ResourceCard = ({
   );
 };
 
+const ResourcePages = ({
+  resourcePages, settingsToggle, siteName
+}) => (
+  <>
+  {/* Display resource cards */}
+  {resourcePages.length > 0
```
kwajiehao: Test for `!_.isEmpty(resourcePages)` if we initialize with empty array

src/layouts/Resources.jsx
prestonlimlianjie: If we initialize resourcePages as [], and there are no existing resourcePages, the page will display “Loading resources” since it will be empty

PR #81, opened 2019-11-21
feat: create a function to read the navigation bar from the tree

kwajiehao:
This PR introduces a function `treeReader` which takes in a flattened tree, and returns a list of objects which represent the items in the navigation bar.

The flattened tree can be obtained by passing the `tree` object through the function `flattenTree` found in `tree-utils`.


kwajiehao:
This pull request was closed and replaced by #82 since we should be branching of `feat/edit-nav` instead of `master'

PR #96, opened 2019-11-28
[EditHomepage] Add carousel and infopic constructors so their errors are stored

kwajiehao:
### Overview
Carousel and infopic sections are not being accounted for in `EditHomepage`, which causes it to crash when it encounters a page with a carousel or an infopic (and the carousel/infopic is not the last section). We should add this just so our `EditHomepage` doesn't crash - development of carousel functionality in `EditHomepage` can come later

### Context
Currently, carousel and infopic sections are read from the front matter, but there is no carousel or infopic constructor so there are no carousel or infopic sections in the errors.section array. This results in the `sectionIndex` being used to reference the section errors being wrong. For example, the `errors.sections[sectionIndex].resources` value here would refer to the wrong index if there is a carousel section before the resources section - the sectionIndex will be larger by 1:

```
<EditorResourcesSection
     title={section.resources.title}
     subtitle={section.resources.subtitle}
     button={section.resources.button}
     sectionIndex={sectionIndex}
     deleteHandler={this.deleteHandler}
     onFieldChange={this.onFieldChange}
     shouldDisplay={displaySections[sectionIndex]}
     displayHandler={this.displayHandler}
     errors={errors.sections[sectionIndex].resources}
/>
```

This PR fixes that by adding a carousel and infopic constructor and pushing an instance of the carousel and infopic constructor object into `errors.section`.

### DIFF
--- a/src/layouts/EditHomepage.jsx
+++ b/src/layouts/EditHomepage.jsx
@@ -46,6 +46,16 @@ const InfobarSectionConstructor = () => ({
   },
 });
 
+const CarouselConstructor = () => ({
+  carousel: {
+    title: '',
+    subtitle: '',
+    description: '',
+    image: '',
+    alt: '',
+  },
+});
+
 const KeyHighlightConstructor = () => ({
   title: '',
   description: '',
@@ -68,6 +78,8 @@ const enumSection = (type) => {
       return ResourcesSectionConstructor();
     case 'infobar':
       return InfobarSectionConstructor();
+    case 'carousel':
+      return CarouselConstructor();
     default:
       return InfobarSectionConstructor();
   }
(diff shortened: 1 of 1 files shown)


lisatjide:
> Are you doing this because one of the sites we’re doing testing for next week has a carousel?

Hey @prestonlimlianjie! I needed this to take a few screenshots and also video to demo to URA :D

kishore03109:
> I needed this to take a few screenshots and also video to demo to URA :D

This seems to suggest that this was used for demo purposes a long time ago. Closing this, but tagging the person with the most context of this as a quick sanity check @alexanderleegs (please correct me if I am wrong about this ya)

PR #97, opened 2019-12-02
feat: add radio button to switch resource types

taufiqmmhd:
This PR replaces the button to switch resource types in the resource settings modal into
radio buttons as users were confused by the button during round 2 of
user tests. Here's a preview of the change:
### Before
image
### After
image

### DIFF
--- a/src/components/ResourceSettingsModal.jsx
+++ b/src/components/ResourceSettingsModal.jsx
@@ -78,9 +78,10 @@ export default class ResourceSettingsModal extends Component {
     }
   }
 
-  permalinkFileUrlToggle = () => {
+  handlePermalinkFileUrlToggle = (event) => {
     const { permalink } = this.state;
-    if (permalink) {
+    const { target: { value } } = event;
+    if (value === 'file') {
       this.setState({ permalink: null, fileUrl: '/file/url/' });
     } else {
       this.setState({ permalink: '/permalink/', fileUrl: null });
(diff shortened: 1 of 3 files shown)


src/components/ResourceSettingsModal.jsx
```diff
@@ -246,7 +247,16 @@ export default class ResourceSettingsModal extends Component {
               <p className={elementStyles.formLabel}>Resource Type</p>
 
               {/* Permalink or File URL */}
-              <button type="button" onClick={this.permalinkFileUrlToggle}>{permalink ? 'Switch to download' : 'Switch to post'}</button>
+              <div className="d-flex">
+                <label htmlFor="radio-post" className="flex-fill">
+                  <input type="radio" id="radio-post" name="resource-type" value="post" onClick={this.handlePermalinkFileUrlToggle} checked={!!permalink} />
```
prestonlimlianjie: Is there a particular reason why you're using `!!`?

taufiqmmhd: It's more of a cleanliness thing as I wanted to make it clear that the `checked` attribute works like a boolean

src/components/ResourceSettingsModal.jsx
```diff
@@ -246,7 +247,16 @@ export default class ResourceSettingsModal extends Component {
...
+                  Post Content
+                </label>
+                <label htmlFor="radio-post" className="flex-fill">
+                  <input type="radio" id="radio-file" name="resource-type" value="file" onClick={this.handlePermalinkFileUrlToggle} checked={!permalink} />
+                  Downloadable File
+                </label>
+              </div>
               {permalink
```
kwajiehao: We can have more sophisticated logic here (maybe in a future PR) so that we can store the state of the permalink instead of resetting it to null every time we toggle the radio button

PR #98, opened 2019-12-04
feat: add homepage infopic section

prestonlimlianjie:
This PR addresses https://github.com/isomerpages/isomercms-frontend/issues/11 by adding the ability to add/modify/remove infopic sections in the homepage.

### DIFF
--- a/src/templates/homepage/InfopicLeftSection.jsx
+++ b/src/templates/homepage/InfopicLeftSection.jsx
@@ -1,0 +1,119 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/* eslint
+  react/no-array-index-key: 0
+  jsx-a11y/anchor-is-valid: 0
+ */
+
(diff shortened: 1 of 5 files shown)


src/templates/homepage/InfopicLeftSection.jsx
```diff
@@ -0,0 +1,119 @@
...
+                <div>
+                  <span>{ button }</span>
+                  <i className="sgds-icon sgds-icon-arrow-right is-size-4" aria-hidden="true" />
+                </div>
+              )
+              : null}
+          </a>
+        </div>
+        <div className="col is-half">
+          <img src={`https://raw.githubusercontent.com/isomerpages/${siteName}/staging${imageUrl}`} alt={imageAlt} />
```
kwajiehao: If you want, you can use the `prependImageSrc` utils function on the whole returned html chunk in `EditHomepage`  - in return you need only put the `imageUrl` in the img src.

PR #102, opened 2019-12-10
[Feat] Revamp images tab

taufiqmmhd:
This PR does the following:

1. Display image cards with a preview of each image
2. Introduce ImageSettingsModal

### Before
image
image
### After
image
image


src/layouts/Images.jsx
```diff
@@ -75,26 +104,23 @@ export default class Images extends Component {
   onImageSelect = async (event) => {
```
kwajiehao: Remove the `settingsToggle` method from line 71-79, as well as the `updateNewPageName` method

src/layouts/Images.jsx
```diff
@@ -75,26 +104,23 @@ export default class Images extends Component {
...
     const imgReader = new FileReader();
     imgReader.imgName = event.target.files[0].name;
```
kwajiehao: What is the point of setting `imgReader.imgName` and then reading from it in line 114? We can simply set `const imgName = event.target.files[0].name` and remove line 114

src/utils/validators.js
```diff
@@ -5,6 +5,7 @@ const PERMALINK_REGEX = '^(/([a-z]+([-][a-z]+)*/)+)$';
 const permalinkRegexTest = RegExp(PERMALINK_REGEX);
 const DATE_REGEX = '^([0-9]{4}-[0-9]{2}-[0-9]{2})$';
 const dateRegexTest = RegExp(DATE_REGEX);
+const fileNameRegexTest = /^[\w,\s-]+\.[A-Za-z]{3}$/;
```
kwajiehao: Perhaps we need to discuss what kind of file name to accept since there are some file names which Github accepts that fail this test such as `Screen Shot 2019-12-10 at 4.40.02 PM.png`

Another thing to consider is if we will run all file names through the same regex tests. If not, we might want to consider renaming this variable

src/layouts/Images.jsx
taufiqmmhd: yeah this is bad code my bad i did this because of some earlier implementation that cause funky scoping issues. Had changed the implementation but didn't know I could switch that out as well.

src/layouts/Images.jsx
```diff
@@ -1,27 +1,56 @@
...
 import Header from '../components/Header';
 import Sidebar from '../components/Sidebar';
 import elementStyles from '../styles/isomer-cms/Elements.module.scss';
 import contentStyles from '../styles/isomer-cms/pages/Content.module.scss';
+import mediaStyles from '../styles/isomer-cms/pages/Media.module.scss';
+import ImageSettingsModal from '../components/ImageSettingsModal';
+
+const ImageCard = ({ image, siteName, onClick }) => (
+  <div className={mediaStyles.mediaCard} key={image.path}>
+    <a href="/" onClick={(e) => { e.preventDefault(); onClick(); }}>
```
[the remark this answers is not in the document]
taufiqmmhd: Initially, I tried placing the `onClick` attribute (that's currently in the anchor tag) into the `div` above it. The linter was throwing multiple errors like `click-events-have-key-events`, `no-static-element-interactions`, etc. The quickest way around it was to implement an anchor tag and place the `onClick` in there.

PR #104, opened 2019-12-11
[Feat] Throttle button onClick events

taufiqmmhd:
This PR throttles all button onClick events that invoke API calls so that end-users will not send multiple API calls if they choose to spam click those buttons

### DIFF
--- a/src/components/ImageSettingsModal.jsx
+++ b/src/components/ImageSettingsModal.jsx
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import axios from 'axios';
+import _ from 'lodash';
 import PropTypes from 'prop-types';
 import mediaStyles from '../styles/isomer-cms/pages/Media.module.scss';
 import elementStyles from '../styles/isomer-cms/Elements.module.scss';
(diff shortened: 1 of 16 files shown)


src/components/ImageSettingsModal.jsx
```diff
@@ -103,8 +104,8 @@ export default class ImageSettingsModal extends Component {
               />
             </div>
             <div className={elementStyles.modalButtons}>
-              <button type="button" className={errorMessage ? elementStyles.disabled : elementStyles.blue} disabled={!!errorMessage} onClick={this.renameImage}>Save</button>
-              <button type="button" className={sha ? elementStyles.warning : elementStyles.disabled} onClick={this.deleteImage} disabled={!sha}>Delete</button>
+              <button type="button" className={errorMessage ? elementStyles.disabled : elementStyles.blue} disabled={!!errorMessage} onClick={_.throttle(this.renameImage, 2000)}>Save</button>
```
prestonlimlianjie: Also I know this won't be relevant if we decide against using throttle, but a note on style: we should define magic numbers like `2000` as a constant at the top of the file and re-use it across the file.

kwajiehao:
> Actually I was thinking about it, throttling at 2000ms doesn't guarantee that the user can't make the same call twice if the first call takes longer than 2000ms.
> 
> I think a better solution is to set the state of the button to disabled, show some feedback message in the button like `Saving` or `Deleting` or `Updating`, and only enable the button again after the various handlers are done running.
> 
> What are your thoughts @kwajiehao @taufiqmmhd ?

Yeah that sounds good to me

src/components/ImageSettingsModal.jsx
taufiqmmhd: yea definitely

taufiqmmhd:
> Actually I was thinking about it, throttling at 2000ms doesn't guarantee that the user can't make the same call twice if the first call takes longer than 2000ms.
> 
> I think a better solution is to set the state of the button to disabled, show some feedback message in the button like `Saving` or `Deleting` or `Updating`, and only enable the button again after the various handlers are done running.
> 
> What are your thoughts @kwajiehao @taufiqmmhd ?

Yeah let's do that

taufiqmmhd:
This is a horrible PR.

PR #107, opened 2019-12-18
Feat/upload image rename modal

kwajiehao:
### Overview
Currently, the `Upload Image` button uploads the selected image to Github automatically. This PR introduces an intermediary modal which allows the user to rename the file before upload.

This PR achieves the following:
- reuse the `ImageSettingsModal` to allow the user to name an image when uploading
- Upload the image file only after the `fileName` has been chosen

Refer to the following GIF for how the modal works:

To upload the image with the default name, simply click `Save` without making any modifications


src/components/ImageSettingsModal.jsx
```diff
@@ -15,38 +15,63 @@ export default class ImageSettingsModal extends Component {
       newFileName: fileName,
       sha: '',
       content: null,
+      isUpload: null,
     };
   }
 
   async componentDidMount() {
-    const { match, image: { fileName } } = this.props;
+    const { match, image, isUpload } = this.props;
```
taufiqmmhd: maybe instead of `isUpload` would it be better to call it `isPendingUpload` or `hasPendingUpload`?

src/layouts/Images.jsx
```diff
@@ -68,18 +69,14 @@ export default class Images extends Component {
...
-      const params = {
-        imageName,
-        content: imageContent,
-      };
-
-      await axios.post(`${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/images`, params, {
-        withCredentials: true,
+      // toggle state so that image renaming modal appears
+      this.setState({
+        uploadedImage: {
```
taufiqmmhd: maybe `pendingImageUpload` instead of `uploadedImage`?

src/components/ImageSettingsModal.jsx
```diff
@@ -15,38 +15,63 @@ export default class ImageSettingsModal extends Component {
...
-    this.setState({ sha, content });
+    if (isUpload) {
+      const { content } = image;
+      this.setState({ content, isUpload });
+    } else {
+      const { fileName } = image;
+      const { data: { sha, content } } = await axios.get(`${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/images/${fileName}`, {
+        withCredentials: true,
+      });
+      this.setState({ sha, content, isUpload });
```
taufiqmmhd: Since you're not modifying the value of `isUpload` when it is passed down as a prop, it's redundant to to set it in state when you can access the same value from props

kwajiehao: Fixed in 3ea87ca

src/layouts/Images.jsx
kwajiehao: Fixed in 3ea87ca

src/components/ImageSettingsModal.jsx
kwajiehao: Fixed in 3ea87ca

src/components/ImageSettingsModal.jsx
```diff
@@ -15,38 +15,62 @@ export default class ImageSettingsModal extends Component {
       newFileName: fileName,
       sha: '',
       content: null,
+      isPendingUpload: null,
```
taufiqmmhd: This state isn't used anywhere in the component

src/components/ImageSettingsModal.jsx
```diff
@@ -15,38 +15,62 @@ export default class ImageSettingsModal extends Component {
...
-    this.setState({ sha, content });
+    if (isPendingUpload) {
+      const { content } = image;
+      this.setState({ content, isPendingUpload });
+    } else {
+      const { fileName } = image;
+      const { data: { sha, content } } = await axios.get(`${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/images/${fileName}`, {
+        withCredentials: true,
+      });
+      this.setState({ sha, content, isPendingUpload });
```
taufiqmmhd: No point setting state for `isPendingUpload`, since it has the same value of a prop with the same name

PR #108, opened 2019-12-18
[Feat] Add deletion warning modals

taufiqmmhd:
This PR adds warning modals for any delete operations that are present in the app.

### Notes
After PR #110, this PR broke due to merge conflicts and so I shifted the `LoadingButton` component into the warning modal. Here are all the components for which I have tested the new implementation in.
- [x] EditPage
- [x] EditCollectionPage
- [x] PageSettingsModal
- [x] CollectionPageSettingsModal
- [x] EditResourcePage
- [x] ResourceSettingsModal
- [x] ResourceCategoryModal
- [x] Homepage elements
- [x] ImageSettingsModal

Here are some examples:
image
image


src/styles/isomer-cms/elements/modal.scss
```diff
@@ -58,3 +60,50 @@
...
+    justify-content: space-between;
+    flex: 1;
+  
+    .modalButtons {
+      padding-top: 20px;
+      display: flex;
+      flex-direction: row;
+      justify-content: flex-start;
+    }
+  }
```
kwajiehao: `.modalContent` and `.modalButtons` are identical to the CSS classes defined higher up in the `modal.scss` file. Instead of the duplication of code, you should abstract these subclasses and use them in both `.modalWarning` and `.modal`

kwajiehao: `.modalContent` and `.modalButtons` are identical to the CSS classes defined higher up in the `modal.scss` file. Instead of the duplication of code, you should abstract these subclasses and use them in both `.modalWarning` and `.modal`

src/components/DeleteWarningModal.jsx
```diff
@@ -0,0 +1,33 @@
...
+    <div className={elementStyles['modal-warning']}>
+      <div className={elementStyles.modalHeader}>
+        <h1>
+          Delete
+          {' '}
+          { type }
+        </h1>
+      </div>
+      <form className={elementStyles.modalContent}>
+        <p>Are you sure you want to delete this?</p>
```
kwajiehao: This should be styled in accordance with the styles used in other modal sections

src/layouts/EditPage.jsx
```diff
@@ -73,14 +75,15 @@ export default class EditPage extends Component {
...
-      const { match } = this.props;
+      const { match, history } = this.props;
       const { siteName, fileName } = match.params;
       const { sha } = this.state;
       const params = { sha };
       await axios.delete(`${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/pages/${fileName}`, {
         data: params,
         withCredentials: true,
       });
+      history.goBack();
```
kwajiehao: Nice

src/styles/isomer-cms/elements/modal.scss
prestonlimlianjie: Actually, is the only diff between a normal modal and the warning modal the difference in height? If so, you can create a new CSS selector for `modalWarning`, where the height is set differently.

taufiqmmhd: yeah you're right i'll just flatten out the classes

PR #109, opened 2019-12-20
feat: permalink setter

prestonlimlianjie:
### Overview
This PR adds a permalink setter feature to improve the user experience of setting permalinks for pages. This PR also adds a validation check to see if the title and permalink are already in use.

### Features
- If the user is creating a new page, when they edit the `title`, the `permalink` is automatically generated
- If the user is creating a new page, when they click on the `Save` button, a check is run to ensure that they did not submit a default `title` and `permalink` that already exists
- When the user edits the `title` or `permalink`, a check is run to ensure that these do not clash with existing titles and permalinks.

> Note:
> This was designed with simple pages in mind. Each simple page lives in the `pages` folder of the repo, and when a user modifies a simple page, its `title` and `permalink` is compared against other simple pages' fields.
> This creates a limitation where if a simple page contains the same `permalink` as a collection page, the checker will not be able to prevent a clash in permalinks. Hence, a decision was made to prevent users from using `/` in the `permalink` field.

### Description
When a user clicks on the `Create New Page` or `Page Settings` button, the following `Page Settings` modal shows up.

Screenshot 2019-12-20 at 12 36 48 PM

When the user clicks on the `EDIT` button, the input appears and the user can modify the permalink.

Screenshot 2019-12-20 at 12 45 00 PM


src/utils.js
```diff
@@ -142,3 +142,7 @@ export function generateCollectionPageFileName(title, groupIdentifier) {
   console.log(slugify(title));
   return `${groupIdentifier}-${slugify(title).replace(/[^a-zA-Z-]/g, '')}.md`;
 }
+
+export function generatePermalink(title) {
```
kwajiehao: `generatePermalink` works fine but the regex for `generatePageFileName` should include numbers as well so it should look like `/[^a-zA-Z0-9-]/g`

src/utils/validators.js
```diff
@@ -1,7 +1,7 @@
 
 // Common regexes and constants
 // ==============
-const PERMALINK_REGEX = '^(/([a-z]+([-][a-z]+)*/)+)$';
+const PERMALINK_REGEX = '^(/([a-z]+([-][a-z]+)*)+)$';
```
kwajiehao: regex should be `^(/([a-z]+([-][a-z]+)*/)+)$` to account for the ending slash

src/components/PageSettingsModal.jsx
```diff
@@ -22,12 +28,38 @@ export default class PageSettingsModal extends Component {
...
+      } = this.props;
+
+      // get settings data from backend
+      const settingsResp = await axios.get(`${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/settings`, {
+        withCredentials: true,
+      });
+      const { settings } = settingsResp.data;
+      const { configFieldsRequired } = settings;
+      const baseUrl = configFieldsRequired.url;
+      this.setState({ baseUrl });
```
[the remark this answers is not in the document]
prestonlimlianjie: I think React batches the setState methods and determines when to render, so each setState call doesn’t result in a re-render, but I’m down to remove that line if it looks verbose. I added it because I thought it would make more sense to set the state next to where baseUrl was defined

src/utils/validators.js
prestonlimlianjie: Ah actually it should be `^(/([a-z]+([-][a-z]+)*)+/)$` right because we only want the forward slashes to be at the start and end

src/components/PageSettingsModal.jsx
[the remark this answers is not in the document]
kwajiehao: Just leaving this here for reference: `setState` batches up the actual setting of state, but every time `setState` is called it will be followed by a rerender ([source](https://reactjs.org/docs/react-component.html#setstate))

src/utils/validators.js
kwajiehao: But that would not allow a permalink like `/abc/123/` where there are forward slashes in between - is that the desired behavior?

prestonlimlianjie: Currently, the PageSettingsModal conducts checks across all simple-pages to prevent repeated permalinks. However, if we allow a permalink like `/abc/123/`, there could be a chance that there exists another file in a different directory, for example collection `abc` with a file `123` that has a permalink `/abc/123`.

Preventing forward slashes will prevent such errors from happening. Thoughts?

kwajiehao: That makes sense, let's stick to your change `^(/([a-z]+([-][a-z]+)*)+/)$`. This regex makes sense for collections too since we'll probably want to pre-fill the collection part of the permalink for collection pages (meaning for collection **abc** the permalink `/abc/` part is pre-filled and non-editable)

PR #110, opened 2019-12-20
Feat/disable save when loading

kwajiehao:
## Overview
This PR introduces a `LoadingButton` component which disables all `Save` and `Delete` buttons when it is clicked. When one of these buttons are clicked, we disable the button and replace the text in the button with a spinning loader. 

The components and pages which are affected include:
- `PageSettingsModal` and `CollectionPageSettingsModal`
- `ImageSettingsModal`
- `ResourceSettingsModal`
- `EditPage` and `EditCollectionPage`
- `EditHomepage`
- `EditResourcePage`
- `Settings`

## To-do 
-  Currently, the button does **not** work with the `ResourceCategoryModal`
- @artylope to adjust the height of the button so that it doesn't change size when the loading spinner appears

### DIFF
--- a/src/components/LoadingButton.jsx
+++ b/src/components/LoadingButton.jsx
@@ -1,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import elementStyles from '../styles/isomer-cms/Elements.module.scss';
+
+export default function LoadingButton(props) {
+  // extract props
+  const {
+    label,
+    disabled,
+    className,
+    callback,
+  } = props;
+
+  // track whether button is loading or not
+  const [isLoading, setButtonLoading] = React.useState(false);
+  return (
+    <button
+      type="button"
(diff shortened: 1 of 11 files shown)


src/components/LoadingButton.jsx
```diff
@@ -0,0 +1,47 @@
...
+  // track whether button is loading or not
+  const [isLoading, setButtonLoading] = React.useState(false);
+  return (
+    <button
+      type="button"
+      className={isLoading ? disabledStyle : className}
+      disabled={isLoading ? true : disabled}
+      onClick={() => {
+        setButtonLoading(true);
+        callback().then(() => setButtonLoading(false));
```
prestonlimlianjie: Maybe use `await` to keep the style consistent with the rest of the code when dealing with async funcs

kwajiehao: done

PR #120, opened 2019-12-26
style: login page

prestonlimlianjie:
This PR makes the login page more aesthetically appealing.

Screenshot 2019-12-26 at 2 58 23 PM

### DIFF
--- a/src/styles/isomer-cms/elements/buttons.scss
+++ b/src/styles/isomer-cms/elements/buttons.scss
@@ -1,5 +1,33 @@
 @import './variable';
 
+@mixin buttonColor($color) {
+  background: $color;
+  color: white;
+  padding: 10px 30px;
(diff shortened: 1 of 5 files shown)


src/styles/isomer-cms/elements/buttons.scss
```diff
@@ -92,6 +92,34 @@ button{
     }
   }
 
+  &.green {
+    background: $login-green;
```
kwajiehao: similarity with `&.blue` means we should probably abstract our some elements into a mixin or have `&.blue` and `&.green` inherit from a common ancestor

kwajiehao: We can use a mixin for `&.blue` and `&.green` and pass the color in as an argument

PR #129, opened 2020-02-19
style: line height for resource card

prestonlimlianjie:
This PR fixes the line height issue in the Resource Card in the Resources view.

**Before**
Screenshot 2020-02-19 at 4 58 32 PM

**After**
Screenshot 2020-02-19 at 4 59 00 PM


prestonlimlianjie:
> This looks good to me CSS wise, but currently the lines are not aligned (for example, in the screenshot you posted, the `Post` label is at different heights for each resource card). We should make sure that each `<p>` tag is in line with each other across `ResourceCard`s

PR #154, opened 2020-10-14
Fix: handle behaviour on logout

alexanderleegs:
This PR fixes an issue with a non-existent endpoint being called when the logout button is clicked. The button now calls the appropriate endpoint, removes the user's existing cookie, and redirects the user to the login page. To be reviewed with PR #[59](https://github.com/isomerpages/isomercms-backend/pull/59) on the isomercms-backend repo.

### DIFF
--- a/src/components/Header.jsx
+++ b/src/components/Header.jsx
@@ -1,42 +1,69 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { Redirect } from 'react-router-dom';
+import axios from 'axios';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


kwajiehao (approved):
lgtm for now, we can revisit this when we want to do auto-logout, since currently it's still possible to view the internals of the CMS even while logged out

kwajiehao:
Also one more thing on branch naming - do you mind naming the branches with a slash instead of a dash? So instead of `Fix-handle-behavior-on-logout`, use `fix/handle-behavior-on-logout` just to be consistent with the coding conventions of the project. Lmk if you feel that we should be using the other convention! thanks

Issue #165, opened 2020-10-15
Add date to `EditResourcePage`


kwajiehao:
**Current behavior**: No date below breadcrumb in `EditResourcePage`

<img width="612" alt="Screenshot 2020-10-15 at 6 23 55 PM" src="https://user-images.githubusercontent.com/31984694/96111434-c6db7b00-0f13-11eb-8c38-f733a5aa6083.png">

**Expected behavior**: There should be a date value below the breadcrumb in `EditResourcePage`

<img width="933" alt="Screenshot 2020-10-15 at 6 24 11 PM" src="https://user-images.githubusercontent.com/31984694/96111445-c9d66b80-0f13-11eb-992f-067ff09695a7.png">

kwajiehao:
Resolved in #167

Issue #137, opened 2020-10-07
Edit Pages display is not being rendered properly


kwajiehao:
In our `EditPage` and `EditCollectionPage` layouts, the display pane is not being rendered properly if the browser window is too narrow, as we can see from the following screenshot (taken from `psd-engineer-whats-next-cms-test` repo on 2020/10/07):

<img width="1458" alt="Screenshot 2020-10-07 at 4 07 35 PM" src="https://user-images.githubusercontent.com/31984694/95304761-aa0dca80-08b7-11eb-8b63-3eea44d54a70.png">

We should look at how it's currently being rendered and fix it/improve the CSS so it's rendered properly.

kwajiehao:
Resolved in #167

PR #167, opened 2020-10-16
Fix: display issue for breadcrumb

alexanderleegs:
This PR fixes the issue raised in Issue #137. Currently, the display issue exists because the styling for those components were mistakenly given a right offset value instead of a left offset value. To decrease the chances of this occurring in future, the styling has also been refactored out into a separate css class, and the offset has been changed to padding instead. This PR also fixes an issue with the display exceeding the width of the screen.

In addition, this PR also adds in the date to the breadcrumb on resource pages, as raised in Issue #165. The date formatting has also been changed to match the display on the actual website.

The title and breadcrumb have also been modified to be truncated if their length is too long.

### DIFF
--- a/src/layouts/EditResourcePage.jsx
+++ b/src/layouts/EditResourcePage.jsx
@@ -99,7 +99,6 @@ export default class EditResourcePage extends Component {
     const { siteName, fileName, resourceName } = match.params;
     const { title, date, type } = prettifyResourceFileName(fileName);
     const { editorValue, canShowDeleteWarningModal } = this.state;
-    console.log(match)
     return (
       <>
         <Header
--- a/src/layouts/EditResourcePage.jsx
+++ b/src/layouts/EditResourcePage.jsx
@@ -97,11 +97,13 @@ export default class EditResourcePage extends Component {
   render() {
     const { match } = this.props;
     const { siteName, fileName, resourceName } = match.params;
+    const { title, date, type } = prettifyResourceFileName(fileName);
     const { editorValue, canShowDeleteWarningModal } = this.state;
+    console.log(match)
     return (
       <>
         <Header
-          title={`${prettifyResourceFileName(fileName).title} in ${resourceName}`}
+          title={`${title} in ${resourceName}`}
           backButtonText="Back to Resources"
           backButtonUrl={`/sites/${siteName}/resources`}
         />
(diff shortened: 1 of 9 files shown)


src/layouts/EditResourcePage.jsx
```diff
@@ -97,11 +97,13 @@ export default class EditResourcePage extends Component {
   render() {
     const { match } = this.props;
     const { siteName, fileName, resourceName } = match.params;
+    const { title, date, type } = prettifyResourceFileName(fileName);
```
kwajiehao, line 100: This was a problem that the original code author introduced, but could we rename this to something that reflects that we're extracting resource metadata, as opposed to just prettifying the file name?

src/layouts/EditResourcePage.jsx
```diff
@@ -97,11 +97,13 @@ export default class EditResourcePage extends Component {
...
     const { editorValue, canShowDeleteWarningModal } = this.state;
+    console.log(match)
```
kwajiehao, line 102: Is this log statement a typo?

src/styles/isomer-template.scss
```diff
@@ -12095,4 +12095,14 @@ a.navbar-item {
 .CodeMirror {
     height: calc(85vh - #{$header-height} - #{$footer-height});
     overflow-y: scroll;
+}
+
+.breadcrumb-container {
+    padding-left: 50px;
```
kwajiehao, line 12101: Can we consider changing the values specified in `.breadcrumb-container` and `.page-content-body` to use `rem` instead of `px`? Here's a good blogpost on the difference between `px` vs `rem`: https://engageinteractive.co.uk/blog/em-vs-rem-vs-px

kwajiehao (changes requested):
Left comments, there are a couple of other things I wanted to point out:

1. The breadcrumb is still scrollable

In practice, the breadcrumb is truncated if the length exceeds a certain number of characters.

_Current_
Screenshot 2020-10-19 at 1 49 35 PM

_Expected_
Screenshot 2020-10-19 at 1 50 34 PM

2. This isn't necessarily in the original scope of the PR, but could you also make the title in the `EditPage`, `EditResourcePage`, `EditLeftNavPage` responsive? This is what it currently looks like when you resize the page:

Screenshot 2020-10-19 at 1 35 29 PM
Screenshot 2020-10-19 at 1 35 36 PM

src/layouts/EditResourcePage.jsx
alexanderleegs: Whoops forgot to remove, resolved in https://github.com/isomerpages/isomercms-frontend/pull/167/commits/e127053b110dfdee0646768b3058e6e61e769924

src/layouts/EditResourcePage.jsx
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/167/commits/1f3f6fc20f6288af47b6c3c5357609a2869b0404

src/styles/isomer-template.scss
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/167/commits/8a670b3672735951ef98d942b1d91f72b9782778

kwajiehao (approved):
lgtm

PR #170, opened 2020-10-20
Feat/prevent navigation if unsaved changes

kwajiehao:
## Overview

This PR introduces a warning modal which asks the user for confirmation if they try to navigate away from an `Edit` page when they have unsaved changes. This prevents accidental loss of data due to accidental clicks on the back button. This is achieved by adding new conditions in the `Header` component which will trigger the warning modal if met.

Screenshot 2020-10-20 at 5 52 49 PM

### Caveat

However, this does not trigger a warning modal if a user tries to navigate away from the page using the URL bar. In such a case, we want to listen for the `window.onbeforeunload` event, make sure to prevent default behavior, and trigger the warning modal to ask if the user is sure they want to navigate away from the page. This is something we can do in a separate PR.

### DIFF
--- a/src/components/GenericWarningModal.jsx
+++ b/src/components/GenericWarningModal.jsx
@@ -4,14 +4,12 @@ import elementStyles from '../styles/isomer-cms/Elements.module.scss';
 import LoadingButton from './LoadingButton';
 
 
-const GenericWarningModal = ({ displayTitle, displayText, onProceed, onCancel, type }) => (
+const GenericWarningModal = ({ displayTitle, displayText, onProceed, onCancel }) => (
   <div className={elementStyles.overlay}>
     <div className={elementStyles['modal-warning']}>
       <div className={elementStyles.modalHeader}>
         <h1>
           {displayTitle}
-          {' '}
-          { type }
         </h1>
       </div>
       <form className={elementStyles.modalContent}>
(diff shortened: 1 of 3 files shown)


src/components/GenericWarningModal.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+
+const GenericWarningModal = ({ displayTitle, displayText, onProceed, onCancel, type }) => (
+  <div className={elementStyles.overlay}>
+    <div className={elementStyles['modal-warning']}>
+      <div className={elementStyles.modalHeader}>
+        <h1>
+          {displayTitle}
+          {' '}
+          { type }
```
alexanderleegs, line 14: What is type used for? Might be good to clarify

alexanderleegs (changes requested):
Left minor comments

src/components/GenericWarningModal.jsx
kwajiehao: Good catch, removed `type` in [4dea0ad](https://github.com/isomerpages/isomercms-frontend/pull/170/commits/4dea0ad27123417ebc7dce3c1c822771f2cc1817)

alexanderleegs (approved):
lgtm

Issue #160, opened 2020-10-15
Revamp homepage preview


kwajiehao:
There are currently quite a few problems with the homepage preview. This issue details them and specifies what the fixes should be.

**1. Missing notification bar**
Users should be given the option of adding a notification bar in the homepage. We should provide a **field** for adding the notification bar, and if filled in, that notification bar should be displayed on the homepage preview.

<img width="1387" alt="Screenshot 2020-10-15 at 5 41 56 PM" src="https://user-images.githubusercontent.com/31984694/96106457-d8218900-0f0d-11eb-9631-334a4895939b.png">

**2. `Title` field is ambiguous**
We should rename this field to be `Website title`

<img width="333" alt="Screenshot 2020-10-15 at 5 43 45 PM" src="https://user-images.githubusercontent.com/31984694/96106619-fe472900-0f0d-11eb-9701-c0920fafa9cf.png">

**3. Dropdown element / Highlight subsection within the Hero section should reflect only the title**

For example, in the following image:
<img width="361" alt="Screenshot 2020-10-15 at 5 40 50 PM" src="https://user-images.githubusercontent.com/31984694/96106714-19199d80-0f0e-11eb-9763-2f52e8aa0c6e.png">

The title should just read `Highlight A` instead of `Highlight 1:Highlight A`. This problem occurs for both the dropdown element and the highlight.

**4. "Add a new section" is misleading**
Currently, the "Add a new section" area looks like a button, but does not actually behave like one. We should remove the dotted boundary, the hover state, and left align it to reduce confusion for our users. 

<img width="415" alt="Screenshot 2020-10-15 at 5 47 22 PM" src="https://user-images.githubusercontent.com/31984694/96107054-84636f80-0f0e-11eb-8fc2-56f5c38f9266.png">

Furthermore, we should include an actual button, so that new sections are only added upon click of that button, instead of upon selection of the dropdown option. Currently, we have a bug where we are unable to add two of the same elements in a row (for example, infopic followed by infopic) because the dropdown does not allow us to select two elements consecutively. Adding a button so that the section is added only upon clicking that button would eliminate that bug.

**5. Add a watermark of dummy resource room elements**

Since it's not clear that the resource room elements in the preview are dummy elements and not actually retrieved from the repo, we can make this clearer by adding a watermark/mask over the resource room cards.

[...]

Issue #146, opened 2020-10-08
Text overflow in homepage preview


kwajiehao:
Currently, the text in the homepage preview overflows and overlaps with the buttons below (this example taken from the edb-osp-cms-test repo):

<img width="1447" alt="Screenshot 2020-10-08 at 1 54 41 PM" src="https://user-images.githubusercontent.com/31984694/95420452-07fce980-096e-11eb-9a09-e64ca4572edb.png">

Expected behavior: there should be a clean separation between text and buttons

alexanderleegs:
Resolved with https://github.com/isomerpages/isomercms-frontend/pull/172

PR #172, opened 2020-10-22
Fix/revamp homepage preview

alexanderleegs:
This PR fixes the existing issues in the homepage preview, as described in issue https://github.com/isomerpages/isomercms-frontend/issues/160. This PR also addresses the text overflow issue raised in https://github.com/isomerpages/isomercms-frontend/issues/146.

In addition to addressing the above issues, several new features have been added in this PR:
- The title of the homepage is now changeable in the home page edit page
- The notification bar can be added to the home page
- The form input for images has been modified to use an image modal instead. The selection of favicon in the settings page has also been modified to utilise this
- The infopic section now saves and displays correctly, where previously the images did not show up due to different field names
- The infopic section now has validation on fields as well

Screenshot 2020-10-22 at 5 10 39 PM


src/components/FormFieldImage.jsx
```diff
@@ -11,24 +12,70 @@ const FormFieldImage = ({
...
+  const onImageClick = (path) => {
+    setIsModalOpen(false)
+      const event = {
+        target: {
+          id: id,
+          value: path,
+          parentElement: {
+            parentElement: {
+              id: '',
+            },
```
kwajiehao, line 29: Is this `parentElement` being used anywhere?

src/layouts/EditHomepage.jsx
```diff
@@ -33,62 +33,62 @@ import DeleteWarningModal from '../components/DeleteWarningModal';
...
-    button: '',
-    url: '',
-    imageUrl: '',
-    imageAlt: '',
+    title: isErrorConstructor ? '' : 'Infopic Title',
+    subtitle: isErrorConstructor ? '' : 'Infopic Subtitle',
+    description: isErrorConstructor ? '' : 'Infopic description',
+    button: isErrorConstructor ? '' : 'Button Text',
+    url: isErrorConstructor ? '' : '/faq/',
+    image: isErrorConstructor ? '' : '/image/',
```
kwajiehao, line 61: `/faq` and `/image` might not necessarily exist on the user's repo. I'm wondering if we can replace these with default text that is more descriptive, such as `No URL specified` and `No image selected`, but I understand it would be a bit weird since in the future we might want to validate these fields. However I did test it locally and the site builds even with nonsense values for `url` and `image`. What are your thoughts?

kwajiehao (changes requested):
Overall, this looks really good, nice job! Code is clean and easy to understand. In addition to the comments I've left, I have one more nit:

Screenshot 2020-10-26 at 5 44 49 PM

Could we match the icon background color with the background color of the notification bar?

The only major change that needs to be made is re: point 4 in issue #160 ("Add a new section" is misleading"), which we've discussed online.

src/components/FormFieldImage.jsx
alexanderleegs: As discussed, this will be handled in a future PR. Filed an issue to address this: https://github.com/isomerpages/isomercms-frontend/issues/177

src/layouts/EditHomepage.jsx
kwajiehao: After discussing with designers, we want to:
- Allow default value of `url` to be an empty string, and set the input element's placeholder value to be `Insert permalink or external URL`
- Allow default value of `image` to be an empty string

alexanderleegs: Modified after discussion with designers. Url is now left blank with different placeholder text, and image is left blank as it is handled by the modal. Resolved in https://github.com/isomerpages/isomercms-frontend/pull/172/commits/56b6b3f0ac39b706900c743c667387e01b184e97

src/components/homepage/InfopicSection.jsx
```diff
@@ -21,8 +22,9 @@ const EditorInfopicSection = ({
   shouldDisplay,
   displayHandler,
   errors,
+  siteName,
 }) => (
-  <div className={elementStyles.card}>
+  <div className={`${elementStyles.card} move`}>
     <div className={elementStyles.cardHeader}>
       <h2>
 Infopic section:
```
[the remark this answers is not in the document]
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/172/commits/58ebe9245cd27a5bf89254bcb681197e60c25c5b

kwajiehao (approved):
Lgtm!

Issue #148, opened 2020-10-08
Left nav in EditCollectionPages is not being rendered correctly


kwajiehao:
The left nav bar in the `EditCollectionPages` layout is not being rendered correctly, as seen in the following screenshot (taken from `isomerpages-hlb-cms-test` repo):

<img width="1010" alt="Screenshot 2020-10-08 at 2 19 00 PM" src="https://user-images.githubusercontent.com/31984694/95422305-a3dc2480-0971-11eb-8972-f5b8b9ce1527.png">

We can also refer to the `mse` repo for further test cases.

kwajiehao:
It turns out that the problem in this particular case was due to misnaming of the files in the collection
<img width="1679" alt="Screenshot 2020-10-21 at 8 16 38 AM" src="https://user-images.githubusercontent.com/31984694/96657848-cc065300-1375-11eb-86a1-c9419bfcd849.png">

However, the issue still stands: we don't currently allow for dropdown chevrons in the `EditCollectionPage`:
_Current_
<img width="942" alt="Screenshot 2020-10-21 at 8 17 36 AM" src="https://user-images.githubusercontent.com/31984694/96657890-e93b2180-1375-11eb-83b4-e60aafa47813.png">

_Expected_
<img width="1268" alt="Screenshot 2020-10-21 at 8 17 41 AM" src="https://user-images.githubusercontent.com/31984694/96657899-efc99900-1375-11eb-8281-4bedbfce236a.png">

kwajiehao:
Resolved with #178

PR #169, opened 2020-10-20
Fix/unstack image upload modal

kwajiehao:
## Overview

This PR resolves a litany of UI issues:
- #161: Shift `Heading` button in markdown editor to the front of the toolbar
- #163: Unstack media modals so that there is only one modal open at a time
- #164: Reposition action buttons (save or delete actions) to the bottom right hand corner
- Refactor to subsume the `EditResourcePage` component into `EditPage` since there is a lot of duplicated code

## Screenshots

Issue #163 
_Before_
Screenshot 2020-10-20 at 4 51 15 PM

_After_
Screenshot 2020-10-20 at 4 50 43 PM

Issue #164 
_Before_
Screenshot 2020-10-20 at 4 52 19 PM
Screenshot 2020-10-20 at 4 52 38 PM

_After_
Screenshot 2020-10-20 at 4 53 02 PM
Screenshot 2020-10-20 at 4 53 17 PM

PR #178, opened 2020-10-28
Fix/revamp EditCollectionPage component

kwajiehao:
## Overview

This pull request makes two major changes:
1. Subsume the `EditCollectionPage` component into `EditPage`
2. Revamp the existing `LeftNav` template so that the accordion is properly separated from the page content, and so that the accordion displays third nav elements properly. For reference, the below image gives an example of what a _normal accordion element_ is, and what a _third nav element_ is:

Screenshot 2020-10-28 at 10 28 40 PM

In accomplishing the above, we also fix issue #148.

## Combining `EditCollectionPage` into `EditPage`

In a previous PR (#169), we had refactored the `EditPage` component so that it could also be used for resources, thus replacing `EditResourcePage`. This PR continues in that vein by refactoring `EditPage` so that it can also be used for collection pages. The main changes that needed to be made were:
- create an additional `isCollectionPage` prop to indicated whether the `EditPage` is used for a collection page
- create a function to choose the API endpoint to hit depending on the `isResourcePage` and `isCollectionPage` props (for example, if `isResourcePage` is `false` but `isCollectionPage` is `true`, we want to choose the collection pages endpoint)
- fetch all pages in the same collection, so that it can be passed to the `LeftNavPage` template to construct the accordion properly

## Revamping the `LeftNav` template

Previously, the `LeftNav` template was unable to properly display third  nav elements. There were also spacing issues that were caused by improper use of inline styles. Refer to issue #148 for an example of the problem.

The `LeftNav` template is complicated mainly because we need to reconstruct the accordion logic. As mentioned above, there are two types of accordion elements, normal elements and third nav elements.

The main factor which dictates an accordion element’s appearance is **whether the current page being displayed is part of a third nav**.

Some other factors to consider for context:
- Accordion dynamism: if the accordion element is a third nav element, then it must come with a third nav header, and the third nav header must be clickable so that it expands to display the third nav elements under the third nav header, or to hide already visible third nav elements
- If the accordion element corresponds to the page being displayed, then it must be displayed as active (bolded/highlighted)
- If the accordion element happens to BOTH be a third nav element AND corresponds to the page being displayed, then the element itself must be bolded and its third nav header must be bolded as well

 ### If current page being displayed is part of a third nav
- The third nav header of the accordion element which corresponds to the current page needs to be active
- The third nav header chevron must point UP
- The third nav accordion element and its sibling third nav elements must be displayed and not hidden
- The third nav accordion element corresponding to the current page must be active, but its sibling third nav elements must not be active
- Other third nav sections (headers + accordion elements) must not be displayed and not be active

**Example**
Screenshot 2020-10-28 at 10 47 43 PM

 ### If current page being displayed is NOT part of a third nav
- The accordion element corresponding to the current page must be active
- None of the other third nav sections should be displayed, and none of them should be active
- Any third nav sections must have their chevron point DOWN

**Example**
Screenshot 2020-10-28 at 10 51 26 PM

### Test cases
The following edge cases were tested:

1. Multiple third nav sections

Normal element
Normal element
Third nav section (consists of third nav header + third nav elements)
Normal
Third nav section

Screenshot 2020-10-28 at 10 45 44 PM

2. Consecutive third nav sections

Normal element
Third nav section
Third nav section
Normal element

Screenshot 2020-10-28 at 10 44 30 PM
Screenshot 2020-10-28 at 10 44 51 PM


alexanderleegs (commented):
Currently, there's a bug where third navs with only a single element do not show up, but third navs are supposed to be able to have any number of elements.

alexanderleegs (approved):
lgtm. One thing to note is that the display of 3rd nav titles is inconsistent on cms vs on the actual website currently, due to the cms retrieving the title from the filename, whereas the website retrieves the title from the title field in the file currently. As per discussion, we can look into this in a future PR.

PR #181, opened 2020-10-29
fix: missing `onSave` prop for MediaSettingsModal in Images layout

kwajiehao:
## Overview
This PR fixes a simple bug where the `onSave` prop was missing for the `MediaSettingsModal` in the `Images` layout. This led to the following error message:

```
Unhandled Rejection (TypeError): onSave is not a function
```

This PR fixes that by defining the `onSave` prop so the `Images` layout is reloaded when a new image is uploaded. It also sets the `onSave` prop as a required prop type.

### DIFF
--- a/src/components/media/MediaSettingsModal.jsx
+++ b/src/components/media/MediaSettingsModal.jsx
@@ -212,5 +212,6 @@ MediaSettingsModal.propTypes = {
   type: PropTypes.oneOf(['image', 'file']).isRequired,
   siteName: PropTypes.string.isRequired,
   onClose: PropTypes.func.isRequired,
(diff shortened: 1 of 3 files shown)


alexanderleegs (changes requested):
Could we add this change to Files as well, since it also suffers from the same bug currently?

kwajiehao:
> Could we add this change to Files as well, since it also suffers from the same bug currently?

Fixed in [285b9ea](https://github.com/isomerpages/isomercms-frontend/pull/181/commits/285b9ea245942582e9f73a90aded1a60a04c052d)

alexanderleegs (approved):
lgtm

PR #182, opened 2020-10-29
fix: image filename validation bug

kwajiehao:
## Overview

This PR fixes a simple image filename validation bug. Currently, file names which contain more than one full stop, such as
'Screenshot 2020-10-28 at 10.51.26 PM.png' are provided with the current error message: `Invalid filename: filename must end with a valid file extension (.JPG, .png, .pdf, etc.)`

Screenshot 2020-10-29 at 6 00 55 PM

However, this is not an accurate descriptor of the problem. This commit creates a separate error message for situations where there are more than one full stop in the filename. The new error message now reads: `Invalid filename: filename can only contain one full stop and must follow the structure {name}.{extension}`

### DIFF
--- a/src/utils/validators.js
+++ b/src/utils/validators.js
@@ -605,7 +605,10 @@ const validateFileName = (value) => {
   }
 
   const fileNameArr = value.split('.')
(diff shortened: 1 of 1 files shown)


alexanderleegs (approved):
lgtm

PR #189, opened 2020-11-02
Feat/automatic logout

kwajiehao:
## Note

This PR is to be reviewed with PR #[63](https://github.com/isomerpages/isomercms-backend/pull/63) in the backend.

## Overview

This PR introduces automatic logout functionality on IsomerCMS. When a user's access token expires, any HTTP request made by the frontend will result in a `401` status response from the backend. When the frontend receives this error response, it will display an alert and log the user out upon acknowledgement of the alert.

Screenshot 2020-11-02 at 11 06 18 AM

## Details

The auto-logout functionality relies on several components:
- `isLoggedIn` state hook which determines page routing
- Updated routing rules dependent on `isLoggedIn`
- axios global interceptor, configured to respond to all `401` status error responses

The main component is the axios interceptor. When a user clicks any button to navigate to any page on our CMS, the frontend sends a request, with the user's access token wrapped in a JWT, to the backend to request the relevant resources on GitHub, so that we can display that information on the frontend. However, when the JWT has expired, we have configured our backend to throw a `401` error. Since we know that this will happen, we configure our axios interceptor such that when a `401` status response is received, we set the `isLoggedIn` state to be `false` and hit the logout endpoint on the backend to remove the isomerCMS cookie. We also create an intermediate step where we display an alert to warn the user that we are logging them out.

### Page routing

This PR also makes updates to page routing, which now depends on the login state. The rules implemented are as follows.

**If logged in:**
- Send all requests to their appropriate components (for example, `/sites/demo-v2/pages` will display the `Pages` component)
- If requested URL is not specified, redirect to the `/sites` page
- If requesting `/` path, redirect to the `/sites` page

**If logged out:**
- Redirect all requests to login page

### Order of events

With the auto-logout functionality, it is important to get the order of events correct. For example, one bug that was encountered during the development process was that upon login, the user would be sometimes be able to access the CMS, and sometimes be alerted and logged out immediately. This was because the login state, `isLoggedIn`, was set to `true` **BEFORE** the frontend had received the user's access token from the backend. 

Since state updates are asynchronous in React, the failure only happens sometimes - sometimes, the state update takes longer than the network call to retrieve the access token, and sometimes it takes less time. If the state update takes less time, then the frontend will attempt to render the `/sites` layout without a valid access token, which results in `401` response from the backend when we attempt to retrieve sites data, and which in turn, results in an automatic logout.

Therefore, we needed to set the state only **AFTER** the backend had performed the redirect to our frontend. Since the original redirect was to the `/sites` layout, this would have unnecessarily complicated that layout's code. The solution was to create an intermediate `AuthCallback` layout, which the backend redirects to instead of `/sites`. Upon mounting of the `AuthCallback` component, we set login state to be `true` and redirect to the `/sites` page.

The new authentication flow is thus:
1. User clicks on GitHub auth link to kick off OAuth flow
2. GitHub sends a callback to the backend at `/auth`
3. Backend receives the access token, and performs a `res.redirect` to `{FRONTEND_URL}/auth#isomercms`
4. Frontend performs the redirect to `/auth` and sets `isLoggedIn` to `true`
5. `/auth` performs a redirect to `/sites`


alexanderleegs (approved):
lgtm

PR #198, opened 2020-11-03
fix: redirect to /sites only after login state changes to true

kwajiehao:
## Overview 
This PR fixes a bug which was introduced in PR #189 

Previously, on the `AuthCallback` component, we set the login state to be true and immediately returned a redirect to the `/sites` page. The problem with this is that it's possible that the redirection occurs before the login state is set.

This PR makes the changes to redirect only AFTER login state has been set by introducing a `shouldRedirectToSites` hook, which the redirection is dependent on.

### Other notes
Note that we currently use a context string, `#isomercms`, to filter access to the `AuthCallback` component. If this filter string is not present, the component will not be accessible.

This leads to a small bug. If the user attempts to access `/auth#isomercms`, they will be presented with the following screen and logged out. We can fix this in a later PR.

Screenshot 2020-11-03 at 5 41 10 PM

### DIFF
--- a/src/components/ProtectedRoute.jsx
+++ b/src/components/ProtectedRoute.jsx
@@ -12,9 +12,7 @@ const ProtectedRoute = ({ component: WrappedComponent, isLoggedIn, ...rest }) =>
         <Route {...rest} render={
             props => {
                 if (rest.location.pathname === '/auth') {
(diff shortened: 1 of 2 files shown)


alexanderleegs (approved):
lgtm

PR #199, opened 2020-11-03
hotfix: add missing `isLoggedIn` prop

kwajiehao:
This PR adds the missing `isLoggedIn` prop to wrapped component inside `ProtectedRoute`.

It also ensures that the `setLogin` callback is run if the login state is false.


alexanderleegs (approved):
lgtm

PR #201, opened 2020-11-04
Fix/remove cookie checker

kwajiehao:
## Overview

This commit removes the use of the cookie checker used to initialize the `isLoggedIn` hook in PR #189 . The cookie checker has proven to be problematic because of how it writes a new cookie if the cookie doesn't yet exist. While it works locally and on the staging-dev environment, it does not work in staging for reasons we have yet to understand.

As a workaround, we are going to use local storage on the browser to set a state on login, and to attempt to retrieve it on application mount. If the state exists in local storage, then we initialize login state to be true. Otherwise, the user is not logged in.

Note that this means that if the user manually deletes the state from local storage, they will be immediately logged out without warning. However, since our users have no reason to do this, it's not a big concern.


alexanderleegs (approved):
lgtm

Issue #179, opened 2020-10-29
Bug: Implement isomer's CSP on page preview


lisatjide:
Currently the page preview will show all the unallowed embedding or components as it does not contain the same CSP as isomer sites. 
<img width="631" alt="Screenshot 2020-10-29 at 3 17 48 PM" src="https://user-images.githubusercontent.com/23736580/97537551-ee454400-19f9-11eb-96f5-b936b57b1971.png">

**Intended behaviour:**
When user upload/embed content that is blocked, it should show an error 

This is important as page preview is an indication on whether certain content is allowed.

kwajiehao:
A high-level approach to this would be to retrieve the `netlify.toml` file from the repo and to parse the CSP, and apply it to the HTML output from the SimpleMDE editor. Currently, we already do this (`prependImgSrc`) so it's definitely doable.

kwajiehao:
Solved with #205

PR #205, opened 2020-11-05
Feat/csp check

gweiying:
This PR introduces utilities to check editor markdown content for unsupported elements in EditPage, resolving #179.

## Implementation
At a high level, this is done by first retrieving and parsing the `netlify.toml` file from the website repo to extract the CSP for the repo. The user-edited markdown content is compiled into html, and inspected for any elements that violate the CSP. The violating elements are replaced with an error message, and the Save button is disabled until the user resolves the error message. An example of this is shown in the screenshot below. 

Screenshot 2020-11-09 at 9 10 24 AM

## Details

The CSP of a website specifies accepted sources for scripts on the website. Only scripts received from domains specified in the respective fetch directives can executed on the website. Currently, only 5 fetch directives and their fallbacks are supported in our CSP checking, `frame-src`, `img-src`, `script-elem-src`, `object-src`, `media-src`.

For each fetch directives, the types of html elements restricted are as follows:

- `frame-src`: `frame`, `iframe`
- `img-src`: `img`
- `media-src`: `audio`, `video`, `track`
- `object-src`: `object`, `embed`, `applet`
- `script-src-elem`: `script`

For example, with a `media-src` policy of `'none'`, `<audio src="animal.mp3" controls></audio>` will be considered a violating element. 

If the `frame-src` is not specified, the checker falls back to the `child-src` policy value ([link1](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src) [link2])(http://csplite.com/csp/test121/), and defaults to `default-src` if `child-src` is not specified either. Similarly, if `script-src-elem` is not specified, `script-src-elem` falls back to `script-src` and defaults to `default-src` if `script-src` is not specified. All other fetch directives default automatically to `default-src` if not specified.

### DIFF
--- a/src/utils/cspUtils.js
+++ b/src/utils/cspUtils.js
@@ -9,21 +9,6 @@ import { isLinkInternal } from '../utils';
 import cheerio from 'cheerio';
 import escapeStringRegexp from 'escape-string-regexp';
 
-/* Helper function to retrieve the netlify.toml from repo */
-async function _parseNetlifyToml(repoName) {
-  // axios get withCredentials false is required https://stackoverflow.com/questions/34078676/access-control-allow-origin-not-allowed-when-credentials-flag-is-true-but/34099399 
-  const tomlUrl = `https://raw.githubusercontent.com/isomerpages/${repoName}/staging/netlify.toml`;
-  const tomlFile = await axios.get(tomlUrl, { withCredentials: false } ); 
-  return toml.parse(tomlFile.data);
-};
-
-export async function getCSP(repoName) {
-  const tomlData = await _parseNetlifyToml(repoName);
-  const csp = tomlData.headers[0].values['Content-Security-Policy'];
-  const cspPolicy = new Policy(csp);
-  return cspPolicy;
-};
-
 function _stringContainsValue(string, value) {
   // regex checks specifically if value is preceded by whitespace or is at the start/ end of the string
   const VALUE_REGEX = `(\\s|^)${value}(\\s|$)`;
--- a/src/utils/cspUtils.js
+++ b/src/utils/cspUtils.js
@@ -1,5 +1,5 @@
-// Supported CSP checks are img-src, frame-src, media-src, object-src script-src
-// TODO: child-src script-src-elem
+// Supported CSP checks are img-src, frame-src, media-src, object-src script-src-elem 
+// Supported fallbacks are child-src, script-src, and default-src
 
 import toml from 'toml';
 import axios from 'axios';
@@ -74,16 +74,22 @@ function _elemAttrSatisfiesPolicies (elemAttr, policy) {
 /* Helper function to get resource policy from csp for given policyType*/
 function _getResourcePolicy(csp, policyType) {
   const resourcePolicyMapping = { 
-    // 'child-src': ['frame', 'iframe'],
     'frame-src': ['frame', 'iframe'],
     'img-src': ['img'],
     'media-src': ['audio', 'video', 'track'],
     'object-src': ['object', 'embed', 'applet'],
-    'script-src': ['script'], 
-    // 'script-src-elem': ['script'],
+    'script-src-elem': ['script'],
   };
   const cspPolicy = new Policy(csp);
-  const resourcePolicy = cspPolicy.get(policyType) === '' ? cspPolicy.get('default-src') : cspPolicy.get(policyType);
+
+  let resourcePolicy
+  if (policyType === 'frame-src') { // from http://csplite.com/csp/test121/, fallback chain: frame-src -> child-src -> default-src
+    resourcePolicy = cspPolicy.get(policyType) ||  cspPolicy.get('child-src') || cspPolicy.get('default-src')
+  } else if (policyType === 'script-src-elem') { // fallback chain: script-src-elem -> script-src -> default-src
+    resourcePolicy = cspPolicy.get(policyType) ||  cspPolicy.get('script-src') || cspPolicy.get('default-src')
+  } else {
+    resourcePolicy = cspPolicy.get(policyType) === '' ? cspPolicy.get('default-src') : cspPolicy.get(policyType);
+  }
   const resourcePolicyElems = resourcePolicyMapping[policyType];
   return { resourcePolicy, resourcePolicyElems } 
 }
@@ -106,7 +112,7 @@ function _checkResourcePolicyElems(resourcePolicyElems, policy, $) {
 export function checkCSP(
   csp,
   chunk,
-  policyTypes = ['img-src', 'media-src', 'frame-src', 'object-src', 'script-src'],
+  policyTypes = ['img-src', 'media-src', 'frame-src', 'object-src', 'script-src-elem'],
   ) {
 
   let $ = cheerio.load(chunk);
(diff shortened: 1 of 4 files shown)


src/utils/cspUtils.js
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/* Helper function to check if resourcePolicyElems for a specific policyType satisfies CSP source specifications */
+function _checkResourcePolicyElems(resourcePolicyElems, policy, $) {
+  let policyViolation = false;
+  resourcePolicyElems.forEach(elemType => { 
+    $(elemType).each((i, elem) => {
+      const checkAttr = elemType === 'object' ? 'data' : 'src' // exception for object html: <object data='abc.html'></object>
+      if (!_elemAttrSatisfiesPolicies($(elem).attr(checkAttr), policy)) {
+        $(elem).replaceWith(`<span style="color:#c91508"> Intended &lt${elemType}&gt content violates Content Security Policy and therefore could not be displayed. Isomer does not support display of any forbidden resources. </span>`);
```
kwajiehao, line 104: Nit: What is the rationale for using `<span>` instead of `<p>`? Minor issue, but if there are subsequent violations, there is no line break between the warning text if we use `span`.

Screenshot 2020-11-10 at 6 25 54 AM

src/utils/cspUtils.js
```diff
@@ -0,0 +1,133 @@
...
+  let $ = cheerio.load(chunk);
+  let isCspViolation = false;
+
+  policyTypes.forEach(policyType => {
+    let policyViolation;
+    const { resourcePolicy, resourcePolicyElems } = _getResourcePolicy(csp, policyType)
+    console.log(`checking resource policy for ${policyType}, for elements ${resourcePolicyElems}: ${resourcePolicy}`);
+    ({ $, policyViolation } = _checkResourcePolicyElems(resourcePolicyElems, resourcePolicy, $));
+    isCspViolation = policyViolation || isCspViolation;
+  })
```
kwajiehao, line 127: You can use a reducer function for this section, with the initial value being the `$` variable loaded from the chunk https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

src/utils/cspUtils.js
```diff
@@ -0,0 +1,133 @@
...
+  const specialValues = ['http:', 'https:', 'data:', 'mediastream:', 'blob:', 'filesystem:', "'self'", "'none'", '*'];
+  const hostsources = policy.split(' ').filter(value => (!specialValues.includes(value)));
+  
+  const hostsourcesSatisfied = hostsources.some(hostsource => (_toRegExp(hostsource).test(elemSrc)));
+  return hostsourcesSatisfied;
+};
+
+function _checkSchemasourcePolicy (elemSrc, policy) {
+  const schemes = ['http:', 'https:'];
+  const dataschemes = ['data:', 'mediastream:', 'blob:', 'filesystem:'];
```
kwajiehao, line 53: nit: convert `dataschemes` variable name to camelCase

src/utils/cspUtils.js
```diff
@@ -0,0 +1,133 @@
...
+
+/* Helper function to get resource policy from csp for given policyType*/
+function _getResourcePolicy(csp, policyType) {
+  const resourcePolicyMapping = { 
+    'frame-src': ['frame', 'iframe'],
+    'img-src': ['img'],
+    'media-src': ['audio', 'video', 'track'],
+    'object-src': ['object', 'embed', 'applet'],
+    'script-src-elem': ['script'],
+  };
```
kwajiehao, line 82: Since this variable never changes, can we put this as a global variable?

kwajiehao (changes requested):
this is very well-researched and i like how the functions are broken down, it was very easy to read. left comments and will follow-up offline

src/utils/cspUtils.js
gweiying: I was following the format for error messages in form fields, for example [here](https://github.com/isomerpages/isomercms-frontend/blob/5daede304bb09c8752748168459afc7eae564d71/src/components/FormFieldMedia.jsx#L126), but I can replace it with `<p>` too, what do you think?

src/utils/cspUtils.js
gweiying: That's possible, but since it's only ever used in this function would it be tidier for this variable to only be in the scope of this function?

src/utils/cspUtils.js
```diff
@@ -0,0 +1,133 @@
...
+  policyTypes = ['img-src', 'media-src', 'frame-src', 'object-src', 'script-src-elem'],
+  ) {
+
+  let $ = cheerio.load(chunk);
+  let isCspViolation = false;
+
+  policyTypes.forEach(policyType => {
+    let policyViolation;
+    const { resourcePolicy, resourcePolicyElems } = _getResourcePolicy(csp, policyType)
+    console.log(`checking resource policy for ${policyType}, for elements ${resourcePolicyElems}: ${resourcePolicy}`);
```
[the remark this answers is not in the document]
gweiying: Good idea, will do

src/utils/cspUtils.js
```diff
@@ -0,0 +1,133 @@
...
+  const tomlData = await _parseNetlifyToml(repoName);
+  const csp = tomlData.headers[0].values['Content-Security-Policy'];
+  return csp;
+};
+
+function _stringContainsValue(string, value) {
+  const VALUE_REGEX = `(\\s|^)${value}(\\s|$)`;
+  const ValueRegexTest = new RegExp(VALUE_REGEX);
+  return ValueRegexTest.test(string);
+};
```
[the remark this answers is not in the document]
gweiying: The regex checks specifically for if these values are preceded by whitespace or are at the start/ end of the line, for example in the case of the scheme-source `https:`, the regex returns true for `https:` and false for `https://www.google.com`. 

Correct me if I'm wrong but this does not seem possible with `includes`, or it seems more complicated with `includes`?

src/utils/cspUtils.js
gweiying: e0b1d3c

src/utils/cspUtils.js
gweiying: 600897b

src/utils/cspUtils.js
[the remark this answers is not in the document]
gweiying: 557f57a

src/utils/cspUtils.js
kwajiehao: Yup line break works

src/utils/cspUtils.js
kwajiehao: Sure thing. If this turns out to affect performance down the line we can review it again

src/utils/cspUtils.js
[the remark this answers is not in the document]
kwajiehao: Ah ok I see what you mean. My only (slight) concern is that the regex isn't as easy to read, but I think it's okay as it is.
An alternative would be to return the policy as an array (split on a whitespace ` `) so that this function would be `_policyContainsValue` and you can simply search `arr.includes(value)`.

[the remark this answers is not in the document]
gweiying: 77011dc

src/utils/cspUtils.js
gweiying: e259976

kwajiehao (approved):
lgtm

PR #206, opened 2020-11-05
Feat/collections layout

### LINKED ISSUE #204: Implement file hierarchy for collections
This issue is similar to #202 but it's complicated by the fact that collection pages can contain third nav sections. At the top-level collection page, we want to allow users to create a page, as well as assign it to a collection and a third nav within that collection. However, to do that, we would need to retrieve all pages from all collections to be able to identify all third navs which can be assigned. This is bad because it means the load speed will not improve from the current code.

Instead, we should implement a bunch of YAML files which store the structural contents of each collection. For example, a collection called `Collection A` would have a `_data/collections/collection_a.yml` file which contains:

```
pages:
  - type: page
     title: ABC
  - type: page
     title: DEF
  - type: third-nav
     title: Sample third-nav
     contents:
       - type: third-nav-page
          title: GHI
```

From these files, we would be able to determine which third nav sections exist. We can also implement them in a lazy manner - meaning, only retrieve this structure file when the user selects a collection to assign a page to.

kwajiehao:
**Note: This PR is to be reviewed with PR #[68](https://github.com/isomerpages/isomercms-backend/pull/68) on the backend**

## Overview

As previously discussed, we want to replace our flat `Pages` layout with a more hierarchical file system-like structure which people are more familiar with (think GDrive). This PR adds a new `Workspace` layout to replace the previous `Pages` layout in accordance with the designs specified in our Figma wireframes ([link](https://www.figma.com/file/4qNhGTBMs521pDvCAIh4ON/IsomerCMS?node-id=159%3A54)). This new layout will allow users to access the homepage, click into collections to access collection pages, and unlinked pages. 

To support the hierarchical system, a new component, `NewCollectionPages`, was also created. When a user clicks on a collection in their workspace, they are routed to the `NewCollectionPages` component, which displays all the pages in that particular collection. This layout can also be reused by resources, so we might want to rename it and generalize it.

**Note: this PR does not implement the full functionality needed for the CMS to work!**

What this PR does:
- Visually implement the file system hierarchy for pages and collection pages. This includes updating app routing, and implementing settings functionality (ensuring the `ComponentSettingsModal` works with the elements in the `Workspace` and `NewCollectionPages` layout), and updating the `Back to <layout>` button in the header to redirect to the proper pages

What this PR does not do:
- It does not add third nav functionality for collection pages
- It does not visually implement the file system hierarchy for resources

This PR partially resolves #204 - more needs to be done in a separate PR to implement functionality for third nav items.

## Other changes

This PR makes a few other notable changes:
- update the logic for disabling the category dropdown in `ComponentSettingsModal` so that it is disabled when users try to create a new page **within** a collection
- tweak the styles of `FolderCard` so that it's a horizontal flexbox instead of a vertical one

## Screenshots

**Before**
Screenshot 2020-11-06 at 4 40 25 AM

**After**
Screenshot 2020-11-06 at 4 31 48 AM

Screenshot 2020-11-06 at 4 32 17 AM

_After clicking into one of the collections_
Screenshot 2020-11-06 at 4 31 57 AM


src/components/ComponentSettingsModal.jsx
```diff
@@ -272,11 +281,12 @@ export default class ComponentSettingsModal extends Component {
                 {/* Category */}
                 <p className={elementStyles.formLabel}>Category Folder Name</p>
                 <div className="d-flex">
-                  <select className="w-100" id="category" value={category} onChange={this.changeHandler} disabled={!isNewFile}>
-                    {
+                  <select className="w-100" id="category" value={category} onChange={this.changeHandler} disabled={isCategoryDropdownDisabled(isNewFile, category)} >
```
alexanderleegs, line 284: Using `category` to determine whether this field is disabled causes a bug when creating new files: once the user selects a category, the file is locked into that choice, so the user is unable to make a change afterwards, even though the file has not been created. We should use `prevCategory` from the state instead.

src/layouts/Resources.jsx
```diff
@@ -255,7 +255,10 @@ export default class Resources extends Component {
     return (
       <>
         {/* Resource Room exists */}
-        <Header />
+        <Header
+          backButtonText="Back to My Workspace"
+          backButtonUrl={`/sites/${siteName}/workspace`}
+        />
```
alexanderleegs, line 261: Should Resources be redirecting to Workspace, since all other items in the nav bar redirect to Sites?

alexanderleegs (changes requested):
Left comments. 

In addition, currently inside the collection folder, none of the items in the nav bar are highlighted. Should we leave `My Workspace` highlighted inside the folder as well?
Screenshot 2020-11-06 at 11 38 04 AM

src/layouts/Resources.jsx
kwajiehao: fixed in [bf2adb0](https://github.com/isomerpages/isomercms-frontend/pull/206/commits/bf2adb0d48e4dcb163b47abdb186a7b720fc75d7)

kwajiehao: fixed in [7ddfd3](https://github.com/isomerpages/isomercms-frontend/pull/206/commits/7ddfd3a49c5f3a43cc378819625b54c356e025f6)

src/components/ComponentSettingsModal.jsx
kwajiehao: fixed in [bf2adb0](https://github.com/isomerpages/isomercms-frontend/pull/206/commits/bf2adb0d48e4dcb163b47abdb186a7b720fc75d7)

kwajiehao:
> In addition, currently inside the collection folder, none of the items in the nav bar are highlighted. Should we leave `My Workspace` highlighted inside the folder as well?
> Screenshot 2020-11-06 at 11 38 04 AM

Made the fixes, I also highlighted `My Workspace` if `CategoryPages` is accessed as a collection folder, and highlight `Resources` if it's accessed as a resource category [f9b328d](https://github.com/isomerpages/isomercms-frontend/pull/206/commits/f9b328db8a3bc1d4cb41d27332b94242f9237b4e). I also set the back button text and url for `CategoryPages` to adjust depending on whether it's accessed by a collections route or a resource category route

src/components/CollectionPagesSection.jsx
```diff
@@ -0,0 +1,111 @@
...
+                            _.chain(pages)
+                                .map((page) => page.fileName)
+                                .value()
+                        }
+                    />
+                )
+            }
+            <div className={contentStyles.contentContainerBoxes}>
+                {/* Display loader if pages have not been retrieved from API call */}
+                { pages
```
alexanderleegs, line 65: The check should be `pages && pages.length > 0` as the pages state is initialised to an empty array

alexanderleegs (changes requested):
Changes look good, but one more part I missed out earlier

src/components/CollectionPagesSection.jsx
kwajiehao: Let me check with the designers on whether we should render the "Add new page" button first before the other `OverviewCard`s are rendered, or if we should render them together at the same time

kwajiehao: Actually there's no need to ask them on something so minor, I made the fix in [1f1b783](https://github.com/isomerpages/isomercms-frontend/pull/206/commits/1f1b7833c729b1123cac0718b475311efb700f43)

alexanderleegs (approved):
lgtm

Issue #202, opened 2020-11-05
Implement file hierarchy for resources 


kwajiehao:
We should implement the new file hierarchy components for resources. What this means is that when a user clicks into resources, they'll see only the available resource categories as folders to click into. When they click into a specific category/folder, they will then see the pages that are part of that category. This data can be stored at the top-level resources layout, meaning that a setter function is passed into each category component.

When creating a new resource at the top-level resource layout, the user should be able to select which category the resource should be a part of, or create a new resource category if needed. However, within a resource category, one will not have the option of selecting category when creating a resource.

PR #207, opened 2020-11-09
Feat/resource file hierarchy

alexanderleegs:
This PR implements a file hierarchy system for resources, similar to the existing system for pages, in order to resolve https://github.com/isomerpages/isomercms-frontend/issues/202. 

As Resources and Workspace share similar functionality, the existing CollectionPagesSection and CategoryPages have been modified slightly to allow Resources to reuse these classes. 

In addition, the ComponentSettingsModal has been expanded to allow for custom category names, to allow users to create new files.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -110,7 +110,7 @@ function App() {
                   <ProtectedRouteWithProps exact path='/auth' component={AuthCallback} setLogin={setLogin} />
                   <ProtectedRouteWithProps exact path="/" component={Home} />
                   <ProtectedRouteWithProps path="/sites/:siteName/collections/:collectionName/:fileName" component={EditPage} isCollectionPage={true} isResourcePage={false} />
-                  <ProtectedRouteWithProps path="/sites/:siteName/collections/:collectionName" component={CollectionPages} />
+                  <ProtectedRouteWithProps path="/sites/:siteName/collections/:collectionName" component={CollectionPages} isResource={false}/>
                   <ProtectedRouteWithProps path="/sites/:siteName/files/:fileName" component={EditFile} />
                   <ProtectedRouteWithProps path="/sites/:siteName/files" component={Files} />
                   <ProtectedRouteWithProps path="/sites/:siteName/images/:fileName" component={EditImage} />
@@ -119,6 +119,7 @@ function App() {
                   <ProtectedRouteWithProps path="/sites/:siteName/workspace" component={Workspace} />
                   <ProtectedRouteWithProps path="/sites/:siteName/homepage" component={EditHomepage} />
                   <ProtectedRouteWithProps path="/sites/:siteName/resources/:resourceName/:fileName" component={EditPage} isCollectionPage={false} isResourcePage={true} />
+                  <ProtectedRouteWithProps path="/sites/:siteName/resources/:collectionName" component={CollectionPages} isResource={true}/>
                   <ProtectedRouteWithProps path="/sites/:siteName/resources" component={Resources} />
                   <ProtectedRouteWithProps path="/sites/:siteName/menus/main-menu" component={EditNav} />
                   <ProtectedRouteWithProps path="/sites/:siteName/menus" component={Menus} />
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -13,7 +13,7 @@ import Home from './layouts/Home';
 import Sites from './layouts/Sites';
 import Workspace from './layouts/Workspace';
 import EditPage from './layouts/EditPage';
-import CollectionPages from './layouts/CategoryPages';
+import CategoryPages from './layouts/CategoryPages';
 import Images from './layouts/Images';
 import EditImage from './layouts/EditImage';
 import Files from './layouts/Files';
@@ -110,7 +110,7 @@ function App() {
                   <ProtectedRouteWithProps exact path='/auth' component={AuthCallback} setLogin={setLogin} />
                   <ProtectedRouteWithProps exact path="/" component={Home} />
                   <ProtectedRouteWithProps path="/sites/:siteName/collections/:collectionName/:fileName" component={EditPage} isCollectionPage={true} isResourcePage={false} />
-                  <ProtectedRouteWithProps path="/sites/:siteName/collections/:collectionName" component={CollectionPages} isResource={false}/>
+                  <ProtectedRouteWithProps path="/sites/:siteName/collections/:collectionName" component={CategoryPages} isResource={false}/>
                   <ProtectedRouteWithProps path="/sites/:siteName/files/:fileName" component={EditFile} />
                   <ProtectedRouteWithProps path="/sites/:siteName/files" component={Files} />
                   <ProtectedRouteWithProps path="/sites/:siteName/images/:fileName" component={EditImage} />
@@ -119,7 +119,7 @@ function App() {
                   <ProtectedRouteWithProps path="/sites/:siteName/workspace" component={Workspace} />
                   <ProtectedRouteWithProps path="/sites/:siteName/homepage" component={EditHomepage} />
                   <ProtectedRouteWithProps path="/sites/:siteName/resources/:resourceName/:fileName" component={EditPage} isCollectionPage={false} isResourcePage={true} />
-                  <ProtectedRouteWithProps path="/sites/:siteName/resources/:collectionName" component={CollectionPages} isResource={true}/>
+                  <ProtectedRouteWithProps path="/sites/:siteName/resources/:collectionName" component={CategoryPages} isResource={true}/>
                   <ProtectedRouteWithProps path="/sites/:siteName/resources" component={Resources} />
                   <ProtectedRouteWithProps path="/sites/:siteName/menus/main-menu" component={EditNav} />
                   <ProtectedRouteWithProps path="/sites/:siteName/menus" component={Menus} />
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -245,7 +245,7 @@ export default class ComponentSettingsModal extends Component {
     }
   }
 
-  selectChangeHandler = (newValue) => {
+  categoryDropdownHandler = (newValue) => {
     let event
     if (!newValue) {
       // Field was cleared
@@ -311,7 +311,7 @@ export default class ComponentSettingsModal extends Component {
                   <CreatableSelect
                     isClearable
                     className="w-100"
-                    onChange={this.selectChangeHandler}
+                    onChange={this.categoryDropdownHandler}
                     isDisabled={isCategoryDropdownDisabled(isNewFile, prevCategory)}
                     defaultValue={{value:category,label:category ? category : "Select a category or create a new category..."}}
                     options={
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -148,6 +148,9 @@ export default class ComponentSettingsModal extends Component {
         permalink: null,
         fileUrl: originalFileUrl ? originalFileUrl : '',
         isPost: false,
+        errors: {
+          permalink: '',
+        },
       });
     } else {
       this.setState({
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import axios from 'axios';
+import CreatableSelect from 'react-select/creatable'
 import { Base64 } from 'js-base64';
 import PropTypes from 'prop-types';
 import * as _ from 'lodash';
@@ -74,7 +75,7 @@ export default class ComponentSettingsModal extends Component {
         const resourcesResp = await axios.get(`${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/resources`);
         const { resources: allCategories } = resourcesResp.data;
         this.setState({
-          category: category ? category : allCategories[0].dirName,
+          category: category ? category : '',
           allCategories: allCategories.map((category) => category.dirName),
         })
       } else if (type === "page") {
@@ -82,7 +83,7 @@ export default class ComponentSettingsModal extends Component {
         const { collections: collectionCategories } = collectionsResp.data;
         const allCategories = [''].concat(collectionCategories)
         this.setState({
-          category: category ? category : allCategories[0],
+          category: category ? category : '',
           allCategories,
         })
       }
@@ -201,7 +202,6 @@ export default class ComponentSettingsModal extends Component {
       }
       
       const newPageUrl = await saveFileAndRetrieveUrl(fileInfo)
-      console.log(newPageUrl)
       // Refresh page
       !isNewFile && window.location.reload();
       this.setState({ redirectToNewPage: true, newPageUrl })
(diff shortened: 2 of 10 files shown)


src/layouts/Resources.jsx
```diff
@@ -1,378 +1,166 @@
...
 
-        this.setState({ resourceRoomName, resourceCategories, resourcePages });
+  useEffect(() => {
+    try {
+      const fetchData = async () => {
+        // Get the resource categories in the resource room
+        const resourcesResp = await axios.get(`${BACKEND_URL}/sites/${siteName}/resources`);
+        const { resourceRoomName, resources: resourceCategories } = resourcesResp.data;
+
+        if (resourceRoomName) {
```
kwajiehao, line 40: Check that `resourceCategories` has length > 0

src/layouts/Resources.jsx
```diff
@@ -1,378 +1,166 @@
...
+              resourceRoomName
+              ? <>
+                  {/* Category title */}
+                  <div className={contentStyles.segment}>
+                    Resource Categories
+                  </div>
+                  {/* Info segment */}
+                  <div className={contentStyles.segment}>
+                    <i className="bx bx-sm bx-info-circle text-dark" />
+                    <span><strong className="ml-1">Note:</strong> Categories cannot be empty, create a resource first to create a Category.</span>
```
kwajiehao, line 92: This should only be displayed if a resource room exists

src/layouts/Resources.jsx
```diff
@@ -1,378 +1,166 @@
...
+  const createResourceRoom = async () => {
     try {
-      const { match } = this.props;
       const { siteName } = match.params;
-      const { newResourceRoomName } = this.state;
       const params = { resourceRoom: newResourceRoomName };
-      await axios.post(`${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/resource-room`, params, {
-        withCredentials: true,
-      });
+      await axios.post(`${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/resource-room`, params);
```
kwajiehao, line 62: Either reload the page after the resource room is created, or use a hook so that the page is re-rendered. Currently, after a room is created, nothing happens on screen and there is no user feedback

src/layouts/Resources.jsx
```diff
@@ -1,378 +1,166 @@
...
+                        : 'Loading Resource Categories...'
+                      }
+                    </div>
+                  </div>
+                  {/* Segment divider  */}
+                  <div className={contentStyles.segmentDividerContainer}>
+                    <hr className="invisible w-100 mt-3 mb-5" />
+                  </div>
+                  {/* Pages */}
+                  <CollectionPagesSection
```
kwajiehao, line 123: There is a missing `isResource` prop that isn't being passed into the `CollectionPagesSection` component currently from the `Resources` layout. This results in a a situation where collections, instead of resource categories, are retrieved when trying to create a new resource

Steps to reproduce:
1. Delete all resource pages and resource room
2. Create a new resource room (i'm using the `a-test` repository)
3. Try to create a new resource. In the category dropdown, there should be no options, but the `ComponentSettingsModal` displays `test`, which is in fact a collection and not a resource category

src/layouts/Resources.jsx
```diff
@@ -1,378 +1,166 @@
...
+      <div className={elementStyles.wrapper}>
+        <Sidebar siteName={siteName} currPath={location.pathname} />
+        {/* main section starts here */}
+        <div className={contentStyles.mainSection}>
+          {/* Page title */}
+          <div className={contentStyles.sectionHeader}>
+            <h1 className={contentStyles.sectionTitle}>Resources</h1>
+          </div>
+          {
+            isLoading ? 'Loading Resources...'
```
kwajiehao, line 81: To be consistent with the `Workspace` layout, we can have some elements which are painted first (such as the `Resource Categories` title and the `Note: Categories cannot be empty, create a resource first to create a Category.` text, and paint the dynamic elements later, instead of loading them together all at once

src/App.jsx
```diff
@@ -110,7 +110,7 @@ function App() {
                   <ProtectedRouteWithProps exact path='/auth' component={AuthCallback} setLogin={setLogin} />
                   <ProtectedRouteWithProps exact path="/" component={Home} />
                   <ProtectedRouteWithProps path="/sites/:siteName/collections/:collectionName/:fileName" component={EditPage} isCollectionPage={true} isResourcePage={false} />
-                  <ProtectedRouteWithProps path="/sites/:siteName/collections/:collectionName" component={CollectionPages} />
+                  <ProtectedRouteWithProps path="/sites/:siteName/collections/:collectionName" component={CollectionPages} isResource={false}/>
```
kwajiehao, line 113: Can we rename `CollectionPages` to `CategoryPages` in this file? I think I did it in the other PR but it's not showing up

kwajiehao (changes requested):
Left comments

src/layouts/Resources.jsx
alexanderleegs: As discussed, we will leave this as is currently. The reason the behaviour differs from Workspace is because of the ResourceRoom, which causes something different to show up - since resources cannot be created without first providing a ResourceRoom name. However, we can only check the resource room name through an API call, so painting some elements first might be jarring if no resource room exists yet, since a user would see static elements first and suddenly have it disappear to be replaced with the resource room elements.

src/layouts/Resources.jsx
alexanderleegs: This is nested within the `resourceRoomName` check, so it only displays if a resource room exists

src/layouts/CategoryPages.jsx
```diff
@@ -27,16 +30,33 @@ const getBackButtonInfo = (pathname) => {
   }
 }
 
-const Collections = ({ match, location }) => {
+const Collections = ({ match, location, isResource }) => {
   const { backButtonLabel, backButtonUrl } = getBackButtonInfo(location.pathname)
   const { collectionName, siteName } = match.params;
 
-  const [collectionPages, setCollectionPages] = useState([])
+  const [collectionPages, setCollectionPages] = useState()
```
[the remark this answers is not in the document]
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/31dd245159815cafda93bfa43ccba598cea34c96

src/layouts/CategoryPages.jsx
```diff
@@ -27,16 +30,33 @@ const getBackButtonInfo = (pathname) => {
   }
 }
 
-const Collections = ({ match, location }) => {
+const Collections = ({ match, location, isResource }) => {
```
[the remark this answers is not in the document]
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/31dd245159815cafda93bfa43ccba598cea34c96

src/App.jsx
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/31dd245159815cafda93bfa43ccba598cea34c96

src/layouts/Resources.jsx
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/5f587ba310528d7d11a8acedf63b5e1620d4a241

src/layouts/Resources.jsx
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/0571ba015d848b32415ba10447d4753fdbeabccb

src/layouts/Resources.jsx
```diff
@@ -1,378 +1,166 @@
...
+    </>
+  );
 }
 
+export default Resources
+
 Resources.propTypes = {
   match: PropTypes.shape({
     params: PropTypes.shape({
-      siteName: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
alexanderleegs: Whoops must have accidentally removed it. Added back in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/84bccd7481a389cf1646dbc53481954fb940fd94

src/layouts/Resources.jsx
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/d91a4e0458b27fa5f2223db4d1df4ebcf78cccac

kwajiehao (changes requested):
A few other issues I discovered upon re-review.

1. When there are no resource categories, we still attempt to set state by accessing a `dirname` attribute.

```
 category: category ? category : allCategories[0].dirName,
```

This fails when `allCategories` is an empty array, as it should be when someone is creating a resource room for the first time.

2. Again, the context is when a resource room has just been created, but there are no categories. When you first attempt to create a resource, there is no permalink default value. But if you switch post type to `Downloadable file` and then back to `Post content`, the default value of `permalink` is restored.

Furthermore, if there is an error in the permalink field (for example, if you only type 2 characters, so it's below the min. char limit and displays and error message), then even if you switch from `Post content` to `Downloadable file`, the error message persists. Error messages should be cleared when the form field value is cleared - if an erroneous value for permalink is reinstated, the error message should only be reintroduced then.

3. I get the following warning message when modifying form fields in `ComponentSettingsModal`:

```
Warning: Failed prop type: The prop `errorMessage` is marked as required in `FormFieldMedia`, but its value is `undefined`.
```

I'm probably missing something - what is the rationale for marking `errorMessage` as an `isRequired` prop?

4. Since the concept of unlinked pages exist only for pages and not for resources, we should disable save if a category is not selected when creating a new resource.

alexanderleegs:
Have made the changes, along with a few extra changes, most notably the introduction of an editable option for the dropdown to allow new categories to be created.

> 1. When there are no resource categories, we still attempt to set state by accessing a `dirname` attribute.
> 
> ```
>  category: category ? category : allCategories[0].dirName,
> ```
> 
> This fails when `allCategories` is an empty array, as it should be when someone is creating a resource room for the first time.

Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/8aa5e66bf8673f00f731d164f44c0d23502b1215 

> 1. Again, the context is when a resource room has just been created, but there are no categories. When you first attempt to create a resource, there is no permalink default value. But if you switch post type to `Downloadable file` and then back to `Post content`, the default value of `permalink` is restored.

This problem was actually caused due to the above error, so fixing that also sets the fields to their proper values when initialised. Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/8aa5e66bf8673f00f731d164f44c0d23502b1215

> Furthermore, if there is an error in the permalink field (for example, if you only type 2 characters, so it's below the min. char limit and displays and error message), then even if you switch from `Post content` to `Downloadable file`, the error message persists. Error messages should be cleared when the form field value is cleared - if an erroneous value for permalink is reinstated, the error message should only be reintroduced then.

Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/9cd2d7b6f3eda6879a8597647a2a1fb1050c6f18

> 1. I get the following warning message when modifying form fields in `ComponentSettingsModal`:
> 
> ```
> Warning: Failed prop type: The prop `errorMessage` is marked as required in `FormFieldMedia`, but its value is `undefined`.
> ```
> 
> I'm probably missing something - what is the rationale for marking `errorMessage` as an `isRequired` prop?

Accidentally added in the requirement, has been removed in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/b4251adb5b180e7d5633fdeac9f7eb304ba2ae7a

> 1. Since the concept of unlinked pages exist only for pages and not for resources, we should disable save if a category is not selected when creating a new resource.

Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/8aa5e66bf8673f00f731d164f44c0d23502b1215

src/components/ComponentSettingsModal.jsx
```diff
@@ -281,21 +307,26 @@ export default class ComponentSettingsModal extends Component {
...
-                <div className="d-flex">
-                  <select className="w-100" id="category" value={category} onChange={this.changeHandler} disabled={isCategoryDropdownDisabled(isNewFile, prevCategory)} >
-                  {
-                    allCategories
+                <div className="d-flex text-nowrap">
+                  <CreatableSelect
+                    isClearable
+                    className="w-100"
+                    onChange={this.selectChangeHandler}
+                    isDisabled={isCategoryDropdownDisabled(isNewFile, prevCategory)}
```
kwajiehao, line 315: Currently, since the `isDisabled` prop is dependent on `prevCategory`, which is a calculated state variable, when we're in a collection, the dropdown is actually temporarily active before being disabled. This results in bad UX if a user eagerly clicks on it and manages to type something before it gets disabled.

Since the `prevCategory` is actually the original `category` prop, what we can do instead is to pass an `isCatgeoryDropdownDisabled` prop from `CollectionPagesSection` to `ComponentSettingsModal`, which takes the value of `isCategoryDropdownDisabled(createNewPage, collectionName)` (moving the `isCategoryDropdownDisabled helper function to the `CollectionPagesSection` component). This way the dropdown will be disabled from the start when we're in a category.

We should also set the default value based on the `category` prop instead of the `category` state, because `category` state represents the new category values, which results in `"Select a category or create a new category..."` being displayed inside the `CategoryPage` layout.

kwajiehao (changes requested):
left comments, we can also reconsider renaming `prevCategory` to `originalCategory` to be in line with the other state variables `originalPermalink` and `originalFileUrl`

src/components/ComponentSettingsModal.jsx
kwajiehao: This might be a bit confusing, so lmk if you need clarification

src/utils/validators.js
```diff
@@ -551,6 +557,14 @@ const validateResourceSettings = (id, value) => {
       }
       break;
     }
+    case 'category': {
+      errorMessage = validateCategoryName(value, 'resource')
+      if (value === '') {
+        errorMessage = `The resource category cannot be empty.`;
+      }
+      console.log(errorMessage)
```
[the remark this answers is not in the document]
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/82a982cfbaee61fea61a3eea518d0e551c7e90d6

src/components/ComponentSettingsModal.jsx
```diff
@@ -240,9 +243,32 @@ export default class ComponentSettingsModal extends Component {
         [id]: value,
       });
     }
-    
   }
 
+  selectChangeHandler = (newValue) => {
```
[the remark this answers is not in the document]
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/e088f82a1889372a3a5f74ec526b78a4c9b1ca0c

alexanderleegs:
> left comments, we can also reconsider renaming `prevCategory` to `originalCategory` to be in line with the other state variables `originalPermalink` and `originalFileUrl`

Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/1a7c6d5347fc9755b242a6ab2f70d2d6047ec6c0

src/components/ComponentSettingsModal.jsx
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/207/commits/b24c6bca57ad078e586e66c5c79013b858a5071f

kwajiehao (approved):
lgtm!

Issue #194, opened 2020-11-03
Modal for creating hyperlink on pages


lisatjide:
Current state:
When user click on the hyperlink icon when editing their page, a markdown formatting will appear that looks like this.
![Screenshot 2020-11-03 at 12 34 26 PM](https://user-images.githubusercontent.com/23736580/97950145-ebb76580-1dd0-11eb-81e4-66fbf4d3f238.png)

Ideal state:
Show a modal to allow users to insert the word to be hyperlinked and the actual link. It need not be over the actual word. e.g below (ref from google docs)
![Screenshot 2020-11-03 at 12 35 50 PM](https://user-images.githubusercontent.com/23736580/97950222-24efd580-1dd1-11eb-98fe-e7362b84a242.png)

Other notes:
This UI should also be implemented when inserting image. Especially to prompt users to insert alt text (see below)
![Screenshot 2020-11-03 at 12 39 30 PM](https://user-images.githubusercontent.com/23736580/97950382-a5163b00-1dd1-11eb-8bec-ab8d4ed962ed.png)

gweiying:
@NatMaeTan @sarahsalim 

1) I've done up a quick hyperlink modal that pops up when users press the hyperlink button, shown below. Unlike Google, the modal isn't inline, but centred in the middle of the page, and greys out everything else when it pops up. Would something like this be alright? 
<img width="1792" alt="Screenshot 2020-11-09 at 3 43 34 PM" src="https://user-images.githubusercontent.com/39231249/98555529-5588b100-22a2-11eb-8f7f-9bae7f35f401.png">

2) Could you weigh in on how a similar UI for prompting users to add alt-text could look like? Currently, users are prompted by this image selection modal when they press button for images. Would it be intuitive if I just added the alt-text field below in the same modal?
<img width="1792" alt="Screenshot 2020-11-09 at 3 47 56 PM" src="https://user-images.githubusercontent.com/39231249/98556074-f0818b00-22a2-11eb-91fa-9eda266e2109.png"> 

Thanks!

NatMaeTan:
This is alright for now!

kwajiehao:
Fixed with #209

PR #209, opened 2020-11-09
feat: adds hyperlink modal to prompt users for input text and link

gweiying:
This PR addresses issue #194, creating a popup modal when users click on the link button for the users to populate. The popup modal is previewed as below, and will take selected text (if any) as input to the text form field, otherwise the form field has a default value of "".

Screenshot 2020-11-09 at 3 43 34 PM

The look of the hyperlink modal above has been approved by the designers, but we are still discussing whether something similar can be applied to inserting alt-text for images. That change can be applied in a future PR as this seems like it would take more significant research.

### Future considerations
Depending on the discussion for image alt-text, it might be worth considering if we should create a more general popup modal to subclass.

### DIFF
--- a/src/components/HyperlinkModal.jsx
+++ b/src/components/HyperlinkModal.jsx
@@ -10,25 +10,25 @@ axios.defaults.withCredentials = true
 
 export default class HyperlinkModal extends Component {
   constructor(props) {
-		super(props);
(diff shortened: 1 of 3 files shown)


alexanderleegs (approved):
lgtm

PR #213, opened 2020-11-10
chore: increase hero subtitle max length to 160 chars

kwajiehao:
This PR increases hero subtitle max length to 160 chars in line with the recommended length for the `og:description` meta tag, which is what the hero subtitle is used for.

source: https://moz.com/learn/seo/meta-description#:~:text=Optimal%20length,descriptions%20between%2050%E2%80%93160%20characters.


lisatjide (approved):
lgtm

PR #225, opened 2020-11-12
fix: rename prevCategory to originalCategory

kwajiehao:
## Overview 
In a previous change, the `prevCategory` variable was renamed to `originalCategory`. This variable is passed down as a function argument to the `saveAndRetrieveUrl` function, but we forgot to change the reference to `prevCategory` from within the function, resulting in failed API calls.

This commit fixes that typo by renaming all instances of `prevCategory` to `originalCategory`.


lisatjide (approved):
lgtm

PR #230, opened 2020-11-15
Fix/edit homepage ui fixes

kwajiehao:
## Overview

This PR makes miscellaneous UI fixes to the `EditHomepage` layout as requested by the designers:
- If there is a button name, but button URL is empty, we should raise an error message
- Implement same behavior on EditPage (warning before exit) on EditHomepage
- Make the click area for the EditHomepage sections larger (just a bit more to the left of the chevron icon)

### DIFF
--- a/src/layouts/EditHomepage.jsx
+++ b/src/layouts/EditHomepage.jsx
@@ -247,14 +247,14 @@ export default class EditHomepage extends Component {
           // This needs to be done separately because it relies on the state of another field
           if (
             field === 'url' && !value && this.state.frontMatter.sections[sectionIndex][sectionType].button
-            && !(!this.state.frontMatter.sections[sectionIndex][sectionType].button && !this.state.frontMatter.sections[sectionIndex][sectionType].url)
+            && (this.state.frontMatter.sections[sectionIndex][sectionType].button || this.state.frontMatter.sections[sectionIndex][sectionType].url)
           ) {
             const errorMessage = 'Please specify a URL for your button'
             newSectionError = _.cloneDeep(errors.sections[sectionIndex])
             newSectionError[sectionType][field] = errorMessage
           } else if (
             field === 'button' && !this.state.frontMatter.sections[sectionIndex][sectionType].url
-            && !(!this.state.frontMatter.sections[sectionIndex][sectionType].button && !this.state.frontMatter.sections[sectionIndex][sectionType].url)
+            && (this.state.frontMatter.sections[sectionIndex][sectionType].button || this.state.frontMatter.sections[sectionIndex][sectionType].url)
           ) {
             const errorMessage = 'Please specify a URL for your button'
             newSectionError = _.cloneDeep(errors.sections[sectionIndex])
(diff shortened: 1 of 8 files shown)


src/layouts/EditHomepage.jsx
```diff
@@ -241,6 +242,34 @@ export default class EditHomepage extends Component {
 
           sections[sectionIndex][sectionType][field] = value;
 
+          // Set special error message if hero button has text but hero url is empty
+          // This needs to be done separately because it relies on the state of another field
+          if (sectionIndex === 0 && sectionType === 'hero' && field === 'url' && !value && this.state.frontMatter.sections[0].hero.button) {
```
alexanderleegs, line 247: We could consider also adding this check for when the button is changed - currently users are still allowed to save if the url is empty but there is button text (happens most frequently if the hero section is changed from a dropdown to highlights + buttons)

alexanderleegs (changes requested):
We should also implement the checks for url for the infobar and infopic sections, which also have buttons.

src/layouts/EditHomepage.jsx
kwajiehao: Good point. Fixed with [904044a](https://github.com/isomerpages/isomercms-frontend/pull/230/commits/904044a605acf85a6109f481e3fd34b0ed465399) and extended to infobar and infopic sections in [9910ee4](https://github.com/isomerpages/isomercms-frontend/pull/230/commits/9910ee4548157caf6377783fbe5d80feefb60626)

src/layouts/EditHomepage.jsx
```diff
@@ -238,13 +239,39 @@ export default class EditHomepage extends Component {
...
+          if (
+            field === 'url' && !value && this.state.frontMatter.sections[sectionIndex][sectionType].button
+            && !(!this.state.frontMatter.sections[sectionIndex][sectionType].button && !this.state.frontMatter.sections[sectionIndex][sectionType].url)
+          ) {
+            const errorMessage = 'Please specify a URL for your button'
+            newSectionError = _.cloneDeep(errors.sections[sectionIndex])
+            newSectionError[sectionType][field] = errorMessage
+          } else if (
+            field === 'button' && !this.state.frontMatter.sections[sectionIndex][sectionType].url
+            && !(!this.state.frontMatter.sections[sectionIndex][sectionType].button && !this.state.frontMatter.sections[sectionIndex][sectionType].url)
```
alexanderleegs, line 257: Nit: the double negatives here are a bit confusing, could we change them to (this.state.frontMatter.sections[sectionIndex][sectionType].button || this.state.frontMatter.sections[sectionIndex][sectionType].url) instead?

alexanderleegs (approved):
Minor nit, but otherwise lgtm

src/layouts/EditHomepage.jsx
kwajiehao: fixed in [267650b](https://github.com/isomerpages/isomercms-frontend/pull/230/commits/267650b7fac8e2dbed211cd544dfbd3b76974a3a)

PR #231, opened 2020-11-18
Fix: prevent re-render of third nav component

alexanderleegs:
This PR fixes a bug where editing pages with third navs was extremely laggy. This was due to the way we generated the third nav data - editing anything in the page caused a rerender of the third nav as well, which was computationally expensive. This PR fixes this by generating the the third nav data once upon loading only, to prevent unnecessary rerendering.


kwajiehao (approved):
lgtm

Issue #233, opened 2020-11-18
Misc. UI fixes for `EditPage`


kwajiehao:
1. Currently, we can type in the markdown editor while it's loading the page content (including left-nav). However, 
once the content is loaded, it replaces whatever content we have typed, which is a bad user experience. We should disable the markdown editor until content has been retrieved.

2. We can also consider adding a loading spinner on the page preview until content is loaded

PR #240, opened 2020-11-18
feat: mask markdown editor until page content has loaded

kwajiehao:
This PR addresses issue #233 

## Overview
Currently, when users enter the EditPage layout, they can enter text into the EasyMDE editor right away. However, whatever content they enter will be replaced when the page content finally loads. This is a bad UI experience and this commit seeks to change that by blocking access to the editor until all page content has loaded.

There is no native way of masking the EasyMDE editor that we're using (refer to the issue on their repo https://github.com/Ionaru/easy-markdown-editor/issues/49) so we need to manually apply a mask to prevent the user from editing content until the page content is loaded.

**Loading**
Screenshot 2020-11-19 at 1 12 26 AM

**After load**
Screenshot 2020-11-19 at 1 12 20 AM

### Follow-up 

Do we need to add some kind of feedback on the page preview side? Or is the loading spinner on the editor sufficient?

### DIFF
--- a/src/layouts/EditPage.jsx
+++ b/src/layouts/EditPage.jsx
@@ -107,6 +107,7 @@ export default class EditPage extends Component {
       selectionText: '',
       isFileStagedForUpload: false,
       stagedFileDetails: {},
+      isLoadingPageContent: true,
     };
     this.mdeRef = React.createRef();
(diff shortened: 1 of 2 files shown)


alexanderleegs (approved):
lgtm

Issue #235, opened 2020-11-18
Misc. issues for resource room creation


kwajiehao:
When creating a resource room
- Validate name to include only alphanumeric characters and `-` (dash)
- Convert to lowercase on submission

When creating a resource room, we also want to switch the submission button to our `LoadingButton` component so that a loading spinner is displayed when the request is made.

PR #241, opened 2020-11-18
Fix/create resource room

kwajiehao:
## Overview

This PR resolves issue #235 by adding:
- validation for resource room name 
- replacing the button with the `LoadingButton` component

**_Resource room name validation_**
Screenshot 2020-11-19 at 2 34 10 AM

**_Replacement of button with `LoadingButton`_**
Screenshot 2020-11-19 at 2 59 30 AM

### DIFF
--- a/src/utils/validators.js
+++ b/src/utils/validators.js
@@ -11,7 +11,7 @@ const fileNameRegexTest = /^[a-zA-Z0-9" "_-]+$/;
 const fileNameExtensionRegexTest = /^[a-zA-z]{3,4}$/;
 const RESOURCE_CATEGORY_REGEX = '^(([a-zA-Z0-9_-]+([\\s][a-zA-Z0-9_-]+)*)+)$';
 const resourceCategoryRegexTest = RegExp(RESOURCE_CATEGORY_REGEX);
-const resourceRoomNameRegexTest = /^[a-zA-Z-]+$/;
+const resourceRoomNameRegexTest = /([a-zA-Z]+([-][a-zA-Z]+)?)+$/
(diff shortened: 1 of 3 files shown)


src/utils/validators.js
```diff
@@ -11,6 +11,7 @@ const fileNameRegexTest = /^[a-zA-Z0-9" "_-]+$/;
 const fileNameExtensionRegexTest = /^[a-zA-z]{3,4}$/;
 const RESOURCE_CATEGORY_REGEX = '^(([a-zA-Z0-9_-]+([\\s][a-zA-Z0-9_-]+)*)+)$';
 const resourceCategoryRegexTest = RegExp(RESOURCE_CATEGORY_REGEX);
+const resourceRoomNameRegexTest = /([a-zA-Z]+([-][a-zA-Z]+)?)+$/
```
alexanderleegs, line 14: This regex doesn't quite work - it's not enforcing the start of the string, and also not allowing numerals, which conflicts with the error message.
I think we could use `^([a-zA-Z0-9]+-)*[a-zA-Z0-9]+$`, which is what I'm using in another PR to check for category/collection names

alexanderleegs (changes requested):
Left a comment

src/utils/validators.js
kwajiehao: fixed with [75f8df8](https://github.com/isomerpages/isomercms-frontend/pull/241/commits/75f8df811ed43f7e62a53ac7d7e1886fe132b4f2)

alexanderleegs (approved):
lgtm

Issue #228, opened 2020-11-13
Memory leak bug because of unmounted Loading button


gweiying:
Because our modals unmount before the Loading button completes its state update, we receive the memory leak error message shown below:

<img width="1792" alt="Screenshot 2020-11-13 at 10 25 10 AM" src="https://user-images.githubusercontent.com/39231249/99056373-831f7400-259a-11eb-9b23-6f22d078f21d.png">

This happens for all modals (mediaModal, hyperlinkModal etc) that close once the Loading button is clicked because of the following lines in the `LoadingButton.jsx` component [link](https://github.com/isomerpages/isomercms-frontend/blob/467adeb593dd5fcbc6c4a1cb609c38f03015c78f/src/components/LoadingButton.jsx#L22):

```
      onClick={async () => {
        setButtonLoading(true);
        await callback();
        setButtonLoading(false);
      }}
```

kwajiehao:
So it turns out that the issue isn't to do with the lines specified:

```
      onClick={async () => {
        setButtonLoading(true);
        await callback();
        setButtonLoading(false);
      }}
```

but from the fact that we are attempting to set the button loading state to `false` even after the `LoadingButton` component has unmounted (because the modal has closed). The fix is to add a condition to update state only when the component is mounted (reference: https://stackoverflow.com/questions/53949393/cant-perform-a-react-state-update-on-an-unmounted-component)

PR #242, opened 2020-11-18
Fix: memory leaks in application

kwajiehao:
## Overview

This PR fixes instances in the app where we attempt to perform state updates on unmounted components. It resolves issue #228.

Occasionally, when we navigate away from a layout on the CMS, or close a modal before it's done loading something, we receive the following error message in the console:

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function
```

The problem arises when we are awaiting some async operation, and setting state after the async operation completes. For example, I might be in the `EditPage` layout, and after I retrieve the page's content, I set that content into the component state.

The problem is that if I navigate away from the `EditPage` layout while it is retrieving the content (the async operation), the `EditPage` component unmounts, but the `componentWillMount` function, or the `useEffect` function will continue to run and attempt to set state after the async operation completes. This is why we get the error message about updating state on an unmounted component.

### Solution

The solution is to only set state when we can be sure that the component is mounted. We can do this by introducing an `_isMounted` local variable, which is set to `true` when the async operation is run, but also define a cleanup function (either within the `useEffect` or as `componentWillUnmount`) which sets the `_isMounted` variable to `false`. Since state updates are only performed when `_isMounted` is `true`, we no longer attempt to update state after a component is unmounted. The caveat is that `_isMounted` needs to be defined as a normal variable, and not as state, since it needs to be updated synchronously for this to work.


alexanderleegs (approved):
lgtm

Issue #232, opened 2020-11-18
Misc. `EditHomepage` UI issues


kwajiehao:
1. Delete warning modal
- Applies to “Delete highlights”, “Delete resources”, etc.
- Buttons should be shifted to the bottom right of the modal, instead of the bottom left

<img width="1124" alt="Screenshot 2020-11-18 at 5 06 13 PM" src="https://user-images.githubusercontent.com/31984694/99509156-60d58e00-29c0-11eb-8258-863b2c1fb983.png">

2. Currently, hero dropdown options are labelled “Dropdown elem <number>”. The word “elem” isn’t easily understandable - we should change it to “option”, so that it reads “Dropdown option <number>”

<img width="344" alt="Screenshot 2020-11-18 at 5 06 41 PM" src="https://user-images.githubusercontent.com/31984694/99509221-75198b00-29c0-11eb-8165-51603fae82dc.png">

3. Reorder form fields, since the subtitle is above the title, but the title input field is above the subtitle input field. This applies for Resources section, as well as infobar and infopic

<img width="1195" alt="Screenshot 2020-11-18 at 5 07 12 PM" src="https://user-images.githubusercontent.com/31984694/99509312-95e1e080-29c0-11eb-9853-d7eafba30c55.png">

PR #219, opened 2020-11-12
Feat/add page moving functionality

alexanderleegs:
This PR enhances the existing OverviewCard component to add in a dropdown menu and a page moving modal, in order to resolve https://github.com/isomerpages/isomercms-frontend/issues/203. This PR also resolves a bug involving the creation of new collections from the settings modal.

One change to the existing UX of the OverviewCard is that the transform on hover has been removed: this was due to an issue with transform not working well with z-index values, causing certain elements to overlap the dropdown menu. We could investigate a way to reimplement the hover effect in future.

The main complication when getting the dropdown menu to work was the behaviour to determine when the menu should be closed. Ideally, we would want to be able to force only one dropdown to be open at a time, and clicking outside of the dropdown should automatically close the modal. To implement this, we use the focus of elements to determine the behaviour of the menu. When the dropdown is expanded, we artificially force focus onto the menu element, and trigger the closing of the menu when focus is lost. However, this caused some issues with text boxes inside the menu, as the focus of the input element was not the same as the menu. We eventually got around this problem with the solution suggested here: https://stackoverflow.com/a/60094794

One issue we should address is the UX involved when transferring a file from one category to another - currently there is no indication that the process is taking place. We should implement something and disable the remaining buttons while this is taking place.

**Items to test for this feature:**
Workspace
- Move unlinked page into existing collection
- Move unlinked page into new collection

Collection folder
- Move page with no 3rd nav into unlinked page
- Move page with no 3rd nav into different collection
- Move page with no 3rd nav into new collection
- Move 3rd nav page into unlinked page
- Move 3rd nav page into different collection
- Move 3rd nav page into new collection

Resources folder
- Move resource into different category

PR #246, opened 2020-11-20
Fix/misc homepage ui issues and minor general bug fixes

alexanderleegs:
This PR fixes some misc homepage ui issues, to resolve https://github.com/isomerpages/isomercms-frontend/issues/232.

**Bug fixes**
---
- This PR also fixes a bug involving the saving of files in PR https://github.com/isomerpages/isomercms-frontend/pull/219 - I accidentally changed one of the arguments when generating collection file names, which broke the flow for collection pages. 
- It also fixes an existing bug where renaming 3rd nav pages was not being handled.
- Also, it fixes a bug where clicking on the text to open the ComponentSettingsModal in OverviewCard would not work.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -224,6 +224,7 @@ export async function saveFileAndRetrieveUrl(fileInfo) {
     // Creating a collection page
     if (category) {
       newFileName = await generateNewCollectionFileName({
+        fileName,
         originalThirdNavTitle,
         thirdNavTitle,
         thirdNavOptions,
@@ -291,6 +292,7 @@ export async function saveFileAndRetrieveUrl(fileInfo) {
 
 // Generate new filename when creating page in a collection
 const generateNewCollectionFileName = async ({
+  fileName,
   originalThirdNavTitle,
   thirdNavTitle,
   thirdNavOptions,
@@ -333,7 +335,8 @@ const generateNewCollectionFileName = async ({
     }
   } else if (originalThirdNavTitle === thirdNavTitle && collectionPageData) {
     // Case: renaming/changing details of existing file in collection
-    const groupIdentifier = await generateGroupIdentifier(collectionPageData, thirdNavTitle ? true : false, baseApiUrl, false, thirdNavTitle)
+    // We can resuse the existing identifier
+    const groupIdentifier = fileName.split('-')[0]
     newFileName = generateCollectionPageFileName(title, groupIdentifier);
   } else {
     // Case: Creating a new page from the workspace and assigning to collection BUT not third nav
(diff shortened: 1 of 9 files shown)


src/utils.js
```diff
@@ -333,6 +331,10 @@ const generateNewCollectionFileName = async ({
 
       newFileName = generateCollectionPageFileName(title, groupIdentifier);
     }
+  } else if (originalThirdNavTitle === thirdNavTitle && collectionPageData) {
+    // Case: renaming/changing details of existing file in collection
+    const groupIdentifier = await generateGroupIdentifier(collectionPageData, thirdNavTitle ? true : false, baseApiUrl, false, thirdNavTitle)
+    newFileName = generateCollectionPageFileName(title, groupIdentifier);
```
kwajiehao, line 337: This fails when trying to rename a collection third nav page to the same third nav section (for example, if I want to rename a page with title `abc` to `def` but to keep it within the same third nav section, `5a-abc.md` should become `5a-def.md`). Currently, it returns `undefined`, or it generates a non-third-nav group identifier by adding it as the last page to the collection (for example, `8-abc.md`)

kwajiehao: I added the following two cases in the excel sheet https://docs.google.com/spreadsheets/d/1-U69XpMmSHwtOXSYkWGc5gHpwrvA4i23cSQgOA4j5_s/edit#gid=2010487752:

- Rename an existing third nav page while keeping it in the same third nav section
- Rename an existing (normal) collection page while keeping it as a normal collection page

kwajiehao (changes requested):
Left comments

One more issue: currently, the templates for `infopic-left` and `infopic-right` are wrong. The subtitle should come above the title for the templates to remain faithful to the original Isomer templates. So what we need to do is:
- correct the templates to switch `title` and `subtitle` for `infopic-left` and `infopic-right`
- correct `EditHomepage` so that we switch `title` and `subtitle` fields for `infopic-left` and `infopic-right`

src/utils.js
alexanderleegs: Good catch, we should just reuse the existing group identifier in these cases. Resolved in https://github.com/isomerpages/isomercms-frontend/pull/246/commits/4c310d2136e2a1242c4df3e398e65a2c8a98cbba

alexanderleegs:
> One more issue: currently, the templates for `infopic-left` and `infopic-right` are wrong. The subtitle should come above the title for the templates to remain faithful to the original Isomer templates. So what we need to do is:
> 
> * correct the templates to switch `title` and `subtitle` for `infopic-left` and `infopic-right`
> * correct `EditHomepage` so that we switch `title` and `subtitle` fields for `infopic-left` and `infopic-right`

Resolved in https://github.com/isomerpages/isomercms-frontend/pull/246/commits/8fdbf1017f7aa235297fce5563f3d31681cb7f0b

kwajiehao (approved):
lgtm

Issue #234, opened 2020-11-18
Misc. issues for `ComponentSettingsModal`


kwajiehao:
### ~1. Sanitize Title field input to prevent HTML input~
Canceled since React sanitizes input for us

### 2. Validation of category folder name
- Only allow alphanumeric characters with `-` (dash)
- Convert to lowercase on submission

### 3. Validation of third nav section name
- Only allow alphanumeric characters and space
- Convert to lowercase on submission

### ~4. Permalink~ Superceded by #255
~- Make category folder name lowercase~
~- Slugify third nav section name and make lowercase~

### 5. When creating a new collection, also update `navigation.yml` to include the new collection
- This is an update to the backend endpoint
- Deslugify the collection name to come up with the collection title when writing to `navigation.yml`

alexanderleegs:
Point 4 (permalink) is superseded by https://github.com/isomerpages/isomercms-frontend/issues/255, given that we no longer want to enforce permalink prefixes

Issue #255, opened 2020-11-23
Do not overwrite user permalinks


kwajiehao:
## Overview

In the current implementation of the CMS, the CMS imposes structure on the users' page permalinks by fixing subsections of the permalink to correspond with the collection name or resource category name. For example, if the page belongs in a collection called `alphabet`, but the current permalink is `/abc/def`, the rendered permalink on the `ComponentSettingsModal` is `/alphabet/def` instead of `/abc/def`. If they decide to make any changes to the page settings using the modal, we will overwrite the previous permalink so that it is now `/alphabet/def` even if they did not actually want to change their permalink. 

However, in a team discussion, we agreed that it was not right to impose this structure for several reasons:
- We are simply displaying the wrong permalink
- Users might want to have a particular permalink for specific reasons, and they would not be able to do this
- We are changing user permalinks without their permission. This could result in broken links and they would have no idea

**Verdict**: Overall, the team was in favor of handing responsibility for permalink naming back to the user.

## Moving pages from one collection or resource category to another

When moving pages from one collection or resource category to another, there are issues associated both with automatically changing user permalinks, and with preserving the original permalink.

Issue with automatically updating user permalink to reflect new collection: 
- For example, if I'm moving a page from collection `abc` to collection `def`, and the original permalink is `/abc/title`, then we automatically update a page's permalink to be `/def/title` when we move it to collection `def`
- However, this leads to broken links since the original `/abc/title` link no longer links to an existing page 

Issue with keeping original permalinks:
- For example, if I'm moving a page from collection `abc` to collection `def`, and the original permalink is `/abc/title`, we keep the page's permalink to be `/abc/title`
- The problem with this is that when website users see the page, they will see that they are in collection `def` but the url contains `abc`, which is a cognitive disconnect

**Verdict**: Since the current version of Isomer places the responsibility of maintaining permalinks with the user, the team agreed that we should, for now, leave the responsibility of maintaining permalinks with the user. This means that we will keep the original permalinks, and not automatically update it. While this might lead to bad UX, this is a problem that's not unique to Isomer: any time someone decides to changes a page's URL, they run the risk of breaking links. By leaving the choice in the hands of the user, they can take responsibility for their decision.

## Proposed solution

Permalinks should be displayed as-is if a page already has an existing permalink.
- When creating a new collection page, we can pre-fill the permalink text input with the collection name. This provides a suggestion to the user, but they would still be able to modify the permalink if they so wished.

[...]

PR #247, opened 2020-11-23
Feat/add error handling for same file name

alexanderleegs:
This PR adds error handling when moving and renaming files, in order to resolve https://github.com/isomerpages/isomercms-frontend/issues/237. To be reviewed in conjunction with ~~PR [#78](https://github.com/isomerpages/isomercms-backend/pull/78)~~ (_already merged_) and PR [#84](https://github.com/isomerpages/isomercms-backend/pull/84) on the isomercms-backend repo. It also adds a new component for toast notifications. In addition, it introduces styling changes to the position of dropdown modals present in OverviewCard and FolderCard, to resolve https://github.com/isomerpages/isomercms-frontend/issues/261 and resolve https://github.com/isomerpages/isomercms-frontend/issues/243.

This PR also changes the util function involved when saving files to attempt to create a new file first before deleting the existing one. Originally, users could experience the loss of file data in the cases where a file name conflict exists for the new file, but the old file had already been deleted. Now, an appropriate error message is shown, and the old file is preserved.

As the OverviewCard lacks an error message field for each of the folder choices when moving files, a Toast notification has been created to appear when a file name conflict occurs.

Screenshot 2020-11-27 at 10 55 46 AM

PR #256, opened 2020-11-24
Fix/misc component settings issues

alexanderleegs:
This PR addresses several issues with the current component settings modal, and is meant to resolve https://github.com/isomerpages/isomercms-frontend/issues/234 and partially address https://github.com/isomerpages/isomercms-frontend/issues/255. To be reviewed with PR [#80](https://github.com/isomerpages/isomercms-backend/pull/80) on the isomercms-backend repo. 

The main change is the reverting of automatically generated permalinks for users - we have decided that users should manage their own hyperlinks. This PR also enforces the conversion of category names to lowercase.

### DIFF
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -14,7 +14,6 @@ import {
   generatePageFileName,
   generateCollectionPageFileName,
   generateResourceFileName,
-  retrieveCollectionAndLinkFromPermalink,
   saveFileAndRetrieveUrl,
 } from '../utils';
 import elementStyles from '../styles/isomer-cms/Elements.module.scss';
@@ -153,7 +152,7 @@ const ComponentSettingsModal = ({
               // Set default values for new file
               if (_isMounted) {
                 setTitle('Title')
-                setPermalink('permalink')
+                setPermalink(`${originalCategory ? `/${originalCategory}` : ''}/permalink`)
                 if (type === 'resource') setResourceDate(new Date().toISOString().split("T")[0])
               }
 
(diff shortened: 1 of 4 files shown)


kwajiehao (changes requested):
The changes made here look good but there are a couple of missing things:
1. When using the page-moving modal, a confirmation modal should pop up asking the user to confirm their selection, and also providing some warning/info text (refer to the **Proposed solution** section in #255)
2. When moving a non-third-nav page from one collection to another, the group identifier is not correctly generated; the page keeps its group identifier.

For example: if I'm trying to move `1-title.md` from collection **abc** to collection **def**, which has the following files:
- `1-first-page.md`
- `2-second-page.md`
- `3-third-page.md`

Then I would expect the newly-moved file to be renamed as `4-title.md`. Instead, it retains its group identifier and the file name is still `1-title.md` even in the new collection. We don't have this problem when moving a third-nav page using the page-moving modal to move it to a new collection

alexanderleegs:
> 1. When using the page-moving modal, a confirmation modal should pop up asking the user to confirm their selection, and also providing some warning/info text (refer to the **Proposed solution** section in #255)

I think this one might be better handled by https://github.com/isomerpages/isomercms-frontend/pull/247, since I'm currently making changes to the OverviewCard in that PR - i'll update this one so that it doesn't automatically close the issue

> 2. When moving a non-third-nav page from one collection to another, the group identifier is not correctly generated; the page keeps its group identifier.

Good catch, resolved in https://github.com/isomerpages/isomercms-frontend/pull/256/commits/e368947810758acfca15ea85da23438f16ffa98f

kwajiehao (approved):
lgtm

PR #266, opened 2020-11-26
Feat/contact us page 2

gweiying:


src/utils/validators.js
```diff
@@ -76,12 +76,12 @@ const HERO_DROPDOWN_MAX_LENGTH = 30;
...
 // ===============
 // Contacts
-const CONTACT_TITLE_MIN_LENGTH = 2;
+const CONTACT_TITLE_MIN_LENGTH = 0;
 const CONTACT_TITLE_MAX_LENGTH = 30;
 const CONTACT_DESCRIPTION_MAX_LENGTH = 400;
 
 // Locations
-const LOCATION_TITLE_MIN_LENGTH = 2;
+const LOCATION_TITLE_MIN_LENGTH = 0;
```
alexanderleegs, line 84: Nit: I feel like these should be min_length 1 and the checks left as `value.length < CONTACT_TITLE_MIN_LENGTH` so that the wording is clearer

gweiying: good idea eb4b82c

kwajiehao (approved):
Lgtm

Couple of nits as discussed offline:
1. Font size of CMS sidebar changed 

_After_



2. Reduce the spacing in a multi-line title for each subsection in the EditContactUs layout
Screenshot 2020-12-03 at 9 57 55 PM

Issue #267, opened 2020-11-27
EditHomepage doesn't work when page has no sections


alexanderleegs:
Steps to reproduce:
- Delete all sections in the homepage
- Save page
- Upon reloading the page, user cannot open hero section to edit or add new sections to homepage

Upon further investigation, this may be an issue because we don't initialise errors properly if the homepage is lacking both dropdown and keyHighlights - will look into this further

PR #268, opened 2020-11-27
Fix: proper intialization of section errors

alexanderleegs:
This PR fixes the bug raised in issue https://github.com/isomerpages/isomercms-frontend/issues/267. The reason for this issue was our assumption that a page had either a dropdown or highlights - given the relaxation of compulsory fields in the hero section, it became possible for a page to have neither, in which case our error initialization was not performed correctly. This PR fixes that issue by checking for this case.


kwajiehao (approved):
lgtm

Issue #272, opened 2020-11-30
Loading collections on workspace does not disappear if there are no collections


alexanderleegs:
If a repo has no collections, on the workspace, the `Loading collections...` text stays because we currently check for the length of collections. We should differentiate between the case where we are still retrieving collections and the case where there are none.

Issue #254, opened 2020-11-23
Display resource page type


kwajiehao:
## Overview

Currently, we are unable to distinguish between resource pages which are of the `downloadable file` type (a link to a file) and those which are of the `post` type (contains text content). This means that users are able to click into these resource pages where they will be presented with the `EditPage` layout and they can start adding content to the post. This is misleading because this page content will not be displayed anywhere, thus this should not be allowed.

## Proposed solution

The proposed solution to this problem is to include resource page type as part of the resource page file name. Currently, the resource page file names are of the format `<YYYY-MM-DD>-<title>.md`. By changing this format to `<YYYY-MM-DD>-<type>-<title>.md`, we would be easily able to retrieve the file type from the file name and thus display it on the Resource category layout without any additional API calls.

What we need to do to achieve this:
- Migrate existing repositories' resource pages to the new file name format using a script
- Modify the CMS code to read from and write to the new file name format `<YYYY-MM-DD>-<type>-<title>.md`
- Resource cards should display the type of resource, either FILE (downloadable file) or POST (post). The card label will be modified to include file type (from `<date>` to `<date> / <type>`). 
- At the resource category layout, we will disable clicking into resource pages of type FILE.
- Modify `EditPage` layout to disable input if there is a `file_url` attribute in the front matter, since users can still attempt to access the link through the URL

## Operational to-dos

- We need to coordinate with agencies when migrating their content since we need to ensure that they do not touch their repos or the CMS during migration
- We will also need to manually inspect their content to ensure that the migration went well - in the event of a bad migration, we should roll back to the last working commit 
- Update the V1 to V2 migration instructions to include instructions for migrating resource page file names

kwajiehao:
@alexanderleegs to begin work on the migration script itself

PR #274, opened 2020-12-01
Feat/implement resource type handling from file name

alexanderleegs:
This PR re-implements resource file naming and identification of resource type through the file name, to resolve https://github.com/isomerpages/isomercms-frontend/issues/254. The major changes are as follows:
- The util functions for reading and writing resource file names have been changed to the new file name format `<YYYY-MM-DD>-<type>-<title>.md`
- Resource cards display the type of resource, either FILE (downloadable file) or POST (post). The card label has been modified to include file type.
- Resource pages of type FILE can no longer be clicked in CategoryPages.
- The EditPage layout disables input if the resource is of type FILE.

In addition, this PR also resolves https://github.com/isomerpages/isomercms-frontend/issues/272, by fixing a small bug involving collections.

### DIFF
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -182,7 +182,7 @@ const ComponentSettingsModal = ({
                 setFileUrl(frontMatter.file_url)
                 setOriginalFileUrl(frontMatter.file_url)
 
-                setResourceDate(type === 'resource' ? retrieveResourceFileMetadata(fileName).date : frontMatter.date)
+                setResourceDate(type === 'resource' ? retrieveResourceFileMetadata(fileName).date : '')
                 setOriginalThirdNavTitle(frontMatter.third_nav_title)
                 setThirdNavTitle(frontMatter.third_nav_title)
               }
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -16,6 +16,7 @@ import {
   generateCollectionPageFileName,
   generateResourceFileName,
   saveFileAndRetrieveUrl,
+  retrieveResourceFileMetadata,
 } from '../utils';
 import elementStyles from '../styles/isomer-cms/Elements.module.scss';
 import { validatePageSettings, validateResourceSettings } from '../utils/validators';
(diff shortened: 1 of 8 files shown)


kwajiehao:
Leaving this PR here until we're ready to accept resource page migrations for CMS users

src/components/ComponentSettingsModal.jsx
```diff
@@ -181,7 +182,7 @@ const ComponentSettingsModal = ({
                 setFileUrl(frontMatter.file_url)
                 setOriginalFileUrl(frontMatter.file_url)
 
-                setResourceDate(frontMatter.date)
+                setResourceDate(type === 'resource' ? retrieveResourceFileMetadata(fileName).date : frontMatter.date)
```
kwajiehao, line 185: Are there any cases other than resource pages where the `date` attribute in the front matter is used?

src/layouts/EditPage.jsx
```diff
@@ -367,55 +372,65 @@ export default class EditPage extends Component {
...
+                    ],
+                  }}
+                />
+              </div>
+            :
+              <div className={`${editorStyles.pageEditorSidebar}`}>
+                <div className={`text-center`}>
+                  Editing is disabled for downloadable files.
+                </div>
+              </div>
```
kwajiehao, line 432: What do you think about something similar to the loading styles but with lower opacity? My concern is that I still want to preserve the format of the `EditPage` and that removing the text editor is a different view that introduces a small bit of cognitive load. Let me know what you think

Screenshot 2020-12-16 at 3 21 26 PM

kwajiehao (commented):
left comments - let's merge this in on thursday 17 dec after we complete the migration

src/layouts/EditPage.jsx
alexanderleegs: Makes sense, made the change in https://github.com/isomerpages/isomercms-frontend/pull/274/commits/4964587a7b727403ea0aff55908f40658890fb38

src/components/ComponentSettingsModal.jsx
alexanderleegs: You're right, there aren't any others - I've removed this from other cases in https://github.com/isomerpages/isomercms-frontend/pull/274/commits/56280f359168d893548e6547c92be6f292e71566

kwajiehao (approved):
lgtm

Issue #271, opened 2020-11-30
Provide more site configuration options


kwajiehao:
In addition to issues #269 and #270, we need to provide the following site configuration options for users:
- whether the government masthead should be displayed (https://github.com/isomerpages/isomerpages-template/blob/next-gen/_includes/masthead.html - the `is_government` attribute in `_config.yml`)
- Google analytics
- Facebook pixel
- shareicon (the image displayed when sharing a link on fb, ig, whatsapp, etc.) and shareicon description (`og:meta` tags https://neilpatel.com/blog/open-graph-meta-tags/)

Issue #270, opened 2020-11-30
Provide option to set agency logo in `Settings`


kwajiehao:
There is a `logo` attribute that can be set in `_data/navigation.yml` (refer to the template file https://github.com/isomerpages/isomerpages-template/blob/next-gen/_includes/agencyhead.html) which determines the logo of the navbar. We should provide an option for the user to change the navigation bar logo in the `Settings` layout

Issue #269, opened 2020-11-30
Provide option to edit title in `Settings`


kwajiehao:
Previously, we had `Title` fields in the `EditHomepage` and `Settings` layout pages, which we removed in PR #197 as they were deemed confusing. However, new user requirements dictate that we should reinstate the `Title` field. To avoid confusion, there should only be one `Title` field which dictates the value of both the **browser title** as well as the **footer header** (we can retrieve the value from `_config.yml`, but we need to write to both `index.md` and `_config.yml`). 

We can create this `Title` field either in `Settings` or `EditHomepage`, but either way, we would need to update two separate files (`index.md` and `_config.yml`) to sync the titles selected by the user.

kwajiehao:
resolved with #275

PR #197, opened 2020-11-03
Fix/form field image modal and minor homepage and settings changes

alexanderleegs:
Changes
--
- This PR fixes the issue with the FormFieldImage when uploading images (issue https://github.com/isomerpages/isomercms-frontend/issues/196), which occurred due to the changes made to MediasModal made in https://github.com/isomerpages/isomercms-frontend/pull/169. FormFieldImage has been updated to follow the same unstacking behaviour of modals.
- The title fields have been removed from the homepage and settings pages, as it was determined that they should be set automatically for users in the generation script.
- A grey mask has been applied to the homepage background (issue https://github.com/isomerpages/isomercms-frontend/issues/195)
- Remove unnecessary parentElement field from FormFieldImage (issue https://github.com/isomerpages/isomercms-frontend/issues/177)
- Adjust zIndex of colour picker modal

PR #275, opened 2020-12-01
Feat/add settings fields

kwajiehao:
Note: This is to be reviewed with PR #[85](https://github.com/isomerpages/isomercms-backend/pull/85) in the backend.

## Overview
This PR adds new fields and logic to the `Settings` layout, resolving issues #269, #270, and #271 in the process.

Firstly, it adds the following fields to the `Settings` layout:
- `shareicon`
- `facebook_pixel`
- `google_analytics`
- `is_government` (whether government masthead should be displayed)
- `logo` (agency logo)

It also reintroduces the `title` field, which was previously removed in PR #197. The reason for reintroduction is explained in more detail in issue #269.

Note that these changes need to be accompanied by changes in the backend, since we need to retrieve and update the agency logo from a new file (`_data/navigation.yml`), and also because the reintroduced `title` field now represents both the browser tab title (found in `index.md`) as well as the footer title (found in `_data/footer.yml`).

## Style changes for dropdown fields

There are certain fields in `Settings`, such as `show_reach` or `is_government`, which should only accept `true` or `false` values. In such cases, we should limit the user's options instead of allowing them to enter plain text. This PR introduces a `FormFieldDropdown` component which does just that.

Screenshot 2020-12-02 at 11 58 35 AM

### DIFF
--- a/src/components/FormFieldDropdown.jsx
+++ b/src/components/FormFieldDropdown.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import CreatableSelect from 'react-select/creatable';
+import Select from 'react-select';
 import elementStyles from '../styles/isomer-cms/Elements.module.scss';
 
 const calculateBoolean = (booleanVal) => {
@@ -30,7 +30,7 @@ const FormFieldDropdown = ({
         <div className={`d-flex align-items-center ${elementStyles.formDropdownLabel}`}>
             <span>{`${title}:`}</span>
         </div>
-        <CreatableSelect
+        <Select
             className={elementStyles.formDropdownInput}
             onChange={(event) => creatableSelectHandler(onFieldChange, event, id)}
             value={{
@@ -54,13 +54,7 @@ export default FormFieldDropdown;
   
 FormFieldDropdown.propTypes = {
     title: PropTypes.string.isRequired,
-    defaultValue: PropTypes.string,
     value: PropTypes.string.isRequired,
     id: PropTypes.string.isRequired,
     onFieldChange: PropTypes.func.isRequired,
-};
-  
-FormFieldDropdown.defaultProps = {
-    defaultValue: 'false',
-    style: undefined,
 };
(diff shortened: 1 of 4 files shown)


src/components/FormFieldDropdown.jsx
```diff
@@ -0,0 +1,74 @@
...
+    isRequired,
+    style,
+}) => (
+    <>
+      <div className={elementStyles.formDropdown}>
+        <div className={`d-flex align-items-center ${elementStyles.formDropdownLabel}`}>
+            <span>{`${title}:`}</span>
+        </div>
+        <CreatableSelect
+            isClearable
```
alexanderleegs, line 38: We should remove the `isClearable` parameter - clearing the option causes an error currently

src/components/FormFieldDropdown.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+const FormFieldDropdown = ({
+    title,
+    defaultValue,
+    value,
+    id,
+    errorMessage,
+    onFieldChange,
+    isRequired,
+    style,
```
alexanderleegs, line 30: Some parameters here are unused

alexanderleegs (changes requested):
Seems mostly good, left minor comments

src/components/FormFieldDropdown.jsx
kwajiehao: Resolved with [bc17015](https://github.com/isomerpages/isomercms-frontend/pull/275/commits/bc1701505ed202660f3e6e472dd03d66507858fb)

src/components/FormFieldDropdown.jsx
kwajiehao: Resolved with [bc17015](https://github.com/isomerpages/isomercms-frontend/pull/275/commits/bc1701505ed202660f3e6e472dd03d66507858fb)

kwajiehao:
> Seems mostly good, left minor comments.
> 
> Also is it obvious what needs to be filled in for the Facebook Pixel and Google Analytics fields? We could consider modifying the placeholder text for those if it isn't

Updated placeholder text to indicate they should add their ID in [cff9b51](https://github.com/isomerpages/isomercms-frontend/pull/275/commits/cff9b51030817ab25264b97f8f0511b9affaf2bd)

alexanderleegs (approved):
lgtm

Issue #239, opened 2020-11-18
modifying state directly without using setState


gweiying:
`EditHomepage` currently has multiple places where state is directly modified without using setState, for example not limited to:

1. `sections[sectionIndex][sectionType][field] = value;` ([link](https://github.com/isomerpages/isomercms-frontend/blob/6f8f4b3171cb2c79ea949656874c8715433262e2/src/layouts/EditHomepage.jsx#L242))
2. `highlights[highlightsIndex][field] = value;` ([link](https://github.com/isomerpages/isomercms-frontend/blob/6f8f4b3171cb2c79ea949656874c8715433262e2/src/layouts/EditHomepage.jsx#L300))
3. Here below we're making a shallow copy of the state, and modifying the reference to the state, which may also cause undesirable behavior. ([link](https://github.com/isomerpages/isomercms-frontend/blob/6f8f4b3171cb2c79ea949656874c8715433262e2/src/layouts/EditHomepage.jsx#L710))
```
          const { displaySections } = this.state;
          const sectionId = idArray[1];
          const newDisplaySections = displaySections;
          newDisplaySections[sectionId] = !newDisplaySections[sectionId];
```

Directly modifying the state without using setState is problematic because this is likely to introduce bugs / unintentional behavior that's difficult to trace.

PR #277, opened 2020-12-02
fix: set state directly in EditHomepage

kwajiehao:
This PR addresses issue #239.

Currently, there are multiple instances in the `EditHomepage` layout where state was being indirectly set by referencing the state object directly. Refer to issue #239 for more details. This PR fixes that by ~referencing a deep-copied version of the state object instead of the state object itself~ using `immutability-helper` to update the state instead.

### DIFF
--- a/src/layouts/EditHomepage.jsx
+++ b/src/layouts/EditHomepage.jsx
@@ -251,13 +251,21 @@ export default class EditHomepage extends Component {
         case 'section': {
           // The field that changed belongs to a homepage section config
           const { sections } = state.frontMatter;
-          const copiedSections = _.cloneDeep(sections);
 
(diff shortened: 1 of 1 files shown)


gweiying (changes requested):
as discussed offline, we should use the immutability-helper library for state updates to standardise with newErrors and reduce memory footprint

kwajiehao:
> as discussed offline, we should use the immutability-helper library for state updates to standardise with newErrors and reduce memory footprint

Fixed in [a9fd696](https://github.com/isomerpages/isomercms-frontend/pull/277/commits/a9fd6961103fc9968c2be279159f7f5d74fea803)

gweiying (approved):
lgtm

Issue #278, opened 2020-12-02
Error-handling: 404 and error page


kwajiehao:
Previously, issue #253 brought up the need for 404 pages and error pages. This issue goes into more detail on the implementation details of these pages.

## Error page

Any error caught by the error boundary should render the error page as the fallback component. The Figma design for the error page can be found here: https://www.figma.com/file/4qNhGTBMs521pDvCAIh4ON/IsomerCMS?node-id=474%3A6821

## 404 page

Currently, when we attempt to access a layout for a non-existent file, the API call fails and throws an error, but the component does not handle the error. In such a case, we should throw a 404 error from our backend, and the frontend should handle the 404 error by redirecting the user to the 404 page, where they will have the option to go back to their Workspace.

We will need to do this for the following layouts/components:
- `EditPage`
- `EditContactUs`

We do not need to do this for `CategoryPages` because we have built in automatic redirection rules back to the Workspace if the collection/resource category doesn't exist.

## Other notes

Currently, we need to implement the 404 redirection and error boundary fallback component separately, since the native error boundary provided by React does not catch certain events (see PR #262). We can use an alternative error boundary library, `react-error-boundary`, to unify these approaches. However, since we only need to implement the 404 page for a small number of components, we will leave this as an option to explore in the future if needed.

PR #280, opened 2020-12-03
Feat/add 404 and generic error pages

alexanderleegs:
This PR attempts to resolve https://github.com/isomerpages/isomercms-frontend/issues/278 by adding 404 and generic error pages.

The 404 page currently also handles the case where users attempt to edit a page they do not have access to - although users only see sites they have access to on '/sites', it is still possible for them to retrieve page information for public repos by accessing the url directly. In these cases, we redirect them to the 404 page once we find that they do not have the appropriate permissions.

We wrap our routes with a check if users are accessing a site - for any route (workspace, files, images etc), if the user does not have the appropriate permissions for that repo, or if the repo does not exist, they are redirected to the 404 page. In this case, the return button redirects them to `/sites`. Separately, if the user attempts to access the Contact Us page when it does not exist in the repo, or any page which does not exist by modifying the url directly, they are also redirected to the 404 page, and the return button on this page redirects them to the workspace of the site they are on.

To be reviewed with PR [#87](https://github.com/isomerpages/isomercms-backend/pull/87) and [#89](https://github.com/isomerpages/isomercms-backend/pull/89) on the isomercms-backend repo.


src/layouts/Files.jsx
```diff
@@ -19,6 +20,7 @@ export default class Files extends Component {
       files: [],
       chosenFile: null,
       pendingFileUpload: null,
+      shouldRedirect: false,
```
kwajiehao, line 23: What situation are we expecting to redirect to `Not Found` for files and images? When the entire directory doesn't exist?

src/layouts/Workspace.jsx
```diff
@@ -36,15 +38,14 @@ const Workspace = ({ match, location }) => {
...
-        } catch (e) {
-          setCollections(undefined)
-          console.log(e)
+        try {
+            const collectionsResp = await axios.get(`${BACKEND_URL}/sites/${siteName}/collections`);
+            if (_isMounted) setCollections(collectionsResp.data?.collections)
+        } catch (err) {
+            setCollections(undefined)
+            console.log(err)
+            if (err?.response?.status === 404) setShouldRedirect(true)
```
kwajiehao, line 47: What is the rationale for redirecting to `Not Found` only if we cannot find collections, but not when we cannot find unlinked pages?

kwajiehao (changes requested):
Left some comments, will follow up with a discussion offline

Would also be helpful if we can make a list of every situation where we are performing a redirect due to an error

kwajiehao (approved):
lgtm

src/layouts/EditHomepage.jsx
```diff
@@ -126,6 +126,7 @@ export default class EditHomepage extends Component {
       },
       savedHeroElems: '',
       savedHeroErrors: '',
+      shouldRedirect: false
```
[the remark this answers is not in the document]
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/280/commits/69cc7f5fbaf9936f7ecf819d4bf0b3c9e1d66e43

src/layouts/EditHomepage.jsx
```diff
@@ -1420,6 +1423,14 @@ export default class EditHomepage extends Component {
             />
           </div>
         </div>
+        {
+          this.state.shouldRedirect &&
```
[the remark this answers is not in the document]
alexanderleegs: Good idea, after the reworking of the 404 page redirection, the split for redirection is easier to handle. Resolved in https://github.com/isomerpages/isomercms-frontend/pull/280/commits/288df089a2d6174c1c800e3aafb14bacb3e380ca

src/layouts/Files.jsx
alexanderleegs: As discussed offline, this was to prevent users from accessing any route in a site they did not have access to. No longer necessary after the rework in https://github.com/isomerpages/isomercms-frontend/pull/280/commits/29101878097a4e7a4c7a80df5dec76b18b9cb6e0

src/layouts/Workspace.jsx
alexanderleegs: No longer necessary after the rework in https://github.com/isomerpages/isomercms-frontend/pull/280/commits/29101878097a4e7a4c7a80df5dec76b18b9cb6e0

alexanderleegs:
> Left some comments, will follow up with a discussion offline
> 
> Would also be helpful if we can make a list of every situation where we are performing a redirect due to an error

The initial comment has been updated with the list of redirects. As discussed, we reworked the redirect to 404 page to wrap it inside the ProtectedRoute component, rather than individually inside each subcomponent. The only remaining components which need separate handling of redirection are ContactUs and EditPage.

src/layouts/EditContactUs.jsx
```diff
@@ -180,10 +182,11 @@ export default class EditContactUs extends Component {
         errors: {
           contacts: contactsErrors,
           locations: locationsErrors,
-        }
+        },
       });
     } catch (err) {
       console.log(err);
+      this.setState({ shouldRedirectToNotFound: true })
```
kwajiehao, line 189: Currently, error-handling for this layout is lacking in sophistication. We'll handle this in a separate PR, documented in issue #284

kwajiehao (approved):
left a comment, but overall lgtm

Issue #279, opened 2020-12-02
Error-handling: toast notifications


kwajiehao:
Previously, issue #253 brought up the need for toast notifications when our API calls fail. Toast notifications are crucial for conveying feedback to users when something goes wrong, and also to provide them with some information on how to proceed.

We should implement a toast notification for each failed API call in the frontend.

kwajiehao:
Resolved with #281

PR #281, opened 2020-12-03
Feat/toast notification errors

kwajiehao:
## Overview

This PR resolves issue #279 by rendering error toast notifications, which first appeared in #247, when a user-triggered API call fails. The error toasts serve as a form of feedback to let the user know that their action has failed and informs them of other possible actions they could take (reload the page, check internet connection etc.).

Previously, we displayed error toasts when a user attempted to create/save a file but there already existed another file with the same file name (PR #247). This PR expands on that to display error toasts for **all** errors, not just for the cases where there is a file name conflict.

This PR also performs a minor bug fix where we attempt to access the `status` attribute of an error (`err.response.status`) in an attempt to display an error toast. This `status` attribute might not exist if the error object is not an HTTP error, which leads to an undefined attribute exception. This is resolved by using optional chaining.

The following list identifies the components and operations where the error toast notification gets displayed.

### Components
**ComponentSettingsModal**
- When a user changes the category dropdown (this triggers an API call to retrieve third nav options)
- When a user attempts to save page settings
- When a user attempts to delete the page

**FolderCard**
- When a user attempts to delete a folder

**FolderModal**
- When a user attempts to rename a folder

**MediaSettingsModal**
- When a user attempts to save media (image/file)
- When a user attempts to delete media

**OverviewCard**
- When a user attempts to move a file
- When a user attempts to delete a file

### Layouts

**EditHomepage**
- When a user attempts to save their homepage data

**EditPage**
- When a user attempts to save page content
- When a user attempts to delete page

**Settings**
- When a user attempts to save their settings

### DIFF
--- a/src/components/CollectionPagesSection.jsx
+++ b/src/components/CollectionPagesSection.jsx
@@ -115,6 +115,7 @@ const CollectionPagesSection = ({ collectionName, pages, siteName, isResource })
                         }
                         collectionPageData={collectionPageData}
                         loadThirdNavOptions={loadThirdNavOptions}
+                        setIsComponentSettingsActive={setIsComponentSettingsActive}
                     />
                 )
             }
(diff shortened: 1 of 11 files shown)


alexanderleegs (approved):
lgtm

PR #285, opened 2020-12-08
Feat/warning if unsaved settings

kwajiehao:
## Overview

Currently, the `EditPage`, `EditHomepage`,  and `EditContactUs` layouts displays warnings if the user attempts to navigate away from the page while there are unsaved changes. This PR extends this functionality to the `Settings` layout.

PR #286, opened 2020-12-08
feat: send only settings field which have been changed

kwajiehao:
**Note 1:** This PR is built on top of PR #285 - please review that first
**Note 2:** This PR is to be reviewed together with PR #[92](https://github.com/isomerpages/isomercms-backend/pull/92) on the backend.

## Overview

This PR modifies the behavior of the `Settings` layout save handler so that only fields which have been changed are sent in the request. The payload we send consists of three attributes: `configSettings`, `footerSettings`, and `navigationSettings`. For example, if only the `title` field has been changed, then the payload sent with the request is:

```
{
  configSettings: {
    title: <new title>
  },
  footerSettings: { },
  navigationSettings: { },
}
```

Previously, we would send all state fields, regardless of whether the fields have changed in value. The problem is that this results in wasteful writes to the GitHub repo on the backend, since we end up writing to all three settings files (`_config.yml`, `_data/footer.yml`, and `_data/navigation.yml`) instead of just the one (with the above example, only the `_config.yml` file would be updated since only the `title` field was modified). It also leads to slow performance since the writes to the GitHub repo are sequential - we are unable to update multiple files on a repo at once because the Git tree we write to would end up being out-of-date.

With this update, we can reduce the load time by more than half (from preliminary test results).

### DIFF
--- a/src/layouts/Settings.jsx
+++ b/src/layouts/Settings.jsx
@@ -71,6 +71,10 @@ const stateFields = {
   },
 };
 
+const CONFIG_FIELDS = ['colors', 'favicon', 'facebook_pixel', 'google_analytics', 'is_government', 'shareicon', 'title']
+const FOOTER_FIELDS = ['otherFooterSettings', 'socialMediaContent']
+const NAVIGATION_FIELDS = ['navigationSettings']
+
 export default class Settings extends Component {
   _isMounted = false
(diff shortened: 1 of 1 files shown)


alexanderleegs (approved):
lgtm

Issue #284, opened 2020-12-08
Error handling for `EditContactUs` page


kwajiehao:
Currently, error handling for the `EditContactUs` page is lacking in sophistication.

1. We have multiple async API calls in the same try-catch block, so when one of them fails, we don't actually know which one is the cause.

```
      const settingsResp = await axios.get(`${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/settings`)
      const { footerContent, footerSha } = settingsResp.data.settings;
      
      const contactUsResp = await axios.get(`${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/pages/contact-us.md`);
      const { content, sha } = contactUsResp.data;
```

2. In the catch block, we simply redirect to the `Not Found` page if there are any errors (PR #280). We should redirect to the `Not Found` page only when the Contact Us file cannot be found (`404` from the Contact Us page api call only). Any other error that arises should be dealt with with an error toast notification (PR #279)

Issue #252, opened 2020-11-23
Accordion behavior in EditHomepage Hero section


kwajiehao:
Currently in the hero section of the `EditHomepage` layout, users are allowed to open multiple sections of the key highlights subsections. If users open all subsections, this would result in bad UX because it makes it harder for them to find another subsection they're trying to edit. The solution is to ensure that only one subsection is open at any one time.

Issue #185, opened 2020-10-30
Permalink regex inconsistent with error message


alexanderleegs:
Currently our permalink regex allows numbers as well as lowercase letters and dashes. This is inconsistent with our error message, and it causes some issues with the generation on the backend. We should update the regex to reflect this.

Issue #175, opened 2020-10-26
Investigate hero subtitle, infopic/infobar description max length


kwajiehao:
Currently, the max length allowed for hero subtitle, infopic/infobar description is 30 characters, which seems a bit scant. We should check with the designers what the max length should be for these fields.

lisatjide:
Just did a quick check, subtitle max length is still 30. Can we put it at 160 characters? This is a recommended meta descriptions for good SEO

PR #290, opened 2020-12-14
Fix/edit error handling and misc fixes

alexanderleegs:
This PR updates the error handling when loading pages and addresses several miscellaneous bugs.

- Resolves https://github.com/isomerpages/isomercms-frontend/issues/284: EditContactUs has been modified to redirect to the 404 page only if the contact-us page failed to load - any other case will only throw a toast notification. EditPage has been modified in the same way.
- EditHomePage only has a single API call for loading, but a toast notification has been added if it fails to load.
- Contact us has been removed from the list of unlinked pages, since it has a separate card now.
- Resolves https://github.com/isomerpages/isomercms-frontend/issues/185: Updated the permalink regex failure message to be consistent with what is allowed.
- Previously, the Contact Us page was still hitting both endpoints to change contact us and settings even if only one of those files were edited. Additional checks have been added to only call the appropriate endpoint if changes have been made.
- Resolves https://github.com/isomerpages/isomercms-frontend/issues/252: Resets the display state of other elements in the accordion whenever an element is opened, similar to how EditContactUs works.
- Fixes a bug where active elements were not collapsed when creating a new item in EditContactUs.
- Resolves https://github.com/isomerpages/isomercms-frontend/issues/175: increased the allowed length of the infobar description and infopic description.
- An additional line has been added in images and files to inform the user of the maximum file size.


src/layouts/EditContactUs.jsx
```diff
@@ -185,8 +196,13 @@ export default class EditContactUs extends Component {
         },
       });
     } catch (err) {
+      if (!this.state.shouldRedirectToNotFound) {
```
kwajiehao, line 199: We shouldn't rely on a state check here since state updating is asynchronous and it's possible that by the time the code reaches this block, the state has not yet been updated.

kwajiehao (changes requested):
Everything looks good, left one comment on the handling of the contact us Not Found page error

src/layouts/EditContactUs.jsx
alexanderleegs: Good point, resolved in https://github.com/isomerpages/isomercms-frontend/pull/290/commits/ba0aab2a90d109ca05267c110db67a6efd3b9baf

kwajiehao (approved):
lgtm

Issue #292, opened 2020-12-15
Bug: `$` sign in file title


kwajiehao:
**Problem**
When making updates to a page / resource page, we generate the file name using the **title** attribute from the file's front matter. However, certain special characters get converted wrongly in the process. In this case, we encountered issues with the `$` sign. For example, `$title` becomes `dollartitle.md`. 

**Solution**
It's actually okay for the file name to take that form, but we want to correct it so that whatever is displayed on the CMS is correct. For example, even if the file name is `dollartitle.md`, when we render on the CMS we want to display `$title`.

For the case of `$`, we can specifically use regex to detect the word `dollar` and its subsequent words. In such cases, we can convert the title back to the `$` sign. For other characters, we will probably need to wait and see whether there are other characters with this problem.

PR #294, opened 2020-12-15
Fix/update resource name display and fix misc bugs

alexanderleegs:
This PR resolves https://github.com/isomerpages/isomercms-frontend/issues/292. It updates our resource name display in preparation for the resource name migration to rename existing resource files. The title of resources are now prettified, similarly to what is being done for pages. Additionally, we check for the existence of the special string "dollar" followed by any alphanumeric character, as this indicates that the title of the file contains a $ sign. The reason we require this is due to the generation and prettifying of resource file names - the filename will still contain the dollar string, but we would like the display the title on the cms as the user first inputted it.

In addition, the following bugs have been fixed:
- A typo in the endpoint being called when deleting folders has been corrected.
- Previously, we were not distinguishing correctly between new and existing collections, causing an error when we attempted to retrieve collection data for a new collection. A new variable to check for new collections has been added to OverviewCard.
- The title of the delete warning modal when deleting pages has been corrected.

### DIFF
--- a/src/components/OverviewCard.jsx
+++ b/src/components/OverviewCard.jsx
@@ -47,6 +47,7 @@ const OverviewCard = ({
   const [canShowGenericWarningModal, setCanShowGenericWarningModal] = useState(false)
   const [chosenCategory, setChosenCategory] = useState()
   const [isOutOfViewport, setIsOutOfViewport] = useState()
+  const [isNewCollection, setIsNewCollection] = useState(false)
   const baseApiUrl = `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}${category ? isResource ? `/resources/${category}` : `/collections/${category}` : ''}`
 
   useEffect(() => {
@@ -74,7 +75,7 @@ const OverviewCard = ({
       } = frontMatter;
 
       let collectionPageData
-      if (!isResource && chosenCategory) {
+      if (!isResource && !isNewCollection) {
         // User selected an existing page collection
         const { collectionPages } = await retrieveThirdNavOptions(siteName, chosenCategory, true)
         collectionPageData = collectionPages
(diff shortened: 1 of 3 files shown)


src/components/OverviewCard.jsx
```diff
@@ -74,7 +75,7 @@ const OverviewCard = ({
       } = frontMatter;
 
       let collectionPageData
-      if (!isResource && chosenCategory) {
+      if (!isResource && !isNewCollection) {
         // User selected an existing page collection
         const { collectionPages } = await retrieveThirdNavOptions(siteName, chosenCategory, true)
         collectionPageData = collectionPages
```
kwajiehao, line 81: Currently, this fails to account for the case when we attempt to move a page from a collection back to unlinked page

kwajiehao (changes requested):
left comments

src/utils.js
```diff
@@ -107,7 +107,16 @@ export function retrieveResourceFileMetadata(fileName) {
...
+  const titleTokenArray = tokenArray.slice(3);
+  const prettifiedTitleTokenArray = titleTokenArray.map((token) => {
+    if (token.length < 2) return token.toUpperCase()
+    return token.slice(0,1).toUpperCase() + token.slice(1)
+  });
+  const prettifiedTitle = prettifiedTitleTokenArray.join(' ')
+
+  // We search for special characters which were converted to text
+  // Convert dollar back to $ if it is followed by any alphanumeric character
+  const title = prettifiedTitle.replaceAll(/dollar(?=([a-zA-Z0-9]))/gi, '$')
```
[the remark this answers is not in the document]
alexanderleegs: Yeah that makes more sense actually, fixed in https://github.com/isomerpages/isomercms-frontend/pull/294/commits/335ab12b00f3afe5b46206ac92bb62fbfb0fdf60

src/components/OverviewCard.jsx
alexanderleegs: Good catch, resolved in https://github.com/isomerpages/isomercms-frontend/pull/294/commits/79bb5f447355ece52ed299aed6e8578a2bf94378

kwajiehao (approved):
lgtm

Issue #297, opened 2020-12-16
Bug: `EditContactUs` layout warning modal bug


kwajiehao:
Currently, we truncate the information on the `EditContactUs` layout if it's too long. For example, if an address is 4 lines long,

```
["45 Maxwell Road", "The URA Centre, Tower", "13-01", "Singapore 069118"]
```

We truncate it so it's only three lines long.

```
["45 Maxwell Road", "The URA Centre, Tower", "13-01"]
```

The problem with this is that we currently make a comparison between the original state of the page with its current state, so that we can notify the user if they attempt to navigate away without saving their changes:

```
    const hasChanges = JSON.stringify(originalFrontMatter) === JSON.stringify(frontMatter) && JSON.stringify(footerContent) === JSON.stringify(originalFooterContent);
```

The problem is that because of this truncation, even if the user has not done anything, when they attempt to navigate away from the page, they will be presented with the warning modal which tells them `You have unsaved changes. Are you sure you want to navigate away from this page?`. **This warning should not be shown** since the user has not actually made any changes.

kwajiehao:
resolved with #298

PR #298, opened 2020-12-18
Feat/contact us changes

gweiying:
## Summary
This PR completes the following requested changes for the Contact Us page. 
- [x] Validation on mount (8dfb463, e940c3d, 000f672, c800d01, 62a2504)
- [x] bug: No warning should be displayed if navigating away without changing anything (Addresses #297) (0ab94a6)
- [x] Display comparison of sanitisation changes (36e9fb0, e970948, 4a8fffd, bfb500c, 
7f2371b, ff056bd, 4f18207)

## Details
When a page loads, both sanitisation and validation is run on a user's `contact-us.md` page contents. Santisation removes fields that are not supported in our Contact-Us editor page format (e.g. we support maximum 3 address lines and 5 fields for operating hours), whereas validation highlights fields that are supported but erroneous under our current validation rules (e.g. we cap the maximum length of the cards titles to 30). 

To highlight to our users what content is removed, we show a modal summarising the changes made to the fields once the page renders, examples displayed below: 

Screenshot 2020-12-28 at 1 00 37 PM
Screenshot 2020-12-30 at 10 19 13 AM

To highlight to our users which of their content fields are erroneous, we highlight the relevant field with red outlines, example displayed below:
Screenshot 2020-12-28 at 1 08 47 PM

To make the Contact Us page most useful, we should take into account how our users use the page (e.g. the average length of content fields, the number of address fields ). A short analysis is [here](https://docs.google.com/presentation/d/1kZ9O94DjDNtnXa9IX8LcTIFlMj2U-t81dKGWOvgPTY4/edit#slide=id.gb1dd37e217_0_82).


src/layouts/EditContactUs.jsx
```diff
@@ -76,6 +78,34 @@ const enumSection = (type, args) => {
...
+  locations.map((location, i) => {
+    if (!isEmpty(location)) {
+      displayText += `<br/>In Location <code>${i+1}</code>: <br/>`
+      Object.entries(location).forEach(([key, value]) => {
+        if (value?.length) {
+          displayText += `    <code>${key}</code> field, <code>${JSON.stringify(value)}</code> is removed </br>`
+        }
+      })
+    }
+  })
```
alexanderleegs, line 105: Since the order of fields on the contact us page is `locations` followed by `contacts`, I think it might be more consistent to show the deleted items in that order as well

alexanderleegs (changes requested):
Code is well written and easy to understand, good job!

A couple of comments - for the demo-v2 site, the default join our mailing list section is outlined in red, but no error message is shown anywhere - I think this might have something to do with incorrect error setting. I believe the error should be with the phone number not being valid, but it is not indicated.

Another issue I encountered which might be related is that contacts section fields other than title cannot be edited - I think this might also be related to error handling.

src/layouts/EditContactUs.jsx
gweiying: That's a good idea, I've swapped it here 5ca1033

gweiying:
> Code is well written and easy to understand, good job!
> 
> A couple of comments - for the demo-v2 site, the default join our mailing list section is outlined in red, but no error message is shown anywhere - I think this might have something to do with incorrect error setting. I believe the error should be with the phone number not being valid, but it is not indicated.
> 
> Another issue I encountered which might be related is that contacts section fields other than title cannot be edited - I think this might also be related to error handling.

Thanks for catching the bug, turns out the format for storing errors was wrong, updated here: 877ce8c

alexanderleegs (approved):
lgtm

Issue #295, opened 2020-12-16
Misc. UI bugs


kwajiehao:
This issue details a couple of minor UI bugs that we should fix.

1. Update error message when uploaded file or image is too large

The new error messages should be `Unable to upload as file size is too large. Please reduce your file size and try again.`

2.  `ComponentSettingsModal` field title

Currently, the field title of `Category Folder Name` is used for both resource categories and collections. We should be more specific on the field title.
- resources: `Category Folder Name` -> `Resource Category Name`
- collections: `Category Folder Name` -> `Collection Name`
- workspace: `Category Folder Name` -> `Collection Name`

kwajiehao:
@alexanderleegs assigning this to you since you've already completed some of it

PR #299, opened 2020-12-21
Feat/error handling when media payload is rejected by nginx

alexanderleegs:
This PR attempts to handle the case where payloads are rejected by nginx. Currently, if a payload is too large, the request is immediately rejected by nginx, preventing our backend error handling from catching and throwing the correct 413 error. To account for this, we check for an empty response, which would indicate that our request did not make it to our backend server. I have simulated this locally by moving our cors initialization after the express limit has been set on the backend to force a network error for large files, since it's not possible to test this properly on local dev, but we should still ensure this is working on staging after merging.

This PR also adds some miscellaneous fixes. The list of these are:
- Change the placeholder text from categories to collections for pages in the page moving modal
- Set the placeholders correctly for the selects in ComponentSettingsModal so they have placeholder styling
- Fixed the error message of category from appearing on the same line as title in the ComponentSettingsModal
- Update the error message for large file size and the field titles for ComponentSettingsModal, to resolve https://github.com/isomerpages/isomercms-frontend/issues/295. Also partially addresses https://github.com/isomerpages/isomercms-frontend/issues/301

### DIFF
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -442,6 +442,7 @@ const ComponentSettingsModal = ({
                     isRequired={true}
                     onFieldChange={changeHandler}
                   />
+                  {errors.title && <br/>}
                   {/* Third Nav */}
                   { 
                     ((type === "page" && originalCategory) || (type === 'page' && !originalCategory && category)) &&
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -416,10 +416,14 @@ const ComponentSettingsModal = ({
                       className="w-100"
                       onChange={categoryDropdownHandler}
                       isDisabled={isCategoryDisabled}
-                      defaultValue={{
+                      placeholder={"Select a category or create a new category..."}
+                      defaultValue={originalCategory ? 
+                        {
                           value: originalCategory,
                           label: generateInitialCategoryLabel(originalCategory, isCategoryDisabled),
-                        }}
+                        }
+                        : null
+                      }
                       options={allCategories}
                     />
                   </div>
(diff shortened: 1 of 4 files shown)


src/components/ComponentSettingsModal.jsx
```diff
@@ -416,14 +416,23 @@ const ComponentSettingsModal = ({
...
                           label: generateInitialCategoryLabel(originalCategory, isCategoryDisabled),
-                        }}
+                        }
+                        : null
+                      }
                       options={allCategories}
                     />
                   </div>
-                  <span className={elementStyles.error}>{errors.category}</span>
+                  { errors.category &&
```
kwajiehao, line 430: We need to do this for the permalink field as well 
Screenshot 2020-12-21 at 2 44 27 PM

kwajiehao (changes requested):
Left a comment on error message rendering bug

src/components/media/MediaSettingsModal.jsx
```diff
@@ -94,10 +94,10 @@ export default class MediaSettingsModal extends Component {
           <Toast notificationType='error' text={`Another ${type === 'image' ? 'image' : 'file'} with the same name exists. Please choose a different name.`}/>, 
           {className: `${elementStyles.toastError} ${elementStyles.toastLong}`}
         );
-      } else if (err?.response?.status === 413) {
-        // Error due to file size too large
+      } else if (err?.response?.status === 413 || err?.response === undefined) {
+        // Error due to file size too large - we receive 413 if nginx accepts the payload but it is blocked by our express settings, and undefined if it is blocked by nginx
         toast(
-          <Toast notificationType='error' text={`The ${type === 'image' ? 'image' : 'file'} is too large. Please choose a smaller ${type === 'image' ? 'image' : 'file'}.`}/>, 
+          <Toast notificationType='error' text={`Unable to upload as ${type === 'image' ? 'image' : 'file'} size is too large. Please reduce your ${type === 'image' ? 'image' : 'file'} size and try again.`}/>,
```
kwajiehao, line 100: "as the file size exceeds 5mb" instead of "size is too large"

alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/299/commits/684120b4c5838f026d524047aa49a05d80aa5645

src/components/ComponentSettingsModal.jsx
alexanderleegs: Good catch, resolved in https://github.com/isomerpages/isomercms-frontend/pull/299/commits/b215b7e3dbfcf6738a61693d140e42d941832afc

kwajiehao (approved):
lgtm

PR #309, opened 2020-12-22
Fix/url check for button and category field title

alexanderleegs:
This PR fixes some issues which surfaced during a usability test conducted. The main bugs fixed are:
- Currently, we have an issue where if the button url is left blank, then the button text is deleted, the error message still persists. This was due to us using state for value before the state actually updated, resulting the error message not being cleared. This PR fixes this issue.
- The field name for resource category in the ComponentSettingsModal has also been updated to reflect that this field is compulsory.


kwajiehao (approved):
lgtm

PR #311, opened 2020-12-23
fix: remove use of enquote string

kwajiehao:
Currently, when saving data for resource pages, we enquote the title in anticipation of special characters in the title. However, this turns out to be unnecessary, and in fact, detrimental because our yaml library automatically enquotes strings which contains special characters. This manual enquoting results in a title string with two double quotes.

**Example**
Screenshot 2020-12-23 at 10 39 33 AM

This PR fixes that by removing the use of enquote string and letting the library do its work.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -143,13 +143,6 @@ export function isEmpty(obj) {
   return isEmptyVal
 }
 
-export function enquoteString(str) {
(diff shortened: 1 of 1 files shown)


LoneRifle (approved):
Lgtm, I trust your reading of the situation

Issue #303, opened 2020-12-21
Implement media search in MediaModal, finding an image for selection after upload


kwajiehao:
### Overview

Currently, in the media uploading flow, after one uploads an image / file, the user is brought back to the media modal, where they are presented with all the images in their repo. This is a bad user experience because they still need to scroll through all their other images to find the image they just uploaded before they can select it. 

This occurs in `EditPage` (when uploading and then selecting an image in the editor), as well as in `EditHomepage` and `Settings`, where image fields need to be selected.

### Proposed solution

Firstly, we can create a search bar for the media modal. This search bar will implement a simple filter of the file names of the images / files retrieved from the repo, so that only the images / files whose filenames match the search term will be displayed in the `MediaModal`.

Secondly, once the search bar is implemented, we can implement a mechanism which pre-fills the search bar with the filename of the newly-uploaded image so that it is immediately displayed and can be selected.

PR #313, opened 2020-12-24
Feat/media modal search bar

kwajiehao:
## Overview

This PR addresses issue #303. 

This PR does two main things:
1. It adds a search bar for the `MediaModal` component (including the corresponding styles). The search bar filters images in the `MediaModal` by file name for images/files which contain the search term as a substring.
2. ~When a user uploads an image/file and they are redirected back from the `MediaSettingsModal` to the `MediaModal`, we pre-fill the search bar so that the image they just uploaded is displayed in the `MediaModal`.~ We automatically insert an image into the editor (in the case of `EditPage`) / automatically select an image (in the case of a `FormFieldMedia`) upon upload of an image

This PR also fixes miscellaneous bugs associated with the file reader:
- unable to retrieve file for upload if an upload operation was previously cancelled ([7a6a550](https://github.com/isomerpages/isomercms-frontend/commit/7a6a55092cc9afb09928a2e5877bf344ed2f278b))
- unable to select the same file twice consecutively for upload ([4a2c745](https://github.com/isomerpages/isomercms-frontend/commit/4a2c745446e97e410326fffcad93ed469a69e2f6))
- fix component prop definitions and misc. style changes

### DIFF
--- a/src/components/media/MediaModal.jsx
+++ b/src/components/media/MediaModal.jsx
@@ -20,23 +20,23 @@ export default class MediaModal extends Component {
   }
 
   async componentDidMount() {
-    const { siteName, type, imageSearchTerm } = this.props;
+    const { siteName, type, mediaSearchTerm } = this.props;
     const mediaRoute = type === 'file' ? 'documents' : 'images';
     try {
       const { data: { documents, images } } = await axios.get(`${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/${mediaRoute}`, {
         withCredentials: true,
       });
       if (_.isEmpty(documents || images)) {
-        this.setState({ medias: [], filteredMedias: [], searchTerm: imageSearchTerm });
+        this.setState({ medias: [], filteredMedias: [], searchTerm: mediaSearchTerm });
       } else {
         let filteredMedias
         if (type === 'file') {
-          filteredMedias = this.filterMediaByFileName(documents, imageSearchTerm)
+          filteredMedias = this.filterMediaByFileName(documents, mediaSearchTerm)
         } else {
-          filteredMedias = this.filterMediaByFileName(images, imageSearchTerm)
+          filteredMedias = this.filterMediaByFileName(images, mediaSearchTerm)
         }
 
-        this.setState({ medias: documents || images, filteredMedias, searchTerm: imageSearchTerm });
+        this.setState({ medias: documents || images, filteredMedias, searchTerm: mediaSearchTerm });
       }
     } catch (err) {
       console.error(err);
@@ -53,11 +53,11 @@ export default class MediaModal extends Component {
   }
 
   searchChangeHandler = (event) => {
-    const { setImageSearchTerm } = this.props
+    const { setMediaSearchTerm } = this.props
     const { medias } = this.state
     const { target: { value } } = event
     const filteredMedias = this.filterMediaByFileName(medias, value)
-    setImageSearchTerm(value)
+    setMediaSearchTerm(value)
     this.setState({ filteredMedias })
   }
 
@@ -68,7 +68,7 @@ export default class MediaModal extends Component {
       onMediaSelect,
       type,
       readFileToStageUpload,
-      imageSearchTerm,
+      mediaSearchTerm,
       selectedFile,
       setSelectedFile,
     } = this.props;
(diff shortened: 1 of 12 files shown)


gweiying:
The search functionality looks good. Regarding the uploading flow, would it be possible to directly upload and insert the image via the Edit Image page and bypass the reopening of the MediaModal + "prefilling" of search term?

kwajiehao:
> The search functionality looks good. Regarding the uploading flow, would it be possible to directly upload and insert the image via the Edit Image page and bypass the reopening of the MediaModal + "prefilling" of search term? The flow seems a little tedious this way.
> 
> 

Yup, that's definitely possible. What do you think @NatMaeTan ?

NatMaeTan:
We've discussed this and considered a few points, and determined that pre-filling the file name in the search bar isn't an ideal UX as:

1. users may be confused when they return to the select image modal because it's not intuitive to delete text input to see all images again
2. this will be a problem for uploading multiple images at once in the future

- In our view, search bar may not be the best component in an image upload modal as compared to a 'sort by' filter. (E.g. by date uploaded, by name)
- A 'sort by' filter is more commonly used than a search bar as users may not remember all their file names.
- A search bar is better suited for resources that are likely to be properly named.

**Our recommended  UX:** Highlight the recently uploaded image and put it as the first image beside the "Upload image" button (i.e. default display is by last uploaded) It's ok to leave the search bar there, but it'll be good to also have the sort by!!

_Refer to screenshot by what we mean by highlight:_
Highlight

NatMaeTan:
We've discussed this and considered a few points, and determined that pre-filling the file name in the search bar isn't an ideal UX as:

1. users may be confused when they return to the select image modal because it's not intuitive to delete text input to see all images again
2. In our view, search bar may not be the best component in an image upload modal as compared to a 'sort by' filter. (E.g. by date uploaded, by name)

A 'sort by' filter is more commonly used than a search bar as users may not remember all their file names.
A search bar is better suited for resources that are likely to be properly named.

**Our recommended UX:** If we are only allowing users to insert 1 image at a time, we recommend auto inserting when user saves an image else there are too many clicks when uploading multiple images. Text button should be changed to "Save & insert" to indicate to users that they are doing 2 actions at a time.

kwajiehao:
Upon discussion with Alexis and designers, code has been updated so that the images are automatically inserted/selected upon upload

src/components/media/MediaModal.jsx
```diff
@@ -6,50 +6,81 @@ import mediaStyles from '../../styles/isomer-cms/pages/Media.module.scss';
...
+        this.setState({ medias: documents || images, filteredMedias, searchTerm: imageSearchTerm });
       }
     } catch (err) {
       console.error(err);
     }
   }
 
+  filterMediaByFileName = (medias, filterTerm) => {
+    const filteredMedias = medias.filter((media) => {
+      if (media.fileName.toLowerCase().includes(filterTerm.toLowerCase())) return true
```
gweiying, line 48: cleaning, unnecessary `if` check here, can simplify lines 48, 49 to:
`return media.fileName.toLowerCase().includes(filterTerm.toLowerCase())`

src/layouts/Images.jsx
```diff
@@ -149,7 +149,7 @@ export default class Images extends Component {
             media={pendingImageUpload}
             siteName={siteName}
             // eslint-disable-next-line react/jsx-boolean-value
```
gweiying, line 151: cleaning, can remove 151

src/layouts/EditPage.jsx
```diff
@@ -399,13 +427,17 @@ export default class EditPage extends Component {
...
-            <MediasModal
+            <MediaModal
               type="image"
               siteName={siteName}
               onMediaSelect={this.onImageClick}
               toggleImageModal={this.toggleImageModal}
               readFileToStageUpload={this.readFileToStageUpload}
               onClose={() => this.setState({ isSelectingImage: false })}
+              imageSearchTerm={imageSearchTerm}
+              setImageSearchTerm={this.setImageSearchTerm}
```
gweiying, line 438: suggestion: rename `imageSearchTerm` to `searchTerm` (see overall comments)

gweiying (commented):
The PR looks great, no logical flaws or buggy behavior that I can see! 

Just left some small suggestions for cleaning up, also a minor nit that I have was that we are switching between referring to an uploaded item as an 'image' and as a 'file' in EditPage, for example below:

```
imageSearchTerm={imageSearchTerm}
setImageSearchTerm={this.setImageSearchTerm}
selectedFile={selectedFile}
setSelectedFile={this.setSelectedFile}
```
Might be good to rename `imageSearchTerm` to `searchTerm`, `setImageSearchTerm` to `setSearchTerm` etc so that it's clear that the search also works on files.

kwajiehao:
> imageSearchTerm

fixed with [e3a61a2](https://github.com/isomerpages/isomercms-frontend/pull/313/commits/e3a61a211316fce738f0d5821dae1ac87dd09269). I used `mediaSearchTerm` instead because that is more consistent with our naming of our components (`MediaModal`, `MediaSettingsModal`) which we use for both images and files, and also just in case we implement search somewhere else in the CMS

gweiying (approved):
lgtm

PR #314, opened 2020-12-29
Fix: disable check for url for resources section

alexanderleegs:
This PR fixes a bug introduced by https://github.com/isomerpages/isomercms-frontend/pull/309/. Currently, the resources section cannot be saved when the button name is present, as the url for that section is always missing and hence results in an error. This PR fixes that by adding an additional check to ensure the section is not of type resources.


kwajiehao (approved):
Lgtm

PR #316, opened 2020-12-29
Fix/highlight errors on edit homepage

alexanderleegs:
This PR fixes several issues raised during the usability test conducted on 29 Dec. The issues fixed in this PR are:
- Implemented error highlighting, similar to what EditContactUs currently has for minimised sections.
- Loosened social media url regex to allow for periods - facebook and instagram links both allow the use of periods.
- Fixed an issue where images with spaces in their name could not be displayed in EditPage
- Implemented automatic scrolling within EditHomepage. Previously, users were having issues with editing and creating new sections because they were not sure of which section they were editing until they changed a value. This PR changes the snapping to occur when the box is expanded, and also when a new section is created.


gweiying (approved):
lgtm

PR #318, opened 2020-12-29
fix: creates div to wrap formfield

gweiying:
This PR contains a small fix for the bug shown below where the error message runs onto the next form field, by creating a `<div></div>` tag to wrap the form fields. 

Before:
Screenshot 2020-12-29 at 12 28 09 PM

Fixed:
Screenshot 2020-12-29 at 12 29 50 PM

### DIFF
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -437,12 +437,7 @@ const ComponentSettingsModal = ({
                       options={allCategories}
                     />
                   </div>
-                  { errors.category &&
-                  <>
-                    <span className={elementStyles.error}>{errors.category}</span>
-                    <br/>
-                  </>
-                  }
+                  { errors.category && <span className={elementStyles.error}>{errors.category}</span> }
                   {/* Title */}
                   <FormField
                     title="Title"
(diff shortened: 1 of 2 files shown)


kwajiehao:
Previously we added extra line breaks in the `ComponentSettingsModal` when we faced this problem (#299), but I think this makes more sense given that the problem is appearing in `EditHomepage` and ostensibly other places as well. 

1. Can you help remove the line breaks from `ComponentSettingsModal`? Currently, the extra div and line break results in two effective line breaks

Screenshot 2020-12-29 at 8 38 45 PM

2. Do we need to apply the same fix to the other form fields (mainly `FormFieldHorizontal`), or is the current situation ok?  Personally, I think the current situation is fine.

_Current situation_
Screenshot 2020-12-29 at 8 43 10 PM

_With additional line break_
Screenshot 2020-12-29 at 8 47 18 PM

gweiying:
> Previously we added extra line breaks in the `ComponentSettingsModal` when we faced this problem (#299), but I think this makes more sense given that the problem is appearing in `EditHomepage` and ostensibly other places as well.
> 
> Can you help remove the line breaks from `ComponentSettingsModal`? Currently, the extra div and line break results in two effective line breaks
> 
> Screenshot 2020-12-29 at 8 38 45 PM
 
Sounds good, addressed here b938f9c

kwajiehao (approved):
lgtm

Issue #320, opened 2020-12-30
Easier way to edit header size


alexanderleegs:
Currently, it is unclear to users that the header size can be adjusted by clicking on the header button multiple times. We should make this clearer to users. We could either add a dropdown menu allowing users to select header size upon clicking the header button, add more header buttons for each header size available, or mouseover text to the button.

PR #322, opened 2020-12-30
fix: adds multiple header buttons to increase and decrease size of header text

gweiying:
Addresses #320 with addition of new header button to allow users to increase and decrease the size of header text.

Screenshot 2020-12-30 at 12 04 52 PM

Note the initial click to create a large/small header is not very intuitive as clicking on the "large" header creates the smallest header that will increase in size as the user clicks it. Personally, I think this may not be so problematic as the contrast in the two buttons may encourage users to experiment, but please feedback if you have other ideas on the behavior!

### DIFF
--- a/src/layouts/EditPage.jsx
+++ b/src/layouts/EditPage.jsx
@@ -28,7 +28,8 @@ import {
   boldButton,
   italicButton,
   strikethroughButton,
-  headingButton,
+  headingSmallerButton,
+  headingBiggerButton,
   codeButton,
   quoteButton,
   unorderedListButton,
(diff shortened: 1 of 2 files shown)


alexanderleegs (commented):
lgtm, but let's check with the designers whether this sufficiently solves the issue. @NatMaeTan do you have any feedback for this?

seaerchin:
tagging @NatMaeTan and @prestonlimlianjie  - is this still relevant?

NatMaeTan:
> tagging @NatMaeTan and @prestonlimlianjie - is this still relevant?

Think we would rethink the UI of it to be more like a dropdown, but the problem is still relevant. Not really a priority now tho, maybe when we do templates

seaerchin:
@NatMaeTan at present, we can decrease the header size by continually clicking the `H` button until it hits the normal text size (then it goes back to the biggest one again); is this good enough for now? (so that i can close this PR HAHA)

NatMaeTan:
Yea sure but we def need to come back to this!!

seaerchin:
Closed for now!

PR #324, opened 2020-12-31
Fix: show different messages for field for new pages and existing pages

alexanderleegs:
This PR fixes a minor issue with the field title of the category field of the component settings modal.

Previously, we updated the title to reflect to users more clearly that they should be choosing categories to add to - however, this also resulted in the field title changing when editing details, which is a source of confusion. This PR introduces separate field titles depending on whether the user can select a field.

In addition, the placeholder text for the collection name when editing the details of unlinked pages has been changed to `Unlinked Page` instead of the previous `Select a category or create a new category...` to make it clearer to the user that the category cannot be edited from this modal.

### DIFF
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -45,6 +45,14 @@ const generateInitialThirdNavLabel = (thirdNavTitle, originalCategory) => {
     return "Select a third nav section..."
 }
 
+const generateCategoryFieldTitle = (type, isCategoryDisabled) => {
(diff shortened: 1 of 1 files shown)


gweiying (commented):
This looks great, clears up my initial confusion about why I couldn't move the unlinked pages. 

Just one small point to put up for discussion, do you think it might be helpful, in the case `isCategoryDisabled == true`, to also add a "Tip" blue note under the field to tell them to move it via "Move To" in the dropdown?

gweiying (approved):
lgtm

Issue #323, opened 2020-12-31
Slugify new category names for users automatically


alexanderleegs:
Currently, we enforce that new resource category/collection names created through the page moving modal and component settings modal needs to be hyphenated. We could consider automatically slugifying the category names for the users instead.

PR #328, opened 2021-02-01
Fix/slugify collection and resource category name automatically

alexanderleegs:
This PR changes the flow of adding new collections/resource categories, to resolve https://github.com/isomerpages/isomercms-frontend/issues/323. Previously, we would enforce that users slugified their category names when generating them, however, this was not a great user experience. Instead, this PR changes it to allow users to include spaces inside their collection/resource category name, but slugify it under the hood instead. 

In addition, this PR fixes a related bug where the slugified category was not being used to generate the base api url for the file.


kwajiehao (approved):
lgtm

PR #332, opened 2021-02-02
fix: updates syntax for background image size

### LINKED ISSUE #331: EditHomepage background image select bug
**Problem**

When we select a new background image in `EditHomepage`, it is not automatically resized. 

Newly selected image is unresized:
Screenshot 2021-02-02 at 12 55 03 AM

Upon reload however, the image is resized:
Screenshot 2021-02-02 at 12 55 12 AM

**Desired behavior**

Image should be resized upon selection!

gweiying:
Fixes #331


alexanderleegs (approved):
lgtm

PR #334, opened 2021-02-04
Refactor: editHomepage to use hooks

alexanderleegs:
This PR refactors the existing EditHomepage to use hooks instead of a class component. The reason for this is twofold:

- We want to implement a similar page to editHomepage for navigation
- We would like to use context to set site colours instead of how it's being done currently, however, we have a mix of class components and functional components currently, which prevents us from using the useContext hook

Of particular note is the change in rendering behaviour while loading the page data. Previously, we would display a blank version of the editHomepage.

**Previous**

Screenshot 2021-02-04 at 4 24 26 PM

However, I ran into some issues replicating this behaviour. Instead, this PR displays the following while loading the page info.

**After**

Screenshot 2021-02-04 at 4 27 28 PM

Also, this PR refactors the scrollRefs to be a state variable to be more consistent with the behaviour in editContactUs.

### DIFF
--- a/src/layouts/EditHomepage.jsx
+++ b/src/layouts/EditHomepage.jsx
@@ -351,8 +351,7 @@ const EditHomepage = ({ match, siteColors, setSiteColors }) => {
             newSectionError = _.cloneDeep(errors.sections[sectionIndex])
             newSectionError[sectionType]['url'] = errorMessage
           } else {
-            newSectionError = _.cloneDeep(errors.sections[sectionIndex])
-            newSectionError = validateSections(errors.sections[sectionIndex], sectionType, field, value)
+            newSectionError = validateSections(_.cloneDeep(errors.sections[sectionIndex]), sectionType, field, value)
 
             if (field === 'button' && !value) {
               newSectionError[sectionType]['button'] = ''
--- a/src/layouts/EditHomepage.jsx
+++ b/src/layouts/EditHomepage.jsx
@@ -198,36 +198,21 @@ const EditHomepage = ({ match, siteColors, setSiteColors }) => {
           // If this is the hero section, hide all highlights/dropdownelems by default
           if (section.hero) {
             const { dropdown, key_highlights: keyHighlights } = section.hero;
+            const hero = { title: '', subtitle: '', background: '', button: '', url: '' }
             if (dropdown) {
+              hero.dropdown = ''
               // Go through section.hero.dropdown.options
               displayDropdownElems = _.fill(Array(dropdown.options.length), false);
               // Fill in dropdown elem errors array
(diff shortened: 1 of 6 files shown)


src/layouts/EditHomepage.jsx
```diff
@@ -870,73 +859,60 @@ export default class EditHomepage extends Component {
...
+    setErrors(newErrors)
   }
 
-  toggleDropdown = async () => {
+  const toggleDropdown = async () => {
     try {
-      this.setState((currState) => ({
-        dropdownIsActive: !currState.dropdownIsActive,
-      }));
+      setDropdownIsActive(!dropdownIsActive)
```
kwajiehao, line 871: nit: state is not updated synchronously so to be safe we should access the previous hook state value using:
```
setDropdownIsActive((prevState) => !prevState)
```

src/layouts/EditHomepage.jsx
```diff
@@ -97,189 +97,222 @@ const enumSection = (type, isErrorConstructor) => {
...
+              sectionsErrors.push({
+                hero: {
+                  title: '', subtitle: '', background: '', button: '', url: '',
+                },
+              });
+            }
+            if (!dropdown && !keyHighlights) {
+              sectionsErrors.push({
+                hero: {
+                  title: '', subtitle: '', background: '', button: '', url: '',
```
kwajiehao, line 227: nit: there is some repetition on the `sectionsErrors.hero` attribute. Maybe we can do something like:

```
const hero = { title: '', subtitle: '', background: '', button: '', url: '' }

if (dropdown) hero.dropdown = ''
if (keyHighlights) <blah>

sectionsErrors.push({ hero })

kwajiehao (approved):
lgtm, class component features were meticulously converted to hooks

slight concern over file length, but let's take that discussion offline

kwajiehao:
Update: although file length is a slight concern in terms of readability, it might take a substantial rethink of the component to further break the `EditHomepage` component down into sub-components so we will hold off on that for the foreseeable future

src/layouts/EditHomepage.jsx
alexanderleegs: Good point, resolved in https://github.com/isomerpages/isomercms-frontend/pull/334/commits/a1c4f55b818f4d2f66c353b5c67fa5d9016f3068

src/layouts/EditHomepage.jsx
alexanderleegs: Good catch, resolved in https://github.com/isomerpages/isomercms-frontend/pull/334/commits/fba319b93f26be5223019069dcbc6a784a70bdc7

kwajiehao (approved):
lgtm

PR #335, opened 2021-02-04
refactor/change editContactUs to use hooks

alexanderleegs:
This PR refactors editContactUs to use hooks. For the motivations and side-effects, refer to https://github.com/isomerpages/isomercms-frontend/pull/334, which has similar changes.


kwajiehao (approved):
lgtm

PR #336, opened 2021-02-04
Fix/double auto logout

kwajiehao:
## Overview

Currently, when we automatically log users out of the CMS, we trigger a browser window alert which warns users that they are being logged out. However, the browser alert is sometimes displayed multiple times, which is a bad user experience. This PR aims to fix this bug by ensuring that the browser alert is only triggered once upon auto-logout.

An example of the multiple auto-logout warnings:
ezgif com-gif-maker

This PR also accomplishes the following:
- Store login information (the `isLoggedIn` state variable) in the `LoginContext` instead of passing it as a prop to minimize re-renders
- Clean up possible memory leaks by preventing state updates upon component dismount in couple of components (namely, `ProtectedRoute`, `Resources`, `CollectionPagesSection`)

### Cause of multiple alerts

After inspecting the auto-logout flow with printed log statements, we realized that the `useEffect` blocks that we use in our auto-logout flow were being triggered more than once. This goes against the explicit designed intention of the auto-logout flow, which is to run each of the `useEffect` blocks once to log the user out and return the app to its initial state. The logout browser alert was being triggered multiple times because the conditional logic which was used in the auto-logout flow was flawed and triggered subsequent state updates in the wrong order.

For example, the initial condition in the `useEffect` to trigger the logout alert was dependent only on the `shouldBlockNavigation` variable:
```
  useEffect(() => {
    if (shouldBlockNavigation) {
      alert('Warning: your token has expired. Isomer will log you out now.')
      const logout = async () =>  {
        console.log('Logging out...')
        await axios.get(`${BACKEND_URL}/auth/logout`)
        setLogoutState()
      }
      logout()
    }
  }, [shouldBlockNavigation])
```
where `setLogoutState` sets both the `isLoggedIn` and `shouldBlockNavigation` states to be `false`. However, with just a check on the `shouldBlockNavigation` variable, it's possible for the `shouldBlockNavigation` variable to change and trigger this `useEffect` block even if `isLoggedIn` is `false`. This does not make sense, since we should only be displaying the alert and logging people out if they are not already logged out! Updating the condition to check for `(shouldBlockNavigation && isLoggedIn)` helps to eliminate this possibility.

More stringent checks were also applied to other sections of the auto-logout flow. This also works to prevent unnecessary re-renders since we limit access to blocks of code which are changing the state of the app and ensuring they are run only once. Thus, we eliminate a large source of unnecessary state updates, and subsequently, re-renders.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -113,7 +113,7 @@ function App() {
   const ProtectedRouteWithProps = (props) => {
     return (
       <Sentry.ErrorBoundary fallback={FallbackComponent}>
-        <ProtectedRoute {...props} isLoggedIn={isLoggedIn} siteColors={siteColors} setSiteColors={setSiteColors} />
+        <ProtectedRoute {...props} siteColors={siteColors} setSiteColors={setSiteColors} />
       </Sentry.ErrorBoundary>
     )
   }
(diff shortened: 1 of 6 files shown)


alexanderleegs (approved):
lgtm

PR #340, opened 2021-02-09
Feat/update site colors without re-render

kwajiehao:
## Overview

Currently, our app re-renders with a distracting flash when we retrieve an update a repo's site colors. 

_Gif demonstrating the problem_
ezgif com-gif-maker (1)
_Upon reload of the page, it displays the default color palette and then flashes to the new color palette_

This is probably* because of a couple of reasons:
- site colors was stored as a state variable at the root app level, which causes a re-render of the root app
- even if we do not pass site colors as a prop, we use the `render` prop in our `ProtectedRoute` HOC, which causes a re-render of the layout component being passed to the HOC

This PR solves this problem by storing site colors in local storage instead. The site color reading and writing processes are abstracted into a custom hook, which greatly reduces the number of lines of code needed to implement it for each component.

*I say probably because I am not 100% sure that these are the exact causes

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -38,7 +38,8 @@ import elementStyles from './styles/isomer-cms/Elements.module.scss';
 import { defaultSiteColors } from './utils/siteColorUtils';
 
 // Import contexts
-const { LoginContext } = require('./contexts/LoginContext')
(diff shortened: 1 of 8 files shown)


alexanderleegs (approved):
lgtm

Issue #342, opened 2021-02-16
Implement new Folder layout


kwajiehao:
## Overview
We have already established that we need to revamp the CMS layout for rendering collection pages. From our discussions (refer to CMS working doc - [link](https://docs.google.com/document/d/1YhAzIWGU_rCKWv7SItTwreWWVOh761r74rsoQ7clpHs/edit#)), we need a layout in which:
- the collection contents are displayed in one-dimension (along a single line, instead of a grid), so that the idea of ordering pages is more clear
- third nav sections, or sub-folders, are rendered as folders for clarity (again, refer to CMS working doc for more details)

This is the mock-up provided by our designers:
<img width="714" alt="Screenshot 2021-02-16 at 10 14 38 PM" src="https://user-images.githubusercontent.com/31984694/108074621-649eac80-70a4-11eb-8a84-a1079d4bf72c.png">

The link to the Figma designs is here - [link](https://www.figma.com/file/4qNhGTBMs521pDvCAIh4ON/IsomerCMS?node-id=780%3A7215).

## Deliverables

- new Folder layout, with navigation functionality only (when a user clicks on a third nav section/sub-folder, the files within that sub-folder are displayed)

This issue is concerned with simply _visually_ implementing the layout. Functionality such as drag-and-drop to rearrange will be addressed in future issues. For a full list of the frontend tasks related to page-reordering, refer to the design document ([link](https://docs.google.com/document/d/1YhAzIWGU_rCKWv7SItTwreWWVOh761r74rsoQ7clpHs/edit#heading=h.fui2kjaoq0et))

Subsequent issues: #343 (page-adding modal), #344 (drag-and-drop functionality), #345 (update `Edit Details` modal to modify frontend behavior on how pages are moved in and out of third nav sections)

PR #350, opened 2021-02-22
Feat/new folder layout

kwajiehao:
## Overview

This PR address issue #342. It implements two new frontend routes:
- `/sites/:siteName/folder/:folderName`
- `/sites/:siteName/folder/:folderName/subfolder/:subfolderName`

Both of which use the same layout, the new `Folders` layout which has been created according to the Figma specifications ([link](https://www.figma.com/file/4qNhGTBMs521pDvCAIh4ON/IsomerCMS?node-id=792%3A10680)). Note that this PR implements the layout visually; none of the functionality for the options buttons (`Rearrange items`, `Create new page`, etc. have been implemented) but the navigation functionality has been implemented (clicking on a page will bring you to the `EditPage` layout, for example).

Screenshot 2021-02-22 at 4 58 35 PM

As part of the layout, we also implemented new helper functions for parsing the directory file retrieved from the backend. 

This PR also includes a fix for the back button redirection in the `Header` component. 

### Fix for `Header` component redirection

Previously, we encountered a bug where in a subfolder, when we attempted to click the back button in the Header component to go back to the parent folder, we ended up redirecting all the way back to the Workspace layout. The reason for this is because we are using the same Folder layout to render both folders and subfolders, hence the Header component does not unmount. This means that the
`shouldPerformBackNav` state remains true and automatically performs the back nav redirection to the Workspace upon loading the parent folder.

The solution to this is to reset our `Header` state. Since the `shouldRedirect` and `shouldPerformBackNav` states should only be `true` ephemerally, we can set up a useEffect which sets them back to false once they are set to true. This would have the effect of resetting the `Header` component state when the user is redirected back to the parent folder.

### To-do
- Add proptypes


src/App.jsx
```diff
@@ -122,43 +127,47 @@ function App() {
...
+              of them to render at a time
+            */}
+              <LoginContext.Provider value={{isLoggedIn, setLogin, setLogoutState}}>
+                <Switch>
+                    <ProtectedRouteWithProps exact path='/auth' component={AuthCallback} />
+                    <ProtectedRouteWithProps exact path="/" component={Home} />
+                    <ProtectedRouteWithProps path="/sites/:siteName/collections/:collectionName/:fileName" component={EditPage} isCollectionPage={true} isResourcePage={false} />
+                    <ProtectedRouteWithProps path="/sites/:siteName/collections/:collectionName" component={CategoryPages} isResource={false}/>
+                    <ProtectedRouteWithProps exact path="/sites/:siteName/folder/:folderName" component={Folders} />
+                    <ProtectedRouteWithProps exact path="/sites/:siteName/folder/:folderName/subfolder/:subfolderName" component={Folders} />
```
gweiying, line 147: will this artificially restrict us to 1 level of nesting?

alexanderleegs (changes requested):
Code is clear and easy to understand, just have one point regarding the subfolder button

gweiying (commented):
code lgtm! 

As discussed, let's have a discussion with the designers whether Isomer might ever plan to support more nesting than 3 layers, since the current effort needed to generalise the endpoint is only marginally more, but could save us tons of future work.

src/utils.js
```diff
@@ -524,4 +535,66 @@ export const getObjectDiff = (obj1, obj2) => {
...
   return difference
+}
+
+export const parseDirectoryFile = (folderContent) => {
+  const decodedContent = yaml.safeLoad(Base64.decode(folderContent))
+  const collectionKey = Object.keys(decodedContent.collections)[0]
+  return decodedContent.collections[collectionKey].order
+}
+
+export const convertFolderOrderToArray = (folderOrder) => {
```
gweiying, line 546: Screenshot 2021-02-23 at 5 37 15 PM
Didn't investigate it in detail, but seems like this function isn't working as expected, see ^

src/layouts/Folders.jsx
```diff
@@ -0,0 +1,132 @@
...
+              {/* Segment divider  */}
+              <div className={contentStyles.segmentDividerContainer}>
+                <hr className="w-100 mt-3 mb-5" />
+              </div>
+              {/* Collections title */}
+              <div className={contentStyles.segment}>
+                <span>
+                    My workspace >
+                    {
+                        folderName && !subfolderName
```
gweiying, line 83: design suggestion - would it be more intuitive to users if this was a link that allowed users to navigate between folders, rather than having to look in the top nav bar?

src/utils.js
gweiying: Files belonging to a subfolder are not properly added to the children array, children arrays currently have a length of 1.

src/App.jsx
kwajiehao: As discussed, since we do not foresee the Isomer template accommodating more than a third-nav level of nesting, we'll stick with the current routing

src/utils.js
kwajiehao: fixed with [9faab87](https://github.com/isomerpages/isomercms-frontend/pull/350/commits/9faab87c8c2f994077d087476964742359a16b83)

src/layouts/Folders.jsx
kwajiehao: I like that idea - implemented in [
5cd1bdd](https://github.com/isomerpages/isomercms-frontend/pull/350/commits/5cd1bdd8a3dbb7f5f3d62e5ac2e73c2644c8f624)

src/layouts/Folders.jsx
```diff
@@ -0,0 +1,132 @@
...
+                        )
+                        : null
+                    }
+                </span>
+              </div>
+              {/* Options */}
+              <div className={contentStyles.contentContainerFolderRowMargin}>
+                <FolderOptionButton title="Rearrange items" isSelected={isRearrangeActive} onClick={toggleRearrange} option="rearrange" />
+                <FolderOptionButton title="Create new page" option="create-page" />
+                <FolderOptionButton title="Create new sub-folder" option="create-sub" />
```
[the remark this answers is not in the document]
kwajiehao: Nice catch, fixed with [f9080df](https://github.com/isomerpages/isomercms-frontend/pull/350/commits/f9080dfd3e5d72df23b27d0edc22aad85bcfa0b8)

kwajiehao:
Resolved bugs and added error-handling for cases where the collection/folder does not exist

alexanderleegs:
> Resolved bugs and added error-handling for cases where the collection/folder does not exist

Should we handle the case for subfolders as well, i.e. if a subfolder does not exist, redirect to the parent folder?

kwajiehao:
> > Resolved bugs and added error-handling for cases where the collection/folder does not exist
> 
> Should we handle the case for subfolders as well, i.e. if a subfolder does not exist, redirect to the parent folder?

Good catch, resolved with [a8f6d79](https://github.com/isomerpages/isomercms-frontend/pull/350/commits/a8f6d79b3ad1bbc021f36ba3d362ee294c7cdb56)

alexanderleegs (approved):
lgtm!

gweiying (approved):
lgtm too

gweiying:
Temporarily restoring for rebasing reference for `feat/drag-drop-folders`

Issue #344, opened 2021-02-16
Implement drag-and-drop functionality for new Folder layout


kwajiehao:
## Overview 

This issue is a follow-up to issue #342. Now that the layout is already implemented, we want to implement the drag-and-drop functionality so that users can reorder the pages in their folder or sub-folder.

<img width="714" alt="Screenshot 2021-02-16 at 10 14 38 PM" src="https://user-images.githubusercontent.com/31984694/108076012-f78c1680-70a5-11eb-9c04-dde08ada0fac.png">

## Deliverables
- Drag-and-drop functionality within both folders and sub-folders, which is triggered only after the users click on `Rearrange items`, and saves the new order to the directory config file

PR #353, opened 2021-02-23
Feat/drag drop folders

gweiying:
This PR adds drag-drop functionality to 1) reorder files and folders in a collection, 2) reorder files in a subfolder. This PR addresses #344, and is a follow-up to #350.

To simplify the logic for now, we have decided to implement the non-live version of reordering, where the backend is only updated after a user completes the "Rearrange Items" button. This PR does not support dragging a file into a subfolder.

#### Bugs: 
~~1. useQuery is disabled while a user reorders a collection, which is ideal behavior because we do not want to be querying and rerendering while the user reorders files. Once the user completes reorder and sends the mutation to modify the file on the backend, useQuery is enabled again. However, currently, it seems that useQuery does not show the updated order immediately, but instead renders the a cached version of the old order for a few seconds.  This may be either to do with 1) some useQuery parameter setting that we can adjust for removing the cache 2) Github taking some time to update the file. Update on bug, seems like 2) is more plausible. I have tried invalidating the query upon mutation success via `queryClient.invalidateQueries()`, but it does not resolve this issue.~~ resolved with e337f4b
~~2. This PR inherits a small bug from `feat/new-folder-layout` where the files belonging to a subfolder is not properly added to the children array; currently all subfolders have a children array length of 1. Once this is resolved, the `convertArrayToFolderOrder functionality` should work as expected. For now, users can test the reordering in subfolders as `convertSubfolderArray` should work as expected without facing the same bugs.~~ resolved with 67ea5d2

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -20,11 +20,16 @@ const getDirectoryFile = async (siteName, folderName) => {
     }
 }
 
+const setDirectoryFile = async (siteName, folderName, payload) => {
+    return await axios.post(`${BACKEND_URL}/sites/${siteName}/collections/${folderName}/pages/collection.yml`, payload);
+}
+
 const getFolderContents = async (siteName, folderName, subfolderName) => {
     return await axios.get(`${BACKEND_URL}/sites/${siteName}/folders?path=_${folderName}${subfolderName ? `/${subfolderName}` : ''}`);
 }
 
 export {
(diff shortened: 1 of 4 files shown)


gweiying:
Closing this branch and pulling a new branch on the latest staging, see #357.

PR #354, opened 2021-02-24
Fix: restriction of image files that can be uploaded

alexanderleegs:
This PR fixes 2 minor bugs involving the uploading of images.

- In the image tab, we placed a restriction on uploadable files to only a certain subset of images. This caused issues for users who wanted to upload a new favicon.
- In places which used `FormFieldMedia` to deal with image uploads, we only restrict the uploadable file to be of any image type. However, due to a typo (`image*` instead of `image/*`), there were no restrictions being placed on files uploaded in this way.

The two places where this could occur have both been changed to allow any type of image (`image/*`).


kwajiehao (approved):
lgtm

PR #357, opened 2021-02-25
Feat/drag drop new

gweiying:
This PR adds drag-drop functionality to 1) reorder files and folders in a collection, 2) reorder files in a subfolder. This PR addresses #344, and is a follow-up to #350. The contents of this PR are identical to that on #353 with the exception of the latest commit c25644a, and just applies the changes on the latest staging.

This PR does not support dragging a file into a subfolder. To simplify the saving logic, the backend is only updated after a user completes the "Rearrange Items" button. The Header backnav will display a warning if a user has not saved their reordered pages.

### DIFF
--- a/src/layouts/Folders.jsx
+++ b/src/layouts/Folders.jsx
@@ -102,7 +102,7 @@ const Folders = ({ match, location }) => {
 
     const toggleRearrange = () => { 
       // if no folder contents, do not enable reordering
-      if (folderOrderArray.length == 0 || !folderContents) return
+      if (folderOrderArray.length === 0 || !folderContents) return
 
       if (isRearrangeActive) { 
         // drag and drop complete, save new order 
@@ -112,7 +112,7 @@ const Folders = ({ match, location }) => {
         } else {
           newFolderOrder = convertArrayToFolderOrder(folderOrderArray)
         }
-        if (JSON.stringify(newFolderOrder)==JSON.stringify(parsedFolderContents)) { 
+        if (JSON.stringify(newFolderOrder) === JSON.stringify(parsedFolderContents)) { 
           // no change in file order
           setIsRearrangeActive((prevState) => !prevState)
           return
--- a/src/layouts/Folders.jsx
+++ b/src/layouts/Folders.jsx
@@ -101,9 +101,6 @@ const Folders = ({ match, location }) => {
     }, [folderContents, subfolderName])
 
     const toggleRearrange = () => { 
-      // if no folder contents, do not enable reordering
-      if (folderOrderArray.length === 0 || !folderContents) return
-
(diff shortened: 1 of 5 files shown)


src/layouts/Folders.jsx
```diff
@@ -56,28 +69,60 @@ const Folders = ({ match, location }) => {
...
+
+      if (isRearrangeActive) { 
+        // drag and drop complete, save new order 
+        let newFolderOrder
+        if (subfolderName) {
+          newFolderOrder = convertSubfolderArray(folderOrderArray, parsedFolderContents, subfolderName)
+        } else {
+          newFolderOrder = convertArrayToFolderOrder(folderOrderArray)
+        }
+        const updatedDirectoryFile = updateDirectoryFile(folderContents.data.content, newFolderOrder)
```
alexanderleegs, line 115: We could consider only updating the directory file if changes have been made

src/layouts/Folders.jsx
```diff
@@ -56,28 +69,65 @@ const Folders = ({ match, location }) => {
...
+            setFolderOrderArray(convertFolderOrderToArray(parsedFolderContents))
           }
         }
     }, [folderContents, subfolderName])
 
-    const toggleRearrange = () => { setIsRearrangeActive((prevState) => !prevState) }
+    const toggleRearrange = () => { 
+      // if no folder contents, do not enable reordering
+      if (folderOrderArray.length == 0 || !folderContents) return
+
```
alexanderleegs, line 106: Nit: we could probably disable if there's only a single item as well. Could also use triple instead of double equality

src/layouts/Folders.jsx
```diff
@@ -56,28 +69,65 @@ const Folders = ({ match, location }) => {
...
+            setFolderOrderArray(convertFolderOrderToArray(parsedFolderContents))
           }
         }
     }, [folderContents, subfolderName])
 
-    const toggleRearrange = () => { setIsRearrangeActive((prevState) => !prevState) }
+    const toggleRearrange = () => { 
+      // if no folder contents, do not enable reordering
+      if (folderOrderArray.length === 0 || !folderContents) return
+
```
alexanderleegs, line 106: As discussed, we might want to move this logic into conditional disabling of the button instead, since currently the button appears clickable even when reordering is disabled

gweiying: fixed in 4987834

src/layouts/Folders.jsx
gweiying: fixed in 6e21dd2

alexanderleegs (approved):
lgtm! We'll shift the base64 encoding to the backend in a future PR

PR #358, opened 2021-02-25
refactor: use react-query for EditNavBar

kwajiehao:
## Overview

This PR refactors the `EditNavBar` layout to use `react-query`, in line with our recent PRs (#350, #357). The GET and POST API calls are now defined as functions within the `api.js` file, and we replace them with `useQuery` and `useMutation` in the `EditNavBar` code.

This PR also creates a new, simplified `LoadingButton` component that is compatible with `react-query`. `react-query` provides us with the loading state of our api call - this means we don't need to handle it ourselves within our `LoadingButton` component. The new button component simply receives an `isLoading` prop and renders a loading spinner and loading styles when `isLoading` is true.


alexanderleegs (approved):
lgtm

PR #364, opened 2021-03-02
Feat/create success and error toast wrapper functions

kwajiehao:
## Overview 

This PR reduces code duplication and unnecessary dependency imports by creating wrapper toast functions `errorToast` and successToast`. These toast functions replace all existing invocations of the `toast` function.

Other things added in this PR:
- Add a success toast as feedback to CMS users upon successful reordering of pages within a `Folder`
- Refactor the `Folders` and `EditNavBar` layouts so that success and error-handling are done with callbacks using the provided `onSuccess` and `onError` config options. @gweiying brought up a good point that we need the flexibility to implement different forms of error-handling needed for each layout/component. Using these config options (instead of putting them in the API functions) reduces complexity while giving us the flexibility we need.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -1,12 +1,15 @@
-import React from 'react';
 import axios from 'axios';
-import { toast } from 'react-toastify';
 
+<<<<<<< HEAD
 import Toast from './components/Toast';
 
 import { DEFAULT_ERROR_TOAST_MSG, parseDirectoryFile, getNavFolderDropdownFromFolderOrder } from './utils';
 
 import elementStyles from './styles/isomer-cms/Elements.module.scss';
+=======
+import { errorToast, successToast } from './utils/toasts';
+import { DEFAULT_RETRY_MSG } from './utils'
+>>>>>>> fix: update api file to use successToast and errorToast functions
 
 // axios settings
 axios.defaults.withCredentials = true
@@ -21,10 +24,7 @@ const getDirectoryFile = async (siteName, folderName, errorCallback) => {
         if (err.response && err.response.status === 404) {
             errorCallback()
         } else {
-            toast(
-                <Toast notificationType='error' text={`There was a problem retrieving data from your repo. ${DEFAULT_ERROR_TOAST_MSG}`}/>,
-                {className: `${elementStyles.toastError} ${elementStyles.toastLong}`},
-            );
+            errorToast()
         }
         throw err
     }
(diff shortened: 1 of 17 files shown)


src/api.js
```diff
@@ -14,25 +14,34 @@ axios.defaults.withCredentials = true
...
+const getDirectoryFile = async (siteName, folderName, errorCallback) => {
     try {
         return await axios.get(`${BACKEND_URL}/sites/${siteName}/collections/${folderName}/pages/collection.yml`);
     } catch (err) {
         if (err.response && err.response.status === 404) {
-            throw {
-                status: 404,
-                message: err.response.data.message,
-            }
+            errorCallback()
```
gweiying, line 22: As discussed offline, we can consider using React Query's onSuccess, onSettled and onError calls to handle these responses directly in Folders.jsx, since many of these responses seem very specific to reordering.

gweiying (changes requested):
To investigate [useQuery](https://react-query.tanstack.com/reference/useQuery) and [useMutation](https://react-query.tanstack.com/reference/useMutation)'s `on...` options further

src/api.js
kwajiehao: Resolved with [05fc9d4](https://github.com/isomerpages/isomercms-frontend/pull/364/commits/05fc9d4a4836c485a3f9ebb04e1c2aea71cda688)

gweiying (commented):
Small bug observed where toasts with the same value may be duplicated (for example by navigating away and back to the page): Screenshot 2021-03-09 at 9 35 21 AM

This bug should have little impact on an average user because a user is likely to refresh the page on encountering an error, but wondering if there can be a simple way to check for whether there's an active toast with the same success/ error value, and if so suppress the new toast.

kwajiehao:
> Small bug observed where toasts with the same value may be duplicated (for example by navigating away and back to the page): Screenshot 2021-03-09 at 9 35 21 AM
> 
> This bug should have little impact on an average user because a user is likely to refresh the page on encountering an error, but wondering if there can be a simple way to check for whether there's an active toast with the same success/ error value, and if so suppress the new toast.

Fixed with [d655a50](https://github.com/isomerpages/isomercms-frontend/pull/364/commits/966a9b8c9fe74b125900fac013f9062ca4f0e399)

alexanderleegs (approved):
lgtm

gweiying:
> > Small bug observed where toasts with the same value may be duplicated (for example by navigating away and back to the page): Screenshot 2021-03-09 at 9 35 21 AM
> > This bug should have little impact on an average user because a user is likely to refresh the page on encountering an error, but wondering if there can be a simple way to check for whether there's an active toast with the same success/ error value, and if so suppress the new toast.
> 
> Fixed with [d655a50](https://github.com/isomerpages/isomercms-frontend/pull/364/commits/966a9b8c9fe74b125900fac013f9062ca4f0e399)

Nice find!

gweiying (approved):
lgtm!

PR #369, opened 2021-03-08
Feat/save new folder functionality

alexanderleegs:
This PR introduces functionality for the creation of new folders. To be reviewed in conjunction with PR #[124](https://github.com/isomerpages/isomercms-backend/pull/124) on the isomercms-backend.

This PR also fixes a bug where duplicate folder names were not being checked correctly.

~~Currently, there is a bug involving subfolder creation, where the redirection to the new subfolder occurs, but the folder data is not updated, resulting in an immediate redirection to the workspace.~~
Fixed by refetching data when the url is changed.

### DIFF
--- a/src/components/FolderCreationModal.jsx
+++ b/src/components/FolderCreationModal.jsx
@@ -57,23 +57,21 @@ const FolderCreationModal = ({
   ]
 
 
-  const { mutate: saveHandler, isLoading, error: mutateError } = useMutation(
+  const { mutate: saveHandler, isLoading } = useMutation(
     () => moveFiles(siteName, [ ...selectedFiles ], slugifyCategory(title), parentFolder),
     { onSuccess: () => {
-      const redirectUrl = `/sites/${siteName}/folder/${parentFolder ? `${parentFolder}/subfolder/${slugifyCategory(title)}` : slugifyCategory(title)}`
-      setRedirectToPage(redirectUrl)
-      setIsFolderCreationActive(false)
-    }}
-  )
-
-  useEffect(() => {
-    if (mutateError) {
-      toast(
-        <Toast notificationType='error' text={`There was a problem trying to save your nav bar. ${DEFAULT_ERROR_TOAST_MSG}`}/>, 
-        {className: `${elementStyles.toastError} ${elementStyles.toastLong}`}
-      );
+        const redirectUrl = `/sites/${siteName}/folder/${parentFolder ? `${parentFolder}/subfolder/${slugifyCategory(title)}` : slugifyCategory(title)}`
+        setRedirectToPage(redirectUrl)
+        setIsFolderCreationActive(false)
+      },
+      onError: () => {
+        toast(
+          <Toast notificationType='error' text={`There was a problem trying to save your nav bar. ${DEFAULT_ERROR_TOAST_MSG}`}/>, 
+          {className: `${elementStyles.toastError} ${elementStyles.toastLong}`}
+        );
+      }
     }
-  }, [mutateError])
+  )
 
   const folderNameChangeHandler = (event) => {
     const { id, value } = event.target;
--- a/src/components/FolderCreationModal.jsx
+++ b/src/components/FolderCreationModal.jsx
@@ -5,16 +5,15 @@ import * as _ from 'lodash';
 import update from 'immutability-helper';
 import { useMutation } from 'react-query';
 import Select from 'react-select';
-import { toast } from 'react-toastify';
 
 import FolderCard from './FolderCard';
-import Toast from './Toast';
 import LoadingButton from '../components/LoadingButtonReactQuery';
 import FolderNamingModal from './FolderNamingModal';
+import { errorToast } from '../utils/toasts';
 import useRedirectHook from '../hooks/useRedirectHook';
 
 import { validateCategoryName } from '../utils/validators';
-import { deslugifyPage, slugifyCategory, DEFAULT_ERROR_TOAST_MSG } from '../utils'
+import { deslugifyPage, slugifyCategory, DEFAULT_RETRY_MSG } from '../utils'
 
 import elementStyles from '../styles/isomer-cms/Elements.module.scss';
 import contentStyles from '../styles/isomer-cms/pages/Content.module.scss';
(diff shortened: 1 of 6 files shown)


src/components/FolderCreationModal.jsx
```diff
@@ -54,17 +56,29 @@ const FolderCreationModal = ({
...
+    () => moveFiles(siteName, [ ...selectedFiles ], slugifyCategory(title), parentFolder),
+    { onSuccess: () => {
+      const redirectUrl = `/sites/${siteName}/folder/${parentFolder ? `${parentFolder}/subfolder/${slugifyCategory(title)}` : slugifyCategory(title)}`
+      setRedirectToPage(redirectUrl)
+      setIsFolderCreationActive(false)
+    }}
+  )
+
+  useEffect(() => {
+    if (mutateError) {
```
gweiying, line 70: can use `onError: () => function()` callback from useMutation instead, similar to `onSuccess()`

src/components/FolderCreationModal.jsx
```diff
@@ -54,17 +56,29 @@ const FolderCreationModal = ({
     }
   ]
 
-  const baseApiUrl = `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}${parentFolder ? `/collections/${parentFolder}` : ''}`
 
-  const saveHandler = async () => {
-    // TODO
-    console.log('saving')
-  }
+  const { mutate: saveHandler, isLoading, error: mutateError } = useMutation(
```
gweiying, line 60: As opposed to using `isLoading`, try using `mutateAsync` to replace `mutate` - that should return a promise which can be awaited, which is probably compatible with our `await callback()` in `LoadingButton`

kwajiehao: Previously, the `LoadingButton` stored its own loading state, which is why we needed to `await` the callback. We found this unnecessarily complicated and removed the state from the loading button by creating a separate `LoadingButtonReactQuery` so that the button component itself doesn't store any state but rather accepts an `isLoading` prop which determines the loading state of the button

gweiying (commented):
### Note
As discussed offline, we might have to consider inserting `third_nav_title:` (either on backend or frontend) to maintain compatibility with our v4 repos before we rewrite the templates.

src/components/FolderCreationModal.jsx
alexanderleegs: Good point, resolved in https://github.com/isomerpages/isomercms-frontend/pull/369/commits/c479ccafc75503d0676e4e72e1ee24614e0e01f4

src/components/FolderCreationModal.jsx
alexanderleegs: As discussed offline, moving forward we will be using `LoadingButton` instead of `LoadingButtonReactQuery`. Fixed in ce4a432

src/utils/validators.js
```diff
@@ -24,6 +25,8 @@ const RESOURCE_CATEGORY_REGEX = '^([a-zA-Z0-9]+[- ])*[a-zA-Z0-9]+$';
 const resourceRoomNameRegexTest = /^([a-zA-Z0-9]+-)*[a-zA-Z0-9]+$/
 const resourceCategoryRegexTest = RegExp(RESOURCE_CATEGORY_REGEX);
 
+const ISOMER_TEMPLATE_PROTECTED_DIRS = ['data', 'includes', 'site', 'layouts', 'files', 'images', 'misc', 'pages']
```
gweiying, line 28: Suggestion: might be a good idea to display this in the folder naming modal

alexanderleegs: Good point, as discussed, we will put this in the error message for now as this should not be a common occurrence. Changed in https://github.com/isomerpages/isomercms-frontend/pull/369/commits/9a5868f843966298a92e41359e195c980eedaad8

src/components/FolderCreationModal.jsx
```diff
@@ -1,23 +1,24 @@
...
 import * as _ from 'lodash';
 import update from 'immutability-helper';
+import { useMutation } from 'react-query';
 import Select from 'react-select';
-import { toast } from 'react-toastify';
 
 import FolderCard from './FolderCard';
-import LoadingButton from './LoadingButton';
-import Toast from './Toast';
+import LoadingButton from '../components/LoadingButtonReactQuery';
```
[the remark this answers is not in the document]
alexanderleegs: Whoops nice catch, resolved in https://github.com/isomerpages/isomercms-frontend/pull/369/commits/c56913c68df0f8145d7859bbfcdfdc6f4d2f6d3b

gweiying (approved):
lgtm

src/components/FolderCreationModal.jsx
```diff
@@ -42,29 +43,37 @@ const FolderCreationModal = ({
...
+    { onSuccess: () => {
+        const redirectUrl = `/sites/${siteName}/folder/${parentFolder ? `${parentFolder}/subfolder/${slugifyCategory(title)}` : slugifyCategory(title)}`
+        setRedirectToPage(redirectUrl)
+        setIsFolderCreationActive(false)
+      },
+      onError: (error) => {
+        if (error.response.status === 409) {
+          errorToast(`The name chosen is a protected folder name. Please choose a different name.`)
+        } else {
+          errorToast(`There was a problem trying to save your nav bar. ${DEFAULT_RETRY_MSG}`)
```
kwajiehao, line 68: The error toast message should be about folder creation and not nav bar

src/components/FolderCreationModal.jsx
```diff
@@ -185,6 +189,7 @@ const FolderCreationModal = ({
                   disabledStyle={elementStyles.disabled}
                   className={`${selectedFiles.size === 0 ? elementStyles.disabled : elementStyles.blue}`}
                   callback={saveHandler}
+                  isLoading={isLoading}
```
kwajiehao, line 192: If we're back to using `LoadingButton`, then we should remove the `isLoading` variable from `useMutation` and also remove it as a prop

kwajiehao (changes requested):
Left comments

src/components/FolderCreationModal.jsx
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/369/commits/28a1bf897b7b5ec22117538ee4b4ed727466c9f3

src/components/FolderCreationModal.jsx
alexanderleegs: Resolved in https://github.com/isomerpages/isomercms-frontend/pull/369/commits/319ab8186837bb72b6043d5247cfeb0cc6fb7942

kwajiehao (approved):
lgtm!

Issue #361, opened 2021-03-01
Update logic for retrieving page data from subfolders


alexanderleegs:
Currently, due to the navigation file revamp, our existing method of retrieving file information does not work due to the file path containing slashes. To fix this, we should modify the endpoints to make use of encode/decodeURIComponent, to allow path names to be handled properly.

kwajiehao:
Adding the following for more clarity:

### Problem

Currently, we can't access subfolder pages through the `EditPage` layout. This is because of the changes we've made to our Isomer repos when we migrated to Jekyll v4 - third-nav sections/subfolders are now stored in actual sub-directories, whereas before all pages in a collection/folder were flat (no hierarchy). An example of a collection directory before (where `1a-page2.md` and `1b-page3.md` are in the same third-nav section, as seen by the shared number prefix in `1a` and `1b`):

```
Collection A
- 0-page1.md
- 1a-page2.md
- 1b-page3.md
```

However, with the new repo structure, this collection would no longer be flat:

```
Collection A
- page1.md
- subfolder 1
    - page2.md
    - page3.md
```

Therefore, when we access these folders, the URL ends with `<collection name>/<subfolder name>/<page name>`, vs before, when it was simply `<collection name>/<page name>`. To solve this issue, we need to URI encode `<subfolder name>/<page name>` to fit our current frontend routing.

kwajiehao:
Fixed with PR #379

Issue #356, opened 2021-02-25
Add "Create new page" functionality for Folders layout


kwajiehao:
Currently the `Folders` layout visually renders the collections/folders, but we have not implemented all of the layout's functionality yet. This issue tracks the `Create new page` functionality for the layout. Note that after this week's discussion (refer to meeting minutes here: https://docs.google.com/document/d/1br6T6wVX0KrcA3nwQEo7OhUrcT4veLnaz0vByEXjVvo/edit#heading=h.gh7fe7q87hnx) we agreed that the `Edit Details` modal should no longer have the `Collection title` and `Add to third nav` fields, so this should be taken into account when implementing this functionality.

<img width="954" alt="Screenshot 2021-02-25 at 1 43 15 PM" src="https://user-images.githubusercontent.com/31984694/109109021-93d1af80-776f-11eb-83b1-f288e6df13c8.png">

PR #371, opened 2021-03-08
Feat/new page folder

gweiying:
This PR adds functionality to create folder, subfolder and unlinked pages, based on the related issue #356. The functionality to edit these page settings will be added in a later PR. This PR should be reviewed with [#122](https://github.com/isomerpages/isomercms-backend/pull/122) on the backend.

Preview of Page Setting modal when used for creating new page with duplicated title as current files:
Screenshot 2021-03-08 at 6 48 26 PM

## Notes
- While folder, subfolder and unlinked pages can all be created right now, the redirection to subfolder pages will not work as we currently do not support editing subfolder pages via EditPage on the backend. We should add this functionality in a separate PR. Issue #361
- This PR also does not load the website url from the config page, as we plan to add the website url to the display on the leftbar. Once that is completed, we will have access to this value throughout the site. Issue #372 
- ~Our collection page data retrieved from the `collection.yml` via `convertFolderOrderToArray` and `convertSubfolderArray` has a different format (`{title, type, path}`) from our unlinked page data retrieved via `Directory.list()` on the backend ({`name`, `path`, `sha`, `size`, `type`}). We should consider standardising the name of attributes with the same value i.e. `title`, and `name`.~ fixed with f86253f

### DIFF
--- a/src/components/PageSettingsModal.jsx
+++ b/src/components/PageSettingsModal.jsx
@@ -5,6 +5,7 @@ import * as _ from 'lodash';
 import FormField from './FormField';
 import {
   DEFAULT_ERROR_TOAST_MSG,
+  generatePageFileName,
 } from '../utils';
 
 import { createPage } from '../api'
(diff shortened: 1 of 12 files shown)


src/components/PageSettingsModal.jsx
```diff
@@ -0,0 +1,179 @@
...
+    const { setRedirectToPage } = useRedirectHook()
+
+    const idToSetterFuncMap = {
+      title: setTitle,
+      permalink: setPermalink,
+    }
+
+    const { data } = useQuery(
+      'page',
+      () => console.log('getting page data'),
```
alexanderleegs, line 58: Nit: could remove log statement

alexanderleegs (changes requested):
Left a comment on a log statement. There is also a slight display issue with the permalink field:
Screenshot 2021-03-10 at 12 32 14 PM

src/components/PageSettingsModal.jsx
gweiying: Fixed with 7ce1332

gweiying:
> Left a comment on a log statement. There is also a slight display issue with the permalink field:
> Screenshot 2021-03-10 at 12 32 14 PM

Good catch, fixed with d69811a

alexanderleegs (approved):
lgtm!

PR #378, opened 2021-03-11
Add linting and formatting tools

kwajiehao:
## Overview
This PR is the frontend equivalent of the backend PR for adding linting and formatting tools. Refer to backend PR #[127](https://github.com/isomerpages/isomercms-backend/pull/127) for more details on setup and rationale. As with the backend PR, this PR is only concerned with **introducing** these tools to the repo. Actual linting and formatting will be performed in a subsequent PR as that requires more extensive review. As such, the `.eslintIgnore` file is specified to ignore all files since `react-scripts` automatically runs eslint and fails the build if there are any errors.

The main change in this PR is that we upgrade our `react-scripts` version from `3.4.4` to `4.0.3`. This is because CRA (create-react-app) v3 uses an outdated version of eslint (https://github.com/facebook/create-react-app/issues/8849). This introduced a bug related to the `css-loader` library, which can no longer resolve assets in the `public` folder ([link 1](https://github.com/facebook/create-react-app/issues/9870#issuecomment-719823931), [link 2](https://github.com/webpack-contrib/css-loader/issues/1136#issuecomment-664984703)). This was fixed by moving the referenced image to the relevant sub-directory in the `src` directory.

Other differences between that PR and this one are:
- eslint config is optimized for browser use
- eslint config now contains plugins for React
- Prettier config for sorting imports

The sorting order is:
- third-party imports
- layouts
- components
- templates
- hooks
- contexts
- utils
- styles

### Testing details

To run eslint, run the following command in your command line:

```
npx eslint <file path, or `.` for everything>
```

To run Prettier, run the following command in your command line:

```
npx prettier --write <file path, or `.` for everything>
```


prestonlimlianjie (approved):
lgtm

PR #385, opened 2021-03-23
Fix/clean up of `ComponentSettingsModal` and `OverviewCard`

kwajiehao:
## Context

Over the past couple of months, we've made some strategic decisions which have rendered some parts of the codebase irrelevant.

**1. Our move to use Jekyll 4, and using collection config files to determine collection/folder page order**

Previously, since collection page order was determined by the file name prefixes, we had complicated util functions (`generateNewCollectionFileName`, `generateGroupIdentifier`, etc.) which helped us to calculate the file names when we create, delete, update, or move collection pages. With the new collection config files, these util functions are no longer needed since we no longer need to generate file name prefixes.

**2. Our decision to more closely mimic Google Drive's page-moving UI**

Previously, users were able to create pages in a separate collection/folder from the workspace, using the `ComponentSettingsModal`. With this decision, users can only create pages in the collection/folder/resource category they are in. This decision also meant that the only place where the user can move pages is through the `MenuDropdown` component's `Move to` tab. 

This meant that we no longer have to deal with third nav-related logic for the `ComponentSettingsModal`, which allows us to rip out a huge chunk of code from the component. 

**3. New backend endpoints for moving pages** 

Refer to PR #[124](https://github.com/isomerpages/isomercms-backend/pull/124) in the backend. This allows us to eliminate complex, multi-step frontend logic (such as the `moveFile` function in `OverviewCard`) and replace it with a single API call.

## Changes made in this PR

1. Remove irrelevant code chunks from `ComponentSettingsModal`. The component was also renamed to `ResourcePageSettingsModal` since that ended up being the only use case left
2. Remove unused util functions (`saveFileAndRetrieveUrl`, as well as the other file name prefix generation util functions)
3. Replace `OverviewCard` file-moving logic with a single API call

## Changes to make in subsequent PRs

- Refactor `ResourcePageSettingsModal` and `OverviewCard` to use `react-query` (PR #386)


kwajiehao:
Closing because of overlaps with #388

PR #379, opened 2021-03-15
Feat/fix edit page to work with new format

alexanderleegs:
This PR is to be reviewed in conjunction with PR #[131](https://github.com/isomerpages/isomercms-backend/pull/131) on the isomercms-backend repo. It resolves issues #349 and #361. A summary of the changes made in this PR are as follows:

- Adds a new endpoint for subfolder pages to be edited
- Fixes functionality for editPage to be able to retrieve subfolder pages
- Refactors EditPage to use react-query instead
- Changes the EditPage endpoints to use `/folder` instead of `/collection`
- Fixes a bug involving the retrieval of collection order data inside EditNavBar
- Fixes several bugs related to the change in attribute from title to name made in PR #371, when converting folder order to array and checking for duplicate subfolder names
- Removed several unused classes `Collections`, `EditNav`, `Menus`

In addition, this PR also introduces a constants file for reusing keys for react-query, to allow for the caching of results and reuse of the results across different components.

Issue #139, opened 2020-10-07
Missing "Back to Sites" button on Settings page


kwajiehao:
The "Back to Sites" button which is present on every other page on the top left hand corner is not present on the Settings page

<img width="858" alt="Screenshot 2020-10-07 at 4 10 15 PM" src="https://user-images.githubusercontent.com/31984694/95305524-aaf32c00-08b8-11eb-8cbf-0158acce655a.png">

kwajiehao:
Resolved with #153

PR #388, opened 2021-03-23
Feat/page dropdown api

gweiying:
This PR adds the functionality to support:
- Move
- Delete
- Edit settings

For folder/subfolder pages, unlinked pages and resources. 

This PR is a continuation of #380, and cleans up the `getPage`, `createPage`, `updatePage` APIs added in #380 to resemble those in #379 for better uniformity. 

### Background
This PR focuses on the API integrations above, and should be reviewed with [#139](https://github.com/isomerpages/isomercms-backend/pull/139) on the backend. The visual appearance and UX of the Dropdown modal will be refined in a subsequent PR. ~There is also some buggy behavior on the Dropdown back button behavior, where clicking this series of buttons (`Move To >` > `[folderName] >` > `← Move To` ) will lead to Edit Page rather than triggering the previous Move To dropdown menu as expected. This will also be fixed in a subsequent PR.~ Fixed in 
d697a01

As part of this PR, `OverviewCard`, `ComponentSettingsModal` and `CollectionPagesSection` and `Resources` have been refactored to remove unused functionality and to use ReactQuery. 

~Lastly, the APIs introduced may require some cleanup as there is quite a fair bit of reused variables and flag checks. This will also be done in a subsequent PR.~ Fixed in 7911b8d

### Details
Below, we summarise some key changes to each file. 

#### `ComponentSettingsModal` 
This component is now only used for settings on resource pages, so all functionality for collections and third-nav retrieval has been removed. The only API handler that runs here is Save, while the delete API handler is abstracted into `CollectionPagesSection`. The query call to retrieve the page details is abstracted into `CollectionPagesSection` so that the page details can be reused for deleting functionality. 

Since we plan for the resource creation behavior to match that of page creation behavior, users will be expected to navigate to the category before creating the resource. Hence, the resource category dropdown has also been removed. See below for the new UX:
Screenshot 2021-03-24 at 6 41 37 PM

#### `PageSettingsModal`
Similar to `ComponentSettingsModal`, the only API handler that runs here is Save, while the delete API handler is abstracted into `CollectionPagesSection`. The delete, create, and update APIs from #380 have been updated. The query call to retrieve the page details is abstracted into `Folders` so that the page details can be reused for  deleting functionality. 

The UX has been updated to match that of `ComponentSettingsModal`. See below for the new UX:
Screenshot 2021-03-24 at 7 41 08 PM
#### `CollectionPagesSection` and `Folders`
The query call to retrieve the page details is added into `CollectionPagesSection` and `Folders`. The query fires when the user clicks on the ⋮ Options dropdown for any page.  The new delete and move API ReactQuery handlers have been added to `CollectionPagesSection` and to `Folders`. 

Note: `CollectionPagesSection` is actually used for all sections (unlinked pages and resources) apart from collections, so we should rename this during cleanup.

#### `OverviewCard` and `FolderContent`
The delete and the move handlers are abstracted into `CollectionPagesSection` and `Folders` respectively, and the logic for accessing `Move To` folders and subfolders have been added. This logic will be further refined based on designers input.

#### Changes to `Resources` 
Since we plan for the resource creation behavior to match that of page creation behavior, users will be expected to navigate to the category before creating the resource. Hence, the option to create a new file from the Resources folder is removed, and should be replaced by the functionality to create a new resource folder #384. See below for the new UX:
Screenshot 2021-03-24 at 6 43 09 PM

### Subsequent work
To highlight the changes that will be accomplished in a separate PR:

1) Improve visual appearance and UX behavior of the Dropdown modal, waiting on designers' finalised input
- For Move To, dropdown should start from location of file e.g. if it is a file in a subfolder, the dropdown should start at that subfolder
- Users should not be able to move the file to the same subfolder as the file is currently in, so disabled? vs filtered out?
2) ~Fix buggy behavior on the Dropdown back button behaviour with Back button (depending on UX flow, may not be necessary, waiting on designers' input)~ Fixed in d697a01
3) ~Clean up introduced APIs for duplicated flag checking and redundant variables~  Fixed in 7911b8d
4) Fix Resource `ComponentSettingsModal` to automatically generate `file_url` when resource is a file

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -152,14 +152,6 @@ const createPage = async (endpointUrl, content) => {
     return await axios.post(`${BACKEND_URL}/sites/${endpointUrl}`, { content });
 }
 
-const getPage = async (pageType, siteName, collectionName, pageName) => {
-    const endpointUrl = (pageType === 'collection') 
-                      ? `${siteName}/collections/${collectionName}/pages/${pageName}`
-                      : `${siteName}/pages/${pageName}`
-    const resp = await axios.get(`${BACKEND_URL}/sites/${endpointUrl}`);
-    return resp.data
-}
-
 const updatePage = async(endpointUrl, content, sha) => {
     return await axios.post(`${BACKEND_URL}/sites/${endpointUrl}`, { content, sha });
 }
@@ -196,7 +188,6 @@ export {
     getEditNavBarData,
     updateNavBarData,
     createPage,
-    getPage,
     updatePage,
     deletePage,
     moveFiles,
--- a/src/api.js
+++ b/src/api.js
@@ -82,12 +82,12 @@ const createPageData = async ({folderName, subfolderName, newFileName, siteName,
     
     // redirect to new page upon successful creation
     if (folderName) {
-        return `/sites/${siteName}/folder/${folderName}/${encodeURIComponent(`${subfolderName ? `${subfolderName}/` : ''}${newFileName}`)}`
+        return `/sites/${siteName}/folder/${folderName}/${subfolderName ? `subfolder/${subfolderName}/` : ''}${newFileName}`
     } 
     if (resourceName) {
-        return `/sites/${siteName}/resources/${resourceName}/${encodeURIComponent(newFileName)}`
+        return `/sites/${siteName}/resources/${resourceName}/${newFileName}`
     }
-    return `/sites/${siteName}/pages/${encodeURIComponent(newFileName)}`
+    return `/sites/${siteName}/pages/${newFileName}`
 }
 
 const renamePageData = async ({folderName, subfolderName, fileName, siteName, resourceName, newFileName}, content, sha) => {
--- a/src/api.js
+++ b/src/api.js
@@ -21,7 +21,10 @@ const getFolderContents = async (siteName, folderName, subfolderName) => {
 }
 
 // EditPage
-const getPageApiEndpoint = ({folderName, subfolderName, fileName, siteName, resourceName}) => {
+const getPageApiEndpoint = ({folderName, subfolderName, fileName, siteName, resourceName, newFileName}) => {
+    if (newFileName && fileName) return getRenamePageApiEndpoint({folderName, subfolderName, fileName, siteName, resourceName, newFileName})
+    if (newFileName) return getCreatePageApiEndpoint({folderName, subfolderName, siteName, resourceName, newFileName})
+
     if (folderName) {
         return `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/collections/${folderName}/pages/${encodeURIComponent(`${subfolderName ? `${subfolderName}/` : ''}${fileName}`)}`
     }
@@ -29,6 +32,26 @@ const getPageApiEndpoint = ({folderName, subfolderName, fileName, siteName, reso
         return `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/resources/${resourceName}/pages/${fileName}`
     }
     return `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/pages/${fileName}`
+}
+
+const getCreatePageApiEndpoint = ({folderName, subfolderName, fileName, siteName, resourceName, newFileName}) => {
+    if (folderName) {
+        return `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/collections/${folderName}/pages/new/${encodeURIComponent(`${subfolderName ? `${subfolderName}/` : ''}${newFileName}`)}`
+    }
+    if (resourceName) {
+        return `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/resources/${resourceName}/pages/new/${newFileName}`
+    }
+    return `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/pages/new/${newFileName}`
+}
+
+const getRenamePageApiEndpoint = ({folderName, subfolderName, fileName, siteName, resourceName, newFileName}) => {
+    if (folderName) {
+        return `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/collections/${folderName}/pages/${encodeURIComponent(`${subfolderName ? `${subfolderName}/` : ''}${fileName}`)}/rename/${encodeURIComponent(`${subfolderName ? `${subfolderName}/` : ''}${newFileName}`)}`
+    }
+    if (resourceName) {
+        return `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/resources/${resourceName}/pages/${fileName}/rename/${newFileName}`
+    }
+    return `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/pages/${fileName}/rename/${newFileName}`
 }
 
 const getEditPageData = async ({folderName, subfolderName, fileName, siteName, resourceName}) => {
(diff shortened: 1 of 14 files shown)


src/components/OverviewCard.jsx
```diff
@@ -154,9 +77,9 @@ const OverviewCard = ({
   const handleBlur = (event) => {
```
kwajiehao, line 77: Delete unused `handleBlur` functions in `OverviewCard` and `FolderContent`

src/api.js
```diff
@@ -49,13 +68,38 @@ const getCsp = async (siteName) => {
...
+        return `/sites/${siteName}/folder/${folderName}/${encodeURIComponent(`${subfolderName ? `${subfolderName}/` : ''}${newFileName}`)}`
+    } 
+    if (resourceName) {
+        return `/sites/${siteName}/resources/${resourceName}/${encodeURIComponent(newFileName)}`
+    }
+    return `/sites/${siteName}/pages/${encodeURIComponent(newFileName)}`
+}
+
+const renamePageData = async ({folderName, subfolderName, fileName, siteName, resourceName, newFileName}, content, sha) => {
+    const apiEndpoint = getPageApiEndpoint({folderName, subfolderName, fileName, siteName, resourceName, newFileName})
```
kwajiehao, line 87: Can we simplify the logic here and reduce the responsibility of `getPageApiEndpoint`? For example, for `renamePageData`, we can simply use  `getRenamePageApiEndpoint` instead of calling `getPageApiEndpoint`, which then calls `getRenamePageApiEndpoint`. Likewise for `getCreatePageApiEndpoint`

kwajiehao (changes requested):
Left some comments. Additionally, there is another bug related to `Move to` > subfolder > `Move to` for unlinked pages. Steps to reproduce:
1. Go to the `Workspace` layout
2. Scroll down to the Unlinked Pages section. Click on the options button for one of the pages and click on `Move to`
3. Click on the right caret button to view the selected folder's subfolders
4. Click on the `<- Move to` button to go back (this brings you back to the main modal, instead of the modal screen with all the folders)
5. Clicking on `Move to` again brings you directly to the subfolders view, instead of the folders view

src/api.js
```diff
@@ -31,6 +30,26 @@ const getPageApiEndpoint = ({folderName, subfolderName, fileName, siteName, reso
     return `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/pages/${fileName}`
 }
 
+const getCreatePageApiEndpoint = ({folderName, subfolderName, fileName, siteName, resourceName, newFileName}) => {
```
alexanderleegs, line 33: `fileName` is currently unused

src/api.js
```diff
@@ -49,13 +68,38 @@ const getCsp = async (siteName) => {
...
+    await axios.post(apiEndpoint, params)
+    
+    // redirect to new page upon successful creation
+    if (folderName) {
+        return `/sites/${siteName}/folder/${folderName}/${encodeURIComponent(`${subfolderName ? `${subfolderName}/` : ''}${newFileName}`)}`
+    } 
+    if (resourceName) {
+        return `/sites/${siteName}/resources/${resourceName}/${encodeURIComponent(newFileName)}`
+    }
+    return `/sites/${siteName}/pages/${encodeURIComponent(newFileName)}`
```
alexanderleegs, line 83: We shouldn't be encoding the URI component here, since this is the redirection URL, which is handled appropriately by our router. We also use a different url for subfolders - we can substitute this chunk with

```    
if (folderName) {
        return `/sites/${siteName}/folder/${folderName}/${subfolderName ? `subfolder/${subfolderName}/` : ''}${newFileName}`
    } 
    if (resourceName) {
        return `/sites/${siteName}/resources/${resourceName}/${newFileName}`
    }
    return `/sites/${siteName}/pages/${newFileName}`
```

alexanderleegs (changes requested):
Left some comments.

There's a bug involving the page moving modal not closing on blur - if you click Move to, then click anywhere else (even another file's options), the Move to menu persists

Also, when moving pages into subfolders, the page is automatically created as the first item in the collection.yml file.

src/api.js
gweiying: fixed 7911b8d

gweiying:
> Left some comments.
> 
> There's a bug involving the page moving modal not closing on blur - if you click Move to, then click anywhere else (even another file's options), the Move to menu persists
> 
> Also, when moving pages into subfolders, the page is automatically created as the first item in the collection.yml file.


> Left some comments. Additionally, there is another bug related to `Move to` > subfolder > `Move to` for unlinked pages. Steps to reproduce:
> 
> 1. Go to the `Workspace` layout
> 2. Scroll down to the Unlinked Pages section. Click on the options button for one of the pages and click on `Move to`
> 3. Click on the right caret button to view the selected folder's subfolders
> 4. Click on the `<- Move to` button to go back (this brings you back to the main modal, instead of the modal screen with all the folders)
> 5. Clicking on `Move to` again brings you directly to the subfolders view, instead of the folders view
> 
> We don't need to fix this ^ bug in this PR, just wanted to point it out so that we are aware we need to fix this in subsequent PRs

Bugs above fixed in d697a01

src/components/OverviewCard.jsx
gweiying: used in d697a01

src/api.js
gweiying: fixed 7b92c6c

src/api.js
gweiying: thanks for pointing it out! fixed 83aa2f4

src/layouts/Folders.jsx
```diff
@@ -72,63 +74,138 @@ const Folders = ({ match, location }) => {
...
+    useEffect(() => {
+      if (folderContents && folderContents.data) {
+        const parsedFolderContents = parseDirectoryFile(folderContents.data.content)
+        setDirectoryFileSha(folderContents.data.sha)
+        setParsedFolderContents(parsedFolderContents)
+
+        if (subfolderName) {
+          const subfolderFiles = retrieveSubfolderContents(parsedFolderContents, subfolderName)
+          if (subfolderFiles.length > 0) {
+            setFolderOrderArray(subfolderFiles)
```
alexanderleegs, line 89: Rebase issue, should be 
`setFolderOrderArray(subfolderFiles.filter(item => item.name !== '.keep'))`

src/components/folders/FolderContent.jsx
```diff
@@ -56,47 +76,110 @@ const FolderContentItem = ({
...
+                {
+                    isFile
+                    ? <i className={`bx bxs-file-blank ${elementStyles.folderItemIcon}`} />
+                    : <i className={`bx bxs-folder ${elementStyles.folderItemIcon}`} />
+                }
+                <span className={`${elementStyles.folderItemText} mr-auto`} >{deslugifyPage(title)}</span>
+                {
+                    numItems
+                    ? <span className={`${elementStyles.folderItemText} mr-5`}>{numItems} item{numItems === 1 ? '' : 's'}</span>
+                    : null
```
alexanderleegs, line 92: Could we check for `numItems !== null` instead, so that folders with 0 items still have the label show up?

alexanderleegs (changes requested):
Left some comments.

Also one small comment - the cursor on page cards turns into a normal one instead of a pointer - we might want to change them to always show the pointer

kwajiehao (approved):
Lgtm

src/components/folders/FolderContent.jsx
gweiying: 46ee954

src/layouts/Folders.jsx
gweiying: 5213f6a

alexanderleegs (approved):
lgtm!

PR #390, opened 2021-03-26
feat/ replace js-yaml library with yaml library

kwajiehao:
This PR replaces all use of the npm `js-yaml` library with the npm `yaml` ([link](https://www.npmjs.com/package/yaml)) library. This PR is accompanied by PR #[140](https://github.com/isomerpages/isomercms-backend/pull/140) on the backend.

The main reason for replacing `js-yaml` is because it is unable to handle certain characters, and throws a YAMLException error. An example string which it is unable to handle is: `Stairwells, walkways and fresh
food – making today’s cities healthier cities`, or in unicode, `Stairwells, walkways and fresh food â making todayâs cities healthier cities`

However, the `yaml` library is able to handle such strings. Although it does not have as many weekly downloads as the `js-yaml` library (27 million weekly downloads), it is still widely used (11 million weekly downloads) and actively maintained. It also has an easy interface to use, and substitutes our existing usage of the js-yaml library without any problems.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
     "html-react-parser": "^1.1.1",
     "immutability-helper": "^3.0.1",
     "js-base64": "^2.5.1",
-    "js-yaml": "^3.13.1",
     "lodash": "^4.17.15",
     "marked": "^0.7.0",
     "node-sass": "^4.13.1",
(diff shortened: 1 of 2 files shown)


alexanderleegs (approved):
lgtm

PR #391, opened 2021-03-29
Feat/add ordering to new files when moving

alexanderleegs:
This PR adds in functionality to display the order in which new files are added to newly created folders/subfolders. A number has been added to the top right corner of the file cards to indicate the order in which each file will show up in the new folder/subfolder. This order is determined by the order in which the user selects the files.

Screenshot 2021-03-29 at 1 42 54 PM


kwajiehao (approved):
lgtm, nice and concise change

PR #396, opened 2021-04-06
Fix/resource url

### LINKED ISSUE #395: Fix validation for Resource post url & auto-generation for file url
Currently, our Resource Settings modal does not automatically populate and grey out the file url box when a user uploads a file. We should make changes to automatically populate this field

Another related bug is we should add a backslash before the automatically generated post url, and throw validation errors if the url does not have a backslash.

gweiying:
This PR fixes resource url bugs and other miscellaneous resource bugs. Bugs fixed:

- If a user selects a resource of type `file`, the `file_url` is automatically generated upon upload/ selection (in `FormFieldMedia`) and shown in the field. The field is disabled for editing as users should not be able to edit the file url.
- Missing `/` for examplePermalink when a user creates a new post.
- When toggling between File and Page for new resource, permalink is reset to `permalink` instead of `/${category}/permalink`
- Redirect to EditPage only for new post, not for new file
- Title validation error message disappears when other error messages appear
- Error when deleting file before pageData is retrieved
- Disables Save button if no file selected

Closes #395

### DIFF
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -208,9 +208,9 @@ const ComponentSettingsModal = ({
                   <FormFieldHorizontal
                     title={`https://abc.gov.sg`}             
                     id="permalink"
-                    value={permalink ? permalink : ''}
+                    value={isPost ? permalink : fileUrl}
                     errorMessage={errors.permalink}
-                    isRequired={true}
+                    isRequired={isPost}
                     onFieldChange={changeHandler}
                     disabled={!isPost}
                     placeholder=' '
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -103,7 +103,7 @@ const ComponentSettingsModal = ({
         }
         if (isNewFile) {
           const exampleDate = new Date().toISOString().split("T")[0]
-          const examplePermalink = `${category}/permalink`
+          const examplePermalink = `/${category}/permalink`
           let exampleTitle = 'Example Title'
           while (_.find(pageFileNames, (v) => generateResourceFileName(exampleTitle, exampleDate, isPost) === v ) !== undefined) {
             exampleTitle = exampleTitle+'_1'
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -170,7 +170,7 @@ const ComponentSettingsModal = ({
         setErrors((prevState) => ({
             ...prevState,
             [id]: errorMessage,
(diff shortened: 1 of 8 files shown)


src/components/ComponentSettingsModal.jsx
```diff
@@ -227,7 +227,7 @@ const ComponentSettingsModal = ({
                   />
                 </div>
                 <SaveDeleteButtons 
-                  isDisabled={isNewFile ? hasErrors : (hasErrors || !sha)}
+                  isDisabled={isNewFile && !isPost ? (hasErrors || !fileUrl) : isNewFile && isPost ? hasErrors : (hasErrors || !sha)}
```
alexanderleegs, line 230: For resources, we should perform this check regardless of whether it is a new file, i.e.

`isDisabled={!isPost ? (hasErrors || !fileUrl) : isNewFile && isPost ? hasErrors : (hasErrors || !sha)}`

src/components/ComponentSettingsModal.jsx
```diff
@@ -170,7 +170,7 @@ const ComponentSettingsModal = ({
         setErrors((prevState) => ({
             ...prevState,
             [id]: errorMessage,
-            title: titleErrorMessage,
+            title: (id === 'title' || id === 'date') ? titleErrorMessage : prevState.title
         }));
```
alexanderleegs, line 174: This still doesn't work correctly - having a title error, followed by a date error will remove the title error. 

Also for resources, the duplicate file name check never comes into play - since the titles are compared against the actual file name (which has a date-type-title format) in `pageFileNames.filter(file => file !== fileName)`, we never actually encounter duplicates.

Should we be allowing for identical titles if the date is different?

alexanderleegs (changes requested):
Left comments

src/components/ComponentSettingsModal.jsx
gweiying: Discussed offline, we should disable files from having identical names even if they are of different file types or on posted on different dates.

src/components/ComponentSettingsModal.jsx
gweiying: Fixed in acbe42c, `setHasErrors` contains logic to differentiate between errors for resource posts and resources files

src/components/ComponentSettingsModal.jsx
gweiying: Fixed in 7f76d17

alexanderleegs (approved):
lgtm!

gweiying:
Commits 7e1d27e, 183d066, 3fbbb05 reverts rebase errors, as these commits were only meant to be committed on draft PR #397.

PR #397, opened 2021-04-06
Fix/workspace folder pages ux

gweiying:
This PR fixes inconsistent behavior & layout on Workspace between the folders and the unlinked pages. Namely, these bugs are fixed:
- On hover, create button for both folders and pages turns blue
- Create button for pages changed from "Add a new page" to "Create new page" to match folders
- Create button for folders appears before ReactQuery completes. While query is running, loading message will appear. If no folders are retrieved, loading message will disappear. This behavior matches that of unlinked pages below.

### DIFF
--- a/src/layouts/Workspace.jsx
+++ b/src/layouts/Workspace.jsx
@@ -53,7 +53,7 @@ const Workspace = ({ match, location }) => {
 
         try { 
           const collectionsResp = await axios.get(`${BACKEND_URL}/sites/${siteName}/collections`);
-          if (_isMounted) setCollections(collectionsResp.data?.collections)
+          if (_isMounted) setCollections(collectionsResp.data?.collections || [])
         } catch (e) {
           setCollections(undefined)
           console.log(e)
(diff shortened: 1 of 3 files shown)


src/layouts/Workspace.jsx
```diff
@@ -136,30 +136,16 @@ const Workspace = ({ match, location }) => {
...
-                collections && collections.length === 0 &&
-                <div className={contentStyles.segment}>
-                  There are no collections in this repository.
-                </div>
-              }
-              {/* Collections */}
+              {/* Folders */}
               <div className={contentStyles.folderContainerBoxes}>
                 <div className={contentStyles.boxesContainer}>
+                  <FolderOptionButton title="Create new folder" option="create-sub" isSubfolder={false} onClick={() => setIsFolderCreationActive(true)}/>
```
alexanderleegs, line 146: Could we block this option from appearing until collections have been loaded? Otherwise an error occurs if users attempt to click it too early

alexanderleegs (changes requested):
Left a comment

gweiying:
We will let the designers review the different behaviours on staging and decide how to reconcile between them after they've looked at it!

gweiying:
outdated

PR #400, opened 2021-04-12
Fix/original frontmatter

gweiying:
This PR adds functionality to keep original frontmatter in place when page settings are edited. This is useful for repos such as GoBusiness, where pages may have additional custom frontmatter ([example](https://raw.githubusercontent.com/isomerpages/govtech-gobusiness-main-clone-v4/migration/_business-grants-faqs/productivity-solutions-grant-psg/vendors-edited.md?token=AJLJ6EI2O7UYPIIA2MM7WHLAOPJNI)) that is not editable via the CMS. With this functionality, users such as GoBusiness will be able to keep their custom frontmatter when saving their page settings, even if the custom frontmatter is not editable on the CMS.


alexanderleegs (approved):
lgtm!

PR #402, opened 2021-04-12
Fix: disable creation of resource room link if user has no resource room

alexanderleegs:
This PR fixes an issue with EditNavBar regarding the handling of resource rooms for sites which do not have one. Previously, we always assumed that sites contained a resource room, which caused errors when attempting to retrieve the list of categories. This PR changes it such that we first check if a resource room exists. In addition, the option to create a link with a resource room has now been disabled if a site does not contain a resource room.

Also fixes a minor bug involving the `isEditPage` prop passed to `Header` being a string instead of a boolean.


gweiying (commented):
lgtm!

gweiying (approved):
*lgtm!

PR #403, opened 2021-04-12
Fix/check contactus layout

gweiying:
### Bug
If your Contact Page is a plain markdown file, you won’t be able to edit it on Workspace, but neither will Edit Contact Us work

This is because we parse `contact-us.md` as a specific file with yaml input content, and filter it out from Workspace Pages. This means that a repo that has previously used Contact Us as a normal markdown file will not be able to edit the file properly as the Edit Contact Us button will not work for them, but neither will it show up on Workspace pages.

### Solution 
This PR adds an additional layout check for `contact-us.md` on Workspace , so that if it is a plain markdown file, we will add it to Workspace pages and hide the Edit Contact Us button, while if its a yaml input file, we will enable the Edit Contact us button and hide the Workspace page.


alexanderleegs (approved):
lgtm

PR #407, opened 2021-04-13
fix: resource copy changes

gweiying:
This PR makes the Resources copy changes requested by designers in their CMS v1 review, 13 April ([link](https://docs.google.com/document/d/1_aC5FuHqSQqqWx0aOcpDFyw2iEk1vsCpDgt4fLQ1zck/edit#)). 

Additionally, this PR applies the `deslugify` changes to category/ folder names to Resources, Folders and related components such as `ComponentSettingsModal`, `PageSettingsModal`, `FileMoveDropdown` etc.

The list is also copied below for reference.
> Resources Tab
> [Copy changes] Change ‘Resource Categories’ to just ‘Categories’ (done)
> [Copy changes] Button: Change to ‘Create new category’ (done)
> 
> [Copy changes] Can we standardise the category names to Title case instead of all CAPS (done)
> [Copy changes] At the category workspace, can we use the category name without hyphens i.e. ‘Media Release’ instead of ‘media-release’ (done)
> [Copy changes] When moving resources, it should say ‘Resources’ instead of Workspace (done)
> 
> [Copy Changes] Resources settings: Remove ‘My Workspace’ from breadcrumb (done)

### DIFF
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -198,7 +198,7 @@ const ComponentSettingsModal = ({
             && (
             <div className={elementStyles['modal-settings']}>
               <div className={elementStyles.modalHeader}>
-                <h1>{ isNewFile  ? 'Create new page' : 'Resource settings' }</h1>
+                <h1>{ isNewFile  ? 'Create new resource page' : 'Resource settings' }</h1>
                 <button id="settings-CLOSE" type="button" onClick={() => {setSelectedFile(''); setIsComponentSettingsActive(false)}}>
                   <i id="settingsIcon-CLOSE" className="bx bx-x" />
                 </button>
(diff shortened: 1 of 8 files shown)


src/components/ComponentSettingsModal.jsx
```diff
@@ -208,7 +208,7 @@ const ComponentSettingsModal = ({
                   { isNewFile ? 'You may edit page details anytime. ' : ''}
                   To edit page content, simply click on the page title. <br/>
                   <span className={elementStyles.infoGrey}> 
-                    My workspace > Resources > { category } > <u className='ml-1'>{ title }</u><br/><br/>
+                    Resources > { category } > <u className='ml-1'>{ title }</u><br/><br/>
```
alexanderleegs, line 211: Could we also deslugify category here?

alexanderleegs (changes requested):
[Copy changes] At the category workspace, can we use the category name without hyphens i.e. ‘Media Release’ instead of ‘media-release’ (done)

Could we also change the title on OverviewCard? We can replace line 97 with
`<div className={contentStyles.componentCategory}>{category ? deslugifyDirectory(category) : ''}</div>`

To be consistent, we should also change the category titles in the dropdown menu.

src/components/ComponentSettingsModal.jsx
gweiying: Good suggestions! I've applied the deslugifyDirectory to all Folders- and Resources- related components, d7160af.

alexanderleegs (approved):
lgtm!

PR #410, opened 2021-04-14
Fix/folder naming copy ux

gweiying:
This PR addresses a number of miscellaneous copy and UX bugs to do with Folder/Resource Creation & Folder/Resource Renaming, and the Folders/ Resources/ Category pages, based on the designers' review linked [here](https://docs.google.com/document/d/1_aC5FuHqSQqqWx0aOcpDFyw2iEk1vsCpDgt4fLQ1zck/edit#)

A list of the changes accomplished:
- Copy changes 
- Adds validation to check for duplicated names during Folder & Subfolder & Resource renaming
- Suppresses API call for renaming if no change in name
- Fixes bug with creation of Subfolder
- Relaxes Resource & Folder name regex
- Adds Breadcrumb to CategoryPages

### DIFF
--- a/src/components/CollectionPagesSection.jsx
+++ b/src/components/CollectionPagesSection.jsx
@@ -177,7 +177,7 @@ const CollectionPagesSection = ({ collectionName, pages, siteName, isResource, r
                 && (
                     <GenericWarningModal
                         displayTitle="Warning"
-                        displayText="Moving a page to a different collection might lead to user confusion. You may wish to change the permalink for this page afterwards."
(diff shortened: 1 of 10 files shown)


alexanderleegs (changes requested):
> There's a bug involving folder creation on the workspace currently.
Fixed with 98cdb8a


> Also, is the removal of the sort by button performed in this PR? 

My bad, there was a commit I accidentally reset, now pushed 57d210c

alexanderleegs (approved):
lgtm - As discussed offline, we should consider fixing the name attribute of unlinked pages to bring it more in line with regular collection pages, but we can do this in a future PR.

PR #411, opened 2021-04-14
fix: add resourceRoomName and resourceName to breadcrumb

gweiying:
This PR adds `resourceRoomName` and `resourceName` as a breadcrumb for resource pages, so that the `EditPage` preview shows an accurate breadcrumb similar to that on the actual Isomer site. This PR should be reviewed with [#153](https://github.com/isomerpages/isomercms-backend/pull/153) on the backend.


alexanderleegs (approved):
lgtm

PR #427, opened 2021-04-22
Fix/rearrange layout

gweiying:
This PR revises the layout for rearranging folder content.

This is based on the designers' feedback for the UX review of V1 CMS, where the rearrange feature lacks user guidance. Specifically, the feedback given is as follows: 

- *Rearrange Feature* - Need to discuss this
  - Not intuitive for now, feature is very hidden 
  - Not obvious for users to click on button again to get out of the mode/ save changes 
  - [Copy changes] Change ‘Rearrange items’ to ‘Reorder items’
  - Other functions should be disabled in Reorder mode
  - Cursor should change to Grab cursor to make it more obvious users can drag and drop
  - Critical UX changes:
    - Obvious mode change into reordering mode
    - User need a more intuitive way of how to save changes and get out of Reorder mode

This PR revises the layout to be inline with the previously discussed design layout:
https://www.figma.com/proto/4qNhGTBMs521pDvCAIh4ON/IsomerCMS?node-id=780%3A7264&viewport=-257%2C-424%2C0.15383589267730713&scaling=scale-down

Attached are photos of the updated behavior on Staging CMS:
Screenshot 2021-04-26 at 1 51 00 PM

Additionally, this PR refactors `FolderContent` to simplify the logic and parameter passing.

### DIFF
--- a/src/layouts/Folders.jsx
+++ b/src/layouts/Folders.jsx
@@ -10,7 +10,8 @@ import Header from '../components/Header';
 import Sidebar from '../components/Sidebar';
 import FolderCreationModal from '../components/FolderCreationModal'
 import FolderOptionButton from '../components/folders/FolderOptionButton';
-import FolderContent from '../components/folders/FolderContent';
+import FolderReorderingModal from '../components/FolderReorderingModal';
+import { FolderContent } from '../components/folders/FolderContent';
 import FolderModal from '../components/FolderModal';
 import PageSettingsModal from '../components/PageSettingsModal'
 import DeleteWarningModal from '../components/DeleteWarningModal'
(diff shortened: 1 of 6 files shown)


src/layouts/Folders.jsx
```diff
@@ -58,6 +56,7 @@ const Folders = ({ match, location }) => {
     const [directoryFileSha, setDirectoryFileSha] = useState('')
     const [folderOrderArray, setFolderOrderArray] = useState([])
     const [parsedFolderContents, setParsedFolderContents] = useState([])
+    const [parsedFolderOutput, setParsedFolderOutput] = useState(true)
```
alexanderleegs, line 59: Could we rename `parsedFolderOutput`? It's not immediately apparent that it's a boolean value for showing the folder

src/layouts/Folders.jsx
```diff
@@ -91,9 +90,10 @@ const Folders = ({ match, location }) => {
     // parse contents of current folder directory
     useEffect(() => {
       if (folderContents && folderContents.data) {
-        const parsedFolderContents = parseDirectoryFile(folderContents.data.content)
+        const { order: parsedFolderContents, output: parsedFolderOutput } = parseDirectoryFile(folderContents.data.content)
         setDirectoryFileSha(folderContents.data.sha)
         setParsedFolderContents(parsedFolderContents)
+        setParsedFolderOutput(parsedFolderOutput)
```
alexanderleegs, line 96: Nit: could we rename the variables retrieved so that it doesn't have the same name as the state variable?

alexanderleegs (changes requested):
Looks good, only have some minor nits

alexanderleegs (approved):
lgtm!

PR #430, opened 2021-04-23
Fix/resource color

gweiying:
This PR fixes the resource page header colors to use the repo's secondary color.


alexanderleegs (approved):
lgtm!

PR #439, opened 2021-04-30
fix: pass parameters to wrapped components

alexanderleegs:
This PR fixes a bug in which params given to the `ProtectedRoute` were not being passed to the wrapped component.


prestonlimlianjie (approved):
lgtm!

PR #440, opened 2021-04-30
fix: fixes toast popup on item select, folder deletion modal

gweiying:
This PR fixes two issues on the folders layout:

- Success toast pops up on selecting the item for settings even if no settings are adjusted
- Folder deletion does not work as the deletion modal does not appear


alexanderleegs (approved):
lgtm

PR #442, opened 2021-05-03
Feat/files tab revamp

alexanderleegs:
This PR revamps the existing Files layout to bring it to the same state as the Images layout. To be reviewed in conjuction with PR #TBD on the isomercms-backend repo.

Major changes:
- The Images layout has been refactored into a more general Media layout, as the design of the 2 components are largely similar.
- This PR renames the existing endpoint for files. Previously, we were using the path `/sites/:siteName/files` for the files tab, but this path could not be extended to handle nested subfolders, due to the existence of EditPage using the same path. The path for the files tab has thus been renamed to `/sites/:siteName/documents`.

This PR also fixes a minor bug in which images in nested subfolders could not be deleted.


src/routing/RouteSelector.jsx
```diff
@@ -43,9 +42,10 @@ export const RouteSelector = () => (
     <ProtectedRouteWithProps exact path="/sites/:siteName/folder/:folderName/subfolder/:subfolderName" component={Folders} />
     <ProtectedRouteWithProps exact path="/sites/:siteName/navbar" component={EditNavBar} />
     <ProtectedRouteWithProps path="/sites/:siteName/files/:fileName" component={EditFile} />
-    <ProtectedRouteWithProps path="/sites/:siteName/files" component={Files} />
-    <ProtectedRouteWithProps path="/sites/:siteName/images/:customPath" component={Images} />
-    <ProtectedRouteWithProps path="/sites/:siteName/images" component={Images} />
+    <ProtectedRouteWithProps path="/sites/:siteName/documents/:customPath" component={Media} mediaType={'documents'} />
+    <ProtectedRouteWithProps path="/sites/:siteName/documents" component={Media} mediaType={'documents'} />
+    <ProtectedRouteWithProps path="/sites/:siteName/images/:customPath" component={Media} mediaType={'images'} />
+    <ProtectedRouteWithProps path="/sites/:siteName/images" component={Media} mediaType={'images'} />
```
gweiying, line 48: As a suggestion, what do you think of locating the `images` and `documents` root folder routes as `"/sites/:siteName/documents/home"` and `"/sites/:siteName/images/home"`, which would allow us to simplify the `ProtectedRouteWithProps` for images and documents from 4 to 2

src/api.js
```diff
@@ -256,27 +256,27 @@ const moveFile = async ({selectedFile, siteName, resourceName, folderName, subfo
...
+// Media
+const getMedia = async (siteName, customPath, mediaType) => {
+    const resp = await axios.get(`${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/files/${customPath ? encodeURIComponent(`${mediaType}/${customPath}`) : mediaType}`);
     const { directoryContents } = resp.data;
 
-    let respImages = []
+    let respMedia = []
     let respDirectories = []
     directoryContents.forEach((fileOrDir) => {
         const modifiedFileOrDir = { ...fileOrDir, fileName: fileOrDir.name }
```
gweiying, line 267: variable name isn't very clear to me, why is this `modifiedFileOrDir` if this is a `GET` call?

gweiying:
Overall, the revamp looks great, thanks for the quick work! 

Just some bugs to highlight:
- Upload should show success message and refetch (currently doesn’t refetch)
- Uploading/ Saving media with duplicated name should show duplicate error message instead of generic error message
- Rename of directory should refetch, not reload
- Rename of media should refetch, not reload
- Breadcrumb should be displayed to allow for navigating between folder layers
- This is more a UX concern, but *Ungrouped Images/ Files* seems not a very intuitive name especially if you're already in an image/ Files subfolder

src/api.js
alexanderleegs: Renamed in https://github.com/isomerpages/isomercms-frontend/pull/442/commits/641ee75d585935e5503a6409169b336a92669fde

src/routing/RouteSelector.jsx
alexanderleegs: As discussed offline, using `/sites/:siteName/documents/home` will prevent users from having a media subfolder named `home`. We could work around this by prefixing all media folder routes with `home`, but we have elected to leave it as is for now for greater clarity.

alexanderleegs:
> * Uploading/ Saving media with duplicated name should show duplicate error message instead of generic error message
This was a backend issue, fixed in https://github.com/isomerpages/isomercms-backend/pull/169/commits/52d4ddf403815d798002f153753805a98007f41f

> * Rename of directory should refetch, not reload
https://github.com/isomerpages/isomercms-frontend/pull/442/commits/daeaadd4c9d865106b9c30d1588a870724880115

> * Rename of media should refetch, not reload
As discussed offline, this requires a more extensive refactor of the `MediaSettingsModal` class component - we will do this in a future PR.

> * Breadcrumb should be displayed to allow for navigating between folder layers
Good suggestion, resolved in https://github.com/isomerpages/isomercms-frontend/pull/442/commits/9977a0d34ece872613c2a395fccb727ee5264260

> * This is more a UX concern, but _Ungrouped Images/ Files_ seems not a very intuitive name especially if you're already in an Image/ Files subfolder. Maybe we can check with the designers about this, what do you think?

gweiying (approved):
lgtm!

PR #443, opened 2021-05-03
chore: set up basic Cypress E2E tests

prestonlimlianjie:
This PR adds basic end-to-end tests. Please refer to the instructions in the README.md file on how to run Cypress E2E tests.

#### Home page
- Home page should have login button
- Login button should redirect to GitHub login on click

#### Sites page
- Sites page should have sites header
- Sites page should allow user to click into a test site repo

### DIFF
--- a/cypress/integration/homepage.spec.js
+++ b/cypress/integration/homepage.spec.js
@@ -1,0 +1,31 @@
+describe('Login flow', () => {
+  const CMS_BASEURL = Cypress.env('BASEURL')
+  const COOKIE_NAME = Cypress.env('COOKIE_NAME')
+  const LOGIN_BUTTON_TEXT = 'Login with GitHub'
+  const GITHUB_LOGIN_URL = 'https://github.com/login'
+
+  beforeEach(() => {
+    // Before each test, we can automatically preserve the cookie.
+    // This means it will not be cleared before the NEXT test starts.
+    Cypress.Cookies.preserveOnce(COOKIE_NAME)
+  })
+
+  it('Home page should have login button', () => {
+    cy.visit(CMS_BASEURL)
+    cy.get('button').should('have.text', LOGIN_BUTTON_TEXT)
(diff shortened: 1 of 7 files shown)


cypress/integration/homepage.spec.js
```diff
@@ -0,0 +1,25 @@
+describe('Login flow', () => {
+  const CMS_BASEURL = Cypress.env('BASEURL')
+  const COOKIE_NAME = Cypress.env('COOKIE_NAME')
```
gweiying, line 3: nit: unused variable. 

As a suggestion (for a separate issue), the first test is failing right now (correctly) because visiting `/` pings the `whoami` endpoint, which sends a 401 since there's no cookie saved. Since we don't handle the 401, this throws an error and fails the test. As a workaround, we can handle the 401 in a try-catch to prevent this test from failing.

gweiying (approved):
lgtm, just a small comment below

cypress/integration/homepage.spec.js
prestonlimlianjie: Good catch on the 401 error - this should be fixed on the routing front (i.e. the LoginContext should not throw a 401 error on the frontend if the user is not logged in on the homepage - because that is the expected behavior anyway). I'll make a separate PR for the fix and rebase this testing PR on top of that.

prestonlimlianjie: So I took a closer look at this problem.

A possible solution is to get the backend to return `200` for all `whoami` calls
- with no cookie: `userId` value of `null`
- with incorrect cookie: `userId` value of `null`
- with correct cookie: `userId` value of `<USER_ID>`

I'll create a separate issue and PR. Will merge this in the meantime.

prestonlimlianjie: Created issue https://github.com/isomerpages/isomercms-backend/issues/170

PR #446, opened 2021-05-07
fix: third nav title on page creation

alexanderleegs:
This PR fixes a bug involving the creation of collection pages. Currently, when we create a new collection page, the third nav title placed into the front matter is unslugified, meaning that the display on the actual site also shows the unslugified version. This can result in 2 separate third nav categories:
Screenshot 2021-05-04 at 5 46 43 PM

To fix this, we unslugify the third-nav-title of the frontmatter on page creation.

Note that this only affects third nav page creation, as page movement correctly unslugifies the third nav title.


gweiying (approved):
lgtm!

PR #454, opened 2021-05-11
[Cleanup] remove unused utils

gweiying:
This PR removes unused util functions that are mostly related to (but not strictly) the now-unused collection pages naming structure. This PR also removes an unused route `"/sites/:siteName/files/:fileName"`, and the unused files `EditImage`, `EditFile`.


gweiying:
> There's a few more references to EditFile inside our `RouteSelector` test that we could also get rid of now: lines 625-654 make references to this route.

Good catch, addressed in e9009e7

alexanderleegs (approved):
lgtm!

PR #458, opened 2021-05-12
Fix/resource name check

alexanderleegs:
This PR fixes a bug in which duplicate resource names were not being checked properly, resolving https://github.com/isomerpages/isomercms-frontend/issues/455.

Previously, we compared our resource name against a prettified version of the original resource title, meaning that only those matching the casing of the prettified title would be identified as a duplicate. We now slugify both the new title and the list of other resource file titles before running the duplicate check.

In addition, this PR also disables the save button for resource files if no changes have been made.


gweiying (approved):
lgtm!

PR #459, opened 2021-05-12
copy: add text to indicate that sections are draggable

gweiying:
This PR adds copy text to indicate that Homepage, NavBar and Contact Us sections are draggable.

### DIFF
--- a/src/components/contact-us/Section.jsx
+++ b/src/components/contact-us/Section.jsx
@@ -45,7 +45,7 @@ const EditorSection = ({
               >
                 {cards && cards.length > 0 && (
                   <>
-                    <div className={contentStyles.segment}>
+                    <div className={`${contentStyles.segment} mb-0`}>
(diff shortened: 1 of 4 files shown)


gweiying:
From @NatMaeTan:
> From the usability tests from last year, people managed to figure out on their own that they could drag items around to rearrange them because of the grab cursor. We should save that pro tip space for other important info, also because the UI of the cards are the same across all the pages, once they figure it out for one page they should understand that the behaviour is the same for the others.

Closing since this seems to be a non-issue for now!

gweiying:
From @NatMaeTan:

> From the usability tests from last year, people managed to figure out on their own that they could drag items around to rearrange them because of the grab cursor. We should save that pro tip space for other important info, also because the UI of the cards are the same across all the pages, once they figure it out for one page they should understand that the behaviour is the same for the others.

Closing since this seems to be a non-issue for now!

PR #467, opened 2021-05-14
feat: add Pages tests

prestonlimlianjie:
This PR adds end-to-end tests for the Pages flow.

Note: this PR does not have the following two tests:
- Edit page (will be tackled in a general Editor test suite)
- Rearrange pages and rearrange folders (we will need to do more research on how to run drag-and-drop operations in Cypress)

---
### Create page, delete page, edit page settings in Workspace
- [ ] Test site workspace page should have unlinked pages section
- [ ] Should be able to create a new page with valid title and permalink
- [ ] Should not be able to create page with invalid title or permalink
- [ ] Should be able to edit existing page details with valid title and permalink
- [ ] Should be able to delete existing page on workspace

### Create folder/subfolder, rename folder/subfolder, delete folder/subfolder
- [ ] Test site workspace page should have folders section
- [ ] Should be able to create a new folder with valid folder name with no pages
- [ ] Should be able to create a new folder with valid folder name with page
- [ ] Should not be able to create a new folder with invalid folder name
- [ ] Should be able to create a new sub-folder within a valid folder name with no pages
- [ ] Should be able to create a new sub-folder within a valid folder name with one page
- [ ] Should not be able to create a new sub-folder with invalid sub-folder name
- [ ] Should not be able to create a nested sub-folder within a sub-folder
- [ ] Should be able to rename a sub-folder
- [ ] Should be able to rename a folder
- [ ] Should be able to delete a sub-folder
- [ ] Should be able to delete a folder

### Create page, delete page, edit page settings in folder
- [ ] Should be able to create a new page with valid title and permalink
- [ ] Should not be able to create page with invalid title or permalink
- [ ] Should be able to edit existing page details with valid title and permalink
- [ ] Should be able to delete existing page in folder

### Create page, delete page, edit page settings in subfolder
- [ ] Should be able to create a new page with valid title and permalink
- [ ] Should not be able to create page with invalid title or permalink
- [ ] Should be able to edit existing page details with valid title and permalink
- [ ] Should be able to delete existing page in subfolder

### Move pages
- [ ] Should be able to move page from workspace to folder
- [ ] Should be able to move page from workspace to subfolder
- [ ] Should be able to move pages from subfolder to workspace
- [ ] Should be able to move pages from folder to workspace

### DIFF
--- a/cypress/integration/pages.spec.js
+++ b/cypress/integration/pages.spec.js
@@ -402,13 +402,13 @@ describe('Pages flow', () => {
       cy.contains(EDITED_PRETTIFIED_PAGE_TITLE)
     })
 
-    it('Should be able to delete existing page on workspace', () => {
+    it('Should be able to delete existing page in folder', () => {
       // User should be able to remove the created test page card
       cy.contains(EDITED_PRETTIFIED_PAGE_TITLE)
       cy.get('.bx-dots-vertical-rounded').parent().click()
       cy.get('div[id^=delete-]').first().click()
       cy.contains('button', 'Delete').click()
-      cy.wait(4000)
+      cy.wait(8000)
 
       cy.contains(EDITED_PRETTIFIED_PAGE_TITLE).should('not.exist')
     })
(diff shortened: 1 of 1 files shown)


cypress/integration/pages.spec.js
```diff
@@ -0,0 +1,639 @@
...
+      // Reset the page title to previous title
+      editedTestPageCard.children().within(() => cy.get('[id^=settings-]').click())
+      cy.get('div[id^=settings-]').first().click() // .first() is necessary because the get returns multiple elements (refer to MenuDropdown.jsx)
+      cy.get('#title').clear().type(TEST_PAGE_TITLE)
+      cy.contains('button', 'Save').click()
+      cy.wait(4000)
+
+      // Page title should be reset in the Workspace
+      cy.contains(PRETTIFIED_PAGE_TITLE)
+      cy.wait(4000) // Wait needed because the next test fails otherwise - the rename call takes some time to propagate
```
kwajiehao, line 120: If we're using `cy.wait()`  pretty often, should we set the default timeout to be a larger value? Part of the appeal of using Cypress is to avoid having to code for these edge cases (reference: [Cypress docs](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress#Cypress-is-Not-Like-jQuery)]). Let me know if I'm missing something

prestonlimlianjie: @kwajiehao I read a little more about using `cy.wait()` and it turns out to not be good practice. This is because as you've pointed out, if the wait time is too short, the tests are likely to flake. But on the other hand, if the wait time is too long, the time to run the entire test suite increases significantly.

Instead, I've made a change in 
b8fae3d to do the equivalent of wait by using `cy.should('exist')` with a very long timeout. `.should()` will keep retrying until it hits the timeout or until the assertion is met ([docs](https://docs.cypress.io/api/commands/should#Timeouts)). This way, if the awaited step completes early, the test will move on.

alexanderleegs (approved):
lgtm!

PR #468, opened 2021-05-17
Fix: disable save button for folder modal on error

alexanderleegs:
This PR fixes a minor bug where users could still save even if the renamed folder title was invalid.


gweiying (approved):
lgtm!

PR #473, opened 2021-05-18
Fix images in private repos not displaying in CMS

rlyss:
Make a call to the backend API to retrieve the b64 encoded image blob if
the link to the raw file from GitHub cannot be found. Converts image
blob to blob url to be set as image src in the MediaSettingsModal,
MediaCard and EditPage components. Removes cached image urls upon image
deletion and image update.


src/utils.js
```diff
@@ -310,4 +343,43 @@ export const convertSubfolderArray = (folderOrderArray, rawFolderContents, subfo
...
+    byteArrays.push(byteArray);
+  }
+
+  const blob = new Blob(byteArrays, {type: contentType});
+  return blob;
+}
+
+export async function fetchImageURL(siteName, filePath) {
+  var rawURL = `https://raw.githubusercontent.com/isomerpages/${siteName}/staging/${filePath}${filePath.endsWith('.svg') ? '?sanitize=true' : ''}`
+  const response = await fetch(rawURL);
```
alexanderleegs, line 370: Attempting to retrieve the image here requires an extra call, which might not be very efficient - we could look into a one off check if the repo is public or private and run the blob generation code if the repo is private.

alexanderleegs: Also for future reference, we mostly use axios, so we should try to standardise our request library as far as possible

src/utils.js
```diff
@@ -310,4 +343,43 @@ export const convertSubfolderArray = (folderOrderArray, rawFolderContents, subfo
...
+  //If the image is public, return the link to the raw file, otherwise make a call to the backend API to retrieve the image blob
+  if (response.ok) {
+    return rawURL;
+  } else {
+    const fileName = filePath.slice(filePath.lastIndexOf('/')+1)
+    const customPath = filePath.slice(filePath.indexOf('/')+1, filePath.lastIndexOf('/')+1).replaceAll('/','')
```
alexanderleegs, line 376: This section could be broken up - the use of these indexes might be difficult to understand without comments or intermediate variables

src/utils.js
```diff
@@ -59,18 +61,49 @@ export function isLinkInternal(url) {
...
+//     // check for whether the original image source is from within Github or outside of Github
+//     // only modify URL if it's a permalink on the website
+//     if (isLinkInternal($(elem).attr('src'))) {
+//       $(elem).attr('src', `https://raw.githubusercontent.com/isomerpages/${repoName}/staging${$(elem).attr('src')}?raw=true`);
+//     }
+//     // change src to placeholder image if images not found
+//     $(elem).attr('onerror', "this.onerror=null; this.src='/placeholder_no_image.png';")
+//   });
+//   return $.html();
+// }
```
alexanderleegs, line 76: We can remove the commented out code here

src/utils.js
```diff
@@ -59,18 +61,49 @@ export function isLinkInternal(url) {
...
+export async function prependImageSrc(repoName, chunk) {
+  const $ = cheerio.load(chunk)
+  const imagePromises = []
+  const elementsToUpdate = []
+  const queryClient = new QueryClient({
+    defaultOptions: {
+      queries: {
+        staleTime: 1000 * 60 // 60 seconds
+      }
+    }
```
alexanderleegs, line 87: Could we use the existing queryclient `useQueryClient` instead?

src/layouts/EditPage.jsx
```diff
@@ -280,12 +280,17 @@ const EditPage = ({ match, isResourcePage, isCollectionPage, history }) => {
...
+
+  async function loadChunk() {
     const html = marked(editorValue)
     const { isCspViolation: checkedIsCspViolation, sanitisedHtml: processedSanitisedHtml } = checkCSP(csp, html)
-    const processedChunk = prependImageSrc(siteName, processedSanitisedHtml)
+    const processedChunk = await prependImageSrc(siteName, processedSanitisedHtml)
     setIsCspViolation(checkedIsCspViolation)
     setChunk(processedChunk)
-  }, [editorValue])
+  }
```
alexanderleegs, line 292: We can move this chunk into the useEffect, so it's clear that it's only used in that section

alexanderleegs (changes requested):
Good work on this feature - the functionality works great, even though the feature was quite a lot more complicated than we originally thought it would be

I've left a few comments mostly on styling - it would also be good if you could leave spaces between operators for better readability. We're planning to introduce linting in the future to standardise coding style as well

src/utils.js
```diff
@@ -310,4 +343,43 @@ export const convertSubfolderArray = (folderOrderArray, rawFolderContents, subfo
...
+    const byteArray = new Uint8Array(byteNumbers);
+    byteArrays.push(byteArray);
+  }
+
+  const blob = new Blob(byteArrays, {type: contentType});
+  return blob;
+}
+
+export async function fetchImageURL(siteName, filePath) {
+  var rawURL = `https://raw.githubusercontent.com/isomerpages/${siteName}/staging/${filePath}${filePath.endsWith('.svg') ? '?sanitize=true' : ''}`
```
prestonlimlianjie, line 369: We should use bias towards using `let` since it is block-scoped while `var` is function-scoped

src/utils.js
```diff
@@ -59,18 +61,49 @@ export function isLinkInternal(url) {
...
+    }
+  })
+
   $('img').each((i, elem) => {
     // check for whether the original image source is from within Github or outside of Github
     // only modify URL if it's a permalink on the website
     if (isLinkInternal($(elem).attr('src'))) {
-      $(elem).attr('src', `https://raw.githubusercontent.com/isomerpages/${repoName}/staging${$(elem).attr('src')}?raw=true`);
+      const filePath = $(elem).attr('src').substring(1) // remove leading '/'
+      imagePromises.push(queryClient.fetchQuery(filePath, () => fetchImageURL(repoName, filePath)))
```
prestonlimlianjie, line 95: We should break this into two lines so it is more intuitive to the reader
```
const imagePromise = queryClient.fetchQuery(filePath, () => fetchImageURL(repoName, filePath))
imagePromises.push(imagePromise)
```

src/utils.js
```diff
@@ -59,18 +61,49 @@ export function isLinkInternal(url) {
...
+      elementsToUpdate.push(elem)
     }
     // change src to placeholder image if images not found
-    $(elem).attr('onerror', "this.onerror=null; this.src='/placeholder_no_image.png';") 
+    $(elem).attr('onerror', "this.onerror=null; this.src='/placeholder_no_image.png';")
   });
-  return $.html();
+
+  const imageURLs = await Promise.allSettled(imagePromises)
+  for (var i = 0; i<imageURLs.length; i++) {
```
prestonlimlianjie, line 103: We should iterate over the `elementsToUpdate` array using the `.forEach` method so that it is more semantically meaningful.

That is, a reader immediately knows that the intention is "to iterate over the `elementsToUpdate` array", as opposed to "to iterate over an array of equal length to the `elementsToUpdate` array"

(We tend to bias towards using `.map` and `.forEach` array methods as opposed to the general iterator methods because they provide extra context/semantic meaning to the code)

src/utils.js
rlyss: Hooks can't be used outside of component function bodies apparently so it throws an error when `useQueryClient` is called in utils. I've stuck with creating a new `QueryClient` instance for now

src/utils.js
```diff
@@ -310,4 +330,56 @@ export const convertSubfolderArray = (folderOrderArray, rawFolderContents, subfo
...
+ *    The leading '/' is optional. The filePath parameter should be URI decoded. Examples:
+ *    images/test-folder/image sample.png
+ *    /images/test.svg
+ *    /images/folder 1/folder2/folder3/names.jpg
+ * @returns {Promise<string>}
+ */
+export async function fetchImageURL(siteName, filePath) {
+  const cleanPath = filePath.replace(/^\//, '') //Remove leading / if it exists e.g. /images/example.png -> images/example.png
+  //If the image is public, return the link to the raw file, otherwise make a call to the backend API to retrieve the image blob
+  if (!window.sitesIsPrivate[siteName]) {
```
prestonlimlianjie, line 371: Should use `localStorage.getItem()` to retrieve the map. 

Also let's remove this storage item when the user logs out in `LoginContext.jsx` (`logout` function)

src/layouts/Sites.jsx
```diff
@@ -22,6 +22,12 @@ export default class Sites extends Component {
         withCredentials: true,
       });
       const { siteNames } = resp.data;
+
+      window.sitesIsPrivate = siteNames.reduce((map, siteNames) => {
+        map[siteNames.repoName] = siteNames.isPrivate
+        return map
+      })
```
prestonlimlianjie, line 29: I took a look at the value of `window.sitesIsPrivate` in the browser console and discovered that there were some unexpected fields in the map. We see `lastUpdated`, `permissions`, `repoName`, `isPrivate` as fields of the map, when we should only be seeing `<repoName>` as the fields.

Additionally, I couldn't find the value that corresponded to the `a-test` repo.

Screenshot 2021-05-21 at 4 11 51 PM

The main reason this is happening is because we currently don't provide the `Array.reduce` method with an `initialValue`. And if we don't provide `initialValue`, `the first element in the array will be used as the initial accumulator value and skipped as currentValue` [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce).

As a result, `map` is instantialized as 
```
{
  "lastUpdated": "Updated 1 day ago",
  "permissions": {
      "admin": true,
      "push": true,
      "pull": true
  },
  "repoName": "a-test",
  "isPrivate": false
}
```

And subsequent iterations will continue adding to `map` as such
```
{
  "lastUpdated": "Updated 1 day ago",
  "permissions": {
      "admin": true,
      "push": true,
      "pull": true
  },
  "repoName": "a-test",
  "isPrivate": false,
  "a-test-v4": "true",
  // ... rest of sites
}
```

The fix is to provide the `reduce` function with an `initialValue` of `{}`, i.e. line 29 should be:
```
}, {})
```

prestonlimlianjie (changes requested):
Some changes required - but apart from these, the rest of the code looks good

prestonlimlianjie (approved):
lgtm

src/components/media/MediaCard.jsx
```diff
@@ -1,10 +1,19 @@
...
 
 const MediaCard = ({
   type, siteName, onClick, media, isSelected, canShowEditIcon,
-}) => (
+}) => {
+
+  const {data: imageURL, status} = useQuery(media.path, () => fetchImageURL(siteName, media.path), {
+    refetchOnWindowFocus: false,
+    staleTime: 1000 * 60 // 60 seconds
+  })
```
alexanderleegs, line 14: Could we specify this query to only run on image type MediaCards? Currently we make a lot of failed calls to the backend for other types of media

alexanderleegs (changes requested):

alexanderleegs (commented):
Also, there's a very minor bug where is users are able to bypass the `/sites` page completely, their images will not load, regardless of whether the repo is a private or public one. I think this is a relatively minor bug that isn't very urgent - we can fix in a future pr

src/components/media/MediaCard.jsx
rlyss: Added a `shouldLoad` parameter to the `fetchImageURL` method. Tried setting `enabled: type === 'images'` in the `useQuery` options but the MediaCards in Files somehow still trigger failed `fetchImageURL` calls after navigating from Files to Images.

alexanderleegs (approved):
Great work, lgtm!

PR #474, opened 2021-05-18
fix: duplicate resource title check for example_title

gweiying:
This bug fixes a small bug where upon file creation with title "Example Title", "Example Title" should be automatically postpended with "_1" if the title already exists. 

The bug was caused by checking for the uniqueness of the full resource post filename `{date}-{filetype}-{name}.md` as opposed to the uniqueness of just the `name` as we've now decided upon.


alexanderleegs (approved):
lgtm

PR #478, opened 2021-05-21
Feat/front matter changes for resources and pages

### LINKED ISSUE #476: [bug] layout type not set for resource pages
Currently, newly created resource pages use the same template as regular collection pages, instead of having secondary colours, due to the layout not being set for resource pages. We should include this in the front matter when creating new pages.

### LINKED ISSUE #477: [bug] Title in page details should be taken from front matter
Currently, when we open up the settings modal for pages, we retrieve the title of the file from its name on github, instead of the title in the file's front matter. The difference in the two is that the title in the front matter can contain special characters not allowed in the name of a file, while the name on github is the slugified version of the front matter title.

As we are already reading the file details when we open up the settings modal, we should use the title in the front matter instead, to prevent users from unintentionally modifying the title of the file when editing the file details.

alexanderleegs:
This PR introduces several changes to resource file handling and page settings.

- Previously, newly created resource post pages on the cms were using the same layout as those for pages - this was due to us not setting the layout properly for newly created resource pages. This PR adds in the additional layout field.
- The layout field is also enforced now when changing any resource page settings, regardless of its previous value
- The default permalink for newly created resource pages now also includes the resource room, to bring it closer to the way users access the page on the actual site (resource room -> filter by resource category -> click on post)
- The page title for existing pages is now retrieved from the frontmatter, instead of the file name - this is to prevent unintentional changes to the title when editing other portions of the settings, since the file name on github does not contain special characters while the title in the front matter can.

It resolves #476 and resolves #477.

### DIFF
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -111,8 +111,8 @@ const ComponentSettingsModal = ({
         if (pageData !== undefined) { // is existing page
           const { pageContent, pageSha } = pageData
           const { frontMatter, mdBody: pageMdBody } = frontMatterParser(pageContent)
-          const { file_url: originalFileUrl, permalink: originalPermalink } = frontMatter
-          const { title: originalTitle, type: originalType, date: originalDate } = retrieveResourceFileMetadata(fileName)
+          const { title: originalTitle, file_url: originalFileUrl, permalink: originalPermalink } = frontMatter
+          const { type: originalType, date: originalDate } = retrieveResourceFileMetadata(fileName)
           if (_isMounted) {
             setSha(pageSha)
             setMdBody(pageMdBody)
--- a/src/components/ComponentSettingsModal.jsx
+++ b/src/components/ComponentSettingsModal.jsx
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from 'react';
-import { useMutation, useQueryClient } from 'react-query';
+import { useMutation, useQuery, useQueryClient } from 'react-query';
 import axios from 'axios';
 import PropTypes from 'prop-types';
 import * as _ from 'lodash';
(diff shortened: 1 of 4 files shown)


src/components/ComponentSettingsModal.jsx
```diff
@@ -178,6 +178,7 @@ const ComponentSettingsModal = ({
         const frontMatter = isPost 
           ? { ...originalFrontMatter, title, date: resourceDate, permalink }
           : { ...originalFrontMatter, title, date: resourceDate, file_url: fileUrl }
+        if (isNewFile && isPost) frontMatter.layout = 'post'
```
gweiying, line 198: Just to keep things clean, can we subsume this line into line 179, as in
`{ ...originalFrontMatter, title, date: resourceDate, permalink, layout: post }`

src/components/ComponentSettingsModal.jsx
```diff
@@ -87,9 +88,25 @@ const ComponentSettingsModal = ({
         fileUrl: setFileUrl,
     }
 
+    useQuery(
+      [RESOURCE_ROOM_NAME_KEY, siteName],
+      () => getResourceRoomName(siteName),
+      {
+        retry: false,
+        onSuccess: (retrievedName) => {
+          setResourceRoomName(retrievedName)
```
gweiying, line 97: nit, what do you think of unpacking the data from the useQuery, as in:
`const { data: resourceRoomName } = useQuery(...)` so that we don't need a state hook?

gweiying (changes requested):
thanks for investigating the issues causing bugs! nice work, I've just added some small nits for cleanup but mostly looks good to me

src/components/ComponentSettingsModal.jsx
alexanderleegs: Great suggestion, I've made the change in https://github.com/isomerpages/isomercms-frontend/pull/478/commits/c008d3f9be261d6048df85984c322193c7676767

src/components/ComponentSettingsModal.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/478/commits/e31ae39d3045482f94fe131d11cd484b7ad627ee. As discussed offline, we should also check if users use custom layouts in resources, since this would replace their custom layout when editing page settings.

gweiying (approved):
lgtm

PR #419, opened 2021-04-19
chore: add husky and commitlint

prestonlimlianjie:
This PR adds the husky and commitlint libraries - please refer to the backend PR for more details https://github.com/isomerpages/isomercms-backend/pull/155

PR #509, opened 2021-06-08
Chore/precommit hook

gweiying:
This PR rebases #419 and adds a pre-commit hook to run eslint and prettier before commit.


alexanderleegs (commented):
I'm thinking it might be better to install and run lint-staged on pre-commit instead - so we don't have to run the linting on every single file each time we make a commit

gweiying:
Closing, superseded by #518

PR #514, opened 2021-06-11
chore: set npm run dev 

prestonlimlianjie:
This PR adds `npm run dev` as a command to make running the CMS frontend consistent with that of the backend. The PR also sets the local env var file to `.env` to match that in the backend repo.


kwajiehao (approved):
lgtm

Issue #494, opened 2021-05-31
[bug] future-dated resource posts should not be supported


gweiying:
Since we do not have a regular cron job to rebuild sites regularly, we should not allow future-dated resource posts on the CMS. To resolve this, we will disable saving when users choose a date in the future for their resource posts.

gweiying:
Completed, closing

![Screenshot 2022-02-08 at 10 30 14 AM](https://user-images.githubusercontent.com/39231249/152906561-c962eac1-82a2-4ba2-aa96-e3188b0ee3a3.png)

PR #516, opened 2021-06-16
fix: validates that resource page dates are not in future

gweiying:
This PR addresses #494.

### Problem
Currently, users are able to choose a resource date that's in the future. Since we currently do not have a cron job running to build sites regularly, we cannot trigger a rebuild when these resource pages are supposed to appear, hence the posts will not appear when users expect them to. 

### Fix
For now, we have decided to disable future resource posts, and explore supporting this in the future.


prestonlimlianjie (approved):
lgtm!

PR #520, opened 2021-06-17
Revert "style: introduce aliasing, sort imports (#511)"

kwajiehao:
## Overview
This PR reverts PR #511, or commit 17276df88446a665832cfb18781747d3cd0e054b.

We are reverting PR #511 because we found out that the outputs of `craco start` and `craco build` were different, resulting in CSS files being imported in an order different to what is intended.

The workarounds we found online were convoluted and would introduce unnecessary complexity to the repo. Furthermore, we also found a library (react-app-rewire-alias) which can help us achieve aliasing with CRA as an alternative to Craco. In addition to the bug associated with Craco here, Craco is also overkill since it's a powerful tool which helps users to extend their default CRA configs for webpack among other things.

As such, we are reverting this commit and will be attempting to reintroduce aliases at a later date.

A sample of how to use react-app-rewire-alias can be found in the vaccinegovsg repo https://github.com/datagovsg/vaccinegovsg-clinic/blob/dev/frontend/config-overrides.js

References:
- https://github.com/gsoft-inc/craco/issues/57
- https://github.com/DocSpring/craco-antd/issues/12


src/components/HyperlinkModal.jsx
```diff
@@ -20,7 +17,7 @@ export default class HyperlinkModal extends Component {
     }
   }
 
-  changeHandler = (event) => {
+  changeHandler(event) {
```
prestonlimlianjie, line 20: We should switch back to arrow function definition because it helps us bind the `changeHandler` to this automatically.

If we wanted to do away with the arrow function definition, we would have to explicitly write `this.handleClick = this.handleClick.bind(this)`

kwajiehao: fixed with [efcbefc](https://github.com/isomerpages/isomercms-frontend/pull/520/commits/efcbefc66441fffbe4d3e055fae809422aa7a386)

prestonlimlianjie (approved):
lgtm!

PR #521, opened 2021-06-17
test: cleanup e2e tests

kwajiehao:
# Overview
This PR patches some of our failing and flaky E2E tests. It fixes the following tests:
1. Failing delete test for `EditPage`
2. Failing delete tests using the `Folders` and `Media` layouts
3. Failing test for subfolder renaming
4. Flaky file-moving breadcrumbs

Other misc. changes:
- Eliminate extraneous `id` attribute in `DeleteWarningModal`

In the course of testing this PR, issue #526 (cleanup of uploaded assets during testing) was also filed, to be resolved in a subsequent PR.

## 1. Failing delete test for `EditPage`
**Problem**
Previously, the "Edit page should allow user to delete page" test failed because Cypress was unable to click on the delete button in the `DeleteWarningModal`. I suspect that this is because the modal takes some time to appear in the DOM.

**Solution**
To resolve this problem, we wait for the `DeleteWarningModal` to come into existence by using a "should" assertion before attempting to click on it.

## 2. Failing delete tests using the `Folders` and `Media` layouts
**Problem**
Even though the snapshots show images/files being deleted successfully, our subsequent assertions (like `cy.contains(IMAGE_TITLE).should("not.exist")`) fail. This is because `ReactQueryDevTools` maintains a reference to the delete API call and so Cypress can still find a reference to the deleted file. Refer to the image below for details:

Screenshot 2021-06-17 at 8 49 41 PM

**Solution**
~We have a few options:~
- ~When running tests, ensure that the `REACT_APP_ENV` environment variable is set to a value that is NOT `LOCAL_ENV`, so that `ReactQueryDevTools` isn't activated.~
- ~Remove `ReactQueryDevTools` from our codebase since we aren't using it much at this point~

~Either approach works for me - please let me know what you think.~

After some discussion, we will ensure that the `REACT_APP_ENV` environment variable is set to a value that is NOT `LOCAL_ENV`, so that `ReactQueryDevTools` isn't activated when running tests.

## 3. Failing test for `"Should be able to rename a sub-folder"`
**Problem**
Previously, we attempted to assert using a custom timeout, but the test failed to find the renamed subfolder despite the successful HTTP request. 

```
      cy.contains("1 item", { timeout: CUSTOM_TIMEOUT }).should("exist")
      cy.contains(PRETTIFIED_EDITED_SUBFOLDER_WITH_PAGES_TITLE).click() // test fails here
```

This is because the UI takes some time to update and the second contain requires a longer timeout as well.

**Solution**
Instead of waiting for a longer time with `CUSTOM_TIMEOUT`, if we intercept the subfolder renaming request and wait for it, the test passes successfully.

This means we can probably also replace the use of `CUSTOM_TIMEOUT` with interceptors throughout our test suite so that our tests can fail more quickly (#524). This is because the slowest actions in the CMS come from network calls, and DOM changes should not take more than the default of 4 seconds set by Cypress .

## 4. Flaky breadcrumb elements in `FileMoveMenuDropdown` component
**Problem**
Cypress sometimes complains that the breadcrumb element is detached from the DOM, despite already asserting that the element is visible, or exists.

**Solution**
We use the `cypress-pipe` library (on recommendation from the Cypress team) to help to click on the correct DOM element when we're trying to move a file from one directory to another.

Previously, Cypress would complain that the breadcrumb element was detached from the DOM, despite already asserting for visibility. This is a case where the app is slow to react and Cypress was fast to act: see [this blog post](https://www.cypress.io/blog/2019/01/22/when-can-the-test-click/) from the Cypress team which addresses the issue. The problem is probably that the DOM element has been created but the event listeners have not been set up.

`cypress-pipe` solves this problem by continuously retrying until “the assertions that follow pass”.

### DIFF
--- a/cypress/integration/editPage.spec.js
+++ b/cypress/integration/editPage.spec.js
@@ -187,10 +187,12 @@ describe("Edit unlinked page", () => {
 
     // Cancel works properly in modal
     cy.get("#modal-cancel").click()
+    cy.get("#modal-cancel").should("not.exist")
 
     cy.contains(":button", "Delete").click()
 
     // Test delete in modal
+    cy.get("#modal-delete").should("exist")
     cy.get("#modal-delete").click()
 
     // Assert: page no longer exists
--- a/cypress/integration/editPage.spec.js
+++ b/cypress/integration/editPage.spec.js
@@ -27,7 +27,7 @@ describe("Edit unlinked page", () => {
     TEST_UNLINKED_PAGE_TITLE
   )}.md`
 
-  const DEFAULT_IMAGE_TITLE = "hero-banner.png"
+  const DEFAULT_IMAGE_TITLE = "isomer-logo.svg"
   const ADDED_IMAGE_TITLE = "balloon.png"
   const ADDED_IMAGE_PATH = "images/balloon.png"
 
--- a/cypress/integration/files.spec.js
+++ b/cypress/integration/files.spec.js
@@ -363,7 +363,12 @@ describe("Files", () => {
       cy.visit(`/sites/${TEST_REPO_NAME}/documents`)
       cy.contains(FILE_TITLE) // file should be contained in directory
 
-      cy.deleteMedia(FILE_TITLE) // cleanup
+      cy.intercept({
+        method: "DELETE",
+        url: `/v1/sites/${TEST_REPO_NAME}/documents/${FILE_TITLE}`,
+      }).as("deleteFile")
+      cy.deleteMedia(FILE_TITLE)
+      cy.wait("@deleteFile")
     })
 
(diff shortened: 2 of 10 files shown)


alexanderleegs (commented):
The images tests might require the same fixes as the files tests - I'm still receiving errors when attempting to move images. Other than that, the tests look great!

kwajiehao:
> The images tests might require the same fixes as the files tests - I'm still receiving errors when attempting to move images. Other than that, the tests look great!

Good catch, fixed with [b107465](https://github.com/isomerpages/isomercms-frontend/pull/521/commits/b107465311015d1ebacfb9c7b30635029a953637)

alexanderleegs (approved):
lgtm!

PR #522, opened 2021-06-18
fix: oauth flow

kwajiehao:
To be reviewed with https://github.com/isomerpages/isomercms-backend/pull/206


lamkeewei (approved):
lgtm! If `REACT_APP_CLIENT_ID` is not used elsewhere could be good to remove it as an env var also.

prestonlimlianjie (approved):
lgtm

PR #525, opened 2021-06-21
test: reset e2e repo state

kwajiehao:
## Overview

This PR uses a node script to reset the `e2e-test-repo` state before running our e2e tests so that we don't have to do it manually.

As it stands, the repo is reset once before all e2e tests are run. This should be sufficient since our tests should not be dependent on side effects to pass (i.e. we shouldn't rely on the outputs of one of our tests for a subsequent test to pass).


alexanderleegs (approved):
lgtm!

PR #537, opened 2021-07-14
fix: update logout to use delete

gweiying:
This addresses [isomer-product/issues/12](https://github.com/isomerpages/isomer-product/issues/12). To be reviewed with backend PR [isomercms-backend/pull/217](https://github.com/isomerpages/isomercms-backend/pull/217).


gweiying:
Converted to draft - as @alexanderleegs pointed out, we need to handle the case where the token naturally expires but the local storage values are not automatically cleared.

alexanderleegs (approved):
lgtm!

PR #538, opened 2021-07-15
Vapt/xss

gweiying:
This PR addresses [isomer-product/issues/7](https://github.com/isomerpages/isomer-product/issues/7).

The following fixes were implemented in this PR:
- Upgrading `DOMPurify`, installing `@braintree/sanitize-url`
- Sanitize before preview:
  - Sanitizing user input before EditPage preview
  - Sanitizing user input for URLs before EditContactUs preview
  - Disabling URL click-throughs in EditContactUs preview (to standardize with behavior on the rest of the sites)
- Sanitizing user input before Save:
  - Sanitizing user input upon Save for Edit Page 
 

However, this means that the behavior on our site will be slightly inconsistent as we will not be sanitising user input on other pages before saving the pages to Github. This has no security implications on Isomer sites as we do not enable inline scripts on Isomer sites.

To standardize the behavior across the site, we should create a reusable component that will help with sanitising user input before Saving. This will be tackled as part of our refactor and is documented here [isomer-product/issues/20](https://github.com/isomerpages/isomer-product/issues/20). 

This PR addresses GTA-03-002 WP1 and GTA-03-014 WP1.

### DIFF
--- a/.env-example
+++ b/.env-example
@@ -9,3 +9,8 @@ export CYPRESS_BASEURL=''
 export CYPRESS_COOKIE_NAME=''
 export CYPRESS_COOKIE_VALUE=''
 export CYPRESS_TEST_REPO_NAME=''
+
+# Reset e2e-test-repo
+export E2E_COMMIT_HASH=6800d5c2bf40d010e8b2f008ee6abc43df0cd4a2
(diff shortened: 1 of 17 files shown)


alexanderleegs (approved):
lgtm!

PR #560, opened 2021-07-31
fix: dependency upgrades

prestonlimlianjie:
This PR upgrades vulnerable dependencies:
- the majority of deps were upgraded using `npm audit fix`
- `node-sass` upgraded to 6.0.1 using `npm install node-sass@6.0.1`
- `marked` upgraded to 2.1.3 using `npm install marked@2.1.3`

PR #569, opened 2021-08-02
Fix/dependency upgrades except node sass

alexanderleegs:
This PR is a duplicate of https://github.com/isomerpages/isomercms-frontend/pull/560 minus the upgrade of node-sass, as it was causing issues with the netlify build.

In addition, react-scripts and node-sass have been moved into devDependencies to prevent them from being flagged as an issue by snyk.


prestonlimlianjie (approved):
lgtm

PR #572, opened 2021-08-03
Fix: adjust simple mde styling

alexanderleegs:
This PR fixes a display issue with the editor section in editPage causing by the bumping of easymde from 2.9.0 to 2.15.0.


prestonlimlianjie (approved):
lgtm - thanks for investigating!

PR #575, opened 2021-08-04
feat: validate and sanitize media files

prestonlimlianjie:
This PR fixes https://github.com/isomerpages/isomer-product/issues/27. It specifically does the following:
- update the `<input accept="<ACCEPTED INPUT>">` file types to match the file extensions explained in the issue
- sets the file restriction message for our admin users


alexanderleegs (approved):
lgtm!

PR #599, opened 2021-08-17
Refactor/service class conversion

gweiying:
This PR is a followup to #590, introducing an abstraction service layer for API calls to the backend. 

## Background
During our refactor discussion (12 Aug), we decided to introduce a service class layer on the frontend to perform API calls to the backend. Using service classes allows us to perform dependency injections, which means that we are also able to do unit tests more easily in the future as we are able to mock the apiClient. 

To make the services accessible to the hooks, we introduce a context and provider for storing the services and making them accessible to the custom hooks.

The key changes accomplished in this PR:
- Creating the PageService class
- Creating the ServicesContext and ServicesContextProvider
- Updating the hooks to consume from the ServicesContext


prestonlimlianjie (approved):
lgtm! did we also want to delete `utils/apiEndpoints.js` since the functions are being replaced by the `PageService`?

gweiying:
> lgtm! did we also want to delete `utils/apiEndpoints.js` since the functions are being replaced by the `PageService`?

Good catch, fixed in 0d7bb2f

PR #636, opened 2021-09-23
Fix/handle 409

gweiying:
This PR catches the 409 error thrown when users try to save conflicting versions of the same file, and implements a Warning modal to confirm if users want to override the previous changes. The Warning modal is shown below.

Screenshot 2021-09-23 at 3 15 19 PM

Refer to [#isomercms-backend/pull/303/files](https://github.com/isomerpages/isomercms-backend/pull/303/files) for the corresponding backend PR.

## Problem
When 2 users edit the same file, and one user saves their changes before the second user does, the second user will not be able to save their changes as there'll be a conflict error. 
Screenshot 2021-09-23 at 12 22 26 PM

This conflict error is a 409 error thrown from Github when the second user tries to save their file, as the page `sha` that they have stored locally when they first opened the file is different from the current `sha` of the page in Github.

## Solution
We implement a Warning modal when the 409 error is caught. If the user cancels and clicks "Back to Editing", they will still get the Warning message on any further attempts to save, unless the user refreshes the page and discards their changes. If the user clicks "Override and Save", their new changes will override the existing file.

### DIFF
--- a/src/hooks/pageHooks/useUpdatePageHook.jsx
+++ b/src/hooks/pageHooks/useUpdatePageHook.jsx
@@ -15,9 +15,11 @@ export function useUpdatePageHook(params, queryParams) {
   const { pageService } = useContext(ServicesContext)
   return useMutation((body) => pageService.update(params, body), {
     ...queryParams,
-    onSuccess: () => {
+    onSettled: () => {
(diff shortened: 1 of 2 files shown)


alexanderleegs (approved):
lgtm!

PR #647, opened 2021-10-06
Fix: disable manual user input for colours

### LINKED ISSUE #646: [Enhancement] No error catching if colour code were input wrongly
If users using CMS input in the colour code wrongly, they would still be able to save and this would caused netlify to have failure built from here onwards.

We should put in error checker to prevent users from saving.

image

Putting in as P1 since it can be easily rectified (?) instead of making huge changes

alexanderleegs:
This PR resolves #646. It fixes an issue which allowed users to manually modify the site colours, which could result in the site breaking if the colours were not specified properly. The input has been disabled in this PR, such that users can only use the colour picker to enforce correct colours being chosen.

### DIFF
--- a/src/components/FormFieldColor.jsx
+++ b/src/components/FormFieldColor.jsx
@@ -26,6 +26,7 @@ const FormFieldColor = ({
         className={elementStyles.formColorInput}
         style={style}
         onChange={onFieldChange}
+        disabled
       />
       <div
         className={elementStyles.formColorBox}


gweiying:
Instead of disabling, should we validate/ throw errors if wrong? 

Just thinking of users who may consult with designers and be given specific hex values to key in.

gweiying (approved):
lgtm!

PR #618, opened 2021-09-01
refactor: backend settings v2

kwajiehao:
This PR is modifies frontend components which hit the backend GET `/{siteName}/settings` endpoint in response to changes to the settings response format made in the backend PR #[282](https://github.com/isomerpages/isomercms-backend/pull/282). 

The biggest change to note is the removal of the `footerSha` state variable in the `EditContactUs` layout. The GET `/{siteName}/settings`  endpoint pulls the existing footer file to check for differences in the request footer fields and the current repo value - in doing so, it retrieves the existing footer file sha. Thus, we no longer need to submit a sha value when updating the footer. 

Additional refactoring to bring the settings component in line with our existing refactor is in a future PR : https://github.com/isomerpages/isomercms-frontend/pull/666

PR #666, opened 2021-11-01
Refactor/settings

alexanderleegs:
This PR refactors the settings flow. Builds on and supersedes #618.

**Key Changes**
--
Settings format update:
- We have combined the config settings from the three files (config, navigation, footer) into a single object, rather than three separate objects, to simplify the flow of the settings pages. The appropriate changes to EditContactUs and GetPageColorsHook have been made to support this change.

Settings flow:

- Refactors settings to use hooks instead
- Changes all Settings React Query calls to use query hooks
- Shifts Color Picker fields out into separate component
  - The logic for the Color Picker is significantly more complicated than the other fields, which are mostly just inputs, due to the presence of keydown event listeners and the picker modal

**Other Changes**
--
- Shift settings into hooks into separate settings component folder
  - After discussion with @gweiying, we have decided to leave the axios calls in settings hooks with only `GET` operations (e.g. getSiteColors), rather than moving them to a `SettingsService`, as we feel that these services should support multiple CRUD operations.
- Fix settings tests
- Rearrange position of site preview

**Future Work**
--
Currently, there’s some repeated work in both the frontend and backend in determining which settings fields have been changed. We could potentially move this to be purely on the backend side, but the frontend still has use for a diff checker, since this is how we determine whether to overwrite the existing contents of the settings page if changes are made by another user. Other components like EditContactUs, which only need to change a single field, also have use for the backend updating only the relevant field. We have opted to leave it as is for now, as [this discussed change to the location of config files](https://github.com/isomerpages/isomercms-frontend/issues/686) would also affect this behaviour, but we can revisit this in future.


src/layouts/Settings.jsx
```diff
@@ -88,45 +92,45 @@ const CONFIG_FIELDS = [
...
+  useEffect(() => {
+    if (settingsData && !hasChanges) {
+      const {
+        configSettings,
+        footerSettings,
+        navigationSettings,
+      } = settingsData
 
-      const originalState = {
+      const retrievedState = {
```
gweiying, line 133: nit - is there a problem with using the `settingsData`'s `configSettings`, `footerSettings` and `navigationSettings` as is, instead of unpacking the values individually into new objects `otherFooterSettings`, `socialMediaContent` etc? there's quite a lot of state renaming and unpacking going on here, which makes it hard to follow what is happening to each value.

gweiying: If necessary, it could be useful to coordinate the backend data structure with the frontend so that we don't have to unpack and rename all our values multiple times.

src/components/settings/ColorPickerSection.jsx
```diff
@@ -0,0 +1,218 @@
...
+      if (event.key === "Escape") {
+        setColorPickerToggle(false)
+      }
+    },
+    [setColorPickerToggle]
+  )
+
+  // event listener callback function that resets ColorPicker modal
+  // when mouse clicks on area outside of modal while modal is active
+  const clickFunction = useCallback(
```
gweiying, line 35: Nice feature. Can I make a suggestion that we don't change the color in the background until the user actually presses "Select"? Otherwise it's slightly unintuitive to me that it doesn't "autosave", as in the color in the Settings screen is changing when I change the color in the ColorPicker modal, but when I click outside it doesn't "autosave" and instead reverts back. Maybe this should only change when the user presses "Select" maybe?

https://user-images.githubusercontent.com/39231249/141722461-df5579f7-2ebd-447f-8bfa-cbd3c6298f26.mov

rt

gweiying: @NatMaeTan thoughts?

src/components/settings/ColorPickerSection.jsx
```diff
@@ -0,0 +1,218 @@
...
+      }
+    }
+    return () => {
+      document.removeEventListener("keydown", escFunction)
+      document.removeEventListener("click", clickFunction)
+    }
+  }, [colorPickerToggle, escFunction, clickFunction])
+
+  // toggles color picker modal
+  const activateColorPicker = (event) => {
```
gweiying, line 72: Is it possible to move these handlers (`activateColorPicker`, `onColorSelect`, `setRealTimeColor`, `changeHandler` etc) into `FormFieldColor` and make each component responsible for its own update?

src/layouts/Settings.jsx
```diff
@@ -768,39 +478,62 @@ export default class Settings extends Component {
...
                   }
-                  callback={this.saveSettings}
-                />
+                )}
               </div>
+              <br />
+              <br />
+            </div>
+            <div className={elementStyles.formSave}>
+              <LoadingButton
```
gweiying, line 495: Just a small feature request, the designers/ Nat has previously requested to pin the Save button to the bottom of the Settings screen (since this is done for all the rest of our pages), would it be possible to implement this in this refactor?

gweiying (changes requested):
Nice work on the services and hooks refactor! Everything looks clean and functional.

Some comments inline about how to simplify the `ColorPicker` component and the `Settings` screen. 
- For `ColorPickerSection`, I'm wondering if some of the handlers can be moved into `FormFieldColor`, so that each component is responsible for its own update. 
- For `Settings` screen, there is alot of logic used to rename and repack the `settingsData` in the onData Load handler, change handler and save handler. I think this can be greatly simplified through some combination of rethinking the backend structuring of data response, and introducing useForm. We can introduce useForm at a later date if that's too complex, but the backend structuring of data response could be a quicker fix. Happy to discuss more offline to clarify if needed!

src/layouts/Settings.jsx
alexanderleegs: Thanks for the reminder - changed in https://github.com/isomerpages/isomercms-frontend/pull/666/commits/4b946ad083cbd1a47d2a9949ad7d1b70a4e483b3

src/components/settings/ColorPickerSection.jsx
alexanderleegs: Good suggestion, updated in https://github.com/isomerpages/isomercms-frontend/pull/666/commits/1dd482ca0214a79ba30f7dee0fd993e9427efe91

src/components/settings/ColorPickerSection.jsx
alexanderleegs: After discussion with @NatMaeTan, we have decided to leave it as is because the colour picker modal does not show the selected colour clearly enough.

src/layouts/Settings.jsx
alexanderleegs: Good suggestion - updated in https://github.com/isomerpages/isomercms-frontend/pull/666/commits/2c15260967dcd709236a71ade747192e47415212

gweiying (approved):
looks good!

Thanks for the work on this, this is much cleaner than previously. I think there can still be more cleanup done on the Settings state setting, getting and saving, but is probably blocked by template-side refactor and `useForm` integration, so that can be a separate refactor.

That aside, everything looks good!

PR #669, opened 2021-11-02
Feat/use form

gweiying:
## Context

This PR adds the following changes:
- Reorganizes component into component folder (currently contains component and schema, can include stories in the future)
- Replace all `useState` hooks with `useForm` to manage form state
- Adds `Yup` object schema with `@hookform/resolvers` to provide validation in `useForm`

## Background
Currently, we have no form management tool, so we manually manage state and update via `useState` and `changeHandler` for each field that we have. To simplify this, this PR introduces `useForm`, which helps with managing form state and validation. 

This PR also reorganises component files and breaks up our validation logic into smaller component-specific schemas, so that each component can be composed in the future with type, component, schema and stories, similar to FormSg V2.


alexanderleegs (changes requested):
Great suggestion, this will help a lot with state management for most of our components!

One issue with the current implementation though is that the errors don't seem to be propagated until the user clicks out of the text box they are editing - this results in a scenario where a user trying to change the file name has the save button become invalid only when attempting to click on it, or after receiving an error message, a user doesn't know that their title is valid until they have made changes and clicked outside of the box

gweiying:
> Great suggestion, this will help a lot with state management for most of our components!
> 
> One issue with the current implementation though is that the errors don't seem to be propagated until the user clicks out of the text box they are editing - this results in a scenario where a user trying to change the file name has the save button become invalid only when attempting to click on it, or after receiving an error message, a user doesn't know that their title is valid until they have made changes and clicked outside of the box

Checked with @NatMaeTan, the designers would like to start moving towards validation on Blur (delaying the validation), so we will stick w the current UI!

alexanderleegs (approved):
lgtm!

PR #673, opened 2021-11-03
Style/module aliases and import sorting

alexanderleegs:
This PR introduces the following style updates:

- Absolute imports, for simpler local dependency imports
- Eslint import order sorting

**Main changes**
- eslintrc.json: Additional configuration for specifying import order of local dependencies. Currently these are set to alphabetical order, but I am open to alternative sorting methods.
- jsconfig.json: Used to enable intellisense on the absolute paths.
- cypress/plugins/index.js: Used to configure cypress to read absolute paths used in tested components
  - To support this, we also added 2 dev dependencies, `find-webpack` and `@cypress/webpack-preprocessor`

The other files have been changed to update paths and import ordering.

Our import order in the config has been configured to match that of the backend, namely:
- built-in packages
- external packages
- our aliases (whose relative orders are defined above)
- parent and sibling paths
in that order.


gweiying (approved):
lgtm

PR #677, opened 2021-11-10
Hotfix to develop: update fonts for span

gweiying:
This PR fixes the font on EditPage. 

Currently, there's a style override that causes the EditPage font to render as `monospace`, and on highlight, render as `Inter`. We remove the style override so that `Inter` is the default font.


kathleenkhy (approved):
lgtm

PR #689, opened 2021-11-19
Refactor/resource room

gweiying:
This PR refactors the Resource Room flow. 

## Key Changes
This PR refactors the Resource Room flow: 
* replaces `Resources` screen with `ResourceRoom` screen
  * replaces existing components with v2 components:
    *  `FolderCardV1` -> `FolderCard`
    * `FolderNamingModal` -> `DirectoryCreationModal`
    * `FolderModal` -> `DirectorySettingsModal`
  * replaces existing query logic with query hooks
* update services 
  * update backend api endpoints
  * update query invalidation logic
* update frontend routes 
* update tests

## Notes
This refactor only covers ResourceRoom. ResourceCategory will be covered in another refactor, so the resource category screen continues to use the v1 endpoints.


alexanderleegs (commented):
Currently, the resources in the sidebar isn't highlighted when in the resources page

gweiying:
> Currently, the resources in the sidebar isn't highlighted when in the resources page

Good catch! Fixed here 19e9ed1

alexanderleegs (changes requested):
The refactor looks great! Just one other outstanding issue - currently we don't handle the case where users visit a resource room of a different name through the url, e.g. `/sites/a-test-v4/resourceRoom/not-real` - we should catch the 404 error that's being thrown by the backend and redirect users to their actual resource room/the page to create a resource room

gweiying:
> The refactor looks great! Just one other outstanding issue - currently we don't handle the case where users visit a resource room of a different name through the url, e.g. `/sites/a-test-v4/resourceRoom/not-real` - we should catch the 404 error that's being thrown by the backend and redirect users to their actual resource room/the page to create a resource room

Good suggestion! Fixed in 40c2f77

alexanderleegs (approved):
lgtm!

PR #705, opened 2021-12-02
Fix: increase allowable permalink length

alexanderleegs:
This PR increases the length of allowable permalinks to 100 characters, up from 50 characters. To be merged after design approval.


gweiying (approved):
lgtm!

PR #709, opened 2021-12-13
Fix: unset permalink for files

alexanderleegs:
See https://github.com/isomerpages/isomercms-frontend/pull/708 for details.


alexanderleegs:
Opting to do a regular merge rather than a hotfix, see https://github.com/isomerpages/isomercms-frontend/pull/710

PR #710, opened 2021-12-13
Fix/unset permalink for files

alexanderleegs:
This PR fixes the updating of resource files of type file.

Currently, a default permalink is erroneously added to these files when updated, causing the links on the actual site to link to
a page which does not exist, rather than the file. We remove the permalink for these files in this PR.


gweiying (approved):
lgtm!

PR #715, opened 2021-12-17
Media/refactor hooks

gweiying:
This PR adds custom React query hooks for the media CRUD operations, and updates the existing directory hooks with media directory invalidation and caching logic.

## Implementation 
The media CRUD React query hooks differ from previous hooks.

As the media retrieval is slower and more resource-heavy, we use React Query's `setQueryData` [details](https://react-query.tanstack.com/reference/QueryClient#queryclientsetquerydata)  to immediately create cached copies of the media content. We use this caching for the following scenarios:
- creating a media file
- updating the name of a media file
- deleting a media file
- retrieving the directory of media files (from the media directory list, caching the content of each individual media file)

We can also consider in the future extending this to the renaming of media directories as that is also fairly slow, however this is not currently supported on the backend as the backend does not return the new directory name. As this is unlikely to be a common issue, we can consider this in the future.

## Others
This PR does not provide any standalone functionality and is meant to be reviewed with #716, #717, #718, and #719. However this PR is backward compatible and **does not** break existing functionality.


alexanderleegs (approved):
lgtm!

PR #735, opened 2022-01-27
chore(nvmrc): adds nvmrc to pin node version

seaerchin:
## Problem
Isomer doesn't pin the node version. This might lead to incompatible versioning and unexpected build errors. This occurs when node version is 17 but `node-sass` is version 4.xx, where node cannot build the package.

## Solution
1. Pin the node version using a `.nvmrc` file. This was preferred as it serves 2 purposes: first, it allows devs developing locally to use the correct version of node, assuming that they have nvm installed; next, it allows [netlify](https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript) to use the `.nvmrc` to build the site.
2. Node version was set to 14.xx because 14.xx is a [LTS version](https://nodejs.org/en/about/releases/), which guarantees support till 2023 (odd numbered versions only have 6 months worth of support). We could use 16.xx as the LTS version but because my local install of `node-sass` with 17.xx failed, i opted for 14.xx instead.


seaerchin:
will also do this on the backend repo once approved

alexanderleegs (approved):
Good catch, thanks for setting this up!

prestonlimlianjie (approved):

PR #739, opened 2022-01-31
test(ci): adds recording to cypress

seaerchin:
## Problem
Currently, we don't have CI; this is an issue (refer to jiehao's [doc](https://docs.google.com/document/d/13k746vGrdmkdIc_ZSH3rPIspkG5sJ0muZ1CLIdf-xik/edit#) for more deets) as manual testing takes engineering time and it can be automated. moreover, having CI on pull requests allows engineers to be sure that their changes are working fine and reduce (not eliminate sadly) the possibility of runtime errors popping up and causing trouble.

This PR is part 1 of adding ci, where a recording option is added to stream the tests and their output to the cypress dashboard. This will then allow the github bot plugin to integrate with cypress dashboard to report tests outcome. 

invite link [here](https://dashboard.cypress.io/invitation/1804b63d-c25d-4f5b-9b2a-5807bbad07bb)

## Solution
1. Creates cypress dashboard and links it to our cypress test runner

## Tests

1. run `npm run test-e2e` with the `CYPRESS_RECORD_KEY` env var set. this should record the run on the dashboard.

## Deploy Notes

- `CYPRESS_RECORD_KEY` : this key is required to record down test runs. 

**Notes**
1. we require github integration for an additional reason: netlify, by default, reports success whenever the site is built and deployed successfully. this means that even if tests **fail**, netlify will still report it as a successful deploy (see [here](https://github.com/cypress-io/netlify-plugin-cypress), under failing the deploy)
2. not sure if `cypress open` will stream the tests by default because `run` has the `--record` option but `open` does not.
3. we can also add slack integration (which we should for prod branches) so that we are informed when tests fail


prestonlimlianjie (approved):
lgtm

seaerchin:
not merging till #742 is approved

PR #740, opened 2022-01-31
chore(.github/templates): adds github PR template

seaerchin:
## Problem
Right now we don't have a structured way of writing PRs. This means that important information might be lost (eg: changing build scripts/env vars/breaking deps upgrade) as the author forgot while writing their PRs. Adding a github template would solve this issue and extra sections can be deleted if unused.

## Solution
1. copy formsg's pr template 

## Note
1. feel free to suggest edits as necessary! i'll make the changes and raise a corresponding PR on BE when this is merged in with the final version of the PR template

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,64 @@
+## Problem
+
+<!-- What problem are you trying to solve? What issue does this close? -->
+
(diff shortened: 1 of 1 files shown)


gweiying (approved):
lgtm!

PR #742, opened 2022-02-03
test(ci): github integration with cypress 

seaerchin:
# Problem
This is a follow-up to part 1, located [here](https://github.com/isomerpages/isomercms-frontend/pull/739). This PR tackles the same issue of lacking CI. Initially this was done through `netlify-cypress-plugin`, but due to the extremely long build + testing time, it got timed out by netlify. 

An alternative solution was to use github actions, which is done here.

# Solution 
1. Use github actions to define a workflow for CI
2. Define a `staging` environment within the `isomercms-frontend` repo to store our secrets. This is then pointed to by the github actions workflow, which allows us to retrieve our secrets.

# Required changes 
- [ ] change env vars on isomer prod 
- [ ] (might have more will add later) 

# Expected behaviour
- [ ] bot comments on this (and any subsequent PR when merged) 
- [ ] bot shows status

# GHA setup 
1. defined an environment within the `isomeradmin` github account and used a personal access token with limited permissions.
2. the current CI set up has 3 portions: 
  - `install`: this installs all dependencies for our project. this is done using [`npm ci`](https://docs.npmjs.com/cli/v8/commands/npm-ci) to ensure faster run times.
  - `lint`: this fixes linting and formatting issues using the predefined `lint-fix` and `format-fix` commands. 
  - `build`: this builds the site
  - `test`: this runs ci using a **new** command. the new command is functionally the same as `npm run test-e2e`, but it doesn't `source .env` as we should obtain the env vars from the ci provider. 

for GHA, we are able to use github actions runner (basically the machine/os which our workflow runs on) but an important caveat here is that our netlify config runs on `ubuntu 16.04` which github actions [doesn't support by default](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). instead, i opted for `ubuntu latest`. 

# Follow-ups
1. isomer does true e2e testing, which involves actual network requests hitting the API. This leads to complications during ci (like now), when we cannot spin up a local instance of the backend, as they are located in 2 separate repos. Hence, the ci config here is only run on a few branches/PRs against `master`

# notes 
1. recording is done using an env var - right now, i'm unsure if `cypress open` has the recording option (don't think so as [docs](https://docs.cypress.io/guides/guides/command-line#cypress-open) omit this)
2. recording can be done **locally** - opted for this because we might wanna share runs. to avoid recording, use `run-e2e open` locally.
3. ~if we have flaky tests, cypress dashboard should catch it. we should review this once a suitable period has passed to see if firstly, flaky tests are due to the way the spec code is written or the method in which we test and act accordingly~ 
     only for team plans :(
4. not sure if multiple concurrent tests will flake due to how we run e2e tests. consider: A pushes and we run ci, then halfway through running tests, B pushes and ci runs again (resetting the repo). the repo reset might affect A's ci run
5. github actions can have certain workflows running on certain branches. this might be useful in the future where we want to run stuff specifically for production branch


seaerchin:
Leaving this here for future engs: our tests + streaming to cypress to use github status takes forever (>30 mins), which is not allowed using `netlify-cypress-plugin`. This is not a short term problem as we are going to add more e2e and unit tests in the future and hence, the plugin is unable to fit our ci needs. Hence, i'm going to use alternative ci providers (eg: GHA/circle). Leaning towards GHA for now because it's cheaper

seaerchin:
A follow up: testing still takes forever (~45 mins according to GHA). This will be an impediment to running the full test suite on every push/pr opened. 

A workaround might be to: 

1. ensure that actual integration with API requests is only run on deployment to prod branches
2. mock our dependencies/calls in the other branches

.github/workflows/ci.yml
```diff
@@ -0,0 +1,98 @@
+name: ci
+
+on:
```
prestonlimlianjie, line 3: Some thoughts on the CI configs for the various branches  - please share your thoughts on this:
- PR to all branches except master: lint, build
- PR to master: lint, build, **and test**
- (optional) push to master, develop, or staging: lint, build

### Lint and build PRs to all branches
We want to ensure that PRs to all branches have basic code hygiene. This is especially so for PRs to develop and master branch, but we can also imagine that developers make PRs from feat-branch-A-continued into feat-branch-A-base for multi-part PRs.

### Also lint and build on push to master, develop, or staging branches
These are actually optional, since Netlify does the build and deploy for us; but no harm running them again on GitHub Actions.

### Only run e2e test on PRs to master
The Cypress e2e test has a lot of side effects (resets the git history of the `e2e-test-repo`, the entire e2e test suite cannot be run concurrently by two different CI runners), so we want to use it sparingly.

It is quite tempting to use the e2e test in many places - but really, the e2e test should be mainly used as a final release test to ensure that our release is functional and doesn't break any existing features.

The main reason why we are so tempted to run the e2e test in other places, e.g. PR to develop, is because we don't have other integration/unit tests at the moment. So that's something that we can work on next.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,98 @@
...
+  push:
+    branches:
+      - master
+      - "dependabot/**"
```
prestonlimlianjie, line 7: Is your intention to run build and lint if a new dependabot or snyk branch is opened? 

If so, we could alternatively configure this by targeting **all PRs to develop branc**h; this would account for all _dependabot -> develop_ and _snyk -> develop_ PRs, as described in the previous review comment.

.github/workflows/ci.yml
prestonlimlianjie: We could implement the above by either:
- trying out the GitHub Actions [on.workflow_run.<branches|branches_ignore>](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_runbranchesbranches-ignore)
- a gatekeep method as used by [some of our other products](https://github.com/opengovsg/checkfirst/blob/develop/.github/workflows/ci.yml#L71-L86) - it feels very clunky to me, so the first is preferable

.github/workflows/ci.yml
```diff
@@ -0,0 +1,98 @@
...
+      - "snyk-**"
+  pull_request:
+    branches: [master]
+    types: [opened, reopened]
+
+jobs:
+  install:
+    runs-on: ubuntu-latest
```
prestonlimlianjie, line 15: nit (please feel free to ignore if you deem fit): could we leave a note in the ci.yml file explaining why we chose ubuntu-latest (exactly as you've described in the PR)?

prestonlimlianjie (changes requested):
The PR looks good overall - I'd like to propose a different strategy for applying `build`, `lint`, and `test` to the branches and would love to hear your thoughts

.github/workflows/ci.yml
seaerchin: **deciding workflow**
Actually i think i got muddled with this PR - the points you've raised are excellent but an additional distinction could be made between e2e/non-e2e tests. 

This distinction is necessary to decide what tests to run on which branch and i think this whole workflow (the e2e tests) workflow should only be run on PRs against `master` ie, releases. This is because this workflow is both expensive but realistic, which means that it should only be used sparingly in high stakes PRs.

conversely, against every other branch/push, what could happen is we do the workflow that you mentioned with the addition of `npm test`. this command doesn't do anything right now but is a guarantee that when (not if!) we add unit/integration tests in the future, we will get safety on pushes!

 **implementing workflow**
i agree that the gatekeep method is clunky! i think we can just have 2 workflow files using the `branches|branches_ignore` pattern and it should work. a BIG caveat here is that in the period between now till writing non-e2e tests, we lose safety on snyk/major version pushes on dependabot. this is non-ideal because for dependabot, major version pushes have breaking changes. i tried to mitigate this originally with the `push` event + running e2e but i think a suitable alternative might be to just run tests manually for the branch or to allow `develop` as a base and run only if the head ref matches a regex.

.github/workflows/ci.yml
seaerchin: good catch, fixed!

.github/workflows/ci.yml
seaerchin: addressed in the prev comment!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
...
+        with:
+          node-version: "14.x"
+      - name: Load Node.js modules
+        uses: actions/cache@v2
+        with:
+          # npm cache files are stored in `~/.npm` on Linux/macOS
+          path: ~/.npm
+          key: ${{ runner.OS }}-node-${{ hashFiles('**/package-lock.json') }}
+      - run: npm ci
+      - run: npm run test:ci
```
seaerchin, line 100: also, something i'm not entirely sure about is _how_ this works - cypress works by obtaining a website url and then invoking events on the dom. this seems like it should fail (and it does on local) as hte website is not running so i'm not sure how it worked previously. 

if anyone can explain this, it'd be greatly appreciated!

prestonlimlianjie: This is a good catch - I don't think GitHub Actions runners have a browser by default. We might want to try using the [cypress-io/github-action@v2](https://docs.cypress.io/guides/continuous-integration/github-actions#Cypress-GitHub-Action) plugin to run the tests.

seaerchin: i actually saw this before and i was under the impression that it did the same thing as what we're doing here, which is why i omitted it. 

in fact, the runner [does](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md) have chrome!

prestonlimlianjie: You're right - it does. So actually we can proceed with the current CI configs (without cypress-io/github-action@v2)

prestonlimlianjie: Also for summary, @seaerchin and I had an offline discussion. We clarified that the e2e tests work on the live staging CMS, and so there is no need for the CI runner to run a local version of the CMS.

prestonlimlianjie (approved):
lgtm

PR #745, opened 2022-02-07
chore: add mergify

seaerchin:
## Problem
We should automate some stuff on isomer, which would otherwise either slip by or require man hours. 

## Solution
Add mergify to the repo to automate the boring stuff away. The mergify config is relatively conservative (taken from `ts-template`) with a few added configurations. The added configs are: 

1. close PRs without any recent (>1 month) activity 
2. delete branches after their PR is merged (we can still restore branch so this is not information destroying) 
3. mark a PR as draft if [wip] (case insensitive) is in the title.

## Note
1. the mergify config only runs `lint` and does **not** test to see if the changes pass (for dependabot/snyk). this is due to the extremely long run time for ci tests - we should test them once we have lightweight unit tests up.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,9 +1,9 @@
 pull_request_rules:
   - name: Approve and merge non-major version dependabot upgrades
     conditions:
-      - author~=^dependabot(|-preview)\[bot\]$
+      - author~=^dependabot\[bot\]$
       - check-success~=lint
-      - title~=^build\(deps[^)]*\). bump [^\s]+ from ([\d]+)\..+ to \1\.
+      - title~=^Bump [^\s]+ from ([\d]+)\..+ to \1\.
     actions:
       review:
         type: APPROVE
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,42 @@
+pull_request_rules:
+  - name: Approve and merge non-major version dependabot upgrades
+    conditions:
+      - author~=^dependabot(|-preview)\[bot\]$
+      - check-success~=lint
+      - title~=^build\(deps[^)]*\). bump [^\s]+ from ([\d]+)\..+ to \1\.
+    actions:
+      review:
+        type: APPROVE
+      merge:
+        method: squash
+
+  - name: Approve and merge Snyk.io upgrades
+    conditions:
+      - author=snyk-bot
+      - check-success~=lint
+      - title~=^\[Snyk\]
+    actions:
+      review:
+        type: APPROVE
+      merge:
+        method: squash
+
+  - name: Close stale PRs (>1 month since last activity)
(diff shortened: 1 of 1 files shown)


.github/mergify.yml
```diff
@@ -0,0 +1,42 @@
+pull_request_rules:
+  - name: Approve and merge non-major version dependabot upgrades
+    conditions:
+      - author~=^dependabot(|-preview)\[bot\]$
```
prestonlimlianjie, line 4: Should just be `author=dependabot[bot]`?

Dependabot preview is shut down now:
https://github.blog/2021-04-29-goodbye-dependabot-preview-hello-dependabot/

.github/mergify.yml
```diff
@@ -0,0 +1,42 @@
...
+      - author=snyk-bot
+      - check-success~=lint
+      - title~=^\[Snyk\]
+    actions:
+      review:
+        type: APPROVE
+      merge:
+        method: squash
+
+  - name: Close stale PRs (>1 month since last activity)
```
prestonlimlianjie, line 24: Let's hold off on this one for now while we clean up the PR/issues

prestonlimlianjie (changes requested):

.github/mergify.yml
```diff
@@ -0,0 +1,42 @@
+pull_request_rules:
+  - name: Approve and merge non-major version dependabot upgrades
+    conditions:
+      - author~=^dependabot(|-preview)\[bot\]$
+      - check-success~=lint
+      - title~=^build\(deps[^)]*\). bump [^\s]+ from ([\d]+)\..+ to \1\.
```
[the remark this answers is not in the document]
seaerchin: you're absolutely right - i got this from [ts-template](https://github.com/opengovsg/ts-template/blob/develop/.github/mergify.yml) and assumed it was correct...

we should also make a corresponding PR upstream!

.github/mergify.yml
seaerchin: didn't know about this - will do! (same as above, should also fix upstream)

.github/mergify.yml
seaerchin: unsure if you're referring to just this rule (closing stale PRs) or mergify in general. 

if it's just the rule, i can split the PR and make this rule its separate PR so that we can get mergify in first and the stale PR rule can be merged when the clean up is done!

prestonlimlianjie: Just the rule, as discussed earlier! :)

seaerchin: done [here](https://github.com/isomerpages/isomercms-frontend/pull/745/commits/0ce96f8ab1f05810f23a07ed4976694099257f84)

prestonlimlianjie (approved):
lgtm

PR #753, opened 2022-02-08
refactor(formFieldMedia): splits components into smaller pieces

seaerchin:
## Problem
`FormFieldMedia` (and similar `Formxxx` components) have alot of shared logic/formatting. However, as the component has too many responsibilities, this means that it is difficult to code split and share similar logic across different components, resulting in code repetition and subtle differences between similar pieces (eg: in `FormField`, the title is given as `<label>` but in others, it is just `<p>`.

This also made it difficult to write stories as components have to be purely visual, which led to this refactor.

## Solution
The solution was to chunk `FormFieldMedia` down into smaller pieces, with components that can be shared between different form components residing in the `Form` folder and specialized `FormFieldMedia` components. 

This made use of the [compound components](https://www.smashingmagazine.com/2021/08/compound-components-react/) pattern to encapsulate shared state and compose the split up pieces to still have the same functionality.  

## Tests
**please take extra care in checking `resourceFormField` and `pageSettingsModal` as they have unexpected behaviour**
- [ ] open the cms. visit the settings page. attempt to change favicon, shareicon and the agency icon. they should all succeed. 
- [ ] open the cms. visit the workspace tab (first on top). next, under my workspace, click on homepage. click on hero section and then choose image. next, select an image and click save. this should succeed.
- [ ] open the cms. visit the workspace tab (first on top). next, under my workspace, click on homepage. click on add new section and add a new infopic section. next, under infopic image, click choose image and click save. this should succeed.
- [ ] open the cms. visit the resources tab (second from top). next, click on create new category and enter resource as the title and click next. after, click on resource (the newly created one) and click add on new page.
    - [ ] under meta image url, click on select image and upload a image. this should succeed.
    - [ ] under resources detail, select downloadable file then click select file. you should be able to upload a pdf file.
    
## Notes
1. didn't know how to force an error in the website, so i checked the error message by forcing the `hasError` prop to be true and setting a custom error message.
2. we should prefer using default arguments over `defaultProps` if we are planning to shift to ts due to less work required.
3. Some call sites of `FormFieldMedia` pass `siteName` as a prop (eg: `HeroSection`) but it appears to do nothing and hence, has been removed
4. `style`/`defaultValue` was never used and also has been removed. 
4. there was a comment [here](https://github.com/isomerpages/isomercms-frontend/pull/753/files#diff-65723075a13eb2f27e7e7716b517f5b1820d623473ffbf78a67dfae03c7dbe65L190) about a temporary workaround which i've removed - unsure what it was for ._.


prestonlimlianjie (approved):
This PR looks great!!

gweiying:
Late comments but great work!! Looks much more reusable and simpler to understand

Issue #746, opened 2022-02-07
[feature] Add chromatic and storybook to isomercms-frontend


seaerchin:
note: this is a tracking issue, written primarily for my benefit so some stuff might not be clear. do ping me here for any questions/clarifications!

# Problem 
we should set up storybook and then chromatic so that we can test for visual regression easily and have a sandbox to try out new components, easing design/eng handoff.

# Considerations
1. are we fine with minor UI changes? we are currently using manual sass styling, which might be costly to port to the new system - it would be alot easier to simply use the design system's components as-is, if functionality is maintained

# Solution
- [x] add storybook
- [x] add chromatic
- [x] (possibly add typescript in here) 
- [x] add ogp design system (note that this is in ts + requires chakra) 
- [x] write a story involving a page that uses `FormFieldMedia`
- [ ] (maybe write unit tests here if required)

PR #757, opened 2022-02-09
feat(design-system): adds storybook/chromatic and ogp design system 

seaerchin:
**THIS PR REQUIRES A FRESH INSTALL OF `NODE_MODULES` WHEN MERGED**

## Problem
Storybook provides an automated way to test components in isolation. moreover, storybook enforces that our components are purely visual in nature and leads to better separation of components in our codebase. chromatic is a UI review platform for storybook and allows for designers to leave comments on the implemented components

ogp design system builds upon storybook and chakra-ui to provide a unified set of components that are reusable and accessible. as the design system is going to be the standard within ogp, isomer should make the switch to the design system.

This PR also adds typescript (hehe).

part 1 of #746 

## Solution
1. Adds `storybook`
2. Adds ogp design system

The PR is here due to a number of roadblocks encountered in setting up the design system (which will be documented below) so that the build changes can be reviewed in isolation.

**Breaking Changes**

This PR will require package versions to be upgraded on the prod site. I don't think it's a breaking change but unsure if there are any caveats involved

  - Details ...

**Features**:

- Installed `storybook`
- Installed the design system

During installation of the design system, there were 2 major roadblocks.
1. Unable to import named exports - see [here](https://stackoverflow.com/questions/69769360/error-importing-framer-motion-v5-in-react-with-create-react-app) for more detail. Took the approach outlined [here](https://github.com/reactioncommerce/reaction-component-library/issues/399) except that i used `craco` as it was more actively maintained + formsg also uses `craco` so i had something to refer to. This also required an installation of `storybook-preset-craco` along with some config updates to the `.storybook/main.js` file to ensure that the `.mjs` files are parsed in the same way.
2. Can't resolve `react/jsx-runtime`. This happens due to a version upgrade in react, mentioned [here](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html). The fix is 2-fold: first, upgrade react to a version that supports the new transform (went with latest stable). next, tell the ts compiler that we are using the new jsx-transform. (see [here](https://www.typescriptlang.org/tsconfig#jsx)).
3. We need to install all peer dependencies of chakra - the design system doesn't install chakra's peer dependencies and this leads to odd behaviour
 
## Tests

- [ ] run `npm run <start | dev | build | test>`. This should work without any issue.
- [ ] run `npm run storybook`. This should display the base storybook page.

**New scripts**:

- `storybook` : starts the storybook website

**New dependencies**:

- ogp design system and all peers
- `craco`

**New dev dependencies**:

- `storybook` and associated plugins
- typescript and associated typings


prestonlimlianjie (approved):
lgtm

PR #758, opened 2022-02-09
refactor(mediamodal): removed nested ternary

seaerchin:
## Problem

Nested ternaries make expressions difficult to read. `MediaModal` has a couple of nested ternaries, which make it difficult to understand the component

## Solution
1. Define a `Modal` component within `MediaModal` to simplify the flow.


prestonlimlianjie (approved):
lgtm

PR #759, opened 2022-02-09
feat: add story for formFieldMedia

### LINKED ISSUE #746: [feature] Add chromatic and storybook to isomercms-frontend

seaerchin:
## Problem
`FormFieldMedia` does not have an associated story. It is also written in javascript, which lacks static typing.

Closes #746

## Solution
This PR has more stories than normal. This is because isomer has not decided on a direction for the migration strategy and the stories laid out here are meant to provide context and information on what migration strategy might best suit us.

1. `v1/Form Field Media` is the component as-is.
2. `v1/Form Field Media With Theming` is the component as-is but with an arbitrary `ThemeProvider`. This is the config that would be adopted if we were to adopt the design system gradually (ie, minor UI changes but no functionality changes) 
3. `v2/Form Field Media` is the component with mixed usage of our (themed) components and unthemed portions swapping to design system (here, the `input` and the `div` and `button` is from isomer).
4. `v2/Design System Form Field Media` is a fully design system based `FormFieldMedia`.

Note that this component has been refactored also in a prior PR, which aided in the shift to ts + the design system. this cost has to be kept in mind as the design system components are purely visual, whereas our components might have non-UI logic baked into them. 

**Breaking Changes**

  - Details ...

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,6 +1,3 @@
-import "@opengovsg/design-system-react/build/fonts/inter.css"
-import { ThemeProvider } from "@opengovsg/design-system-react"
-
 export const parameters = {
   actions: { argTypesRegex: "^on[A-Z].*" },
   controls: {
@@ -10,11 +7,3 @@ export const parameters = {
     },
   },
 }
-
-export const decorators = [
-  (Story) => (
-    <ThemeProvider>
-      <Story />
-    </ThemeProvider>
-  ),
-]
--- a/src/components/FormFieldMedia/FormFieldMedia.scss
+++ b/src/components/FormFieldMedia/FormFieldMedia.scss
@@ -1,0 +1,11 @@
+// NOTE: This file is used solely for the component's stories.
+// This is because the base styling for the component is done in bootstrap, which is included in the index.html file.
+// The styles here are obtained by looking at the element in the console and copying the styling there.
+
+.d-flex {
+  display: flex;
+}
+
+.border {
+  border: 1px solid #dee2e6;
+}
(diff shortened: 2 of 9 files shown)


prestonlimlianjie (approved):
lgtm - ~will add the chakra deps installation fix in https://github.com/isomerpages/isomercms-frontend/pull/780 instead, since this specific PR doesn't use chakra in the wild, i.e. specific pages/layouts~(I see that this has been done in the prev branch)

seaerchin:
> lgtm - ~will add the chakra deps installation fix in #780 instead, since this specific PR doesn't use chakra in the wild, i.e. specific pages/layouts~(I see that this has been done in the prev branch)

yeap! so the previous branch is for scaffolding - all the setup is already done there so that devs have less dependencies/rebase issues! i think you should be able to safely rebase off onto the `feat/add-design-system` branch without any issues too for #780

PR #768, opened 2022-02-14
Refactor/edit homepage hooks + services

### LINKED ISSUE #751: [refactor - frontend] update EditHomepage with refactored query hooks and components

gweiying:
## Problem

This PR renames the existing Settings services and query hooks to a more general umbrella term Config services/ hooks, so that they can be reused for the following metadata endpoints:

- Homepage
- ContactUs
- Navbar
- Settings

These endpoints only have two operations (get and update) and do not have business-specific logic (e.g. error handling or on settled logic) unique to one particular endpoint, hence lends itself to sharing services and query hooks easily.

Closes #751 

## Solution

Rename existing services and hooks from `SettingsXX` to `ConfigXX`. 

**Breaking Changes**

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


src/services/ConfigService.jsx
```diff
@@ -0,0 +1,20 @@
+export class ConfigService {
+  constructor({ apiClient }) {
+    this.apiClient = apiClient
+  }
+
+  getEndpoint({ siteName, isHomepage, isSettings }) {
+    if (isHomepage) return `/sites/${siteName}/homepage`
+    if (isSettings) return `/sites/${siteName}/settings`
+  }
```
seaerchin, line 9: i think this is p prone to error because callers have to explicitly pass in a `isSettings` or `isHomepage`, failing which this returns undefined. 

Something that we might wanna consider here is who decides the final API endpoint - if it's the caller of this function (indirectly through the hook), we can simplify and just pass in an endpoint directly like so: 

```js
getEndpoint({siteName, endPoint}) {
    return `/sites/${siteName}/${endpoint}`
}
```

this makes it clearer and also hands control back to the caller of the hook, which seems to be what is happening already.

src/services/ConfigService.jsx
```diff
@@ -0,0 +1,20 @@
...
+
+  async update(apiParams, { configConfig }) {
```
seaerchin, line 11: wdyt about renaming `configConfig` to something liek `endpointConfig` or something more descriptive/specific? i feel like it would help with readability/understanding!

src/services/ConfigService.jsx
```diff
@@ -0,0 +1,20 @@
...
+    await this.apiClient.post(this.getEndpoint(apiParams), configConfig)
+  }
+
+  async get(apiParams) {
+    return await this.apiClient
```
seaerchin, line 16: nit: extra `await`

src/services/ConfigService.jsx
gweiying: Yes! Good catch, haha I think I didn't look over the naming after I used the replace functionality. I refactored this in d1ec478

src/services/ConfigService.jsx
gweiying: This makes sense in this context, and I'm open to changing it here, but for our other endpoints e.g. [page services](https://github.com/isomerpages/isomercms-frontend/blob/develop/src/services/PageService.jsx), the parameterisation of the endpoint is passed directly from the url parameters. 

For example, for page: https://staging-cms.isomer.gov.sg/sites/a-test-v4/folders/default/editPage/Lorem%20Ipsum.md, the url parameters will be as follows:
```
{
  siteName: 'a-test-v4',
  collectionName: 'default',
  fileName: 'Lorem%20Ipsum.md',
}
```

And this url parameterisation directly populates the `getEndpoint(params)` computation. By matching the url parameters with the endpoint parameterised callbacks, we can simplify most of endpoint calling on the frontend as there's no messy logic on computation what the endpoint is on each hook caller. In such cases, I would argue against passing back the control over the endpoint to the caller of the hook as it would create more messy logic in the caller. 

I wonder if the more elegant solution is to put something in the url parameter to parameterise this endpoint?

gweiying: What do you think!!

src/services/ConfigService.jsx
gweiying: good catch, fixed [2cf2fa2](https://github.com/isomerpages/isomercms-frontend/pull/768/commits/2cf2fa2a7912624c79c7ec860f7ca145ff04c889)

src/services/ConfigService.jsx
seaerchin: sflr!! wouldn't passing the config object amount to returning control back to the caller also? when we call with the url params, it's the caller that decides what gets pinged isn't it? 

fwiw the reason i'm hesitant to put the endpoint into the url params is that the parameter always fixed at this page, whereas url params might change (and break downstream components that rely on it) but i do agree that it will be alot cleaner!

gweiying: fixed in d72b64, d196d3f, 220b6ca

src/services/ConfigService.jsx
```diff
@@ -0,0 +1,19 @@
+export class ConfigService {
+  constructor({ apiClient }) {
+    this.apiClient = apiClient
+  }
+
+  getEndpoint({ siteName, configEndpoint }) {
+    if (configEndpoint) return `/sites/${siteName}/${configEndpoint}`
+  }
```
seaerchin, line 8: when will `configEndpoint` be a falsey value?

seaerchin (approved):
lgtm with 1 clarification - we can probably delete the `if` check or return a default endpoint to avoid returning `undefined` when `configEndpoint` is falsey

seaerchin:
closing for now - we can reopen when the downstream PR is reopened

PR #770, opened 2022-02-14
fix(mergify): update PR rule for dependabot

seaerchin:
## Problem
previous rule asserted that `Bump` was the start of the PR title due to `^`. Our dependabot titles start with `Chore(deps | dev-deps)`

## Solution
remove `^`


gweiying (approved):
lgtm

alexanderleegs (approved):

Issue #771, opened 2022-02-14
[ref] FormXxx components should be composable 


seaerchin:
# Problem
Currently, form-based components on FE have alot of similar logic and differ only in minor areas. With #753 in place, we should build upon the base of atomic form components and compose them together so that there is less developer overhead and repetition.

# Proposed Solution 
- [x] Refactor `FormField`
- [x] Refactor `FormFieldHorizontal`
- [x] Refactor `FormFieldPermalink` (unused and deleted) 
- [x] Refactor `FormFieldToggle` (not changing this - there's very little similarity beyond the name) 
- [x] Refactor `FormFieldColour`

This also helps us in migrating to the design system as the design system provides atomic form components, together with the ones that chakra-ui already provides.

PR #774, opened 2022-02-15
ref(formField): adds formInput and refactors formField

seaerchin:
Part of #771

## Problem
The `FormField` component currently has too much responsibilities. It's setting the title/description etc and has many confusing props, some of which are unused.

After #753, we have more composable form components, which meant that it was possible to refactor this component more cleanly and make it a simpler and easier to use component.

## Solution
The solution outlined here consists of 3 sections.

**remove unused/unneeded props**
the `placeholder` prop, which is given in the [default input api](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) is preferred over the `defaultValue` prop (which is also never used). 

`maxWidth`/`className` serve the same purpose (setting some css styling on the component) and have been combined into the `className` prop, which is more general. This included 1 new `classNames` package, which allows for easier classname combination.

some oft repeated props (`autocomplete = "off", type = "text"`) were shifted into a base `FormInput` component.

**reusing basic form level components**
The previously defined `FormXxx` components in the form folder was used here to lend semantic understanding to the component. A slight change was also made to separate out the `FormInput` from the `FormField` (this might be superfluous now after the refactor).

The `FormField` component should **not** take a `register` prop as consumers often use it like so: `{...register(id)}`, so a deliberate design choice was made to accept the return values of `register` instead.

**conforming to input spec**
Because the `FormField` is a visual component that wraps an input, we should conform as closely as possible to how an `input` behaves and have minimal additional props on top (or at least hide them). This is in line with the reasoning above in not passing a `register` prop.

**Breaking Changes**

  - Details ...

## Before & After Screenshots

there was a visual change made in `FormMediaInput` and also, an odd lack of the input being `maxWidth`, which will be illustrated below.

### FormMediaInput
**BEFORE**:
image

**AFTER**:
image

(technically the slight double line is during the error state because of this statement: ```className={hasError ? `${elementStyles.error}` : "border-0"``` but due to the difficulty of throwing an error in prod, i chose to just use the default state instead.

### FormField's `maxWidth`
the `maxWidth` prop should make the input take all the width available but it seems to have no effect (and has been removed).
 
**dev**
image

note that this is with the prop omitted.

**prod**
image

This is with the `maxWidth` prop

## Tests
Due to the huge amount of refactored components, it was not feasible to manually check each and every one of them. The ones that have been checked are those with interesting behaviour. There was no visual breakage on those checked (in the success state). 

For those that were not manually checked, a simple code inspection was done to ensure that the logical component is still the same.

Testing steps: 
1. Open up the CMS 
2. navigate to your localhost
3. open a page containing the component
4. verify that the title is still in the same place. 
5. verify that the error message still shows
6. verify that with no input, the placeholder is still identical
6. verify that the behaviour is still identical 

- [x] directorysettingsmodal
- [x] Hyperlinkmodal
- [x] mediasettings
- [x] pagesettings
- [x] Resourceformfields
- [x] contactcard
- [x] Contactfields
- [x] Locationcard
- [x] locationfields 
- [x] hero button 
- [x] hero dropdown
- [x] herosection 
- [x] Infobar section 
- [x] Info pic section
- [x] keyhighlight 
- [x] resources section	
- [x] media alt text
- [x] nav section
- [x] nav sublink section	
- [x] Editcontactus
- [x] Resourceroom

**Notes**
1. having `onFieldChange` in `FormContext` seems superfluous (as usually each controllable component will have their own controller/handler) and it seems like a mistake to place it there. As we are migrating over to the design system soon, i think it's still acceptable for now to leave it (a band-aid) as design system will not have a `onFieldChange` in context.
2. the difference between `alwaysDisabled` and `isDisabled` lies in how the input component should be used. For wrappers around `FormInput` that **do not** require the input be interactive but only as a visual display of information, `alwaysDisabled` should be true. This is to allow us to have an interactive component, while having the input serve as a source of information, as per `FormFieldMedia`. If we did not have this prop, the input would be interactive, when `isDisabled` is false, which is not desired behaviour


src/components/PageSettingsModal/PageSettingsModal.jsx
```diff
@@ -122,13 +122,15 @@ export const PageSettingsModal = ({
...
-                    errorMessage={errors.title?.message}
-                    isRequired
-                  />
+                  <FormContext isRequired hasError={!!errors.title?.message}>
+                    <FormTitle>Page Title</FormTitle>
+                    <FormField
+                      placeholder="Page Title"
+                      id="title"
+                      {...register("title", { required: true })}
+                    />
```
alexanderleegs, line 131: Nit: old one was `Page title` - but I'm fine with capitalisation also. The actual figma shows all titles as having capitalisation only on the first word (e.g. `Meta description` - we might want to modify the rest if we uncapitalise it. Alternatively, we should also capitalise `Page Details` if we swap to capitalisation

alexanderleegs (approved):
The PR looks great, just a minor nit

src/components/PageSettingsModal/PageSettingsModal.jsx
seaerchin: i actually noticed this and thought that the capitalised version looks better but if it's in the figma then i'll change it! thanks for letting me know!

gweiying (approved):
Component refactor looks great. Went ahead and ran some of the remaining manual tests, checked them off above.

PR #776, opened 2022-02-15
Fix: do not return layout for unlinked pages

### LINKED ISSUE #775: Issue with Colour Settings in CMS
Currently the resource and unlinked page colours are inconsistent 

custom layout for pages have `layout:post` so normal pages were using resource colours but resources homepage colour is supposed to take from primary colour, and resource page colour is supposed to be secondary colour

affects newly created unlinked pages

image

image

Linked to #776

alexanderleegs:
## Problem

Newly created unlinked pages are created with `layout:post` in the frontmatter. This causes an issue where unlinked pages take on the secondary colour of a repo, instead of the primary colour as expected. These pages also display a date , which is undesired behaviour.

## Solution

This issue was caused by an incorrect retrieval of default front matter - we relied on collectionName to determine whether the page was a resource page or not, rather than the absence of resourceRoomName.

Resolves #775 

**Breaking Changes**

- No - this PR is backwards compatible

## Tests

Attempting to add a new unlinked page will create one without `layout:post` in the front matter

## Notes
Unlinked pages created since the pages refactor prior to this fix will still have the incorrect colours

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -498,7 +498,7 @@ export const getDefaultFrontMatter = (params, existingTitles) => {
     }`
   }
   examplePermalink += `permalink`
-  if (collectionName)
+  if (!resourceRoomName)
     return {
       title: exampleTitle,
       permalink: examplePermalink,
--- a/cypress/integration/workspace.spec.js
+++ b/cypress/integration/workspace.spec.js
@@ -68,7 +68,20 @@ describe("Workspace Pages flow", () => {
         .click()
       cy.get("#title").clear().type(TEST_PAGE_TITLE)
       cy.get("#permalink").clear().type(TEST_PAGE_PERMALNK)
+      cy.intercept("POST", "/v2/**").as("awaitSave")
       cy.contains("Save").click()
+      cy.wait("@awaitSave")
+        .its("request.body")
+        .should("deep.equal", {
+          content: {
+            frontMatter: {
+              title: TEST_PAGE_TITLE,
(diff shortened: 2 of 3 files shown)


src/utils.js
```diff
@@ -498,7 +498,7 @@ export const getDefaultFrontMatter = (params, existingTitles) => {
     }`
   }
   examplePermalink += `permalink`
-  if (collectionName)
+  if (!resourceRoomName)
```
seaerchin, line 501: wdyt about writing a unit test for this!

alexanderleegs: We currently don't have unit tests for the frontend - i think we might want to consider this when we start adding unit tests. For now, I've added a test for the request body being sent by the create request in https://github.com/isomerpages/isomercms-frontend/pull/776/commits/331e3686d08517c3d95640c2fecda5d62e363e4e, but we also don't do this anywhere else currently

seaerchin: ^actually we do have unit tests! it's under `src/__tests__` - i've updated it in [another](https://github.com/isomerpages/isomercms-frontend/pull/742) PR so it passes.

i guess like the main point of this is just to 1. check correctness 2. prevent regressions in the future!

alexanderleegs: That makes sense, added unit tests for this util function in https://github.com/isomerpages/isomercms-frontend/pull/776/commits/2c118cd3cc904a844424527722485181b4a368db

seaerchin (approved):
lgtm

PR #777, opened 2022-02-16
fix(mediamodal): fixed infinite rendering bug

seaerchin:
## Problem

The `MediaModal` infinitely renders when creating the modal as a component (introduced in #758).

Closes [insert issue #]

## Solution

Change it to a function invocation. However, still unsure why it works as i always thought components and functions are equal (in React).


alexanderleegs (approved):
lgtm

Issue #772, opened 2022-02-14
[feat] Add notification banner to CMS for Identity Phase 1


prestonlimlianjie:
### Notification banner content
> From 31 Mar 2022, all users will have to use an agency-issued email to verify their account before making new edits in the CMS. Read more 

The `Read more` link should point to https://go.gov.sg/isomer-identity

### Creating a story for notification banner
This is to help the engineers (read: me) who aren't as familiar with storybook to understand how it works.

PR #780, opened 2022-02-16
feat: add banner component and identity phase 1 banner

prestonlimlianjie:
### Description
This PR adds a new custom banner component, as well as modifies the Sites layout to use the banner for an announcement about identity phase 1. 

Closes https://github.com/isomerpages/isomercms-frontend/issues/772

### Solution

The PR adds a custom banner component - because, unfortunately, the existing CMS styles interfere with those from the design system.

#### Design system styles do not play nicely with the CMS styles because of poor scoping

There are 2 external icon links, and the left icon in the banner is mis-aligned.

Screenshot 2022-02-16 at 11 12 03 PM

#### Custom banner

As such, we had to create a very simple custom banner that did not have icons (i.e. no left icon and no close button functionality) just to tide us over for now, while we refactor the styles/components in the CMS. We chose to take this hacky approach because it was crucial to get the announcement banner for identity phase 1 out by this week.

Screenshot 2022-02-17 at 11 45 22 AM

#### Breaking Changes

No - this PR is backwards compatible

### Tests
- The banner component can be used in any layout
- The banner should appear on the Sites layout along with the specific announcement message as shown above; clicking on the `Read more` button should open a new tab to the [announcement page](https://guide.isomer.gov.sg/isomer-identity)

### DIFF
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -1,0 +1,26 @@
+import { Box, Flex } from "@chakra-ui/react"
+
+export interface BannerProps {
+  children: string
+}
+
+// TODO: Please get rid of this and use the design system instead!!
+const styles = {
+  banner: { color: "white", bg: "#2B5FCE" },
+  item: {
+    display: "flex",
+    justifyContent: "space-between",
+    px: "1rem",
+    py: ["1rem", "1rem", "0.5rem"],
+  },
+}
+
+export const Banner = ({ children }: BannerProps): JSX.Element => {
+  return (
+    <Box __css={styles.banner}>
+      <Flex sx={styles.item}>
+        <Flex>{children}</Flex>
+      </Flex>
+    </Box>
+  )
+}
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -8,7 +8,6 @@ export interface BannerProps {
 const styles = {
   banner: { color: "white", bg: "#2B5FCE" },
   item: {
-    display: "flex",
     justifyContent: "space-between",
     px: "1rem",
     py: ["1rem", "1rem", "0.5rem"],
@@ -18,9 +17,7 @@ const styles = {
 export const Banner = ({ children }: BannerProps): JSX.Element => {
   return (
     <Box __css={styles.banner}>
-      <Flex sx={styles.item}>
-        <Flex>{children}</Flex>
-      </Flex>
+      <Flex sx={styles.item}>{children}</Flex>
     </Box>
   )
 }
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -5,6 +5,7 @@ export interface BannerProps {
 }
 
 // TODO: Please get rid of this and use the design system instead!!
+// Refer to issue: https://github.com/isomerpages/isomercms-frontend/issues/782
 const styles = {
   banner: { color: "white", bg: "#2B5FCE" },
   item: {
--- a/src/layouts/Sites.jsx
+++ b/src/layouts/Sites.jsx
@@ -56,7 +56,8 @@ export default class Sites extends Component {
     const { siteNames } = this.state
     return (
       <>
-        {/* TODO: Move ThemeProvider to root of app during design system refactor */}
+        {/* TODO: Move ThemeProvider to root of app during design system refactor. 
+        Refer to issue: https://github.com/isomerpages/isomercms-frontend/issues/782 */}
         <ThemeProvider>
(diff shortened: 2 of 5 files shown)


src/styles/isomer-cms/elements/base.scss
```diff
@@ -41,7 +41,6 @@ li {
 .wrapper {
   height: calc(100vh - #{$header-height});
   width: 100%;
-  top: $header-height;
   position: relative;
```
seaerchin, line 44: does this change not affect the styling?

src/layouts/Sites.jsx
```diff
@@ -53,6 +56,20 @@ export default class Sites extends Component {
     const { siteNames } = this.state
     return (
       <>
+        {/* TODO: Move ThemeProvider to root of app during design system refactor */}
```
seaerchin, line 59: we should add an issue and write the issue number within the comment - this allows devs to have more context on the issue and ensure that it's being tracked somewhere.

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,26 @@
+import { Box, Flex } from "@chakra-ui/react"
+
+export interface BannerProps {
+  children: string
+}
+
+// TODO: Please get rid of this and use the design system instead!!
```
seaerchin, line 7: see below comment on adding issue

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,26 @@
...
+const styles = {
+  banner: { color: "white", bg: "#2B5FCE" },
+  item: {
+    display: "flex",
```
seaerchin, line 11: i think this is unneeded as chakra's [flex](https://chakra-ui.com/docs/layout/flex) already has it by default

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,26 @@
...
+    px: "1rem",
+    py: ["1rem", "1rem", "0.5rem"],
+  },
+}
+
+export const Banner = ({ children }: BannerProps): JSX.Element => {
+  return (
+    <Box __css={styles.banner}>
+      <Flex sx={styles.item}>
+        <Flex>{children}</Flex>
```
seaerchin, line 22: i think this can just be `{children}` as the display is already flex from the parent.

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,26 @@
+import { Box, Flex } from "@chakra-ui/react"
+
+export interface BannerProps {
+  children: string
```
seaerchin, line 4: i think this signature should be amended to `ReactNode` as we are using it below with a component (`ChakraLink`).

design system types it as string as the they want it to go through the `useMarkdown` hook to generate the appropriate html, which we are not providing here.

seaerchin (approved):
Approving for release as identity is coming soon (pog) but i think the comments should be addressed ideally!

alexanderleegs (approved):
lgtm

src/styles/isomer-cms/elements/base.scss
prestonlimlianjie: It does - but it's necessary because it fixes badly written CSS. 

The `.header` element was previously set to `position: fixed`, while the `.wrapper` element is `position: relative`. And so the janky `top: $header-height;` was necessary to move the `.wrapper` element down to prevent overlap with the header.

Now that we are setting `.header` to `position: relative` (in order to accommodate the banner), we can remove that janky `top: $header-height;` from `.wrapper`.

src/layouts/Sites.jsx
prestonlimlianjie: Good point - let me create an issue

src/components/Banner/Banner.tsx
prestonlimlianjie: This will be addressed when we entirely replace the custom Banner component with the design system one, so I'm ok overlooking it for now

src/components/Banner/Banner.tsx
prestonlimlianjie: Addressed in 8debc69

src/components/Banner/Banner.tsx
prestonlimlianjie: Addressed in 8debc69

src/components/Banner/Banner.tsx
prestonlimlianjie: Addressed in e831b47

src/layouts/Sites.jsx
prestonlimlianjie: Addressed in e831b47

PR #785, opened 2022-02-17
build(react): downgrade react and react-dom to 16.14.0

alexanderleegs:
We encountered a problem when upgrading react and react-dom to 17.0.2 - the select colour modals in the settings page caused a reload of the page, which we determined to be due to the react and react-dom version change. We downgrade these packages in this PR.


seaerchin (approved):
lgtm soz for trouble

PR #791, opened 2022-02-18
ref(form field horizontal): replace with simpler API

seaerchin:
This PR builds on #774 and is part of #771 

## Problem
The `FormFieldHorizontal` component is extremely similar to `FormField` but has alot of repeated code within it. This PR uses the new `FormInput` component introduced in #774 to refactor this component and make it simpler.

The `FormFieldHorizontal` component additionally had 2 line breaks on error, which it used for formatting. This made it difficult to compose as it now had layout logic in addition to just being a UI input.

## Solution
1. utilized `FormInput` component as a base and pass customized props to it.
2. created a new `Spacing` component (inspired by Chakra's [Stack](https://chakra-ui.com/docs/layout/stack)) that handles layout and removes this concern from `FormInput`. I've checked with Nat on this and the spacing is a constant 24px between fields

**Breaking Changes**

  - Details ...

## Tests 
**page settings modal**
- [ ] Go to workspace and create a page. you should be able to do so successfully. click on a page and then edit details. you should be able to modify the page URL successfully.
- [ ] Go to resources, create a resource category, then create a page. you should be able to do so successfully. click on a page and then edit details. you should be able to modify the page URL successfully.
- [ ] Go to workspace, create a folder, then create a page. you should be able to do so successfully. click on a page and then edit details. you should be able to modify the page URL successfully.

**settings**
- [ ] go to settings. you should be able to edit and save every form field there.

**notes**
1. `pageSettingsModal` has a `siteUrl` prop, which is used to render a description displaying the site url. i couldn't get this behaviour working on prod and when testing on local, `siteUrl` is blank. i've kept the prop in as it might be defined in certain edge cases but ideally we can remove it if everyone is onboard.

### DIFF
--- a/src/components/FormFieldHorizontal.jsx
+++ b/src/components/FormFieldHorizontal.jsx
@@ -1,70 +1,20 @@
-import PropTypes from "prop-types"
-import React from "react"
+import React, { forwardRef } from "react"
 
 import elementStyles from "styles/isomer-cms/Elements.module.scss"
 
-const FormFieldHorizontal = ({
-  title,
-  description,
-  defaultValue,
-  placeholder,
-  value,
-  id,
-  disabled,
-  errorMessage,
-  onFieldChange,
-  isRequired,
-  style,
-  children,
-  register = () => {},
-}) => (
-  <>
-    <p className={elementStyles.formLabel}>{title}</p>
-    {children}
-    <div className={elementStyles.formHorizontal}>
-      <p className={elementStyles.formDescription}>{description}</p>
-      <input
-        type="text"
-        placeholder={placeholder || title}
-        value={value}
-        defaultValue={defaultValue}
-        id={id}
-        autoComplete="off"
-        required={isRequired}
-        disabled={disabled}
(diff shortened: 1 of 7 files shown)


alexanderleegs (approved):
lgtm!

gweiying:
Refactor looks great. Had some questions about the general FormXX refactor about why there are two booleans named similarly used for disabling the form, (`isDisabled` in FormContext` and `alwaysDisabled` in `FormInput`). Synced up with @seaerchin offline, tldr below:

We want to eventually move control of the Form components to the context/ controller layer, hence we have the `isDisabled` value in `FormContext`. However, we currently have components in which the `FormInput` are specifically disabled, without the entire form component being disabled such as `FormFieldColor` and `FormFieldMedia`, hence we need to specifically implement a `alwaysDisabled` boolean in `FormInput`. Moving forward, we should have a new specific component along the lines of `FormDisabledInput`, so that we can remove the `alwaysDisabled` boolean in `FormInput`.

gweiying (approved):
lgtm!

PR #792, opened 2022-02-18
ref(formFieldColour): refactored to use base form components

### LINKED ISSUE #771: [ref] FormXxx components should be composable

seaerchin:
## Problem
`FormFieldColour` has quite a few components inside. It should only be restricted to the the colour picker and the hex display.

Closes #771

## Solution
1. Shift away wrapper component/`FormTitle` into the wrapping component (`ColorPickerSection`)
2. Use base `FormInput` to simplify internal component logic

### DIFF
--- a/src/components/Form/FormTitle.jsx
+++ b/src/components/Form/FormTitle.jsx
@@ -3,11 +3,12 @@ import React from "react"
 
 import elementStyles from "styles/isomer-cms/Elements.module.scss"
 
-const FormTitle = ({ children }) => (
-  <p className={elementStyles.formLabel}>{children}</p>
+const FormTitle = ({ className = elementStyles.formLabel, children }) => (
+  <p className={className}>{children}</p>
 )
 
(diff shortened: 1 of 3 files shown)


alexanderleegs (approved):
lgtm!

gweiying (approved):
lgtm!!

PR #795, opened 2022-02-21
fix(ci): prevent chromatic from being triggered by bots

seaerchin:
## Problem
Chromatic workflow is being triggered by bots, which is not needed.

## Solution
add if condition in the chromatic job so that the workflow is only triggered if the username doesn't end with bot. This PR also adds in a fix for mergify (see #770) that got overridden


alexanderleegs (approved):
lgtm

PR #764, opened 2022-02-11
refactor/edit homepage components

gweiying:
## Problem

This PR adds `useForm` and `Yup` validation schemas to the Homepage editor components (`EditorHeroSection`, `EditorInfobarSection`, `EditorInfopicSection`, `EditorResourcesSection`) to manage the form values for EditHomepage. 

In our pre-refactored code, the creation, updating, deletion, validation and loading of the `EditHomepage` data were all handled by custom handlers, which was **very tedious and difficult to maintain**, as it is nearly [800 lines long](https://github.com/isomerpages/isomercms-frontend/blob/bb982dac2487d1f12a234a3a4c473894951823ce/src/layouts/EditHomepage.jsx#L151-L912), and requires lots of repeated logic to update and validate the same nested object values. If any change was made to the data structure of the form values, then a code maintainer would have to update the create, update, validation and loading handlers separately, causing lots of repeated work.

~CAVEAT: There are slight bugs with the Templates that I have yet to fix that are causing small bugs with the Preview. Will fix soon. See related tasks below.~

Closes #748 

## Solution

To simplify this, this PR adds `useForm` and `Yup` validation schemas to the Homepage editor components to manage the form values of `EditHomepage`. This removes the need for any state management handlers in the parent layout `EditHomepage`, as well as removes the tedious work of maintaining our own custom load, create, update, validate and delete handlers for the form values.

### Considerations
`EditHomepage` has a fairly large and complex form object. A simple example is shown below:

```
{ 
  description: "Brief site description here",
  image: "/images/isomer-logo.svg",
  layout: "homepage",
  notification: "Here's a notification bar you can use!",
  permalink: "/",
  sections: [ {
    hero: {
      background: "/images/hero-banner.png",
      dropdown: {
        options: [ {
          title: 'Hero Dropdown Element Title', 
          url: ''
        }, { 
          title: 'Hero Dropdown Element Title',
          url: ''
        } ],
        title: "",
      },
      subtitle: "Hero subtitle",
      title: "Hero titlZZZZ",
    }
  },
  {
    infobar: {
      button: "Button text",
      description: "About a sentence worth of description here",
      subtitle: "Subtitle",
      title: "Infobar title",
      url: "/faq/",
  },
  {
    infopic: {
      alt: "Image alt text booya",
      button: "Button Text",
      description: "Infopic description",
      image: "/images/hero-banner.png",
      subtitle: "Infopic Subtitlehello therehello therehello therehello therehello therehello therehello there",
[...]

PR #804, opened 2022-02-28
Refactor/edit homepage layouts

### LINKED ISSUE #749: [refactor - frontend] refactor EditHomepage flow
Related tasks:
- [] #747
- [] #748
- [] #750 
- [] #751

gweiying:
## Problem

This PR adds useForm and Yup validation schemas to the EditHomepage layout to manage the form values. This PR also adds query hooks to retrieve and update the homepage data. 

Closes #749 

### Form Management
This PR adds useForm, FormProvider context and Yup validation schemas to the Homepage layout to manage the form values for EditHomepage. The form values are passed down to EditHomepage components via the FormProvider context, and the form object ids are retrieved via the `fieldId` string, which contains the path to the nested value in an object.  For example, if the object has structure as follows:

```
sections: [{
  infobar: {
    title: 'Title of infobar'
  },
}]
```

Then the `fieldId` of the infobar title is `sections.0.infobar.title`.

### Query hooks
This PR uses the existing query settings hooks to retrieve and update the homepage data. A `useEffect` runs when `homepageData` is retrieved to update the homepage values. 

### Toggle state
Previously, the state of the dropdown menus toggles was stored in a list on EditHomepage, and this list was updated each time a user toggled the dropdown menu. This refactor removes the centralised dropdown toggle list, as the EditHomepage editor components have been refactored in #764 to use the `CardContainer` component where each component manages its own dropdown state.

### New section creation
The logic and default objects for creating new sections is moved into `NewSectionCreator`, to simplify the EditHomepage layout.

**Breaking Changes**

Yes - this PR contains breaking changes. This PR should not be merged without #764 and #768 as there's no standalone functionality in this PR.

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

Tests are provided in #807. The test coverage does not include drag and drop, and appearance of site on built Netlify site, so that should be tested outside of Cypress.

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


src/templates/homepage/HeroSection.jsx
```diff
@@ -24,53 +24,59 @@ const HeroDropdownElem = ({ title }) => (
...
+          {options
+            ? options.map((option, index) =>
+                option.title ? (
+                  <HeroDropdownElem
+                    key={`dropdown-${index}`}
+                    title={option.title}
+                  />
+                ) : null
+              )
+            : null}
```
seaerchin, line 74: i think it's not ideal for us to have nested ternaries as it impacts readability. i'm not sure what's the best option here, i usually just write a small component at the top and use it here but maybe a better solution is to just ensure that the stuff passed in is not null (not really sure how to do this)

src/templates/homepage/HeroSection.jsx
```diff
@@ -24,53 +24,59 @@ const HeroDropdownElem = ({ title }) => (
...
-          <i
-            className="sgds-icon sgds-icon-chevron-down is-size-4"
-            aria-hidden="true"
-          />
-        </span>
-      </a>
-    </div>
+const HeroDropdown = ({ title, options }) => {
+  const [dropdownIsActive, setDropdownIsActive] = useState(false)
+  const toggleDropdown = () => setDropdownIsActive((prevState) => !prevState)
```
seaerchin, line 29: if this pattern comes out often we can consider writing a custom hook

seaerchin (commented):
seems good overall! left a few comments regarding semantic html and nested ternaries!

src/components/homepage/NewSectionCreator.jsx
```diff
@@ -5,13 +5,57 @@ import Dropdown from "components/Dropdown"
...
+const enumSection = (type) => {
+  switch (type) {
+    case "resources":
+      return defaultResourcesSection()
+    case "infobar":
+      return defaultInfobarSection()
+    case "infopic":
+      return defaultInfopicSection()
+    default:
+      return defaultInfobarSection()
```
alexanderleegs, line 48: I'm wondering if we should leave out a default entirely, so it's easier to catch if we're providing an invalid section type

src/templates/homepage/HeroSection.jsx
```diff
@@ -24,53 +24,59 @@ const HeroDropdownElem = ({ title }) => (
...
+      <div className="bp-dropdown-trigger">
+        <a
+          className="bp-button bp-dropdown-button hero-dropdown is-centered"
+          role="button"
+          tabIndex="0"
+          aria-haspopup="true"
+          aria-controls="hero-dropdown-menu"
+          onClick={toggleDropdown}
+          onKeyDown={toggleDropdown}
+        >
```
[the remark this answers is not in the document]
gweiying: Think your questions on the template are really valid, but I would suggest not refactoring them in the CMS. 

Currently, the template on the CMS is written these way because this is how it's actually rendered on an Isomer site. By having the CMS template follow the rendered site, we can more closely display a preview of the actual site. 

See below for a side-by-side comparison of the Isomer site html and the CMS templates:

Screenshot 2022-03-15 at 11 24 24 AM

src/templates/homepage/HeroSection.jsx
```diff
@@ -24,53 +24,59 @@ const HeroDropdownElem = ({ title }) => (
...
+          <span>
+            <b>
+              <p>{title}</p>
```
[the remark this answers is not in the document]
gweiying: See comment above!

src/templates/homepage/HeroSection.jsx
gweiying: [a887d80](https://github.com/isomerpages/isomercms-frontend/pull/804/commits/a887d802425f5c2210d70d6be78d60cdc8c169b4) Good idea

src/templates/homepage/HeroSection.jsx
```diff
@@ -147,12 +150,10 @@ const TemplateHeroSection = (
                   </p>
                 ) : null}
                 {/* Hero dropdown */}
-                {hero.dropdown ? (
+                {!_.isEmpty(hero.dropdown) ? (
```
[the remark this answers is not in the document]
gweiying: [79d0a14](https://github.com/isomerpages/isomercms-frontend/pull/804/commits/79d0a140fc4f911f7de520eb0adab17941d8c199) thanks for catching

src/components/homepage/NewSectionCreator.jsx
gweiying: valid point, thanks [b58e761](https://github.com/isomerpages/isomercms-frontend/pull/804/commits/b58e7617fff219d7f3ea19c6708f851f52fbe3db)!

src/templates/homepage/HeroSection.jsx
gweiying: Not tackling this in within this refactor as this pattern isn't that common in our CMS.

alexanderleegs (approved):
lgtm

seaerchin:
closing for now - we can reopen later once the downstream stuff is fixed

PR #818, opened 2022-03-21
chore(app): run code mod to remove all unused React imports

seaerchin:
This is part 1 of a multi-part PR

## Problem
WIth our `.tsconfig`, the explicit `import React from "react"` is [deprecated](https://www.typescriptlang.org/docs/handbook/jsx.html) and can be removed.

## Solution
Run code mod from react (see [here](https://github.com/reactjs/react-codemod#update-react-imports)).


alexanderleegs (approved):
lgtm

PR #819, opened 2022-03-21
chore(eslint): remove only-warn from eslint and refactor errors

seaerchin:
This is part 2 of #818

## Problem
`eslint` had the `only-warn` plugin installed, which demoted errors to warnings. This meant that `eslint` effectiveness was reduced as errors weren't flagged out in the editor and allowed devs to ignore them. Moreover, during code review, reviewers were unable to easily tell which eslint errors were ignored (as they were **all** ignored). 

This PR is separated into 2 (next one is #820) to prevent review fatigue. This PR focuses on **refactors** to our codebase, which might potentially be destructive. Hence, please lmk if i should chop this PR up into even smaller portions.

## Solution
1. Remove `only-warn` plugin
2. Fix all the errors popping up from eslint. As there are alot of errors, only interesting fixes (ie, potential bug causers) will be listed below. 
    - nested ternaries: rewrite to inline component within file
    - no consistent return: add a default case where possible/throw an error if it should not be possible to reach the case
    - non-semantic html: changed to use the correct element and add HTML roles/type/labels etc as appropriate 
        - This is **not** done in `templates/` as the output in preview should mirror jekyll and hence, the errors are silenced there. 
    - `(({unusedVar, ...rest}) => rest)(params)`: used `_.omit`

## Notes
1. components in templates are special and have non-semantic html **by design** to fit the jekyll output better. In this case, i still changed the code so that it's semantic but now the output is (slightly) different from jekyll in that the html tags are changed, for example, from `<a />` to `<button />`
2. array index as key removed -  i double checked e2e and only `settings-/delete-/move-/moveModal-forwardButton-/pageCard-dropDown-/breadcrumbItem-0-/${mediaTitle}-settings-` were used in `cy.get` so it shouldn't affect e2e but if it does, the fix should be to rewrite the tests, not to revert to using the array index as keys.


src/components/contact-us/LocationFields.jsx
```diff
@@ -101,29 +123,16 @@ const LocationAddressFields = ({
...
-        newErrorMessage = errors[0]
-      } else {
-        errors.forEach((error, i) => {
-          newErrorMessage += error ? `Line ${i + 1}: ${error} ` : ""
-        })
-      }
-    }
-    setErrorMessage(newErrorMessage)
-  }, [errors])
+  console.log("ERRORS: ", errors)
```
alexanderleegs, line 126: Extra log statement

src/components/media/MediasSelectModal.jsx
```diff
@@ -203,9 +199,5 @@ export default MediasSelectModal
 
 MediasSelectModal.propTypes = {
   onClose: PropTypes.func.isRequired,
-  siteName: PropTypes.string.isRequired,
   onMediaSelect: PropTypes.func.isRequired,
-  type: PropTypes.oneOf(["files", "images"]).isRequired,
-  readFileToStageUpload: PropTypes.func.isRequired,
-  setUploadPath: PropTypes.func.isRequired,
 }
```
alexanderleegs, line 203: Why are the proptypes here removed?

src/layouts/Sites.jsx
```diff
@@ -11,6 +12,33 @@ import { SITES_IS_PRIVATE_KEY } from "constants/constants"
 import elementStyles from "styles/isomer-cms/Elements.module.scss"
 import siteStyles from "styles/isomer-cms/pages/Sites.module.scss"
 
+const Site = ({ siteNames }) => {
+  if (siteNames && siteNames.length > 0)
+    return siteNames.map((siteName) => (
```
alexanderleegs, line 17: Nit: should be `Sites` since this component encompasses all sites

alexanderleegs (commented):
Left some comments!

src/components/DirectoryCreationModal/DirectoryCreationModal.jsx
```diff
@@ -64,6 +72,44 @@ export const DirectoryCreationModal = ({
     })
   }
 
+  // Sub-component used here for clarity
+  const Folders = () => {
+    if (pagesData && pagesData.length > 0) {
+      return pagesData.map((pageData, pageIdx) => (
+        <FolderCard
```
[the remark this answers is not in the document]
seaerchin: sure, that makes sense!

src/components/DirectorySettingsModal/DirectorySettingsSchema.jsx
```diff
@@ -43,14 +45,11 @@ export const DirectorySettingsSchema = (existingTitlesArray = []) =>
...
-            .matches(
-              slugifyLowerFalseRegexTest,
-              "Title cannot contain any symbols"
-            )
+        return schema
+          .transform((value) => value.trim().replaceAll(" ", "-").toLowerCase())
+          .matches(
+            slugifyLowerFalseRegexTest,
+            "Title cannot contain any symbols"
+          )
```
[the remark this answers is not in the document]
seaerchin: the problem comes when eslint complains about edge cases not being covered cos it can't infer that there's only 3 types @__@ 

the alternative that i can think of is to include the last conditional and returns a default error case that fails validation like so: 

`schema.test("Invalid case", "This is an invalid type of collection!", () => false)` 

wdyt about this approach

src/components/MediaSettingsModal/MediaSettingsSchema.jsx
```diff
@@ -25,19 +26,19 @@ export const MediaSettingsSchema = (existingTitlesArray = []) =>
...
-          return schema.test(
-            "Special characters found",
-            "Title must end with the following extensions: 'pdf'",
-            (value) => filesSuffixRegexTest.test(value)
-          )
+        return schema.test(
+          "Special characters found",
+          "Title must end with the following extensions: 'pdf'",
+          (value) => filesSuffixRegexTest.test(value)
+        )
```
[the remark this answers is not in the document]
seaerchin: see above too AHHA

src/components/MediaMoveModal/MediaMoveModal.jsx
```diff
@@ -11,12 +11,74 @@ import elementStyles from "styles/isomer-cms/Elements.module.scss"
 import { pageFileNameToTitle, getMediaDirectoryName } from "utils"
 
 export const MediaMoveModal = ({ queryParams, params, onProceed, onClose }) => {
-  const [moveQuery, setMoveQuery] = useState(
-    (({ fileName, ...p }) => p)(queryParams)
-  )
+  const [moveQuery, setMoveQuery] = useState((({ ...p }) => p)(queryParams))
```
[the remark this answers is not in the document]
seaerchin: yes we should... this is careless from my end - good catch!

src/components/contact-us/LocationFields.jsx
seaerchin: good catch!

src/components/DirectorySettingsModal/DirectorySettingsSchema.jsx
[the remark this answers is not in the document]
alexanderleegs: That sounds like a good approach! Either that or just leaving a comment directly above this part to explain that it's for `collectionName`

src/components/media/MediasSelectModal.jsx
seaerchin: they're never used

src/layouts/Sites.jsx
seaerchin: good catch! as there's already a `Sites` component below, i'll rename them both to `Sites`/`SitesWrapper`!

src/components/media/MediasSelectModal.jsx
alexanderleegs: Ah i think I saw wrongly - there were missing proptypes for this component but not for the params removed

alexanderleegs (approved):
lgtm - there's some prop type cleanup that we might want to look into, but I think that's out of scope for this PR

seaerchin:
> lgtm - there's some prop type cleanup that we might want to look into, but I think that's out of scope for this PR

Yeap, but I think that it's redundant work if we're shifting to typescript so I'm not gonna pursue it

PR #820, opened 2022-03-21
chore(eslint): fix non-destructive eslint errors

seaerchin:
Part 3 of #819 

## Problem
This is a follow up to #819 and focuses on eslint _chores_, which are code tidiness issues (eg: missing import/unused variables etc) rather than refactoring focused. 

## Solution
1. follow eslint
2. delete/add stuff as appropriate

## Notes
there might be some overlap with the previous part so please still keep your eyes peeled but the majority of code changes here should be relatively easy to review


src/components/Footer.jsx
```diff
@@ -44,11 +43,4 @@ const Footer = ({
   </div>
 )
 
-Footer.propTypes = {
-  isDeleteDisabled: PropTypes.bool,
-  isSaveDisabled: PropTypes.bool,
-  deleteCallback: PropTypes.func,
-  saveCallback: PropTypes.func,
-}
-
```
alexanderleegs, line 53: Could I check why the proptypes here are removed?

src/components/GenericWarningModal.jsx
```diff
@@ -54,7 +54,6 @@ const GenericWarningModal = ({
 GenericWarningModal.propTypes = {
   displayTitle: PropTypes.string.isRequired,
   displayText: PropTypes.string.isRequired,
-  displayImage: PropTypes.string,
```
alexanderleegs, line 57: Same as above

alexanderleegs (approved):
minor nits, but otherwise lgtm

src/components/Footer.jsx
seaerchin: the proptypes are never used as these don't exist as props @__@

src/components/GenericWarningModal.jsx
seaerchin: i removed it as it's unused - there's a similar `displayImg` prop up above though so actually i'll add this `PropType` back in + change the `displayImg` to `displayImage`

PR #822, opened 2022-03-22
test: added e2e test error message if environment variables are not defined.

rc-davis:
## Problem

Cypress tests fail with unhelpful error messages if the required environment variables are not defined.

## Solution

Throw an exception if any variable is undefined (showing the value that was read for each).

**Breaking Changes**

## Tests

npm run test-e2e #should give error messages if `PERSONAL_ACCESS_TOKEN` or `USERNAME` are undefined.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -5,6 +5,11 @@ const btoa = require("btoa")
 
 // login credentials
 const { PERSONAL_ACCESS_TOKEN, USERNAME } = process.env
+if (!PERSONAL_ACCESS_TOKEN || !USERNAME) {
+  throw new Error(
+    `E2E tests require env vars PERSONAL_ACCESS_TOKEN (${PERSONAL_ACCESS_TOKEN}) and USERNAME (${USERNAME}) to be set.`
+  )
+}
 
 const CREDENTIALS = `${USERNAME}:${PERSONAL_ACCESS_TOKEN}`
 const headers = {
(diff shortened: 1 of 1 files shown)


scripts/build.js
```diff
@@ -5,6 +5,11 @@ const btoa = require("btoa")
 
 // login credentials
 const { PERSONAL_ACCESS_TOKEN, USERNAME } = process.env
+if (PERSONAL_ACCESS_TOKEN === undefined) {
```
seaerchin, line 8: wdyt about adding a check for `USERNAME` and using `!PERSONAL_ACCESS_TOKEN` for consistency across the codebase?

scripts/build.js
```diff
@@ -5,6 +5,11 @@ const btoa = require("btoa")
...
+  throw new Error(
+    `E2E tests require env vars PERSONAL_ACCESS_TOKEN (${PERSONAL_ACCESS_TOKEN}) and USERNAME (${USERNAME}).`
```
seaerchin, line 10: nit: add "to be set" to the error message

scripts/build.js
rc-davis: Oops. Careless. Fixed

rc-davis:
I also added a check for `E2E_COMMIT_HASH`, which was still missing. I left this separate from the other two, because I liked the existing separation between code related to login credentials and test repo reset.

seaerchin (approved):
lgtm

PR #830, opened 2022-04-01
chore(sites): remove banner

seaerchin:
## Problem
Remove banner from production site as it is past 31st March.

## Solution
Delete banner. Still keeping banner component in until Chakra/Design system is in and we can use theirs


prestonlimlianjie (approved):
lgtm

PR #840, opened 2022-04-13
feat(design-system): add ThemeProvider in App

seaerchin:
## Problem
Isomer is shifting over to ogp design system and this requires setting up a `ThemeProvider` at the root of our application. This PR sets it up and fixes UI so that it looks roughly identical to the original UI.

## Solution
1. set up `ThemeProvider` in the root of our app 
2. fix css for header 
3. unset global `bg` 
4. add styling for heading
5. add base colour scheme and stories


seaerchin:
@NatMaeTan and @audreytcy, mainly check if there are any visual inconsistencies between this and current prod 

current inconsistencies are: 
 1. dividers disappearing (emotion sets it to 0px) 
 2. heading in body has weight 600 and wrong colour (former from `isomer-template.scss` setting `font-weight: 600 !important`, latter from `Content.module.scss` setting heading colours)

alexanderleegs (approved):
lgtm pending design input!

seaerchin:
> lgtm pending design input!

design is ok with the squashed inputs but i have an upstream PR (#866) that fixes this (and some other styling stuff)

NatMaeTan:
Im ok with the visuals, just smth to note and fix in the future that the input fields take a while to load with content. They appear blank on first load, and a while later then the text they had (If they were alr filled) appears. Will open an issue for it!

PR #848, opened 2022-04-19
fix(test): update e2e tests

seaerchin:
## Problem
Previously, in #819, there was work done to rewrite components in semantic html and this resulted in the html types of elements being changed from `div` to `button` as well as a minor rewrite of some components (most notably, `BreadCrumb` to make them more declarative). 

As a result of this, our e2e tests broke because they were written using selectors for those specific html tags/certain specific text.

This PR fixes the failing e2e tests
## Solution
1. rewrite e2e-tests to use selectors for the new html tags
2. for tests checking the `breadcrumb` component, the cypress tests were updated to check for the _siblings_ rather than the raw text content (as previously it was a string like `Workspace > Something > Some other thing`.) 
3. For certain tests, we require an assertion to guard against re-renders (caused by perhaps `useState`), which remove the element from the dom. To do this, an existential check has been added after the selector and prior to the event trigger. 
For more details, refer [here](https://docs.cypress.io/guides/references/error-messages#cy-failed-because-the-element-you-are-chaining-off-of-has-become-detached-or-removed-from-the-dom)
4. similar to the above, the identity modal pops up on first paint, which then gets removed later. This causes a UI re-render, which might take place between cypress commands. This leads to flakiness in our tests, whicfh is partially resolved by waiting for 1s before tests start (so that the UI stabilizes)

### DIFF
--- a/cypress/integration/editPage.spec.js
+++ b/cypress/integration/editPage.spec.js
@@ -50,7 +50,7 @@ describe("Edit unlinked page", () => {
 
     // Set colour
     cy.visit(`/sites/${TEST_REPO_NAME}/settings`)
-    cy.contains("p", "Primary").siblings("div").click()
+    cy.contains("p", "Primary").siblings("button").click()
     cy.contains(/^r/)
       .siblings()
       .clear()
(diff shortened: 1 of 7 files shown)


alexanderleegs (approved):
lgtm, thanks for fixing the tests!

PR #850, opened 2022-04-20
fix(preview): revert leftnav changes

seaerchin:
## Problem
Previously in #819, this file was refactored to remove eslint errors. however, templates are a special portion of our codebase and should adhere **as closely as possible** to jekyll output.

This change resulted in some stylistic changes, which has been fixed since.

## Solution
1. revert changes to `leftnavGeneration`
2. shift the file to `template/utils` so future devs won't accidentally alter it
3. update call sites

### Prod 
image

### Before fix
image

### After fix 
image


alexanderleegs (approved):
lgtm

Issue #839, opened 2022-04-11
[bug] trailing spaces on names


alexanderleegs:
Users are currently able to create folder/file names with trailing spaces - this causes issues with certain operations such as deleting folders. We should look into stripping trailing whitespace from names.

PR #852, opened 2022-04-20
Fix: trim whitespace

alexanderleegs:
## Problem

This PR fixes an issue where files/folders with trailing whitespace were causing files/folders to not show up properly on the cms. We have opted to strip trailing whitespace from file/folder names to resolve this issue. Resolves https://github.com/isomerpages/isomercms-frontend/issues/839

### DIFF
--- a/src/hooks/pageHooks/utils.js
+++ b/src/hooks/pageHooks/utils.js
@@ -1,4 +1,5 @@
 export function retrievePageInfo({ data, pageData, resourceRoomName }) {
+  if (!data) return pageData
   const trimmedNameData = {
     ...data,
     title: data.title.trim(),
--- a/src/hooks/pageHooks/utils.js
+++ b/src/hooks/pageHooks/utils.js
@@ -1,0 +1,20 @@
+export function retrievePageInfo({ data, pageData, resourceRoomName }) {
+  const trimmedNameData = {
+    ...data,
+    title: data.title.trim(),
+  }
+  return {
+    frontMatter:
+      pageData && pageData.content
+        ? {
+            ...pageData.content.frontMatter,
+            ...trimmedNameData,
+          }
+        : trimmedNameData,
+    sha: pageData?.sha || "",
+    pageBody: pageData?.content?.pageBody || "",
+    newFileName: resourceRoomName
+      ? `${trimmedNameData.date}-${trimmedNameData.layout}-${trimmedNameData.title}.md`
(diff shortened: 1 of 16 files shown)


seaerchin (commented):
i think this fix leaves us open to making the same mistakes again; all these handles use an `onProceed` method call, which is obtained from  `<create | update><page | directory | media>hook`. 

i think making the change on the hook seems more reasonable as future consumers would also get this behaviour by default and avoid having to think about it. also, it seems like some components do this trim exclusively on create whereas others do it also on update - should this also be allowed on update? (ie, when renaming)

alexanderleegs:
> i think this fix leaves us open to making the same mistakes again; all these handles use an `onProceed` method call, which is obtained from `<create | update><page | directory | media>hook`.
> 
> i think making the change on the hook seems more reasonable as future consumers would also get this behaviour by default and avoid having to think about it. also, it seems like some components do this trim exclusively on create whereas others do it also on update - should this also be allowed on update? (ie, when renaming)

That makes sense, moved the logic into hooks for create/update endpoints

src/hooks/pageHooks/utils.js
```diff
@@ -0,0 +1,21 @@
+export function extractPageInfo({ data, pageData, resourceRoomName }) {
```
seaerchin, line 1: not entirely sure if this behaviour is identical to the old one as the `if (!data) return pageData` seems to be additional

src/layouts/EditPage.jsx
```diff
@@ -219,9 +219,11 @@ const EditPage = ({ match, history }) => {
           if (isXSSViolation) setShowXSSWarning(true)
           else {
             updatePageHandler({
-              frontMatter: pageData.content.frontMatter,
-              sha: currSha,
-              pageBody: editorValue,
+              pageData: {
```
seaerchin, line 222: just double checking - this is required because the hook passes the data to `extractPageInfo`, which takes `{ data, pageData, resourceRoomName }` right

seaerchin (approved):
lgtm

src/hooks/pageHooks/utils.js
alexanderleegs: This is for edits within editPage - since both are routed through this point now, we need some way to bypass this method

src/layouts/EditPage.jsx
alexanderleegs: Yep!

PR #857, opened 2022-04-22
Fix/resource category capitalisation

### LINKED ISSUE #856: [Bug] Resource category name
We currently slugify our resource category names upon creation - this causes issues with the title of the category page on the actual site.
Screenshot 2022-04-22 at 11 31 24 AM

alexanderleegs:
## Problem

This PR fixes an issue with our resource category creation - we previously handled resource category creation the same way that we did for collections, which was to slugify the provided name. This resulted in the breadcrumbs of resource categories showing up in their slugified form, as seen in #856. This PR introduces an additional check for resourceRoomName - these directory items should be handled the same way as subcollections in the frontend and only slugified in the backend so that the original desired name can be preserved in the `index.html` file created.

This PR also fixes a minor issue where we were not using the updated v2 endpoint to retrieve the staging url of a site.

Closes #856

### DIFF
--- a/src/components/DirectoryCreationModal/DirectoryCreationModal.jsx
+++ b/src/components/DirectoryCreationModal/DirectoryCreationModal.jsx
@@ -30,7 +30,12 @@ export const DirectoryCreationModal = ({
   onProceed,
   showSelectPages,
 }) => {
-  const { siteName, collectionName, mediaDirectoryName } = params
+  const {
+    siteName,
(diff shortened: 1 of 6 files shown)


seaerchin (approved):
i think the current implementation is starting to get abit messy but i can still live with it as it's not worth the additional effort to refactor, which would involve the parent component also @__@

PR #866, opened 2022-04-28
ref(design-system): replace input and some buttons

seaerchin:
## Problem
Previously, we had no padding on input owing to emotion resetting padding to 0. This caused inputs to look squashed. In addition, the border colour of inputs were wrong (black) rather than grey due to `reset-css`. This also fixes some UI issues with preview (h1-6 on homepage/contact).

## Solution
1. change base components to use chakra/design system components
2. shift h1-6 styling into own css class and other previews extend the class
3. remove reset-css (there's a similar chunk in `isomer-templates.scss` but it's `minireset`)

### DIFF
--- a/craco.config.js
+++ b/craco.config.js
@@ -1,4 +1,14 @@
+const merge = require("lodash/merge")
+
+const customJestConfig = require("./jest.config.ts")
+
 module.exports = {
+  jest: {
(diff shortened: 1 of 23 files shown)


alexanderleegs (approved):
Placeholder text and disabled input currently look the same - other than that lgtm

PR #871, opened 2022-04-29
fix(e2e): release 0.7.0

seaerchin:
## Problem
This PR fixes a flaky e2e test in cypress by changing from a fixed wait time to using an assertion to ensure that the element is no longer in the Dom.

This PR also fixes the check for resource category so that special symbols are no longer allowed.

Also fixes our ui theming from design system
## Solution
1. Changed validation for resources to use the same one as originally
2. Changed wait time to assertion
3. Extend them from ogp design system (I mistakenly thought calling`extendTheme` was sufficient but we need to also add the base design system theme


rc-davis (approved):
LGTM

PR #878, opened 2022-05-10
Fix/add wait for dom

alexanderleegs:
## Problem

This PR introduces a potential fix to some of our flaky e2e tests - some of our tests are currently failing because changes in the DOM have not been fully completed before the check for the new component. A potential fix for this issue, `waitForDom`, has been introduced from [here](https://gist.github.com/xeger/8e37d241ea6c2210a8e0dcc857b4792e).

As this issue is not particularly urgent, and we intend to refactor our e2e tests in the near future, this PR introduces `waitForDom` only to our move tests, which tend to be more flaky - we will continue to monitor whether this reduces the flakiness of our tests.


rc-davis (approved):
Looks awesome.

Issue #861, opened 2022-04-26
[Design System] Migrate existing Footer to ChakraUI


seaerchin:
**work required**
isomer's current `Footer` component is a thin wrapper over 2 buttons; this could be better written using layout components from ChakraUI and pulling some state downwards into the component itself (eg: fix the text labels on buttons; always have 2 buttons on `Footer`) 

This component could either be inlined into its relevant call-sites or to be refactored and presented with a simpler API

Refer to the base figma [here](https://www.figma.com/file/7OtNYkc4s9BkFk05ConEbd/V1.5?node-id=285%3A23615) and an example of the use case [here](https://www.figma.com/file/7OtNYkc4s9BkFk05ConEbd/V1.5?node-id=269%3A30764). Note that the use case lacks the cancel button.

Also, do followup with the designers if it should be a `ButtonGroup` or two separate `Buttons` (eg: if loading -> both load?)

Do also take note that the existing `.sectionFooter/pageEditor` css class **should be replaced** (might have to seek @NatMaeTan for design input here)

seaerchin:
this issue is tangentially solved by the mentioned PR/commits - the component is using chakra UI right now but some components have a `div` wrapping that makes it look like a footer; we have to iron those out

NatMaeTan:
> Also, do followup with the designers if it should be a ButtonGroup or two separate Buttons (eg: if loading -> both load?)

Should be a button group, only 1 button loads at a time

> Do also take note that the existing .sectionFooter/pageEditor css class should be replaced (might have to seek @NatMaeTan for design input here)
Lmk what you'd need

PR #885, opened 2022-05-17
ref(design-system): misc fixes for components

seaerchin:
## Problem
This PR adds in a couple of misc fixes for styling/form validation and removes unused component (clarified with @NatMaeTan on this) 

Related to #861 

## Solution
1. Added extra padding so that buttons don't overflow the sidebar when clicked
2. Changed form validation mode from `onBlur` to `onTouched` (refer to [here](https://react-hook-form.com/api/useform) for more details) - instead of validating when the user clicks away, the validation is comprised of two stages. initial validation is the same (when user clicks away) but subsequent validation is `onChange`
3. For `Footer`, the `Cancel` option was **removed** to keep the footer uniform across every page.
    - this is **temporary**; the footer **should** have a `Cancel` option but this complicates the usage of `Footer` on `Settings` so this is kept for now till `Footer` is refactored


alexanderleegs (approved):
lgtm

Issue #859, opened 2022-04-26
[Design System] Migrate buttons to design system buttons


seaerchin:
**Work required**
remove existing instances of button and replace them all with design system buttons. do keep in mind that the buttons vary across pages (see [here](https://www.figma.com/file/7OtNYkc4s9BkFk05ConEbd/V1.5?node-id=142%3A4022)) and hence, styling has to be adjusted across to fit the page styling.

This does **not** extend to complex components that _behave_ like buttons (eg: cards); those are separated out into their own issue. 

This issue **does** include the existing component(s): 

- [ ] `LoadingButton`

PR #887, opened 2022-05-18
ref(design-system) refactor loading buttons

seaerchin:
## Problem
This PR refactors `LoadingButton` and deletes `SaveDeleteButton`. `LoadingButton` encapsulates the loading state of the button on firing of the callback but this should not be the responsibility of the button as hte caller will already know if the callback is loading. Artificially creating a `useEffect` to encapsulate this loading state means that the button now has some internal logic where there wasn't previously and that the same piece of information is computed twice (eg: in our modals, where we use `mutateAsync`, there's already an `isLoading` prop from `useQuery` that isn't being used). This also means that some components differ in usage (`Footer` uses basic `Buttons` but in some cases, it cannot be used because a `LoadingButton` is used)

This refactor first gets `LoadingButton` to a button like API, and shifts the styling to design system and to typescript. A refactor to get rid of the loading state entirely would be more intricate and will take place in the future.

With this PR, `SaveDeleteButtons` are also gone as they are simply 2 stylized loading buttons.

Part 4 of #859 

## Solution
1. Replace `LoadingButton` props with ChakraUI props
2. Replace base `button` with design system `Button` 
3. migrate to TS
4. update callsites for `LoadingButton`
5. For `SaveDeleteButton`, the callsites were refactored to use either `HStack`/`Flex` component to emulate the deleted component. Styling was also updated (1 clear/1 solid - checked with @NatMaeTan). 

**Notes**
1. For **modals**, the buttons should no longer have a cut-off border. the styling has been adjusted to give slightly more padding (4px/1 unit) as the cut off border is significantly more prominent than a button that's slightly off alignment (fyi @NatMaeTan)


alexanderleegs (approved):
lgtm!

seaerchin:
@alexanderleegs rebasing because i accidentally pushed some commits meant for a new branch

PR #889, opened 2022-05-23
ref(design-system): fix e2e tests

seaerchin:
## Problem
This PR refactors `LoadingButton` and fixes/deletes old e2e.

## Solution
1. refactored `LoadingButton` so that we don't use `useEffect`. This PR also fixes some e2e tests by changing the selection code to select button containing the text; this is to avoid selecting the `p` element containing the text.


alexanderleegs (approved):
lgtm

PR #890, opened 2022-05-23
Fix: prevent operating hours from showing up if blank

alexanderleegs:
This PR fixes a minor bug involving operating hours - we were previously leaving an empty array if there were no operating hours, but this caused an empty operating hours header to show up on the actual site. This PR removes the operating_hours field if no hours are specified.

### DIFF
--- a/src/layouts/EditContactUs.jsx
+++ b/src/layouts/EditContactUs.jsx
@@ -14,8 +14,6 @@ import PropTypes from "prop-types"
 import { createRef, useEffect, useState } from "react"
 import { DragDropContext } from "react-beautiful-dnd"
 
-import "styles/isomer-template.scss"
-
 // Import hooks
 import useRedirectHook from "hooks/useRedirectHook"
 import useSiteColorsHook from "hooks/useSiteColorsHook"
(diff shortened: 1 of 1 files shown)


seaerchin (commented):
being quite annoying here but i think this is a bandaid fix. after looking through the file, the function itself has ~80 lines and it's written quite opaquely - wdyt about rewriting it so that we don't do `cloneDeep/forEach/push/isEmpty` repeatedly for each subfield and instead just extracting this logic out as a function which applies across to teh whole structure? (if so, we can shift this out into a typescript util file too)

alexanderleegs:
> being quite annoying here but i think this is a bandaid fix. after looking through the file, the function itself has ~80 lines and it's written quite opaquely - wdyt about rewriting it so that we don't do `cloneDeep/forEach/push/isEmpty` repeatedly for each subfield and instead just extracting this logic out as a function which applies across to teh whole structure? (if so, we can shift this out into a typescript util file too)

So it appears that it might be more difficult than expected - simply removing all empty fields/objects appears to cause the entire section to not be rendered - some fields need to be present, even if they are blank.

seaerchin (approved):

alexanderleegs:
Closing - resolved on template instead of cms

PR #838, opened 2022-04-10
build(deps): upgrade marked

prestonlimlianjie:
This PR upgrades marked to v4.0.12 and uses the new API (`marked.parse()` vs the old `marked()`).

PR #891, opened 2022-05-24
fix(marked): fixed import

seaerchin:
## Problem
#838 changed `marked` to use a new major version and updated teh API but missed out on the import. this resulted in a bug where we are unable to use the edit page as it will crash.

## Solution
fixes the import (see [here](https://marked.js.org/) under the nodeJS section)


alexanderleegs (approved):
lgtm, thanks for catching!

PR #895, opened 2022-05-26
Refactor/design system/isolate global styles

### LINKED ISSUE #864: [Design system] Migrate Toast to design system
**work required**
This should just be a drop in replacement so minimal work required.

alexanderleegs:
## Problem
This PR address 2 main issues:

- Swaps our toasts to use the design system toasts
- Isolates our imported styling (`sgds` and `isomer-template`) used for preview components.

Resolves #864 

This PR also removes the library `react-toastify` and swaps `node-sass` to `sass`.

## Context
One existing issue with the css in our codebase was the need to import `isomer-template` and `sgds` css in order to accurately replicate the preview of a website. Due to the global nature of css, this resulted in the use of these styles being used across our components. The issue arose from these css having many global type based selectors (e.g. `h1` or `p`) and the frequent use of `!important`, which may have caused these styles to be unknowingly used in multiple different components. Most notably, the styling for the newly introduced design-system toast was extremely different, due to the global styling introduced.

## Solution
The solution to this would be to scope our css to be used only in the relevant components. For our `isomer-template` css, we are unable to swap out all of the relevant css, as we have added in additional css over time which other components rely on. As a compromise, we have isolated all the global type selectors, as the other styling has been scoped by classname, and is unlikely to affect unwanted components. We have almost removed `!important`  from several selectors.

For the imported `sgds` styles, this was a separate issue - since imports were global by nature, importing this inside a stylesheet would still result in the styles being used globally. The solution to this was to use `sass` instead, which allows for the use of `@use` - this allowed us to keep the scope of the sgds styling to within our preview components.

Note that this change has resulted in a few differences in styling, most notably in the homepage and contact us preview components:

**BEFORE**:
Screenshot 2022-05-26 at 11 23 14 AM

Screenshot 2022-05-26 at 11 22 12 AM
Screenshot 2022-05-26 at 11 21 58 AM

**AFTER**:
Screenshot 2022-05-26 at 11 22 57 AM
Screenshot 2022-05-26 at 11 22 20 AM
Screenshot 2022-05-26 at 11 22 39 AM

However, the changed font size/hyperlink colour actually reflects what users would see on the actual site, where it would not before.

## Tests

Tests have been updated where they referred to old selectors which have been scoped, e.g. `bp-notification`


src/layouts/EditHomepage.jsx
```diff
@@ -1405,16 +1406,45 @@ const EditHomepage = ({ match }) => {
...
-                    <div className="col">
-                      <div className="field has-addons bp-notification-flex">
-                        <div className="control has-text-centered has-text-white">
+              <div
+                id="notification-bar"
+                className={getClassNames(
+                  editorStyles,
+                  ["bp-notification", "is-marginless"],
+                  ["bg-secondary"]
+                )}
```
seaerchin, line 1415: could i clarify why we are getting the first two from `editorStyles` but not so for `bg-secondary`?

src/styles/isomer-cms/elements/card.scss
```diff
@@ -194,9 +194,19 @@
...
   }
 
+  &.siteNotificationSection {
+    h2 {
+      font-size: 1.25rem;
+      line-height: 2rem;
+      margin-bottom: 5px;
+    }
+    
+  }
```
seaerchin, line 208: this style is effectively limited to `h2` that are below `.siteNotificationSection` - my worry here is if we add code here in the future and `siteNotificationSection` ends up becoming a wrapper, it could negatively affect nested h2s.

Should we instead scope it so that this is a single classname?

src/styles/isomer-cms/pages/Editor.module.scss
```diff
@@ -1,5 +1,11 @@
+@use "sass:meta";
+
 @import "../elements/variable";
 
+.preview-styles {
+  @include meta.load-css("../../preview-panel.scss");
```
seaerchin, line 6: from the docs located [here](https://sass-lang.com/documentation/modules/meta), this means that the css under `preview-panel.scss` would be scoped under `.preview-styles` right?

src/templates/contact-us/ContactUsHeader.jsx
```diff
@@ -1,22 +1,66 @@
...
+          ["row", "is-inverted"]
+        )}
+      >
+        <div
+          className={getClassNames(
+            editorStyles,
+            ["col", "is-8", "is-offset-2"],
+            ["col", "is-8", "is-offset-2"]
+          )}
+        >
```
seaerchin, line 35: could i check why the styles are repeated twice? are there some styles globally that are not repeated within scope? (if so, can't we just paste them into local scope and delete teh global one?)

src/templates/contact-us/ContactUsHeader.jsx
```diff
@@ -1,22 +1,66 @@
...
+            ])}
+          >
             Get in touch with
             <br />
-            <span className="display has-text-secondary">{agencyName}</span>
+            <span
+              className={getClassNames(
+                editorStyles,
+                ["display"],
+                ["has-text-secondary"]
```
seaerchin, line 59: wdyt about grouping `has-text-secondary` into a local style

src/templates/contact-us/LocationsSection.jsx
```diff
@@ -36,30 +46,61 @@ const LocationAddress = ({ location }) => (
...
         target="_blank"
       >
         <div>
           <span>VIEW MAP</span>
-          <i className="sgds-icon sgds-icon-arrow-right" aria-hidden="true" />
+          <i
+            className={getClassNames(
+              editorStyles,
+              ["sgds-icon", "sgds-icon-arrow-right"],
+              ["sgds-icon"]
```
seaerchin, line 63: we should aim to merge `sgds-icon`

src/templates/homepage/HeroSection.jsx
```diff
@@ -80,17 +109,30 @@ const HeroDropdown = ({ title, options, isActive, toggleDropdown }) => (
 const KeyHighlightElem = ({ title, description }) => (
   <>
     {title || description ? (
-      <div className="col">
-        <div className="is-highlight">
+      <div className={getClassNames(editorStyles, ["col"], ["col"])}>
```
seaerchin, line 112: we should aim to merge `col`

src/templates/homepage/HeroSection.jsx
```diff
@@ -101,9 +143,21 @@ const KeyHighlightElem = ({ title, description }) => (
...
-    <div className="bp-container">
-      <div className="row is-gapless has-text-centered">
+  <section
+    id={editorStyles["key-highlights"]}
+    className={getClassNames(editorStyles, ["bp-section is-paddingless"])}
+  >
+    <div
+      id="key-highlights"
+      className={getClassNames(editorStyles, ["bp-container"])}
+    >
```
seaerchin, line 153: not sure if this will impact e2e but the `section` and and the `div` have different ids after compared to before

src/templates/homepage/HeroSection.jsx
```diff
@@ -136,18 +190,55 @@ const TemplateHeroSection = (
...
-      <section className="bp-hero bg-hero" style={heroStyle}>
-        <div className="bp-hero-body">
-          <div className="bp-container margin--top--lg">
-            <div className="row is-vcentered is-centered ma">
-              <div className="col is-9 has-text-centered has-text-white">
-                <h1 className="display padding--bottom--lg margin--none">
-                  <b className="is-hidden-touch">{hero.title}</b>
-                  <b className="is-hidden-desktop">{hero.title}</b>
+      <section
+        className={getClassNames(editorStyles, ["bp-hero", "bg-hero"])}
```
seaerchin, line 194: could we merge the classname?

src/templates/homepage/InfobarSection.jsx
```diff
@@ -35,13 +62,36 @@ const TemplateInfobarSection = (
...
+              "col",
+              "is-offset-one-third",
+              "is-one-third",
+            ])}
+          >
+            <div
+              className={getClassNames(
+                editorStyles,
+                ["bp-sec-button"],
+                ["bp-sec-button"]
```
seaerchin, line 84: could we merge the classnames?

src/templates/homepage/InfobarSection.jsx
```diff
@@ -35,13 +62,36 @@ const TemplateInfobarSection = (
...
+              )}
+            >
               <div>
                 <span>{button}</span>
                 <i
-                  className="sgds-icon sgds-icon-arrow-right is-size-4"
+                  className={getClassNames(
+                    editorStyles,
+                    ["sgds-icon", "sgds-icon-arrow-right", "is-size-4"],
+                    ["sgds-icon"]
```
seaerchin, line 93: could we merge the `sgds-icon` declaration? this is also repeated above

src/templates/homepage/InfopicLeftSection.jsx
```diff
@@ -37,32 +39,69 @@ const TemplateInfopicLeftSection = (
...
+              >
                 <b>{title}</b>
               </h1>
               <p>{description}</p>
-              <div className="bp-sec-button margin--top padding--bottom">
+              <div
+                className={getClassNames(
+                  editorStyles,
+                  ["bp-sec-button", "margin--top padding--bottom"],
+                  ["bp-sec-button"]
```
seaerchin, line 86: could we merge the `bp-sec-button` classname?

src/templates/homepage/InfopicLeftSection.jsx
```diff
@@ -37,32 +39,69 @@ const TemplateInfopicLeftSection = (
...
+              >
                 {button ? (
                   <div>
                     <span>{button}</span>
                     <i
-                      className="sgds-icon sgds-icon-arrow-right is-size-4"
+                      className={getClassNames(
+                        editorStyles,
+                        ["sgds-icon", "sgds-icon-arrow-right", "is-size-4"],
+                        ["sgds-icon"]
```
seaerchin, line 96: could we merge the `sgds-icon` declaration?

src/templates/homepage/InfopicLeftSection.jsx
```diff
@@ -71,26 +110,64 @@ const TemplateInfopicLeftSection = (
...
+              >
                 <b>{title}</b>
               </h1>
               <p>{description}</p>
-              <div className="bp-sec-button margin--top padding--bottom">
+              <div
+                className={getClassNames(
+                  editorStyles,
+                  ["bp-sec-button", "margin--top", "padding--bottom"],
+                  ["bp-sec-button"]
```
seaerchin, line 144: something something merge declaration

src/templates/homepage/InfopicLeftSection.jsx
```diff
@@ -71,26 +110,64 @@ const TemplateInfopicLeftSection = (
...
+              >
                 {button ? (
                   <div>
                     <span>{button}</span>
                     <i
-                      className="sgds-icon sgds-icon-arrow-right is-size-4"
+                      className={getClassNames(
+                        editorStyles,
+                        ["sgds-icon", "sgds-icon-arrow-right", "is-size-4"],
+                        ["sgds-icon"]
```
seaerchin, line 154: something something merge declaration

src/templates/homepage/InfopicLeftSection.jsx
```diff
@@ -99,26 +176,72 @@ const TemplateInfopicLeftSection = (
...
```
seaerchin, line 229: something something merge declaration

src/templates/homepage/InfopicRightSection.jsx
```diff
@@ -37,32 +39,69 @@ const TemplateInfopicRightSection = (
...
+              >
                 <b>{title}</b>
               </h1>
               <p>{description}</p>
-              <div className="bp-sec-button margin--top padding--bottom">
+              <div
+                className={getClassNames(
+                  editorStyles,
+                  ["bp-sec-button", "margin--top", "padding--bottom"],
+                  ["bp-sec-button"]
```
seaerchin, line 86: something something merge declaration

src/templates/homepage/InfopicRightSection.jsx
```diff
@@ -37,32 +39,69 @@ const TemplateInfopicRightSection = (
...
+              >
                 {button ? (
                   <div>
                     <span>{button}</span>
                     <i
-                      className="sgds-icon sgds-icon-arrow-right is-size-4"
+                      className={getClassNames(
+                        editorStyles,
+                        ["sgds-icon", "sgds-icon-arrow-right", "is-size-4"],
+                        ["sgds-icon"]
```
seaerchin, line 96: something something merge declaration

src/templates/homepage/InfopicRightSection.jsx
```diff
@@ -71,26 +110,63 @@ const TemplateInfopicRightSection = (
...
+              >
                 <b>{title}</b>
               </h1>
               <p>{description}</p>
-              <div className="bp-sec-button margin--top padding--bottom">
+              <div
+                className={getClassNames(
+                  editorStyles,
+                  ["bp-sec-button", "margin--top padding--bottom"],
+                  ["bp-sec-button"]
```
seaerchin, line 158: something something merge declaration

src/templates/homepage/InfopicRightSection.jsx
```diff
@@ -71,26 +110,63 @@ const TemplateInfopicRightSection = (
...
+              >
                 {button ? (
                   <div>
                     <span>{button}</span>
                     <i
-                      className="sgds-icon sgds-icon-arrow-right is-size-4"
+                      className={getClassNames(
+                        editorStyles,
+                        ["sgds-icon", "sgds-icon-arrow-right", "is-size-4"],
+                        ["sgds-icon"]
```
seaerchin, line 168: something something merge declaration

src/templates/homepage/InfopicRightSection.jsx
```diff
@@ -99,26 +175,72 @@ const TemplateInfopicRightSection = (
...
+              >
                 <b>{title}</b>
               </h1>
               <p>{description}</p>
-              <div className="bp-sec-button margin--top padding--bottom">
+              <div
+                className={getClassNames(
+                  editorStyles,
+                  ["bp-sec-button", "margin--top", "padding--bottom"],
+                  ["bp-sec-button"]
```
seaerchin, line 232: something something merge declaration

src/templates/homepage/InfopicRightSection.jsx
```diff
@@ -99,26 +175,72 @@ const TemplateInfopicRightSection = (
...
+              >
                 {button ? (
                   <div>
                     <span>{button}</span>
                     <i
-                      className="sgds-icon sgds-icon-arrow-right is-size-4"
+                      className={getClassNames(
+                        editorStyles,
+                        ["sgds-icon", "sgds-icon-arrow-right", "is-size-4"],
+                        ["sgds-icon"]
```
seaerchin, line 242: something something merge declaration

src/templates/homepage/ResourcesSection.jsx
```diff
@@ -70,13 +122,42 @@ const TemplateResourceSection = (
...
+              "col",
+              "is-offset-one-third",
+              "is-one-third",
+            ])}
+          >
+            <div
+              className={getClassNames(
+                editorStyles,
+                ["bp-sec-button"],
+                ["bp-sec-button"]
```
seaerchin, line 144: something something merge declaration

src/templates/homepage/ResourcesSection.jsx
```diff
@@ -70,13 +122,42 @@ const TemplateResourceSection = (
...
+                  "justify-content-center",
+                ])}
+              >
                 <span>{button || "MORE"}</span>
                 <i
-                  className="sgds-icon sgds-icon-arrow-right is-size-4"
+                  className={getClassNames(
+                    editorStyles,
+                    ["sgds-icon", "sgds-icon-arrow-right", "is-size-4"],
+                    ["sgds-icon"]
```
seaerchin, line 159: something something merge declaration

src/utils/siteColorUtils.js
```diff
@@ -48,6 +48,10 @@ const createPageStyleSheet = (repoName, primaryColor, secondaryColor) => {
     `#key-highlights .col { background-color: ${primaryColor} !important;}`,
     0
   )
+  customStyleSheet.insertRule(
+    `#key-highlights .col:hover { background-color: ${primaryColor} !important;}`,
+    0
+  )
```
seaerchin, line 54: could i double check on why we need this? this seems to be because our preview is stateless right?

seaerchin (commented):
gr8 job in isolating the styles but i feel that having _repeated_ styles (1 for local, 1 for global) is confusing and liable to errors - eg: if we were repeating `sgds-icon` in both local and global styles, it makes sense to inline both into our local declaration and omit the global one. 

if someone were to use `sgds-icon` locally and forget the global declaration, then it'd be a visual bug isn't it? 

i think ideally we'd shift all the global names into a local namespace as well to avoid cluttering and styles interfering with each other but as they are all classnames, i think it's too much work for too little reward

src/utils/siteColorUtils.js
seaerchin: ideally, this should be the job of a theme provider at the root of our preview class - this might be a premature optimization for now so i'm fine with it as-is but wdyt about filing an issue for this?

src/layouts/EditHomepage.jsx
alexanderleegs: `bg-secondary` only affects colours, and we're attempting to override the colours with site specific ones using our site colours hook - so I've opted to leave it out so that we don't run into any css specificity issues, especially since our imported styles make use of `!important` frequently

src/templates/contact-us/ContactUsHeader.jsx
alexanderleegs: Unfortunately, we do use the global styles elsewhere as well (editPage and associated components), and we have a relatively large number of additional such components - I've opted to leave it as is in the interest of time

src/templates/contact-us/LocationsSection.jsx
```diff
@@ -36,30 +46,61 @@ const LocationAddress = ({ location }) => (
               )
         }
         onClick={(event) => event.preventDefault()}
-        className="bp-sec-button has-text-secondary margin--top"
+        className={getClassNames(
+          editorStyles,
+          ["bp-sec-button", "margin--top"],
+          ["bp-sec-button", "has-text-secondary"]
```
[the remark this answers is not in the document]
alexanderleegs: This isn't really possible under our current implementation of `siteColorUtils` - we look for `bp-sec-button` in the custom style sheet we introduce, and we need to specify a class name due to specificity (the sgds css also contains `!important`, so we still need to beat it in specificity)

The alternative would be to introduce id to this element? But I'm not sure if that's something we should be doing at this point

src/templates/contact-us/LocationsSection.jsx
alexanderleegs: See above, `siteColorUtils` issue

src/templates/homepage/HeroSection.jsx
alexanderleegs: See above, siteColorUtils issue

src/templates/homepage/HeroSection.jsx
alexanderleegs: I've verified that e2e tests still work! The `key-highlights` inserted in the div is to make up for `key-highlights` no longer being useable in the section

src/templates/homepage/HeroSection.jsx
alexanderleegs: siteColorUtils issue

src/styles/isomer-cms/elements/card.scss
alexanderleegs: That's a good point, modified in https://github.com/isomerpages/isomercms-frontend/pull/895/commits/1b98237e8c3907bf6661b07cd1c50a42e53ec781

src/templates/contact-us/ContactUsHeader.jsx
alexanderleegs: See above for the discussion on editorStyles and our solution!

src/utils/siteColorUtils.js
alexanderleegs: Yeah I think this would be useful, but out of scope and not as urgent  - I've filed an issue for this! #915

src/utils.js
```diff
@@ -612,3 +627,21 @@ export const getDecodedParams = (params) =>
     }
     return acc
   }, {})
+
+export const getClassNames = (styles, classes) => {
```
seaerchin, line 631: as this function is only used in templates, perhaps we should consider shifting it to `templates/utils/<some_filename>.ts` (and migrate it to ts)

seaerchin: also, if `styles` is going to be `editorStyles` most of the time, we should set it as a default argument so callers can forget about it. this does mean that we need to document (as a JSDoc) when the `styles` might change so that callers are cognizant when using this function

src/utils.js
```diff
@@ -612,3 +627,21 @@ export const getDecodedParams = (params) =>
...
+    } else if (OVERRIDDEN_CLASS_NAMES.includes(className)) {
+      // Styles which affect our inserted styles - require both our css and the imported css
+      styledClasses.push(className)
+      styledClasses.push(styles[className])
+    } else {
+      styledClasses.push(styles[className])
+    }
+  })
+  return classNames(styledClasses)
+}
```
seaerchin, line 647: wdyt about rewriting it so that the intention is clearer?

```ts
  return classes
    .filter((className) => !UNSTYLED_CLASS_NAMES.includes(className))
    .map((className) => styles[className])
    .concat(_.intersection(classes, OVERRIDDEN_CLASS_NAMES))
```

seaerchin (commented):
we should shift the file to the most specific relevant folder so that only those who need it will see it. we should also write it in ts as it's low overhead adn helps with future usage.

src/utils.js
alexanderleegs: Makes sense, https://github.com/isomerpages/isomercms-frontend/pull/895/commits/6184d6c519cb1d039d70f6606228740a773b0bd5

src/templates/utils/stylingUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const getClassNames = (styles: StylesProps, classes: string[]) => {
+  // Generates relevant class, with classes using the provided styles
+  return classNames(
+    classes
+      .filter((className) => !UNSTYLED_CLASS_NAMES.includes(className))
+      .map((className) => styles[className])
+      .concat(_.intersection(classes, OVERRIDDEN_CLASS_NAMES))
+      .concat(_.intersection(classes, UNSTYLED_CLASS_NAMES))
```
seaerchin, line 30: could you help me understand why we still need this? if we omit hte classname, won't our internal classname take precedence by default?

src/templates/utils/stylingUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+  "has-text-secondary",
+  "bg-secondary",
+  "is-secondary",
+]
+
+interface StylesProps {
+  [key: string]: string
+}
+
+export const getClassNames = (styles: StylesProps, classes: string[]) => {
```
seaerchin, line 23: nit: you could use `Record<string, string>` here

seaerchin (approved):
seems good, minor comments about small stuff

src/templates/utils/stylingUtils.ts
seaerchin: resolved offline - `UNSTYLED_CLASS_NAMES` are css classes which are dynamically injected into our stylesheet at runtime (`siteColorUtils`); we'll add a comment + rename the variable for clarity

src/templates/utils/stylingUtils.ts
alexanderleegs: Fixed in https://github.com/isomerpages/isomercms-frontend/pull/895/commits/bd1b4ec8b9b33e76d0d3de23bf049da198b87bfc

src/templates/utils/stylingUtils.ts
alexanderleegs: Renamed in https://github.com/isomerpages/isomercms-frontend/pull/895/commits/c95be74437ce74891ad28a3b66de6d32deaf206b

PR #896, opened 2022-05-26
Refactor/design system/refactor toast

alexanderleegs:
This PR swaps react-toastify for the design system toasts.


src/utils/toasts.jsx
```diff
@@ -1,26 +1,19 @@
...
-export function successToast(message) {
-  toast(<Toast notificationType="success" text={message || `Success!`} />, {
-    className: `${elementStyles.toastSuccess}`,
-    toastId: "success",
+export const useSuccessToast = () =>
+  useToast({
+    description: "Success!",
+    status: "success",
+    duration: 5000,
+    isClosable: true,
```
seaerchin, line 18: could we have a `DEFAULT_TOAST_PROPS` and then use it for these 2? also wdyt about adding a `useWarningToast` as the toast in design system has 3 variants (but it'll be unused so i'm fine w/ not adding also)

seaerchin (approved):
lgtm pending 1 minor comment

PR #899, opened 2022-05-26
fix(package): bump node-sass to 4.14

seaerchin:
## Problem
bump node-sass version to 4.14 so that release builds properly (see #898 on 404)


alexanderleegs (approved):
lgtm

Issue #902, opened 2022-05-27
[FEATURE REQUEST] Add warning banner component to inform users of errors


seaerchin:
# Problem
Isomer requires a code deployment in order to update users of potential changes. this is problematic when github goes down as isomer would then be unable to deploy said code changes.

it also adds overhead as engineers need to manually add code and trigger a deploy 

# Solution
This could be solved by: 

1. adding a `WarningBanner` component that appears on **most** pages (`login` + `workspace` + potentially `fallback/notfound`)
2. Have the mechanism for displaying the text for the `WarningBanner` be **independent of code deploy**
3. Have a mechanism for changing the type of `WarningBanner`

PR #904, opened 2022-05-27
feat(routeselector): add banner

seaerchin:
## Problem
see #902

Related to #902 

## Solution
Add DS banner that reads from env var; follow up PR will have validation and a `useEnv` hook


alexanderleegs (approved):
lgtm - there's css styling issues, but we'll file an issue for this and tackle it another time

PR #913, opened 2022-05-31
Test/add helper methods

alexanderleegs:
This PR introduces helper methods for our cypress tests, by adding a new `clickAndWait` command which makes use of `waitForDom`. We also override the existing visit method to ensure that the verify user modal has disappeared before any actions are taken. Only the existing move tests, which already contain waitForDom, have been replaced with the new `clickAndWait` command, as this action should only be taken if we need to wait for DOM changes to take effect.

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -1,0 +1,32 @@
+import waitForDom from "./waitForDom"
+
+Cypress.Commands.overwrite("visit", (originalFn, url, options) => {
+  // originalFn is the existing `visit` command that you need to call
+  // and it will receive whatever you pass in here.
+  //
+  // make sure to add a return here!
+  originalFn(url, options)
+  cy.wait(2 * 1000)
+  return cy.get(`[data-testid="verify-user-modal"]`).should("not.exist")
+})
+
+Cypress.Commands.add(
+  "clickAndWait",
+  {
+    prevSubject: true,
+  },
+  (subject, options) => {
+    // the previous subject is automatically received
+    // and the commands arguments are shifted
+
+    // allow us to change the console method used
+    cy.wrap(subject).click(options)
+    waitForDom()
+
+    // whatever we return becomes the new subject
+    //
+    // we don't want to change the subject so
+    // we return whatever was passed in
+    return cy.wrap(subject)
+  }
+)
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -7,7 +7,7 @@ Cypress.Commands.overwrite("visit", (originalFn, url, options) => {
   // make sure to add a return here!
   originalFn(url, options)
   cy.wait(2 * 1000)
-  return cy.get(`[data-testid="verify-user-modal"]`).should("not.exist")
+  return cy.get(`[data-cy="verify-user-modal"]`).should("not.exist")
 })
 
 Cypress.Commands.add(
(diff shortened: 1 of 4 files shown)


src/components/VerifyUserDetailsModal.jsx
```diff
@@ -213,7 +213,7 @@ const VerifyUserDetailsModal = () => {
   }
 
   return userId && (!loggedInEmail || !loggedInContactNumber) ? (
-    <div className={elementStyles.overlay}>
+    <div data-testid="verify-user-modal" className={elementStyles.overlay}>
```
seaerchin, line 216: nit: i think we are already using `data-cy` for test-related ids - could we do it here too? if we are removing the `visit` overwrite, we should also remove this

cypress/support/commands.js
```diff
@@ -0,0 +1,32 @@
+import waitForDom from "./waitForDom"
+
+Cypress.Commands.overwrite("visit", (originalFn, url, options) => {
+  // originalFn is the existing `visit` command that you need to call
+  // and it will receive whatever you pass in here.
+  //
+  // make sure to add a return here!
+  originalFn(url, options)
+  cy.wait(2 * 1000)
+  return cy.get(`[data-testid="verify-user-modal"]`).should("not.exist")
```
seaerchin, line 10: could i double-check on why we are over-writing the `visit` command for cypress? it doesn't seem to be required and it seems to impose an additional 2s lag on every `visit`

cypress/support/commands.js
```diff
@@ -0,0 +1,32 @@
...
+
+    // allow us to change the console method used
+    cy.wrap(subject).click(options)
+    waitForDom()
+
+    // whatever we return becomes the new subject
+    //
+    // we don't want to change the subject so
+    // we return whatever was passed in
+    return cy.wrap(subject)
```
seaerchin, line 30: how does this work with being chained? if i click on a button that closes the modal, will cypress successfully return the original DOM reference or will it complain about the element being missing ._. (ideally the former so we can assert its non-existence)

cypress/support/commands.js
alexanderleegs: This is mostly for the assertion that the user modal doesn't exist before proceeding! I included the wait in case the modal took some time to appear, but can remove that if necessary

src/components/VerifyUserDetailsModal.jsx
alexanderleegs: Good point - I've changed it to `data-cy` in https://github.com/isomerpages/isomercms-frontend/pull/913/commits/48e7038677c6b0197796f1ec3839fb3a6c75da5f, though I'm open to removing it if we're removing the visit override

cypress/support/commands.js
seaerchin: hmm i think this might be worth reconsidering as we currently have ~50 invocations of `visit` in our test suite and the modal doesn't seem to be interfering w/ our e2e anyway

alexanderleegs: As discussed offline, we're going to remove this override and just monitor for now - we can come back to this in future if the modal not disappearing after a visit is a consistent problem

src/components/VerifyUserDetailsModal.jsx
alexanderleegs: Removed in https://github.com/isomerpages/isomercms-frontend/pull/913/commits/fc2f482e470441554b09768aa7fa3e807e0abf58!

seaerchin (approved):
lgtm; idk why GHA failed here but could you just re-run it to be safe? ._.

PR #925, opened 2022-06-07
[wip] Ref/design system/dropdown

seaerchin:
ready for review, will add PR description later that's why marked as draft
## Problem

Closes [insert issue #]

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


src/components/PageCard.jsx
```diff
@@ -47,7 +48,7 @@ const PageCard = ({ item, itemIndex }) => {
       return `${url}/editPage/${encodedName}`
     }
     return `/sites/${siteName}/editPage/${encodedName}`
-  }
+  }, [resourceType, resourceRoomName, siteName, encodedName])
```
alexanderleegs, line 51: Should `url` also be a dependency?

src/components/FolderCard/FolderCard.jsx
```diff
@@ -51,7 +51,7 @@ export const FolderCard = ({
       default:
         return ""
     }
-  }, [pageType])
+  }, [pageType, siteName, category])
```
alexanderleegs, line 54: As above, should be dependent on url as well?

alexanderleegs (commented):
Looks great, left a few small comments

Also, minor nit - the button is slightly off centre:
Screenshot 2022-06-09 at 10 37 16 AM

src/components/PageCard.jsx
seaerchin: Yes, good catch! Oddly enough there's no eslint for this rven though there is for `useEffect`

src/components/FolderCard/FolderCard.jsx
seaerchin: See above

src/components/PageCard.jsx
seaerchin: > Looks great, left a few small comments
> 
> Also, minor nit - the button is slightly off centre: Screenshot 2022-06-09 at 10 37 16 AM

Yeap it is and good catch! I'm aware of this but I'm not addressing it for a few reasons:

1. V1.5 figma actually doesn't have the margins for old cards! The button here is absolutely positioned wrt the card itself so we'd need to change either the position of the button or padding of the card which would be fine if not for...
2. New cards are coming up subsequently and those look fine with the drop-down! As the new cards should go out by this month, I'm not particularly concerned for this.

Tagging @NatMaeTan - any thoughts

seaerchin: over slack - no issues

alexanderleegs (approved):
lgtm!

PR #926, opened 2022-06-07
ref(design-system): Cards

seaerchin:
**NOTE: this API is not final, there's an on-going refactor**

**the majority of the lien changes come from installing chakra ui related packages**

## Problem
Isomer uses alot of card components but these components are disparate and not related. This also isn't extensible in the future to new card-like components, which necessitates a refactor into a cleaner API.

Closes [insert issue #]

## Solution
1. This component only implements 2 variations of cards - single line cards w/ a context menu/multi line cards with a context menu. it's also important to note that both of these are **buttons** and future patterns in v1.5 could make these checkboxes/purely display.
3. As we are using nested buttons (the context menu + the card is also a button), the nested button (context menu) needs to have `position: absolute`. 
4. this is done using a [grid API](https://developer.mozilla.org/en-US/docs/Web/CSS/grid) and defining template areas. There is a space of 40x40 (exactly the size of the button) reserved so that text reflows properly around teh button and there's no overlapping

## Notes
1. as this is a first pass review, the focus should be on css styling/potential breakage when using as a drop-in replacement. the wip refactor changed the API to be more complex (i really tried to KISS but cmi)


src/components/Card/Card.tsx
```diff
@@ -5,74 +5,89 @@ import {
...
-export const CardBody = (props: StackProps) => {
+export const Card = ({ children, ...props }: CardProps): JSX.Element => {
+  const variant = props.variant ?? "single"
   const styles = useMultiStyleConfig(CARD_THEME_KEY, props)
+  const buttonBehaviourProps = omitThemingProps(props)
+  const Inner = () =>
+    variant === "multi" ? (
+      <chakra.button __css={styles.container} {...props}>
+        {children}
+      </chakra.button>
```
alexanderleegs, line 47: Just for info, why do we use chakra.button here?

alexanderleegs (approved):
lgtm, had one clarification

src/components/Card/Card.tsx
seaerchin: there's actually 3 kinds of buttons!
 `<Button />` is a component that comes with the styling, so it'll have focus/hover states etc together with padding and what not 

The one above is just a basic HTML button that hooks into chakra, so passing colour tokens work and so does chakra shorthand (p for padding etc). This is also what chakra's `Button` is implemented with 

The last is just `button`, which is basically the above without chakra stuff.

I'm using `chakra.button` here because I want the __behaviour__ but not the styling as the styling should be card specific

PR #927, opened 2022-06-07
[wip] Ref/design system/workspace

seaerchin:
## Problem

Closes [insert issue #]

## Solution

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


alexanderleegs (commented):
The refactored workspace looks great, thanks for all the work on this! Have a few minor comments.

For Section/SectionCaption/SectionHeader, what do you think about consolidating these under their own separate section (maybe `components/sections` or something), so that it's easy to find grouped components? `CreateButton` could be renamed and go in here too if this is specifically used for section creation.

The other issue is a display related one for multi-line cards:
Screenshot 2022-06-09 at 2 23 20 PM

I see the figma does require multi-line cards, but the context menu button for these cases seem to be off (see the second row of cards). I was also wondering if we were okay with uneven positioning of the elements if the length of titles are different.

seaerchin:
> The refactored workspace looks great, thanks for all the work on this! Have a few minor comments.
> 
> For Section/SectionCaption/SectionHeader, what do you think about consolidating these under their own separate section (maybe `components/sections` or something), so that it's easy to find grouped components? `CreateButton` could be renamed and go in here too if this is specifically used for section creation.
> 
> The other issue is a display related one for multi-line cards: Screenshot 2022-06-09 at 2 23 20 PM
> 
> I see the figma does require multi-line cards, but the context menu button for these cases seem to be off (see the second row of cards). I was also wondering if we were okay with uneven positioning of the elements if the length of titles are different.

for the first point, yea, i did it in folders HAHA - i think it's worth relooking at how we want to structure our components but ideally we'd group them by "features"

for the second one - good catch, i actually didn't realise this! i think i need to ask @NatMaeTan for her input - this occurs because we resize the cards (for the grid) but the buttons are absolutely positioned wrt the cards so only hte affected card has proper sizing.

we could resize the whole row of cards or to simply cut text off at 1 line but i'm unsure which is best

alexanderleegs (commented):
Actually another thing I noticed - the sidebar is now scrollable and scrolls with the page, when it should be fixed

alexanderleegs (commented):
E2E tests are also failing due to the refactor - will chat with Kathleen before deciding whether to push this out before tests are fixed (we would have to run through our manual suite)

seaerchin:
@alexanderleegs gonna consolidate replies to avoid spam HAHA
> the sidebar is now scrollable and scrolls with the page, when it should be fixed

fixed in folders PR

> I see the figma does require multi-line cards, but the context menu button for these cases seem to be off (see the second row of cards). I was also wondering if we were okay with uneven positioning of the elements if the length of titles are different.

fixed - clarified with @NatMaeTan on resize vs cutoff; for multiline cards, restrict to 3 lines of text and for single, 1 line.

alexanderleegs (approved):
lgtm

PR #928, opened 2022-06-07
ref(design-system): shift folders to v1.5

seaerchin:
## Problem
Folders has an updated look in v1.5 and could be shifted. This also fixes an existing issue where our header/sidebar are non-sticky. 

As part of this rewrite, the sidebar was also shifted to v1.5 and typescript. This also meant that the hooks used in sidebar was also migrated to TS. 

## Solution
1. Shift floating action button for react-query to be on all screens
2. Add new `Sidebar` component according to v1.5 and deletes old sidebar
3. Add new hooks - `useLocalStorage` and `useLastUpdated`
    - as part of this, `LoginContext` was also rewritten to simplify some boilerplate (encapsulated within the `useLocalStorage` hook)  
5. migrated `Folders` component to v1.5 + ts
6. add new `SiteViewLayout` that composes `Header/Sidebar` for ease of providing a set layout across views


src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,218 @@
...
+  const { selectedTab, setSelectedTab } = useSidebarContext()
+  const [loggedInUser] = useLocalStorage(
+    LOCAL_STORAGE_KEYS.GithubId,
+    "Unknown user"
+  )
+
+  return (
+    <Flex bg="white" h="100vh" flexDir="column" w="15rem">
+      <Box p="1.5rem">
+        <VStack align="flex-start" spacing="0.5rem">
```
alexanderleegs, line 68: The sidebar still scrolls on smaller screens - might need to modify the height to account for the header (and possibly the banner)

alexanderleegs (commented):
Great work on the refactor! Left some comments.

src/layouts/Folders/components/MenuDropdownButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+  BiFileBlank,
+  BiFolder,
+} from "react-icons/bi"
+import { useRouteMatch, Link as RouterLink } from "react-router-dom"
+
+export const MenuDropdownButton = (): JSX.Element => {
+  const { url } = useRouteMatch()
+  const [isCreatePage, setIsCreatePage] = useState(false)
+
+  return (
```
alexanderleegs, line 23: Is this toggle behaviour intended? I was under the impression that clicking on one of the dropdown options would immediately bring up the appropriate modal, but users need to select an option from the dropdown then click the button - just checking whether this is intended

src/components/Sidebar/Sidebar.tsx
seaerchin: changed, but only accounts for header. for banner, i'd need to check to see if multiline banners are a possibility in which case, it's probably unfeasible to do (need to dynamically calc at runtime the height of the banner element)

src/layouts/Folders/components/MenuDropdownButton.tsx
seaerchin: hmm this is a good point - i assumed the current impl as it felt the most natural (otherwise it'd be 2 separate buttons) but will clarify with @NatMaeTan

src/contexts/SidebarContext.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export const SidebarContext = createContext<SidebarContextProps | undefined>(
+  undefined
+)
+
+export const SidebarContextProvider = (
+  props: ChildContextProvider<SidebarContextProps>
+): JSX.Element => {
+  const [selectedTab, setSelectedTab] = useState<TabSection>("workspace")
```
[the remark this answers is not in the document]
seaerchin: good catch, and i'm aware of this - will fix in this PR!

seaerchin:
> Great work on the refactor! Left some comments.
> 
> Also, the back button behaviour has changed on folders - it previously used to return to the level one above it (subfolder -> folder, and folder -> workspace). I see the back to all sites button on the figma, but are the designers aware of this change?

will clarify on this w/ @NatMaeTan  but i strongly prefer this behaviour as it allows us to simplify alot of flows @__@

src/components/Sidebar/Sidebar.tsx
alexanderleegs: This seems somewhat fixed now, but I'm not sure if this issue is related:

https://user-images.githubusercontent.com/22111124/173554086-1b528e36-cafe-4878-939c-ecb2a322d0b4.mov

In the above clip, it looks like there's an inner and outer scroll bar - this is causing the scrolling to snap to the end of one scroll bar, causing it to be less smooth overall. I don't think it's a huge issue, but it would be good if this problem had a quick fix

alexanderleegs (commented):
Also wanted to clarify about the header not being hidden by the modal overlay - the header buttons are now clickable even inside a settings modal, which I think might not be ideal!

src/components/Sidebar/Sidebar.tsx
seaerchin: @alexanderleegs i can fix this actually but this is super confusing because the reason the scrollbar appears is cos there's overflow (as `overflow: auto` is set). 

**however**, the overflow is NOT actually anything?! when i set overflow to visible, the overflow is actually just a grayed out area. i'm not sure what's happening here but we don't seem to be losing anything so technically this is fixable just by doing `overflow: hidden` 

idk why the overflow is just blank space tho ._.

seaerchin:
> Reference in new i

rip i actually just used the default chakra zIndex... this is fixable (band-aid) just by either 1. changing header zindex or 2. changing modal zindex but who would've thought AHHA

src/components/Sidebar/Sidebar.tsx
alexanderleegs: Sure, this looks great - remember to implement this for the remaining refactors later as well

alexanderleegs (approved):
lgtm!

src/layouts/layouts/SiteViewLayout/SiteViewHeader.tsx
```diff
@@ -0,0 +1,112 @@
...
+              </Button>
+              <Skeleton isLoaded={!isLoading}>
+                <ButtonLink href={stagingUrl}>
+                  <Text color="white">Proceed to staging site</Text>
+                </ButtonLink>
+              </Skeleton>
+            </HStack>
+          </ModalFooter>
+        </ModalContent>
+      </Modal>
```
alexanderleegs, line 94: The image for this modal is missing!

PR #932, opened 2022-06-13
Ref/design system/footer

alexanderleegs:
## Problem

This PR refactors our existing Footer component to fully use chakra components instead of styled divs.

Closes https://github.com/isomerpages/isomercms-frontend/issues/861

## Solution

Due to the variations between different footer types (variable number of buttons, button text etc), we have decided to keep the `Footer` component as a simple container, with each component which requires the footer specifying the actual footer content. The existing areas in which the footer is used have been refactored to use this new format, and the other footer styled divs in edit homepage, contact-us, and navbar have also been refactored to use this component.


src/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+interface TemplateArgs {
+  buttonText: string
+}
+
+const SingleButtonFooterTemplate: Story<TemplateArgs> = ({
+  buttonText,
+}: TemplateArgs) => {
+  return (
+    // eslint-disable-next-line react/jsx-props-no-spreading
```
seaerchin, line 19: nit: extra comment

src/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+interface MultiButtonFooterTemplateArgs extends TemplateArgs {
+  button2Text: string
+}
+const MultiButtonFooterTemplate: Story<MultiButtonFooterTemplateArgs> = ({
+  buttonText,
+  button2Text,
+}: MultiButtonFooterTemplateArgs) => {
+  return (
+    // eslint-disable-next-line react/jsx-props-no-spreading
```
seaerchin, line 34: nit: extra comment

src/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+    <Footer>
+      <Button>{buttonText}</Button>
+    </Footer>
+  )
+}
+
+interface MultiButtonFooterTemplateArgs extends TemplateArgs {
+  button2Text: string
```
seaerchin, line 27: could we change the name to avoid mixing numbers/letters in names

src/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+const MultiButtonFooterTemplate: Story<MultiButtonFooterTemplateArgs> = ({
+  buttonText,
+  button2Text,
+}: MultiButtonFooterTemplateArgs) => {
+  return (
+    // eslint-disable-next-line react/jsx-props-no-spreading
+    <Footer>
+      <>
```
seaerchin, line 36: we don't need to wrap in a fragment

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,24 @@
+import { HStack, Flex } from "@chakra-ui/react"
+
+export interface FooterProps {
+  children: JSX.Element
+}
+
+export const Footer = ({ children }: FooterProps): JSX.Element => (
+  <Flex
+    h="80px"
```
seaerchin, line 9: btw, we should specify this in rem probably. this is because rem is actually dynamically computed relative to the root element and allows responsive sizing (see [here](https://chiamakaikeanyi.dev/sizing-in-css-px-vs-em-vs-rem/))

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,24 @@
...
+    w="100%"
+    pr="30px"
+    position="fixed"
+    background="background.action.defaultInverse"
+    border="1px"
```
seaerchin, line 14: can keep border as px fyi

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,24 @@
...
+    borderColor="border.divider.alt"
+    justify="end"
+    align="center"
+    bottom="0"
+  >
+    <HStack>
+      <>{children}</>
```
seaerchin, line 21: nit: don't need to wrap in fragment

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,24 @@
...
+    pr="30px"
+    position="fixed"
+    background="background.action.defaultInverse"
+    border="1px"
+    borderColor="border.divider.alt"
+    justify="end"
+    align="center"
+    bottom="0"
+  >
+    <HStack>
```
seaerchin, line 20: just double checking here - is the intention for the children of `HStack` to be treated as a singular unit and justified to the end of the footer with 1rem spacing? 

this is the current behaviour and if design shows that the footer is consistent liek this across all pages then it's fine

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,24 @@
+import { HStack, Flex } from "@chakra-ui/react"
+
+export interface FooterProps {
+  children: JSX.Element
```
seaerchin, line 4: could we use `PropsWithChildren` from react here for btr typing?

seaerchin (commented):
minor comments but overall looks fine!

seaerchin:
code-wise seems ok, please wait for @NatMaeTan to approve design changes before merging!

src/components/Footer/Footer.stories.tsx
alexanderleegs: Thanks for catching, https://github.com/isomerpages/isomercms-frontend/pull/932/commits/701f438ffdf74c5dc631deb36a01f621f694d103

src/components/Footer/Footer.stories.tsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/932/commits/701f438ffdf74c5dc631deb36a01f621f694d103

src/components/Footer/Footer.stories.tsx
alexanderleegs: Changed in https://github.com/isomerpages/isomercms-frontend/pull/932/commits/88f6867043db46c09fcb43abfad99a66a315a766!

src/components/Footer/Footer.stories.tsx
alexanderleegs: Right, I was having issues with this earlier because of the props in Footer - but swapping to `PropsWithChildren` fixed the issue! https://github.com/isomerpages/isomercms-frontend/pull/932/commits/7faa682fd7af5140926947afef3520d12daa67fb

src/components/Footer/Footer.tsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/932/commits/7faa682fd7af5140926947afef3520d12daa67fb

src/components/Footer/Footer.tsx
alexanderleegs: Thanks for the suggestion, changed in https://github.com/isomerpages/isomercms-frontend/pull/932/commits/7faa682fd7af5140926947afef3520d12daa67fb

alexanderleegs (commented):
Tagging @audreytcy as well!

seaerchin:
@alexanderleegs you gotta assign them (@audreytcy and @NatMaeTan) to be reviewers on chromatic (ui review -> assign reviewers on the RHS); i've done it for you here but just fyi

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { PropsWithChildren } from "react"
+
+export const Footer = ({
+  children,
+}: PropsWithChildren<Record<never, any>>): JSX.Element => (
+  <Flex
+    h="5rem"
+    w="100%"
+    pr="2rem"
+    position="fixed"
```
seaerchin, line 11: i don't see a problem with this but just wanted to point out that fixed removes it from the document flow (see [here](https://www.kevinpowell.co/article/positition-fixed-vs-sticky/)). i think it might be worth asking designers if the page is guaranteed to provide space for footer (when it's positioned) or otherwise, using sticky to ensure that this element is accounted for/using grid layout to reserve space

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,21 @@
+import { HStack, Flex } from "@chakra-ui/react"
+import { PropsWithChildren } from "react"
+
+export const Footer = ({
+  children,
+}: PropsWithChildren<Record<never, any>>): JSX.Element => (
```
seaerchin, line 6: just double checking here - would this not work with `PropsWithChildren<{}>`? it's a very minor change so dw about it but curious

seaerchin (approved):

src/components/Footer/Footer.tsx
alexanderleegs: I guess it does too! But I was getting a linting error, and the suggestion for empty objects was to do something like this instead

src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,23 @@
+import { HStack, Flex, FlexProps } from "@chakra-ui/react"
+import { PropsWithChildren } from "react"
+
+export const Footer = ({
+  children,
+  styles,
```
seaerchin, line 6: what we can do here is to have a rest parameter `...rest` and spread it after; this is so that the API stays identical to `Flex`

seaerchin (commented):
minor nit on props

src/components/Footer/Footer.tsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/932/commits/c62f356a185d6ef76426aee118446b04d78f9e46

seaerchin (approved):

Issue #937, opened 2022-06-14
[Design System] Migrate Generic Warning Modal to design system


seaerchin:
**work required**
see [here](https://docs.google.com/document/d/1J87zrHM1WHdJQHfbkkgAT_N5WDWXfpgHu99310Gr4nU/edit#heading=h.93f1xpf1qwqe)

Note that this modal is only used in a few places now (`EditContactUs/EditNavBar/EditPage`), with extremely similar API. A possible way forward could be to group the editor screens together and define a shared warning modal for those.

PR #941, opened 2022-06-20
Ref/design system/warning modal

alexanderleegs:
## Problem

This PR refactors our existing warning modal to use the design system. We currently have multiple modals with similar functionality, but slightly different layouts - as such, I have converted `GenericWarningModal` to provide the modal wrapper, while components will need to specify the respective buttons and callbacks being used.

In addition, this PR also removes the similar `DeleteWarningModal`, since its usage is similar - existing components which use `DeleteWarningModal` have been refactored to use `GenericWarningModal` instead.

Closes https://github.com/isomerpages/isomercms-frontend/issues/937.


src/components/GenericWarningModal/GenericWarningModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+} from "@chakra-ui/react"
+import { ModalCloseButton } from "@opengovsg/design-system-react"
+import parse from "html-react-parser"
+
+interface ModalProps {
+  isOpen: boolean
+  onClose: () => void
+  displayTitle: string
+  displayText: string
+  children: JSX.Element | JSX.Element[]
```
seaerchin, line 19: prefer using `PropsWithChildren` instead of manual typing as the types themselves don't guarantee that all possible jsx elements that can be children are children (eg: differing typing in `ReactNode` vs `JSX.Element`. [this](https://www.carlrippon.com/react-children-with-typescript/) article provides a brief overview of the many available types but because it's so confusing, the type that react provides should almost always be preferred

src/components/GenericWarningModal/GenericWarningModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+  ModalBody,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
+} from "@chakra-ui/react"
+import { ModalCloseButton } from "@opengovsg/design-system-react"
+import parse from "html-react-parser"
+
+interface ModalProps {
```
seaerchin, line 14: what's our abstraction boundary here? doing this breaks the illusion that this is a modal and can be styled accordingly. if we aim to provide the user with some level of customization (in the future at least), we should aim to extend the according interface exported from chakra. (either `Modal/ModalBody` depending on the intention of this component).

**however**, the caveat here is that if the intention is to limit customizability then this is ok. just need clarification here

src/components/GenericWarningModal/GenericWarningModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+      </ModalHeader>
+      <ModalCloseButton />
+      <ModalBody>
+        <Text textStyle="body-2">{parse(displayText)}</Text>
+      </ModalBody>
+      <ModalFooter>
+        <HStack w="100%" spacing={2} justifyContent="flex-end">
+          {children}
+        </HStack>
+      </ModalFooter>
```
seaerchin, line 43: actually the wrapper here is so thin that we might be better off defining the **sub**-components instead.

src/components/Header.jsx
```diff
@@ -135,28 +152,50 @@ const Header = ({
...
+                  }}
+                  href={stagingUrl}
+                >
+                  <Text color="white">Proceed to staging site</Text>
+                </Button>
+              </Skeleton>
+            </HStack>
+          </ModalFooter>
+        </ModalContent>
+      </Modal>
```
seaerchin, line 198: couldn't we also use `GenericWarningModal` here?

src/components/GenericWarningModal/GenericWarningModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+interface ModalProps {
+  isOpen: boolean
+  onClose: () => void
+  displayTitle: string
+  displayText: string
+  children: JSX.Element | JSX.Element[]
+}
+
+export const GenericWarningModal = ({
```
seaerchin, line 22: nit: could we change the name to be `WarningModal`? somehow having the generic sounds weird AHHAHA

src/layouts/EditContactUs.jsx
```diff
@@ -164,8 +164,17 @@ const EditContactUs = ({ match }) => {
     id: null,
     type: "",
   })
-  const [showDeletedText, setShowDeletedText] = useState(true)
   const errorToast = useErrorToast()
+  const {
+    isOpen: isRemovedOpen,
+    onOpen: onRemovedOpen,
+    onClose: onRemovedClose,
```
seaerchin, line 171: i think this name might invite confusion; wdyt about something like `isRemovedContentWarningOpen` - this is more verbose but the intention is clear.

src/layouts/EditContactUs.jsx
```diff
@@ -164,8 +164,17 @@ const EditContactUs = ({ match }) => {
...
+  } = useDisclosure({ defaultIsOpen: true })
+  const {
+    isOpen: isDeleteOpen,
+    onOpen: onDeleteOpen,
+    onClose: onDeleteClose,
```
seaerchin, line 176: nit: could we just add the `modal` into the names: `isDeleteModalOpen`

src/components/Header.jsx
```diff
@@ -31,9 +46,17 @@ const Header = ({
...
+  const {
+    isOpen: isWarningOpen,
+    onOpen: onWarningOpen,
+    onClose: onWarningClose,
+  } = useDisclosure()
+  const {
+    isOpen: isStagingOpen,
+    onOpen: onStagingOpen,
+    onClose: onStagingClose,
+  } = useDisclosure()
```
seaerchin, line 58: nit: could we just add the `modal` into the variable names? (eg: `isWarningModalOpen/isStagingModalOpen`)

src/layouts/EditContactUs.jsx
```diff
@@ -774,27 +783,49 @@ const EditContactUs = ({ match }) => {
 
   return (
     <>
-      {showDeletedText && !isEmpty(deletedFrontMatter) && (
+      {!isEmpty(deletedFrontMatter) && (
         <GenericWarningModal
+          isOpen={isRemovedOpen}
```
seaerchin, line 788: we should omit the `!isEmpty(deletedFrontMatter)` and inline this into the `isOpen` prop; separately, we might wanna double check if this works as intended.

src/layouts/EditContactUs.jsx
```diff
@@ -774,27 +783,49 @@ const EditContactUs = ({ match }) => {
...
-          onProceed={() => {
-            setShowDeletedText(false)
-          }}
-          proceedText="Acknowledge"
-        />
+        >
+          <Button onClick={onRemovedClose}>Acknowledge</Button>
+        </GenericWarningModal>
       )}
       {itemPendingForDelete.id && (
```
seaerchin, line 798: same as above, we can inline this into the `isOpen` prop

src/layouts/EditContactUs.jsx
```diff
@@ -774,27 +783,49 @@ const EditContactUs = ({ match }) => {
...
             setItemPendingForDelete({ id: null, type: "" })
+            onDeleteClose()
           }}
-          type={itemPendingForDelete.type}
-        />
+          displayTitle={`Delete ${itemPendingForDelete.type} section`}
+          displayText={`Are you sure you want to delete ${itemPendingForDelete.type}?`}
+        >
+          <Button
+            variant="ghost"
```
seaerchin, line 809: this variant is from chakra and might not be what was intended!

src/layouts/EditHomepage.jsx
```diff
@@ -1171,14 +1173,36 @@ const EditHomepage = ({ match }) => {
   return (
     <>
       {itemPendingForDelete.id && (
```
seaerchin, line 1175: inline into the `isOpen` prop

src/layouts/EditHomepage.jsx
```diff
@@ -1171,14 +1173,36 @@ const EditHomepage = ({ match }) => {
...
             setItemPendingForDelete({ id: null, type: "" })
+            onClose()
           }}
-          type={itemPendingForDelete.type}
-        />
+          displayTitle={`Delete ${itemPendingForDelete.type} section`}
+          displayText={`Are you sure you want to delete ${itemPendingForDelete.type}?`}
+        >
+          <Button
+            variant="ghost"
```
seaerchin, line 1186: should this be `clear` instead?

src/layouts/EditHomepage.jsx
```diff
@@ -1171,14 +1173,36 @@ const EditHomepage = ({ match }) => {
...
+              onClose()
+            }}
+          >
+            Cancel
+          </Button>
+          <Button
+            colorScheme="danger"
+            onClick={() => {
+              deleteHandler(itemPendingForDelete.id)
+              setItemPendingForDelete({ id: null, type: "" })
```
seaerchin, line 1199: hmm this pattern seems to be repeated through out the code - why would we not just inline the 
```ts
setItemPendingForDelete({
  id: event.target.id,
  type,
})
```
here?

src/layouts/EditHomepage.jsx
```diff
@@ -1171,14 +1173,36 @@ const EditHomepage = ({ match }) => {
...
-          type={itemPendingForDelete.type}
-        />
+          displayTitle={`Delete ${itemPendingForDelete.type} section`}
+          displayText={`Are you sure you want to delete ${itemPendingForDelete.type}?`}
+        >
+          <Button
+            variant="ghost"
+            colorScheme="secondary"
+            onClick={() => {
+              setItemPendingForDelete({ id: null, type: "" })
```
seaerchin, line 1189: we seem to persistently repeat this pattern of having a bound state, invoking a function with teh bound state or otherwise resetting.

we should aim to create a hook to abstract over this functionality then invoking it only on `GenericWarningModal`'s delete confirmation

src/layouts/EditNavBar.jsx
```diff
@@ -57,9 +56,19 @@ const EditNavBar = ({ match }) => {
...
+    isOpen: isRemovedOpen,
+    onOpen: onRemovedOpen,
+    onClose: onRemovedClose,
+  } = useDisclosure({ defaultIsOpen: true })
+  const {
+    isOpen: isDeleteOpen,
+    onOpen: onDeleteOpen,
+    onClose: onDeleteClose,
+  } = useDisclosure()
+
```
seaerchin, line 71: similar to before, we should aim to include `modal` into the names for clarity

src/layouts/EditNavBar.jsx
```diff
@@ -612,25 +621,47 @@ const EditNavBar = ({ match }) => {
 
   return (
     <>
-      {showDeletedText && !isEmpty(deletedLinks) && (
+      {!isEmpty(deletedLinks) && (
```
seaerchin, line 624: we should aim to inline this into the `isOpen` prop

src/layouts/EditNavBar.jsx
```diff
@@ -612,25 +621,47 @@ const EditNavBar = ({ match }) => {
...
         <GenericWarningModal
+          isOpen={isRemovedOpen}
+          onClose={onRemovedClose}
           displayTitle="Removed content"
           displayText={`Some of your content has been removed as they attempt to link to invalid folders. No changes are permanent unless you press Save on the next page.<br/>${deletedLinks}`}
```
seaerchin, line 629: we could actually skip the `parse` call if this accepts a `Text` as its children

src/layouts/EditNavBar.jsx
```diff
@@ -612,25 +621,47 @@ const EditNavBar = ({ match }) => {
...
             setItemPendingForDelete({ id: null, type: "" })
+            onDeleteClose()
           }}
-          type={itemPendingForDelete.type}
-        />
+          displayTitle={`Delete ${itemPendingForDelete.type}`}
+          displayText={`Are you sure you want to delete ${itemPendingForDelete.type}?`}
+        >
+          <Button
+            variant="ghost"
```
seaerchin, line 645: same as before - intended variant might be `clear`

src/layouts/EditPage.jsx
```diff
@@ -51,8 +53,17 @@ const EditPage = ({ match }) => {
...
+
+  const {
+    isOpen: isXSSOpen,
+    onOpen: onXSSOpen,
+    onClose: onXSSClose,
+  } = useDisclosure()
+  const {
+    isOpen: isOverwriteOpen,
+    onOpen: onOverwriteOpen,
+    onClose: onOverwriteClose,
```
seaerchin, line 65: add `modal`; we should also make the variable name for the XSS modal warning more descriptive - perhaps something like `XSSWarningModal`?

src/layouts/EditPage.jsx
```diff
@@ -128,53 +139,68 @@ const EditPage = ({ match }) => {
...
+        )}
+        <GenericWarningModal
+          isOpen={isOverwriteOpen}
+          onClose={onOverwriteClose}
+          displayTitle="Override Changes"
+          displayText={`A different version of this page has recently been saved by another user. You can choose to either override their changes, or go back to editing.
+            <br/><br/>We recommend you to make a copy of your changes elsewhere, and come back later to reconcile your changes.`}
+        >
+          <Button
+            variant="ghost"
```
seaerchin, line 194: erroneous variant

src/layouts/EditPage.jsx
```diff
@@ -128,53 +139,68 @@ const EditPage = ({ match }) => {
...
+              `<br/><code>${i}</code>: <code>${
+                elem.attribute?.textContent || elem.element?.textContent
+                  ? (
+                      elem.attribute?.textContent || elem.element?.textContent
+                    ).replace("<", "&lt;")
+                  : elem
+              }</code>`
+          )}
+          <br/><br/>Before saving, the editor input will be automatically sanitised to prevent security vulnerabilities.
+          <br/><br/>To save the sanitised editor input, press Acknowledge. To return to the editor without sanitising, press Cancel.`}
```
seaerchin, line 162: i think adding 2 `br`s to format rather than being able to do it naturally is a sign that the abstraction boundary might be wrong here. other cases to consider might be that the warning modal has a fixed size, perhaps sometimes have picture etc; it might be worth refactoring the exposed API to give consumers more flexibility

src/layouts/EditPage.jsx
```diff
@@ -128,53 +139,68 @@ const EditPage = ({ match }) => {
...
+              Acknowledge
+            </Button>
+          </GenericWarningModal>
+        )}
+        <GenericWarningModal
+          isOpen={isOverwriteOpen}
+          onClose={onOverwriteClose}
+          displayTitle="Override Changes"
+          displayText={`A different version of this page has recently been saved by another user. You can choose to either override their changes, or go back to editing.
+            <br/><br/>We recommend you to make a copy of your changes elsewhere, and come back later to reconcile your changes.`}
```
seaerchin, line 191: :pepehands: at double br

src/layouts/EditPage.jsx
```diff
@@ -128,53 +139,68 @@ const EditPage = ({ match }) => {
...
-                    ? (
-                        elem.attribute?.textContent || elem.element?.textContent
-                      ).replace("<", "&lt;")
-                    : elem
-                }</code>`
-            )}
-            <br/><br/>Before saving, the editor input will be automatically sanitised to prevent security vulnerabilities.
-            <br/><br/>To save the sanitised editor input, press Acknowledge. To return to the editor without sanitising, press Cancel.`}
-              onProceed={() => {
+        {isXSSViolation && (
```
seaerchin, line 142: inline

src/layouts/screens/DeleteWarningScreen.jsx
```diff
@@ -34,11 +36,22 @@ export const DeleteWarningScreen = ({ match, onClose }) => {
...
-        onCancel={() => onClose()}
-        type={fileName}
-      />
+      <GenericWarningModal
+        isOpen // Modal is always present for delete screens
+        onClose={onClose}
+        displayTitle={`Delete ${fileName}`}
+        displayText={`Are you sure you want to delete ${fileName}?`}
+      >
+        <Button variant="ghost" colorScheme="secondary" onClick={onClose}>
```
seaerchin, line 45: check variant

src/layouts/screens/DeleteWarningScreen.jsx
```diff
@@ -34,11 +36,22 @@ export const DeleteWarningScreen = ({ match, onClose }) => {
         })
 
     return fileData ? (
-      <DeleteWarningModal
-        onDelete={() => deleteHandler({ sha: fileData.sha })}
-        onCancel={() => onClose()}
-        type={fileName}
-      />
+      <GenericWarningModal
+        isOpen // Modal is always present for delete screens
```
seaerchin, line 40: this is not true - if `fileData` is not present, this whole chunk is null; should we instead change it to `isOpen={!!fileData}` or use lodash's `isEmpty` if it's an object

src/layouts/screens/DeleteWarningScreen.jsx
```diff
@@ -47,11 +60,19 @@ export const DeleteWarningScreen = ({ match, onClose }) => {
...
-      onCancel={() => onClose()}
-      type={deleteItemName}
-    />
+    <GenericWarningModal
+      isOpen // Modal is always present for delete screens
+      onClose={onClose}
+      displayTitle={`Delete ${deleteItemName}`}
+      displayText={`Are you sure you want to delete ${deleteItemName}?`}
+    >
+      <Button variant="ghost" colorScheme="secondary" onClick={onClose}>
```
seaerchin, line 69: check variant

seaerchin (commented):
i think what we want to consider for this component specifically is the abstraction boundary that we present to more consumers (ie, how much power should we offer them). the presented API (only behavioural props + display title/text) is strict but we need to resort to `parse` together with `br`s being passed in display text, which seems to suggest that this API is stricter than required. 

Another (tangentially) related issue is the artificial complexity introduced by using `useState` to track `itemPendingForDelete.id` rather than just calling `deleteHandler(item.id)` at the point of actually clicking the confirm button, making the code easier to understand and reason about.

src/components/GenericWarningModal/GenericWarningModal.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <>
+      <Button onClick={onOpen}>Open Modal</Button>
+      <GenericWarningModal
+        isOpen={isOpen}
+        onClose={onClose}
+        displayTitle={displayTitle}
+        displayText={displayText}
+      >
+        <Button>{buttonText}</Button>
```
seaerchin, line 32: we should aim to provide the most commonly used variant (`danger`)

src/components/GenericWarningModal/GenericWarningModal.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <GenericWarningModal
+        isOpen={isOpen}
+        onClose={onClose}
+        displayTitle={displayTitle}
+        displayText={displayText}
+      >
+        <Button variant="ghost" colorScheme="secondary">
+          {buttonText}
+        </Button>
+        <Button>{secondButtonText}</Button>
```
seaerchin, line 60: see above wrt button variant

src/layouts/EditContactUs.jsx
alexanderleegs: Changed in https://github.com/isomerpages/isomercms-frontend/pull/941/commits/a4e792aaffbb26913a0cc9a40493c17faf42a5d7!

src/layouts/EditHomepage.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/a4e792aaffbb26913a0cc9a40493c17faf42a5d7

src/layouts/EditNavBar.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/a4e792aaffbb26913a0cc9a40493c17faf42a5d7

src/layouts/EditPage.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/a4e792aaffbb26913a0cc9a40493c17faf42a5d7

src/layouts/screens/DeleteWarningScreen.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/a4e792aaffbb26913a0cc9a40493c17faf42a5d7

src/layouts/EditContactUs.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/0721a811504d1b7b38d4d6eb61364e11db356000

src/layouts/EditContactUs.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/0721a811504d1b7b38d4d6eb61364e11db356000

src/components/Header.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/0721a811504d1b7b38d4d6eb61364e11db356000

src/layouts/screens/DeleteWarningScreen.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/commit/a4e792aaffbb26913a0cc9a40493c17faf42a5d7

src/components/GenericWarningModal/GenericWarningModal.tsx
alexanderleegs: Sure, changed in https://github.com/isomerpages/isomercms-frontend/pull/941/commits/89a80125e5a6b606510663d0f3f3003c382fb0e1

src/components/GenericWarningModal/GenericWarningModal.tsx
alexanderleegs: As discussed offline, I think it makes sense to standardise the overall modal layout as much as possible - but we should allow for greater customisability of the modal content as well. The displayText has been converted into a JSX element and the parse has been removed in https://github.com/isomerpages/isomercms-frontend/pull/941/commits/eb973f7885ef6e088e30376fd92c04cfc156f085

src/components/GenericWarningModal/GenericWarningModal.tsx
alexanderleegs: See above!

src/layouts/EditNavBar.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/0721a811504d1b7b38d4d6eb61364e11db356000

src/layouts/EditPage.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/0721a811504d1b7b38d4d6eb61364e11db356000

src/components/GenericWarningModal/GenericWarningModal.tsx
alexanderleegs: Makes sense, changed in https://github.com/isomerpages/isomercms-frontend/pull/941/commits/eb973f7885ef6e088e30376fd92c04cfc156f085

src/components/Header.jsx
alexanderleegs: I left this out originally because of the image that was supposed to be added here, but if we allow for more flexible modal bodies, this makes sense to include here now - changed in https://github.com/isomerpages/isomercms-frontend/pull/941/commits/d209f907ac75192c25af93006168ec91711f3def but without the image - will be added after the PR inserting the image has been added

src/layouts/EditContactUs.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/513eedac744b2fe3c2814aff9b935982147c46d2

src/layouts/EditContactUs.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/513eedac744b2fe3c2814aff9b935982147c46d2

src/layouts/EditHomepage.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/513eedac744b2fe3c2814aff9b935982147c46d2

src/layouts/EditHomepage.jsx
alexanderleegs: These actually handle 2 separate tasks - `deleteHandler` actually handles the removal of the content, while `setItemPendingForDelete` indicates the item being selected. Modification of this behaviour is complicated due to the scope of editHomepage - I don't think this should be tackled in this PR

src/layouts/EditNavBar.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/513eedac744b2fe3c2814aff9b935982147c46d2

src/layouts/EditNavBar.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/bae8cb2823b64f92ac2e40cee947dd907a0baae9

src/layouts/EditPage.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/bae8cb2823b64f92ac2e40cee947dd907a0baae9

src/layouts/EditPage.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/bae8cb2823b64f92ac2e40cee947dd907a0baae9

src/layouts/EditPage.jsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/513eedac744b2fe3c2814aff9b935982147c46d2

src/layouts/screens/DeleteWarningScreen.jsx
alexanderleegs: We actually do a conditional on `fileData` before deciding to render this modal - I've moved it into the `isOpen` param instead in https://github.com/isomerpages/isomercms-frontend/pull/941/commits/513eedac744b2fe3c2814aff9b935982147c46d2 like the others

src/components/GenericWarningModal/GenericWarningModal.stories.tsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/d669dc200631766d1204621f55ffd23099f4829f

src/components/GenericWarningModal/GenericWarningModal.stories.tsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/941/commits/d669dc200631766d1204621f55ffd23099f4829f

src/layouts/EditHomepage.jsx
alexanderleegs: As discussed offline - this will not be the desired final behaviour, and due to the complexity of editHomepage, we are electing to leave this for a future refactor rather than implementing a temporary solution

PR #945, opened 2022-06-24
fix(preview/dropdown): fixed global styles interfering with context menu items

seaerchin:
## Problem
previously, context menu items **that were pdfs** had a pdf file icon appearing on the left side of the item. this was due to global styles in `isomer-templates.scss` not being localized into `preview-panel`. 

## Solution
Delete the style from `isomer-template` and shift it into `preview-panel`


alexanderleegs (approved):
lgtm

PR #949, opened 2022-06-27
fix(v1.5): concurrent editing displays updates/errors properly

seaerchin:
## Problem
See [here](https://docs.google.com/document/d/1EreGS8X1uIOgMrl7yXDF4jCprsIQ25BTVmyGLpT5P9k/edit#), under settings, for an overview of the concurrent editing issues faced.

## Solution
1. use `controller` which react hook forms provides. rather than managing state from within rhf ourselves, the `controller` simplifies these usage patterns and handles the state of the field for us. Hence, using this component allows us to focus on writing simple components that take in values while deferring the lifecycle etc to rhf

### DIFF
--- a/src/layouts/Settings/ColourSettings.tsx
+++ b/src/layouts/Settings/ColourSettings.tsx
@@ -6,6 +6,7 @@ import {
   Flex,
   Modal,
   ModalContent,
+  FormControl,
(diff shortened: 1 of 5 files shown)


alexanderleegs (approved):
lgtm

PR #950, opened 2022-06-27
fix(settings): add warning modal back into settings

seaerchin:
## Problem
previously, the warning modal was removed for ease of manual testing (sadge) - this PR readds the modal back into settings page. 

related to #941 - tbc if need to rebase and use the generic warning modal

## Solution
1. extract `hasDiff` out from `useEffect`
2. conditionally display modal when `hasDiff`


alexanderleegs (approved):
lgtm - we can replace the modal in a later PR

PR #951, opened 2022-06-28
Fix: editpage third nav styling

alexanderleegs:
## Problem

This PR fixes a minor styling issue with third nav previews. With the release of #895, we mistakenly removed some styling of third navs in the editpage preview, as these styles were attached to the globally imported sgds stylesheet - bullet points were showing up in the list where they should not and the chevrons indicating third nav items were missing. This PR adds those styles back into our codebase.

Note that since we do not import sgds styles globally anymore, I have opted to swap the chevron used in the nav preview to bx-chevron rather than sgds-chevron, since the icons are similar enough, rather than having to refactor all editpage components to use scoped styles.

Current (incorrect) display on CMS:
Screenshot 2022-07-01 at 11 39 40 AM

Preview on CMS:
Screenshot 2022-06-28 at 6 02 54 PM

Preview prior to removal:
Screenshot 2022-06-28 at 6 08 15 PM

Actual site:
Screenshot 2022-06-28 at 6 03 07 PM

### DIFF
--- a/src/styles/isomer-template.scss
+++ b/src/styles/isomer-template.scss
@@ -6408,6 +6408,7 @@ a[href$=".pdf"][target="_blank"]:after {
 
 .bp-menu-list {
   line-height: 1.25;
+  list-style: none;
 }
 
 .has-side-nav {
--- a/src/templates/utils/leftnavGeneration.jsx
+++ b/src/templates/utils/leftnavGeneration.jsx
@@ -47,13 +47,13 @@ const calculateThirdNavHeaderChevronState = (
   elementFileName
 ) => {
   if (currentFileThirdNavTitle !== currentThirdNavTitle) {
-    return "sgds-icon-chevron-down"
+    return "bx bx-chevron-down"
   }
   if (
     fileName === elementFileName ||
     currentThirdNavTitle === currentElementThirdNavTitle
   ) {
-    return "sgds-icon-chevron-up"
+    return "bx bx-chevron-up"
   }
   return undefined
 }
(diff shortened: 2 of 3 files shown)


seaerchin:
lgtm but tagging @NatMaeTan for design side changes. main worry is that the preview doesn't match with staging and leads to user confusion -> them writing in.

could we haev a small comparison pic and wait for nat's take to see if this is ok?

alexanderleegs:
> lgtm but tagging @NatMaeTan for design side changes. main worry is that the preview doesn't match with staging and leads to user confusion -> them writing in.
> 
> could we haev a small comparison pic and wait for nat's take to see if this is ok?

Thanks for the reminder, have added comparison pics to the initial post

NatMaeTan:
Is the main difference the color of the first 3rd nav item, and the icon only?

alexanderleegs:
> Is the main difference the color of the first 3rd nav item, and the icon only?

Oops this page was not a good example because of repeated file names, which shouldn't be an issue on normal sites - the colour is different for third nav items matching the page name of the edited page

The only change between the two is the type of the icon (from the sgds icon to boxicons chevron), which should only correspond to the size of the icon!

seaerchin:
@NatMaeTan (just tagging for visibility lol)

NatMaeTan:
How difficult would it be to fix? +1 to @seaerchin's point on having staging + preview be inconsistent. But this isn't the only thing that is inconsistent atm, maybe it's something we can bulk fix tgt with preview in general.

alexanderleegs (commented):
As discussed offline with design, shifting the chevron back to the sgds icon takes significant effort and the effect is relatively minor - we've opted to leave it as is for now.

seaerchin (approved):
lgtm as design is ok

PR #955, opened 2022-06-30
Ref/design system/warning modal

alexanderleegs:
Duplicate of https://github.com/isomerpages/isomercms-frontend/pull/941 - merged into wrong branch


src/layouts/EditPage.jsx
```diff
@@ -51,8 +53,17 @@ const EditPage = ({ match }) => {
...
+  const {
+    isOpen: isXSSOpen,
+    onOpen: onXSSOpen,
+    onClose: onXSSClose,
+  } = useDisclosure()
+  const {
+    isOpen: isOverwriteOpen,
+    onOpen: onOverwriteOpen,
+    onClose: onOverwriteClose,
+  } = useDisclosure()
```
seaerchin, line 66: nit: could we give more descriptive names to these props from `useDisclosure`?

src/components/WarningModal/WarningModal.tsx
```diff
@@ -0,0 +1,43 @@
...
+  ModalBody,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
+} from "@chakra-ui/react"
+import { ModalCloseButton } from "@opengovsg/design-system-react"
+import { PropsWithChildren } from "react"
+
+interface ModalProps {
```
seaerchin, line 14: we probably want to extend `ModalProps` from chakra and call this `WarningModalProps` so this should look something like: 

```ts
interface WarningModalProps extends ModalProps { 
displayTitle: string 
displayText: JSX.Element
}
```

also, could you verify if strings/`undefined` etc is assignable to `JSX.Element`? they should be as consumers might wish to pass only a string/don't pass anything. 

then we can just use `...rest` and spread it on the modal itself.

seaerchin (commented):
seems fine after props naming is fixed; generation of `displayText` in `EditContactUs` is a little wonky but as that's from the original code, it's fine as-is

src/layouts/EditPage.jsx
alexanderleegs: Whoops missed this out, fixed in https://github.com/isomerpages/isomercms-frontend/pull/955/commits/0b96057bf0edf41867940922cae721ef656958d2

src/components/WarningModal/WarningModal.tsx
alexanderleegs: Makes sense, fixed in https://github.com/isomerpages/isomercms-frontend/pull/955/commits/ac3d56c66dedd4ff46f524f52c8ef05d865538bd

Strings/undefined values seem to be fine!

src/components/WarningModal/WarningModal.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const WarningModal = ({
+  isOpen,
+  onClose,
+  displayTitle,
+  displayText,
+  children,
+  ...rest
+}: PropsWithChildren<WarningModalProps>): JSX.Element => (
+  <Modal isOpen={isOpen} onClose={onClose} {...rest}>
```
seaerchin, line 28: very minor nit but idt we need to explicitly specify and pass `isOpen/onClose` as those are part of `ModalProps` and the `...rest` and subsequent spread will capture them correctly

alexanderleegs: Good point,  https://github.com/isomerpages/isomercms-frontend/pull/955/commits/021a8dd53c3dc7063bddd7d279ab4001f8d80b37

seaerchin (approved):
lgtm

PR #958, opened 2022-06-30
Fix/e2e tests

alexanderleegs:
## Problem

This PR fixes some lingering issues with our e2e tests:

- As we changed the wording of some of our buttons from `Delete` to `Yes, delete`, we also changed the respective buttons in the tests here
- We also removed the `.only` in our resource category tests.


seaerchin (approved):
lgtm thanks for updating!

PR #959, opened 2022-07-01
Fix: add logout hook to sidebar button

alexanderleegs:
## Problem

This PR fixes a minor bug involving the logout button on the sidebar not working. This PR adds the onClick event back to the button.

It also refactors the existing LoginContext component to typescript, so as to allow the storybook component for the sidebar to work properly.

### DIFF
--- a/src/components/Sidebar/Sidebar.stories.tsx
+++ b/src/components/Sidebar/Sidebar.stories.tsx
@@ -1,3 +1,4 @@
+import { useToast } from "@chakra-ui/react"
 import { ComponentStory, ComponentMeta } from "@storybook/react"
 import { MemoryRouter, Redirect, Route } from "react-router-dom"
 
@@ -37,11 +38,17 @@ const SidebarMeta = {
 } as ComponentMeta<typeof Sidebar>
 
 const Template: ComponentStory<typeof Sidebar> = () => {
+  const toast = useToast()
   return (
     <LoginContext.Provider
       value={{
         logout: async () => {
-          return undefined
+          toast({
+            title: "User is logged out",
+            status: "success",
+            duration: 9000,
+            isClosable: true,
+          })
         },
         userId: "username",
         email: "user@open.gov.sg",
--- a/src/components/Sidebar/Sidebar.stories.tsx
+++ b/src/components/Sidebar/Sidebar.stories.tsx
@@ -1,6 +1,8 @@
 import { ComponentStory, ComponentMeta } from "@storybook/react"
 import { MemoryRouter, Redirect, Route } from "react-router-dom"
 
+import { LoginContext } from "contexts/LoginContext"
+
 import { handlers } from "../../mocks/handlers"
 
 import { Sidebar } from "./Sidebar"
@@ -35,7 +37,23 @@ const SidebarMeta = {
 } as ComponentMeta<typeof Sidebar>
 
 const Template: ComponentStory<typeof Sidebar> = () => {
-  return <Sidebar />
+  return (
+    <LoginContext.Provider
+      value={{
+        logout: async () => {
+          return undefined
+        },
+        userId: "username",
+        email: "user@open.gov.sg",
+        contactNumber: "98765432",
+        verifyLoginAndSetLocalStorage: async () => {
+          return undefined
+        },
+      }}
+    >
+      <Sidebar />
+    </LoginContext.Provider>
+  )
 }
 
 export const Default = Template.bind({})
--- a/src/components/Sidebar/Sidebar.stories.tsx
+++ b/src/components/Sidebar/Sidebar.stories.tsx
@@ -38,29 +38,7 @@ const SidebarMeta = {
 } as ComponentMeta<typeof Sidebar>
 
 const Template: ComponentStory<typeof Sidebar> = () => {
-  const toast = useToast()
-  return (
-    <LoginContext.Provider
-      value={{
-        logout: async () => {
-          toast({
-            title: "User is logged out",
-            status: "success",
-            duration: 9000,
-            isClosable: true,
-          })
-        },
-        userId: "username",
-        email: "user@open.gov.sg",
-        contactNumber: "98765432",
-        verifyLoginAndSetLocalStorage: async () => {
-          return undefined
-        },
-      }}
-    >
-      <Sidebar />
-    </LoginContext.Provider>
-  )
+  return <Sidebar />
 }
 
 export const Default = Template.bind({})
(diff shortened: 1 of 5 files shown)


seaerchin (approved):
lgtm! the chromatic error looks benign but it might be worth pushing a commit just to check

src/components/Sidebar/Sidebar.stories.tsx
```diff
@@ -35,7 +37,23 @@ const SidebarMeta = {
 } as ComponentMeta<typeof Sidebar>
 
 const Template: ComponentStory<typeof Sidebar> = () => {
-  return <Sidebar />
+  return (
+    <LoginContext.Provider
+      value={{
+        logout: async () => {
+          return undefined
+        },
```
seaerchin, line 45: we should probably change this to show a toast so that it's obvious in tests that the button click works

src/components/Sidebar/Sidebar.stories.tsx
```diff
@@ -35,7 +37,23 @@ const SidebarMeta = {
...
+        userId: "username",
+        email: "user@open.gov.sg",
+        contactNumber: "98765432",
+        verifyLoginAndSetLocalStorage: async () => {
+          return undefined
+        },
```
seaerchin, line 51: see above comment

src/components/Sidebar/Sidebar.stories.tsx
```diff
@@ -35,7 +37,23 @@ const SidebarMeta = {
 } as ComponentMeta<typeof Sidebar>
 
 const Template: ComponentStory<typeof Sidebar> = () => {
-  return <Sidebar />
+  return (
+    <LoginContext.Provider
```
seaerchin, line 41: nit: this isn't important but something to take note of is that such _wrapper_ components should be located in the `decorators` section of the meta (see [here](https://storybook.js.org/docs/react/writing-stories/decorators)) for more detail

src/contexts/LoginContext.tsx
```diff
@@ -1,41 +1,67 @@
...
-const LoginContext = createContext(null)
+interface LoginContextProps {
+  userId: string
+  email: string
+  contactNumber: string
+  logout: () => Promise<void>
+  verifyLoginAndSetLocalStorage: () => Promise<void>
+}
+
+interface loggedInUserProps {
```
seaerchin, line 25: types should be UpperCamelCased; we should also rename this to `LoggedInUser` as this isn't provided as an argument anywhere (we can also update `LoginContextProps` to extend this as well)

src/contexts/LoginContext.tsx
```diff
@@ -1,41 +1,67 @@
...
+    false
+  )
+  const verifyLoginAndSetLocalStorage = useCallback(async () => {
     const { data } = await axios.get(`${BACKEND_URL}/auth/whoami`)
     const loggedInUser = omitBy(data, isEmpty)
 
     setStoredUserId(loggedInUser.userId)
-    setStoredUser(loggedInUser)
-  }
+    setStoredUser((loggedInUser as unknown) as loggedInUserProps)
```
seaerchin, line 63: we can avoid the cast by setting the generic type on `axios.get` like so: `axios.get<LoggedInUser>`

src/contexts/LoginContext.tsx
```diff
@@ -1,41 +1,67 @@
...
 
-  const [, , removeSites] = useLocalStorage(LOCAL_STORAGE_KEYS.SitesIsPrivate)
-  const verifyLoginAndSetLocalStorage = async () => {
+  const [, , removeSites] = useLocalStorage(
+    LOCAL_STORAGE_KEYS.SitesIsPrivate,
+    false
+  )
+  const verifyLoginAndSetLocalStorage = useCallback(async () => {
     const { data } = await axios.get(`${BACKEND_URL}/auth/whoami`)
     const loggedInUser = omitBy(data, isEmpty)
```
seaerchin, line 60: i think the typedef above might be wrong - this seems to suggest that all the properties should be optional >_<

src/routing/ProtectedRoute.jsx
```diff
@@ -11,34 +11,29 @@ import { getDecodedParams } from "utils/decoding"
...
-        )
-      }
-    </LoginConsumer>
+              <WrappedComponent {...rest} {...props} match={newMatch} />
+            )
+          }}
+        />
+      ))
+  ) : (
+    <Redirect to="/" />
```
seaerchin, line 36: skimmed; assuming only change is to use the hook call rather than the render props pattern

src/contexts/LoginContext.tsx
```diff
@@ -44,7 +70,7 @@ const LoginProvider = ({ children }) => {
     removeSites()
     // NOTE: This is REQUIRED (emphasis here) for auto-redirect on removal of stored user id.
     // This is IN ADDITION to removing the value associated with the key.
-    setStoredUserId(null)
+    setStoredUserId("")
```
seaerchin, line 73: this might be worth double checking against e2e - previously when i tried to make a minor change to this line, it broke (and hence the long comment above) and e2e was failing so we might wanna double check if this is safe

src/components/Sidebar/Sidebar.stories.tsx
alexanderleegs: Makes sense! Added in https://github.com/isomerpages/isomercms-frontend/pull/959/commits/0d5ab3ef3354bf5b23af228baaf53fcbecdd072e

src/components/Sidebar/Sidebar.stories.tsx
alexanderleegs: This is actually unused by the sidebar though - it's just required to be defined, so I've left it as is

src/components/Sidebar/Sidebar.stories.tsx
alexanderleegs: Right, thanks for sharing! Changed in https://github.com/isomerpages/isomercms-frontend/pull/959/commits/4f3601e3a2f3de92c0ab286766585057ea5073a4

src/contexts/LoginContext.tsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/959/commits/b4234e06da950f69b585897da54f414caddd7b74

src/contexts/LoginContext.tsx
alexanderleegs: This was actually an issue regardless because of `omitBy` returning a `Partial<object>` - it's been removed in https://github.com/isomerpages/isomercms-frontend/pull/959/commits/021633d04e612010214a625aef12b6522edfa527 so we no longer need to cast after the above suggestion

src/contexts/LoginContext.tsx
alexanderleegs: Ran a few test suites with no issue, so this should be fine!

PR #960, opened 2022-07-05
refactor(settings): change conflict modal to use WarningModal component 

seaerchin:
## Problem
Previously in the v1.5 PR, the `Settings` page uses a newly created `Modal` component rather than using the `WarningModal` component that is used to display warning messages. 

## Solution
This PR changes the `Modal` to use the `WarningModal` component and also, changes the `spacing` in `WarningModal` to 4. This is because the spacing given in figma seems to be 16px, which is `spacing = {4}`.


alexanderleegs (approved):
lgtm!

PR #961, opened 2022-07-06
chore(v1.5): minor UI updates 

seaerchin:
## Problem
Some UI styling differed from what the designers had in mind for v1.5; refer [here](https://docs.google.com/document/d/1EreGS8X1uIOgMrl7yXDF4jCprsIQ25BTVmyGLpT5P9k/edit#) for a comprehensive list of changes. 

This PR addresses the following: 
1. underline on hover for breadcrumbs
2. last breadcrumb colour
3. settings -> colour button's outline colour

## Solution
1. for underline on breadcrumb hover, this is done through overriding the chakra UI styling for breadcrumbs as the `textDecoration` is set by chakra [here](https://github.com/chakra-ui/chakra-ui/blob/main/packages/theme/src/components/breadcrumb.ts). concretely, this means that all future breadcrumbs will **not** have underline on hover. waiting for designers' response on this but if this is the desired behaviour then it should be ok. 
2. changed colours to fit for settings/breadcrumb

### DIFF
--- a/src/layouts/Folders/components/FolderBreadcrumb.tsx
+++ b/src/layouts/Folders/components/FolderBreadcrumb.tsx
@@ -1,5 +1,3 @@
-// NOTE: As Isomer does not have recursively nested folders at present,
-
 import { Breadcrumb, BreadcrumbItem, BreadcrumbLink } from "@chakra-ui/react"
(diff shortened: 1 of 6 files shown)


alexanderleegs (approved):
lgtm

PR #972, opened 2022-07-13
Fix: ignore url in config

alexanderleegs:
## Problem

This PR provides a temporary fix for ignoring `url` in config files, which is a common field - we should eventually add this as an available parameter which can be modified in settings.

It also downgrades @storybook/addon-essentials to 6.4.18, to prevent type errors being caused by storybook packages being of different versions. We will update the storybook packages in a future PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9984,18 +9984,18 @@
       }
     },
     "@storybook/addon-backgrounds": {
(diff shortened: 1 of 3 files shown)


seaerchin (approved):
Lgtm, thanks for catching the storybook issue!

PR #973, opened 2022-07-14
ref(useGetDirectoryHook): migrate giant hook into smaller pieces

seaerchin:
## Problem
`useGetDirectoryHook` is rather inflated and it's used in a couple of `.tsx` files, which hinders page-level stories for those files.

As `useGetDirectoryHook` has to deal with many variants of content in our CMS (resources/media/files/folders and more), the logic involved is extremely complicated and convoluted. moreover, using the same `queryKey` for all these meant that when we update any single type, all other types are also invalidated. 

## Solution
1. Break `useGetDirectoryHook` into various small pieces. 
2. Remove `content` call from the new hooks. 
3. Set new `queryKey` so non-overlapping calls don't cancel 

For these modals/screens (noted below), they are involved in >1 type of content (eg: resources/pages), which meant that they'd involve a more intricate refactor, which we do not have time for atm. what has been done is to deprecate `useGetDirectory` through JSDoc to ensure that future devs don't use it anymore.

The modals/screens are:
1. `PageMoveModal` - `Workspace` pages can be moved and so can `Resource` pages
2. `PageSettingsScreen` - (see above)
3. `DirectorySettingsScreen` - this needs information of all types of folders (workspace/resources/images/files) 
4. `DirectoryCreationScreen` - (see above)


src/hooks/directoryHooks/useUpdateDirectoryHook.jsx
```diff
@@ -42,10 +42,7 @@ export function useUpdateDirectoryHook(params, queryParams) {
             _.omit(params, "collectionName"),
           ])
         else if (params.resourceCategoryName)
-          queryClient.invalidateQueries([
-            DIR_CONTENT_KEY,
-            _.omit(params, "resourceCategoryName"),
-          ])
+          queryClient.invalidateQueries([RESOURCE_ROOM_CONTENT_KEY])
```
alexanderleegs, line 45: We do still need to include `_.omit(params, "resourceCategoryName")`

src/hooks/pageHooks/useUpdatePageHook.jsx
```diff
@@ -33,13 +37,15 @@ export function useUpdatePageHook(params, queryParams) {
...
+        if (params.collectionName)
           queryClient.invalidateQueries([
             // invalidates collection pages or resource pages
             DIR_CONTENT_KEY,
             _.omit(params, "fileName"),
           ])
-        else
+        else if (params.resourceRoomName) {
+          queryClient.invalidateQueries(RESOURCE_CATEGORY_CONTENT_KEY)
+        } else
```
alexanderleegs, line 48: Same as above, we should include `_.omit(params, "fileName")`!

alexanderleegs (commented):
Thanks for the refactor - it's a lot clearer with individual responsibilities for each hook now. Ran through this PR with @kishore03109, have a few minor comments!

src/hooks/directoryHooks/useGetWorkspace.ts
```diff
@@ -1,41 +0,0 @@
...
-import { DirectoryData } from "types/directory"
-import { DEFAULT_RETRY_MSG } from "utils"
-
-export const useGetWorkspace = (
-  siteName: string,
-  queryOptions?: Omit<UseQueryOptions<DirectoryData[]>, "queryFn" | "queryKey">
-): UseQueryResult<DirectoryData[]> => {
-  const { setRedirectToNotFound } = useRedirectHook()
-  const errorToast = useErrorToast()
-
```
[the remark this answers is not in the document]
seaerchin: i actually think i renamed this and combined it cos i realised it's fetching the folders but i'm not entirely sure cos it's been q long .-.

src/hooks/directoryHooks/useUpdateDirectoryHook.jsx
seaerchin: double checking here - this is to have a more precise invalidation strategy for our query client right?

alexanderleegs: Yep - using `invalidateQueries([RESOURCE_ROOM_CONTENT_KEY])` would also invalidate all resource room content on other sites visited by a user, so I think it would be good to invalidate only the specific item that has changed here

seaerchin: i **think** how it works is that the above invalidation will also result in `useGetResourceRoom` triggering a refetch as the query key is more specific (array of length 2 - `[RESOURCE_ROOM_CONTENT_KEY, params]`). 

we also do **not** need to do an omit as `resourceCategoryName` doesn't exist on the param - the type is as follows
```ts 
interface ResourceRoomRouteParams {
  siteName: string
  resourceRoomName: string
}
`
``

File: src/hooks/pageHooks/useUpdatePageHook.jsx (hunk 2 above)
Comment (seaerchin) [reply src/hooks/pageHooks/useUpdatePageHook.jsx:48]: the type here is
```ts 
interface ResourceCategoryRouteParams extends ResourceRoomRouteParams {
  resourceCategoryName: string
}
```
which doesn't have `fileName`!

src/hooks/directoryHooks/useUpdateDirectoryHook.jsx
alexanderleegs: Ah right, but we should still include the params `queryClient.invalidateQueries([RESOURCE_ROOM_CONTENT_KEY, params])`, since we don't want to invalidate the resource room content of other sites

src/hooks/pageHooks/useUpdatePageHook.jsx
alexanderleegs: Same as above, we should still be using `queryClient.invalidateQueries([RESOURCE_CATEGORY_CONTENT_KEY, params])` so we don't invalidate other unaffected resource categories!

src/hooks/directoryHooks/useUpdateDirectoryHook.jsx
seaerchin: oh yea, i missed that - my bad and you're totally right

src/hooks/pageHooks/useUpdatePageHook.jsx
seaerchin: my bad here - missed that and you're totally rgiht

alexanderleegs (approved):
lgtm

PR #976, opened 2022-07-14
Fix: delete resource category

alexanderleegs:
## Problem

This PR fixes an issue where users were unable to delete resource categories. As the logic for this component is complicated and difficult to debug, we will file an issue to tackle a more permanent fix/refactor in future.

Issue: https://github.com/isomerpages/isomercms-frontend/issues/977


seaerchin (approved):
lgtm... this component needs a good refactoring!!!!

PR #980, opened 2022-07-14
ref(storybook): add page-level stories

seaerchin:
## Problem
Isomer uses storybook to track component level changes, but we don't have an equivalent for pages. This means that designers always have to go to staging to look at changes to any particular screen and it is unergonomic for them and inefficient (as they cannot measure precisely etc). 

## Solution
This problem could be solved by adding page level stories for the **new and updated** screens.


alexanderleegs (approved):
lgtm, pending design

PR #981, opened 2022-07-14
ref(storybook): adds workspace stories

seaerchin:
## Problem
See #980 for an overview of the problem

## Solution
1. This PR adds stories for `Workspace` - lmk if i missed out any potentially interesting stories
2. This PR also has minor fixes for loading states: previously, we used a skeleton to wrap the component. this is problematic as the skeleton takes the height of the component, which is 0 when there's nothing loading. This prevents the skeleton from being seen and feedback from being given to the user. This was fixed through a conditional height - display a fixed height when it's loading, let the height fit content after.


alexanderleegs (approved):
lgtm!

PR #982, opened 2022-07-14
ref(storybook): refactor folders and add stories

seaerchin:
## Problem
first part of problem can be seen in #980, where writing page level stories helps to ensure design consistency + give our designers an easier life.

second part of problem is that folders previously did not lend itself well to page level stories. this resulted in the card missing a few UI elements in storybook, which meant that a refactor was needed. moreover, the card component here did not fit the designer's vision in v1.5 (the component itself looked visually different). 

## Solution
1. write stories for `Folders`
2. update api of `useGetFolders` and related types (as it can return both folders/pages)
3. write new `PageCard` and `FolderCard` component for use in `Folders`

### DIFF
--- a/src/layouts/Folders/Subfolder.stories.tsx
+++ b/src/layouts/Folders/Subfolder.stories.tsx
@@ -1,0 +1,53 @@
+import { ComponentStory, ComponentMeta } from "@storybook/react"
+import { MemoryRouter, Route } from "react-router-dom"
+
+import {
+  MOCK_DIR_DATA,
+  MOCK_FOLDER_NAME,
+  MOCK_PAGES_DATA,
+  MOCK_SUBFOLDER_NAME,
+} from "mocks/constants"
+import { buildSubfolderData } from "mocks/utils"
+
+import { handlers } from "../../mocks/handlers"
+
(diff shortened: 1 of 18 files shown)


src/layouts/Folders/Subfolder.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const Template: ComponentStory<typeof Folders> = Folders
+
+export const Subfolder = Template.bind({})
+Subfolder.parameters = {
+  msw: {
+    handlers: [
+      ...handlers,
+      buildSubfolderData([...MOCK_DIR_DATA, ...MOCK_PAGES_DATA]),
+    ],
```
alexanderleegs, line 51: Subfolders shouldn't have directories, we could probably remove `MOCK_DIR_DATA`!

alexanderleegs (approved):
minor nit, but otherwise lgtm

src/layouts/Folders/Subfolder.stories.tsx
seaerchin: good catch, fixed!

PR #983, opened 2022-07-18
ref(storybook): add resources stories

seaerchin:
## Problem
see #980 for an overview of the problem

## Solution
1. write stories for `ResourceRoom` and `ResourceCategory`
2. minor fixes (explicitly specifying last breadcrumb color/add skeleton height etc)

### DIFF
--- a/src/components/Sidebar/Sidebar.tsx
+++ b/src/components/Sidebar/Sidebar.tsx
@@ -22,7 +22,7 @@ import {
   BiBuoy,
   BiCog,
   BiCubeAlt,
-  BiFileBlank,
+  BiFile,
   BiGridAlt,
   BiImage,
   BiLogOutCircle,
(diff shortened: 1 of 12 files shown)


alexanderleegs (commented):
lgtm, but currently, clicking on page settings for resource file cards in the storybook throws an error - i'm not sure if this is something that needs to be fixed though!

seaerchin:
> lgtm, but currently, clicking on page settings for resource file cards in the storybook throws an error - i'm not sure if this is something that needs to be fixed though!

this happens because the card itself is a _link_ to another component (and screen), which isn't rendered in the storybook (only the `ResourceCategory` itself is). this also happens with any other component where the `Route` with the matching path isn't in the story @__@

alexanderleegs (approved):

PR #984, opened 2022-07-18
ref(storybook): add settings stories

seaerchin:
## Problem
see #980 for an overview

## Solution
1. add settings stories


alexanderleegs (approved):
lgtm

PR #985, opened 2022-07-18
ref(storybook): login context for stories

seaerchin:
## Problem
Previously, we were relying on a hard-coded `LoginContext` for our stories as `msw` was not implemented in our codebase and we had no way of providing mock data.

This led to components dependent on `LoginContext` to be unable to update their states accurately to reflect the logged in status. 

Moreover, the `LoginContext` was not being used and components were reading from `localStorage`, which is not reactive and doesn't cause a re-render on change. this led to issues on initial login for `Header`.

Also, this PR adds a new GSIB viewport for visual testing (1920 * 1080).

## Solution
1. Remove `LoginContext` from storybook preview and use `LoginProvider` 
2. Add mocks for the routes used by `LoginProvider`
3. For components relying on `userId`, read directly from the context rather than `localStorage`

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,11 +1,27 @@
 import { ThemeProvider, useToast } from "@opengovsg/design-system-react"
 import { initialize, mswDecorator } from "msw-storybook-addon"
 import { QueryClient, QueryClientProvider } from "react-query"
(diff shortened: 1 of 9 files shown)


alexanderleegs (approved):
lgtm

PR #987, opened 2022-07-18
fix(pagesettingsmodal): use chakra modal and form components

### LINKED ISSUE #986: [BUG] Page settings modal is unscrollable
### Problem
For users on GSIB, the page settings modal might be extremely large and block them from scrolling/making their edits and saving. 

https://user-images.githubusercontent.com/44049504/179457801-03498af6-1577-4068-adc4-b5da286a9b63.MOV

This problem is caused by the hand-written isomer modal not focusing on the modal but instead, still retaining it on the background, causing scrolls to work on the background rather than on the modal. 

### Suggested solution
1. change to chakra ui modal

seaerchin:
## Problem
See #986 for a brief overview of the issue.

Also, as ogp design system is now part of the codebase, we should shift towards using chakra ui and so some of the old `FormXxx` components have been replaced by the respective design system component.

Closes #986

## Solution
1. replace handwritten modal components and styling with chakra modal components
2. replace old `FormXxx` components with ogp design system components

### DIFF
--- a/src/components/PageSettingsModal/PageSettingsModal.jsx
+++ b/src/components/PageSettingsModal/PageSettingsModal.jsx
@@ -1,5 +1,23 @@
-import { CloseButton, HStack } from "@chakra-ui/react"
+import {
+  Input,
+  Modal,
+  ModalOverlay,
+  ModalContent,
+  ModalHeader,
+  ModalFooter,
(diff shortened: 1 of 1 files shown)


alexanderleegs (approved):
lgtm

PR #988, opened 2022-07-18
feat(resourceroom): allow users to update resource room name

### LINKED ISSUE #943: [FEATURE] Allow users to update resource name
### Problem
currently, users are not able to update resource room name by themselves, which results in them having to email isomer support to do this change.

this results in increased ops workload for prod support

### Suggested solution
1. add a modal similar to the existing one to update directory title

seaerchin:
**note: this PR has a dependency on #983**
To be reviewed in conjunction with PR #[481](https://github.com/isomerpages/isomercms-backend/pull/481) on the isomercms-backend repo.
## Problem
see #943 for a description of the problem

Closes #943

## Solution
1. add new hook and directory service method for updating resource room name
2. add new modal and context menu on `ResourceRoom` so that users can update the name
3. add toasts + reset for when update is successful/not


alexanderleegs (commented):
lgtm on the frontend, but going to hold off on approving this for now - the backend endpoint doesn't change the name of the resource room (in `/{resourceRoomName}/index.html`). We should fix that endpoint before merging this in

seaerchin:
> lgtm on the frontend, but going to hold off on approving this for now - the backend endpoint doesn't change the name of the resource room (in `/{resourceRoomName}/index.html`). We should fix that endpoint before merging this in

sounds good - this means that there should be a corresponding PR on the backend, right?

alexanderleegs:
> > lgtm on the frontend, but going to hold off on approving this for now - the backend endpoint doesn't change the name of the resource room (in `/{resourceRoomName}/index.html`). We should fix that endpoint before merging this in
> 
> sounds good - this means that there should be a corresponding PR on the backend, right?

yup, let me do up a quick fix for it

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -96,7 +116,10 @@ const EmptyResourceRoom = () => {
           >
             <Input
               maxW="50%"
-              {...register("newDirectoryName", { required: true })}
+              {...register("newDirectoryName", {
+                required:
+                  "Please ensure that you have entered a resource room name!",
+              })}
```
alexanderleegs, line 122: Actually, could we have the default value of this field be the existing resource room name?

seaerchin: that sounds fine - so for clarity's sake, this means that there will be **no** placeholder and instead, the field will be prefilled w/ the existing resource room name right?

@NatMaeTan wdyt about this change? i think it's reasonable

alexanderleegs: Yup, I think this is more consistent with the existing renaming behaviour of other items (e.g. folders/files come prefilled with the existing name)

NatMaeTan: This is for changing a resource room name right? If so, yup sounds good that it should be prefilled with the existing name for people to edit

seaerchin: done!

kishore03109:
@seaerchin Hey, it seems that if my page does not have any `ResourceRoom` to begin with, the path `sites/:site-name/resourceRoom/` will return an error, instead of the empty state!

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -187,44 +207,137 @@ const ResourceRoomContent = ({
...
+            <ModalHeader>
+              <Text>Edit resource room settings</Text>
+            </ModalHeader>
+            <ModalCloseButton />
+            <ModalBody>
+              <FormControl isRequired>
+                <FormLabel>Resource room title</FormLabel>
+                <Input
+                  placeholder="New resource room name"
+                  {...register("newDirectoryName", { required: true })}
```
kishore03109: Out of curiosity ya, if we as a team agree that we are ok with prop spreading, should we just change this in our `.eslintrc`, feels neater than having multiple comments to ignore this particular rule, wdyt?

kishore03109 (changes requested):
@seaerchin Hey, it seems that if my page does not have any ResourceRoom to begin with, the path sites/:site-name/resourceRoom/ will return an error, instead of the empty state!

I see two possible workarounds: 

1. We change the template to already include a default resource room name (I have a preference for this) 
2. We still display the empty state for resource room - but we hint them with a default name for the room (as previously discussed with Nat in this thread!)

src/layouts/ResourceRoom/ResourceRoom.tsx
seaerchin: hmm so actually the way this panned out was that i was the one that introduced ts and i set a stricter linting than was necessary (so that we get used to having higher standards) and my thinking at that time was that it was easier to go from strict -> loose rather than the other way around (as you won't know what additional checks are useful w/o stumbling upon them). 

so because of that, i found that the prop spreading is a false positive **when using rhf** as the docs advocated the prop spreading way and hence, changed it from error -> warning. the multiple comments to ignore is from when it was an **error** and it **shouldn't** be required now. the reason i didn't remove them was that i felt they were not very confusing + it's drudgery that doesn't really add value but you are correct in that we should remove them...

if you feel that they should be removed, we could file an issue and i could tackle it! wdyt?

seaerchin:
> @seaerchin Hey, it seems that if my page does not have any ResourceRoom to begin with, the path sites/:site-name/resourceRoom/ will return an error, instead of the empty state!
> 
> I see two possible workarounds:
> 
> 1. We change the template to already include a default resource room name (I have a preference for this)
> 2. We still display the empty state for resource room - but we hint them with a default name for the room (as previously discussed with Nat in this thread!)

oops, this was cause there was namespace conflicts in teh msw handlers ._. (the same one you experienced) i rebased the resource room branch off the updated `page-level-stories` branch and subsequently, this branch off the resource room branch. 

thanks for catching!

kishore03109:
> oops, this was cause there was namespace conflicts in teh msw handlers ._. (the same one you experienced) i rebased the resource room branch off the updated page-level-stories branch and subsequently, this branch off the resource room branch.

Hmm it seems that there is another error when I try to go to the resource rooms now! 

Eg. when visiting my [resource room](http://localhost:3000/sites/kishore-test/resourceRoom) at `http://localhost:3000/sites/kishore-test/resourceRoom`
I get this error: 
Screenshot 2022-08-17 at 10 42 50 AM

kishore03109 (changes requested):
Hmm it seems that there is another error when I try to go to the resource rooms now!

Eg. when visiting my [resource room](http://localhost:3000/sites/kishore-test/resourceRoom) at http://localhost:3000/sites/kishore-test/resourceRoom
I get this error:

image
Do let me know if you cannot replicate it!

seaerchin:
> Hmm it seems that there is another error when I try to go to the resource rooms now!
> 
> Eg. when visiting my [resource room](http://localhost:3000/sites/kishore-test/resourceRoom) at http://localhost:3000/sites/kishore-test/resourceRoom I get this error:
> 
> image Do let me know if you cannot replicate it!

you're actually entirely right here and good catch! i didn't notice this as the storybook actually missed the case where the resource room didn't exist ._.

this is due to a faulty conditional where we choose to render either the content or the empty resource room - `!isResourceRoomNameLoading && !resourceRoomName` which should've been 

```ts
!isResourceRoomNameLoading && { resourceRoomName ? EmptyResourceRoom : ResourceRoomContent}``` 
```

kishore03109:
@seaerchin hihi do ping me on slack once the rebasing is done ya

seaerchin:
> @seaerchin hihi do ping me on slack once the rebasing is done ya

thanks for the reminder - this is done!

kishore03109 (approved):
lgtm

PR #991, opened 2022-07-21
test(cypress): upgrade cypress to v10 and migrate to ts

seaerchin:
**note: most line changes are from package-lock**
## Problem
our cypress version is ~4 major versions behind. the v6 ui is not that intuitive and it lacks features as compared to the v10 one (eg: dashboard sync/component testing).

furthermore, our tests are written in js originally, which lacks many intellisense features compared to ts. this is most evident during usage of custom commands, where the types cannot be automatically inferred and the command has no intellisense. 

this PR migrates cypress to v10, together with a migration to ts. 

## Solution
1. upgrade cypress to v10 and followed [migration guide](https://docs.cypress.io/guides/references/migration-guide)
2. removed old settings for cypress
    - most notably, this had the `setUpNodeEvents` to point cypress at the underlying `react-scripts` webpack config but has since been removed as we need a ts step, which is already done by the webpack pre-processor. the alias and such has been replaced using `ts-paths`, in the plugins folder (see [here](https://github.com/cypress-io/cypress/issues/3262#issuecomment-683267360))
3. migrate cypress custom commands to ts. 
    - for **parent** commands, this is now a 2-step process, requiring a function signature in `index.d.ts` (so that the global cypress object will have the correct types for the custom command - note that this should return `Chainable<void>`) and the command definition itself
    - for **child** commands, it's similar to the above except that we have to declare the type correctly as `Chainable<ChildType>` (for commands returning `element`, this is `Chainable<JQuery<HTMLElement>>` and the function definition has to take a `prevSubject` argument). 
4. add custom `tsconfig` for cypress - this is to allow for top level suggestions of hte `cy` object without it being present in app code


cypress/support/commands.js
```diff
@@ -1,22 +0,0 @@
-import waitForDom from "./waitForDom"
```
alexanderleegs, line 1: Just curious, does cypress 10 no longer need waitForDom to reduce test flakiness?

seaerchin: oof i missed this. the `waitForDom` actually didnt seem to have any effect - this might be because i added some assertions. 

just double checking here - the `waitForDom` prevents the `xxx is detached from the dom` error? if so, i think it might be useful for me to undo the deletion and instead, migrate it to ts

alexanderleegs: Yeah this seemed to help with the issue originally - not sure if moving to cypress 10 helps already though. If the tests are less flaky now, maybe it doesn't need to be readded?

alexanderleegs (approved):

Issue #962, opened 2022-07-07
[FEATURE] Empty state for ResourceRoom


seaerchin:
# Problem
currently, our empty state for `ResourceRoom` is not very pleasant to look at and not intuitive (see below). Isomer should have a more descriptive empty state to guide the user along
![Screenshot 2022-07-07 at 4 23 31 PM](https://user-images.githubusercontent.com/44049504/177727351-b9d43dc4-482c-454d-b458-73090b971257.png)
 
# Suggested Solution
Implement the UI as shown in the figma [here](https://www.figma.com/file/7OtNYkc4s9BkFk05ConEbd/V1.5?node-id=1559%3A62055) and show a toast when the resource room is created

PR #993, opened 2022-07-26
feat(resource-room): empty state for empty resource room

kishore03109:
## Problem
Currently, our empty state for ResourceRoom is not very pleasant to look at and not intuitive. Isomer should have a more descriptive empty state to guide the user along

Addresses #962

## Solution
As suggested in [figma here](https://www.figma.com/file/7OtNYkc4s9BkFk05ConEbd/V1.5?node-id=1559%3A62055)

**Breaking Changes**

## Before & After Screenshots

**BEFORE**:

before

**AFTER**:

Screenshot 2022-07-26 at 12 24 38 PM

Other Notes: 
Other changes are linked to outstanding PRs by @seaerchin

### DIFF
--- a/src/assets/images/EmptyBoxImage.tsx
+++ b/src/assets/images/EmptyBoxImage.tsx
@@ -1,6 +1,4 @@
-export const EmptyBoxImage = (
-  props: React.SVGProps<SVGSVGElement>
-): JSX.Element => {
+export const EmptyBoxImage = (): JSX.Element => {
   return (
     <svg
       width="256"
--- a/src/assets/images/EmptyBoxImage.tsx
+++ b/src/assets/images/EmptyBoxImage.tsx
@@ -1,4 +1,6 @@
-export const EmptyBoxImage = (): JSX.Element => {
+export const EmptyBoxImage = (
+  props: React.SVGProps<SVGSVGElement>
+): JSX.Element => {
   return (
     <svg
       width="256"
@@ -6,7 +8,6 @@ export const EmptyBoxImage = (): JSX.Element => {
       viewBox="0 0 256 120"
       fill="none"
       xmlns="http://www.w3.org/2000/svg"
-      // eslint-disable-next-line react/jsx-props-no-spreading
       {...props}
     >
       <path
--- a/src/assets/images/EmptyBoxImage.tsx
+++ b/src/assets/images/EmptyBoxImage.tsx
@@ -8,6 +8,7 @@ export const EmptyBoxImage = (
       viewBox="0 0 256 120"
       fill="none"
       xmlns="http://www.w3.org/2000/svg"
+      // eslint-disable-next-line react/jsx-props-no-spreading
       {...props}
     >
       <path
--- a/src/components/EmptyArea.tsx
+++ b/src/components/EmptyArea.tsx
@@ -1,11 +1,10 @@
 import { Box, Center, VStack, Text, HTMLChakraProps } from "@chakra-ui/react"
-import { ReactFragment } from "react"
 
 import { EmptyBoxImage } from "assets"
 
-export interface EmptyAreaOption extends HTMLChakraProps<"div"> {
+export interface EmptyAreaProps extends HTMLChakraProps<"div"> {
   isItemEmpty: boolean
-  actionButton: ReactFragment
+  actionButton: JSX.Element
   mainText?: string
   subText?: string
 }
@@ -16,13 +15,14 @@ export const EmptyArea = ({
   actionButton,
   children,
   isItemEmpty,
-}: EmptyAreaOption): JSX.Element => {
+  ...rest
+}: EmptyAreaProps): JSX.Element => {
   return (
     <>
       {isItemEmpty ? (
-        <Box as="form" w="full">
+        <Box w="full">
           {/* Resource Room does not exist */}
-          <VStack spacing={5}>
+          <VStack spacing={5} {...rest}>
             <EmptyBoxImage />
             <Center>
               <VStack spacing={0}>
--- a/src/layouts/ResourceCategory/ResourceCategory.tsx
+++ b/src/layouts/ResourceCategory/ResourceCategory.tsx
@@ -62,7 +62,7 @@ export const ResourceCategory = (): JSX.Element => {
         </Section>
         <Section>
           {/* 
-              There is no loading state for this as '!arePagesEmpty' will only evulate to 
+              There is no loading state for this as '!arePagesEmpty' will only evaluate to 
               true after we have received data from the API call.
            */}
           {!arePagesEmpty && (
--- a/src/layouts/ResourceCategory/ResourceCategory.tsx
+++ b/src/layouts/ResourceCategory/ResourceCategory.tsx
@@ -81,7 +81,7 @@ export const ResourceCategory = (): JSX.Element => {
                   to={`${url}/createPage`}
                   leftIcon={<Icon as={BiPlus} fontSize="1.5rem" fill="white" />}
                 >
-                  Create Page
+                  Create page
                 </Button>
               }
               subText="Create a resource page to get started."
--- a/src/layouts/ResourceCategory/ResourceCategory.tsx
+++ b/src/layouts/ResourceCategory/ResourceCategory.tsx
@@ -1,13 +1,5 @@
-import {
-  Box,
-  Text,
-  Skeleton,
-  SimpleGrid,
-  Button,
-  Center,
-  Icon,
-  VStack,
-} from "@chakra-ui/react"
+import { Box, Text, Skeleton, SimpleGrid, Button, Icon } from "@chakra-ui/react"
+import { EmptyArea } from "components/EmptyArea"
 import { BiBulb, BiPlus } from "react-icons/bi"
 import {
   Switch,
@@ -35,7 +27,6 @@ import {
 import { ProtectedRouteWithProps } from "routing/ProtectedRouteWithProps"
 
 // Import utils
-import { EmptyBoxImage } from "assets"
 import { ResourceCategoryRouteParams } from "types/resources"
 import { deslugifyDirectory } from "utils"
 
@@ -82,19 +73,9 @@ export const ResourceCategory = (): JSX.Element => {
             w="100%"
             h={isLoading ? "10rem" : "fit-content"}
           >
-            {arePagesEmpty && (
-              <VStack spacing={5}>
-                <EmptyBoxImage />
-                <Center>
-                  <VStack spacing={0}>
-                    <Text textStyle="subhead-1">
-                      There&apos;s nothing here yet.
-                    </Text>
-                    <Text textStyle="body-2">
-                      Create a resource page to get started.
-                    </Text>
-                  </VStack>
-                </Center>
+            <EmptyArea
+              isItemEmpty={arePagesEmpty}
+              actionButton={
                 <Button
                   as={RouterLink}
                   to={`${url}/createPage`}
@@ -102,8 +83,10 @@ export const ResourceCategory = (): JSX.Element => {
                 >
                   Create Page
                 </Button>
-              </VStack>
-            )}
+              }
+              subText="Create a resource page to get started."
+            />
+
             <SimpleGrid columns={3} spacing="1.5rem">
               {/* NOTE: need to use multiline cards */}
               {(pagesData || []).map(({ name, title, date, resourceType }) => (
--- a/src/layouts/ResourceCategory/ResourceCategory.tsx
+++ b/src/layouts/ResourceCategory/ResourceCategory.tsx
@@ -1,5 +1,14 @@
-import { Box, Text, Skeleton, SimpleGrid } from "@chakra-ui/react"
-import { BiBulb } from "react-icons/bi"
+import {
+  Box,
+  Text,
+  Skeleton,
+  SimpleGrid,
+  Button,
+  Center,
+  Icon,
+  VStack,
+} from "@chakra-ui/react"
+import { BiBulb, BiPlus } from "react-icons/bi"
 import {
   Switch,
   useRouteMatch,
@@ -26,6 +35,7 @@ import {
 import { ProtectedRouteWithProps } from "routing/ProtectedRouteWithProps"
 
 // Import utils
+import { EmptyBoxImage } from "assets"
 import { ResourceCategoryRouteParams } from "types/resources"
 import { deslugifyDirectory } from "utils"
 
@@ -66,7 +76,30 @@ export const ResourceCategory = (): JSX.Element => {
             w="100%"
             h={isLoading ? "10rem" : "fit-content"}
           >
-            {!pagesData || (!pagesData.length && <Text>No content here</Text>)}
+            {!pagesData ||
+              (!pagesData.length && (
+                <VStack spacing={5}>
+                  <EmptyBoxImage />
+                  <VStack spacing={0}>
+                    <Center textStyle="subhead-1">
+                      There&apos;s nothing here yet.
+                    </Center>
+                    <Center textStyle="body-2">
+                      Create a page to get started.
+                    </Center>
+                  </VStack>
+                  <Button
+                    variant="solid"
+                    as={RouterLink}
+                    to={`${url}/createPage`}
+                    leftIcon={
+                      <Icon as={BiPlus} fontSize="1.5rem" fill="white" />
+                    }
+                  >
+                    Create Resource Page
+                  </Button>
+                </VStack>
+              ))}
             <SimpleGrid columns={3} spacing="1.5rem">
               {/* NOTE: need to use multiline cards */}
               {(pagesData || []).map(({ name, title, date, resourceType }) => (
--- a/src/layouts/ResourceCategory/ResourceCategory.tsx
+++ b/src/layouts/ResourceCategory/ResourceCategory.tsx
@@ -48,7 +48,7 @@ export const ResourceCategory = (): JSX.Element => {
   const history = useHistory()
 
   const { data: pagesData, isLoading } = useGetResourceCategory(params)
-  const isPagesEmpty = !pagesData?.length
+  const arePagesEmpty = !pagesData?.length
   return (
     <>
       <SiteViewLayout>
@@ -61,7 +61,11 @@ export const ResourceCategory = (): JSX.Element => {
           </Box>
         </Section>
         <Section>
-          {!isPagesEmpty && (
+          {/* 
+              There is no loading state for this as '!arePagesEmpty' will only evulate to 
+              true after we have received data from the API call.
+           */}
+          {!arePagesEmpty && (
             <Box w="full">
               <SectionHeader label="Resource Pages">
                 <CreateButton as={RouterLink} to={`${url}/createPage`}>
@@ -78,19 +82,20 @@ export const ResourceCategory = (): JSX.Element => {
             w="100%"
             h={isLoading ? "10rem" : "fit-content"}
           >
-            {isPagesEmpty && (
+            {arePagesEmpty && (
               <VStack spacing={5}>
                 <EmptyBoxImage />
-                <VStack spacing={0}>
-                  <Center textStyle="subhead-1">
-                    There&apos;s nothing here yet.
-                  </Center>
-                  <Center textStyle="body-2">
-                    Create a resource page to get started.
-                  </Center>
-                </VStack>
+                <Center>
+                  <VStack spacing={0}>
+                    <Text textStyle="subhead-1">
+                      There&apos;s nothing here yet.
+                    </Text>
+                    <Text textStyle="body-2">
+                      Create a resource page to get started.
+                    </Text>
+                  </VStack>
+                </Center>
                 <Button
-                  variant="solid"
                   as={RouterLink}
                   to={`${url}/createPage`}
                   leftIcon={<Icon as={BiPlus} fontSize="1.5rem" fill="white" />}
--- a/src/layouts/ResourceCategory/ResourceCategory.tsx
+++ b/src/layouts/ResourceCategory/ResourceCategory.tsx
@@ -48,7 +48,7 @@ export const ResourceCategory = (): JSX.Element => {
   const history = useHistory()
 
   const { data: pagesData, isLoading } = useGetResourceCategory(params)
-
+  const isPagesEmpty = !pagesData?.length
   return (
     <>
       <SiteViewLayout>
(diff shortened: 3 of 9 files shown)


src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -82,40 +86,72 @@ const EmptyResourceRoom = () => {
...
+              isRequired
+              isInvalid={!!errors.newDirectoryName?.message}
+            >
+              <Input
+                marginTop={5}
+                placeholder="Resource room name"
+                {...register("newDirectoryName", {
+                  required: "Please enter resource room name",
+                })}
+              />
```
kishore03109, line 133: @seaerchin
The eslint gives and error saying: "Prop spreading is forbidden"

However, I do see this in other areas of the codebase.. Are we ignoring this warning as a team for now?

seaerchin:
@kishore03109 I changed the base branch for you; could you rebase to remove the extra commits?

src/layouts/ResourceRoom/ResourceRoom.tsx
seaerchin: Hmm, this should be a warning and not an error; I'm in favour of ignoring this as this is a design pattern given by the `react-hook-form` docs and we shouldn't need to care about the individual props returned

kishore03109:
@seaerchin I was about to commit to remove the changes for the changelog and package.json... but would it actually be easier for you when we merge this into master? do you want me to remove the or nah?

seaerchin:
> @seaerchin I was about to commit to remove the changes for the changelog and package.json... but would it actually be easier for you when we merge this into master? do you want me to remove the or nah?

We should remove! The merging strategy here is for all of these commits to go into the tracking branch for the page level stories (#980), which would then be merged back into `develop`. 

As `develop` already contains these commits (and the commits are non-overlapping), keeping the changes to `package.json`/`changelog.md` is unnecessary and might wonk up the merge later ._. 

Also, we **don't** merge into `master` normally (this is only during release) and instead, merge into `develop`

src/layouts/ResourceCategory/ResourceCategory.tsx
```diff
@@ -38,7 +48,7 @@ export const ResourceCategory = (): JSX.Element => {
   const history = useHistory()
 
   const { data: pagesData, isLoading } = useGetResourceCategory(params)
-
+  const isPagesEmpty = !pagesData?.length
```
seaerchin, line 51: supernit: `arePagesEmpty` or `hasPages` (but needs conditional change)

src/layouts/ResourceCategory/ResourceCategory.tsx
```diff
@@ -51,22 +61,44 @@ export const ResourceCategory = (): JSX.Element => {
...
+              <SectionHeader label="Resource Pages">
+                <CreateButton as={RouterLink} to={`${url}/createPage`}>
+                  Create page
+                </CreateButton>
+              </SectionHeader>
+              <SectionCaption icon={BiBulb} label="NOTE: ">
+                Pages are automatically ordered by latest date.
+              </SectionCaption>
+            </Box>
+          )}
```
seaerchin, line 75: should we account for the fact that `isPagesEmpty` might change state due to loading? a probable sequence of events might be `true` (as before first fetch it defaults to `undefined`) -> successful load -> `false`, leading to this to be shown.

a quick fix would just be to have a skeleton wrap this (or to subsume it into the skeleton below)

src/layouts/ResourceCategory/ResourceCategory.tsx
```diff
@@ -51,22 +61,44 @@ export const ResourceCategory = (): JSX.Element => {
...
+            {isPagesEmpty && (
+              <VStack spacing={5}>
+                <EmptyBoxImage />
+                <VStack spacing={0}>
+                  <Center textStyle="subhead-1">
+                    There&apos;s nothing here yet.
+                  </Center>
+                  <Center textStyle="body-2">
+                    Create a resource page to get started.
+                  </Center>
```
seaerchin, line 90: this is kinda weird - the `Center` already guarantees that both are vertically/horizontally centered; we should instead structure this like so:

```ts
<Center>
  <Text textStyle = "subhead-1">
    There&apos;s nothing here yet.
  </Text>
  <Text textStyle = "body-2">
    Create a resource page to get started.
  </Text>
</Center>
```

which prevents ambiguity. wdyt about this change?

src/layouts/ResourceCategory/ResourceCategory.tsx
```diff
@@ -51,22 +61,44 @@ export const ResourceCategory = (): JSX.Element => {
...
           <Skeleton
             isLoaded={!isLoading}
             w="100%"
             h={isLoading ? "10rem" : "fit-content"}
           >
-            {!pagesData || (!pagesData.length && <Text>No content here</Text>)}
+            {isPagesEmpty && (
+              <VStack spacing={5}>
+                <EmptyBoxImage />
+                <VStack spacing={0}>
```
seaerchin, line 84: if we combine the `Center`, we shouldn't need this `VStack` anymore

src/layouts/ResourceRoom/ResourceRoom.stories.tsx
```diff
@@ -50,6 +51,34 @@ Empty.parameters = {
...
+    return (
+      <MemoryRouter
+        initialEntries={[
+          `/sites/storybook/resourceRoom/${MOCK_RESOURCE_ROOM_NAME}`,
+        ]}
+      >
+        <Route path="/sites/:siteName/resourceRoom/">
+          <Story />
+        </Route>
+      </MemoryRouter>
```
seaerchin, line 69: could we shift this into the decorator above? it'd still apply to this and it'd be neater

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -80,40 +91,73 @@ const EmptyResourceRoom = () => {
...
+        <VStack spacing={5}>
+          <EmptyBoxImage />
+          <VStack spacing={0}>
+            <Center textStyle="subhead-1">
+              There&apos;s nothing here yet.
+            </Center>
+            <Center textStyle="body-2">
+              Create a resource room to get started.
+            </Center>
+          </VStack>
```
seaerchin, line 109: see prior comment about combining `Center` and using `Text`

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -80,40 +91,73 @@ const EmptyResourceRoom = () => {
     }
   }, [errorToast, isError])
 
+  const { isOpen, onOpen, onClose } = useDisclosure()
+
   return (
     <SiteViewLayout>
-      <Box as="form" w="full">
+      <Box as="form" w="full" mt={100}>
```
seaerchin, line 98: this is slightly weird - why do we need `mt = {100}`

src/layouts/ResourceCategory/ResourceCategory.tsx
```diff
@@ -51,22 +61,44 @@ export const ResourceCategory = (): JSX.Element => {
...
+                  <Center textStyle="subhead-1">
+                    There&apos;s nothing here yet.
+                  </Center>
+                  <Center textStyle="body-2">
+                    Create a resource page to get started.
+                  </Center>
+                </VStack>
+                <Button
+                  variant="solid"
```
seaerchin, line 93: idt we need this prop as the default variant is already `solid`

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -80,40 +91,73 @@ const EmptyResourceRoom = () => {
...
+            <Center textStyle="body-2">
+              Create a resource room to get started.
+            </Center>
+          </VStack>
           <Button
-            type="submit"
-            isLoading={isLoading}
-            onClick={handleSubmit(onSubmit)}
-            isDisabled={!_.isEmpty(errors)}
+            variant="solid"
```
seaerchin, line 111: extra prop

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -80,40 +91,73 @@ const EmptyResourceRoom = () => {
...
+            onClick={onOpen}
+            leftIcon={<Icon as={BiPlus} fontSize="1.5rem" fill="white" />}
           >
             Create Resource Room
           </Button>
-        </Section>
+        </VStack>
       </Box>
+
+      <Modal blockScrollOnMount={false} isOpen={isOpen} onClose={onClose}>
```
seaerchin, line 120: could i check why scrolling is blocked on mount? if this is just to maintain compat w/ current behaviour, we should double check with the designers on the _intended_ behaviour

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -80,40 +91,73 @@ const EmptyResourceRoom = () => {
...
+                })}
+              />
+              <FormErrorMessage>
+                {errors.newDirectoryName?.message}
+              </FormErrorMessage>
+            </FormControl>
+          </ModalBody>
+          <ModalFooter>
+            <Button mr={3} variant="link" onClick={onClose}>
+              <Text textStyle="subhead-1" colorScheme="orange">
```
seaerchin, line 147: supernit: could we change `colorScheme` to `textColor` as chakra actually leaves some `colorSchemes` unimplemented and it might be confusing in the future.

also, [this prop](https://chakra-ui.com/docs/components/text/props) doesn't seem to be implemented so i'm not q sure if this is working as expected

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -80,40 +91,73 @@ const EmptyResourceRoom = () => {
...
+                Cancel
+              </Text>
+            </Button>
+            <LoadingButton
+              isDisabled={!_.isEmpty(errors)}
+              type="submit"
+              onClick={handleSubmit(onSubmit)}
+            >
```
seaerchin, line 155: oops, could we use `Button` and pass the `isLoading` prop? this component is legacy (should've marked it in JSDoc as dep also psps) and it's not ideal to use it.

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -80,40 +91,73 @@ const EmptyResourceRoom = () => {
...
+                  required: "Please enter resource room name",
+                })}
+              />
+              <FormErrorMessage>
+                {errors.newDirectoryName?.message}
+              </FormErrorMessage>
+            </FormControl>
+          </ModalBody>
+          <ModalFooter>
+            <Button mr={3} variant="link" onClick={onClose}>
```
seaerchin, line 146: could we double check if this is supposed to be `link` or `clear`?

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -199,20 +243,40 @@ const ResourceRoomContent = ({
...
-            <CreateButton as={RouterLink} to={`${url}/createDirectory`}>
-              Create category
-            </CreateButton>
-          </SectionHeader>
-          <SectionCaption icon={BiBulb} label="PRO TIP: ">
-            Categories impact navigation on your site. Organise your resources
-            by creating categories.
-          </SectionCaption>
-        </Box>
+        {directoryData.length !== 0 && (
```
seaerchin, line 246: we can inline this below unless the concern is the `Skeleton`

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -199,20 +243,40 @@ const ResourceRoomContent = ({
...
+            <VStack spacing={5}>
+              <EmptyBoxImage />
+              <VStack spacing={0}>
+                <Center textStyle="subhead-1">
+                  There&apos;s nothing here yet.
+                </Center>
+                <Center textStyle="body-2">
+                  Create a resource category to get started.
+                </Center>
+              </VStack>
```
seaerchin, line 270: see above wrt combining

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -199,20 +243,40 @@ const ResourceRoomContent = ({
...
+              <Button
+                variant="solid"
```
seaerchin, line 272: extra prop

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -80,40 +91,73 @@ const EmptyResourceRoom = () => {
...
+          </ModalBody>
+          <ModalFooter>
+            <Button mr={3} variant="link" onClick={onClose}>
+              <Text textStyle="subhead-1" colorScheme="orange">
+                Cancel
+              </Text>
+            </Button>
+            <LoadingButton
+              isDisabled={!_.isEmpty(errors)}
+              type="submit"
```
seaerchin, line 153: does this work? there doesn't seem to be a `form` higher up in the DOM from what i can tell

src/assets/images/EmptyBoxImage.tsx
```diff
@@ -0,0 +1,40 @@
+export const EmptyBoxImage = (
+  props: React.SVGProps<SVGSVGElement>
```
seaerchin, line 2: should we spread this over the `svg`?

seaerchin (changes requested):

seaerchin:
need rebase, feel free to ping me on slack when done

src/assets/images/EmptyBoxImage.tsx
```diff
@@ -0,0 +1,38 @@
+export const EmptyBoxImage = (): JSX.Element => {
```
seaerchin, line 1: we should have a props here that we spread over the `svg` - this allows consumers to style the `svg` if needed

src/layouts/ResourceCategory/ResourceCategory.tsx
```diff
@@ -51,22 +61,49 @@ export const ResourceCategory = (): JSX.Element => {
...
-              <CreateButton as={RouterLink} to={`${url}/createPage`}>
-                Create page
-              </CreateButton>
-            </SectionHeader>
-            <SectionCaption icon={BiBulb} label="NOTE: ">
-              Pages are automatically ordered by latest date.
-            </SectionCaption>
-          </Box>
+          {/* 
+              There is no loading state for this as '!arePagesEmpty' will only evulate to
```
seaerchin, line 65: nit: evaluate

src/layouts/ResourceRoom/ResourceRoom.stories.tsx
```diff
@@ -50,6 +55,19 @@ Empty.parameters = {
...
+}
+
+export const EmptyCategory = Template.bind({})
+EmptyCategory.parameters = {
+  msw: {
+    handlers: {
+      resourceRoomData: buildResourceRoomData([]),
+      resourceRoomName: buildResourceRoomName({
+        resourceRoomName: MOCK_RESOURCE_ROOM_NAME,
+      }),
```
seaerchin, line 70: just an fyi, i actually added these stories upstream also - if your PR gets merged into `develop` first, i'll remove mine!

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -80,40 +92,72 @@ const EmptyResourceRoom = () => {
     }
   }, [errorToast, isError])
 
+  const { isOpen, onOpen, onClose } = useDisclosure()
+
   return (
-    <SiteViewLayout>
-      <Box as="form" w="full">
+    <SiteViewLayout
+      divider={undefined} // to remove bottom divider line
```
seaerchin, line 99: should we prefer using fragments for this use case? ie, 
```ts
<>
<SiteViewLayout />
<Modal />
</>
```

this avoids having to sidestep the divider

seaerchin (changes requested):

src/components/EmptyArea.tsx
```diff
@@ -0,0 +1,45 @@
+import { Box, Center, VStack, Text, HTMLChakraProps } from "@chakra-ui/react"
+import { ReactFragment } from "react"
+
+import { EmptyBoxImage } from "assets"
+
+export interface EmptyAreaOption extends HTMLChakraProps<"div"> {
```
seaerchin, line 6: nit: could we rename this to `EmptyAreaProps`? i think the convention is to use `xxxProps` so we should stick to it!

also, we should probably extend `StackProps` from chakra as the wrapper element is a `VStack`

src/components/EmptyArea.tsx
```diff
@@ -0,0 +1,45 @@
...
+  isItemEmpty: boolean
+  actionButton: ReactFragment
```
seaerchin, line 8: this might be wrong - `ReactFragment` specifically refers to `<> </>` ._.  as an aside, this actually accepts jsx elements, which i didn't know so til! 

the side effect of this is that now, this becomes a required prop, which isn't ideal - wdyt about chagning this to `JSX.Element`?

src/components/EmptyArea.tsx
```diff
@@ -0,0 +1,45 @@
+import { Box, Center, VStack, Text, HTMLChakraProps } from "@chakra-ui/react"
+import { ReactFragment } from "react"
+
+import { EmptyBoxImage } from "assets"
+
+export interface EmptyAreaOption extends HTMLChakraProps<"div"> {
+  isItemEmpty: boolean
```
seaerchin, line 7: this should be an optional prop - this makes the API stricter without offering additional benefits. eg: if an initial call to an API returns `undefined`, the user now has to do a cast, which is additional friction for this component

src/components/EmptyArea.tsx
```diff
@@ -0,0 +1,45 @@
+import { Box, Center, VStack, Text, HTMLChakraProps } from "@chakra-ui/react"
```
seaerchin, line 1: there's nothing wrong with putting this in `src/components` but this implies that the # of use-cases this component has to cover would be much wider than if it was limited. 

this impacts how the component should be structured - eg: presence of button/text. one way people get around this (see: chakra's [menu](https://chakra-ui.com/docs/components/menu#menubutton-props) or [modal](https://chakra-ui.com/docs/components/modal#block-scrolling-when-modal-opens)) is by having many related components grouped together and exported under a common prefix (or as per design system, under a grouped export).

the rationale to allow for more flexibility here is good (by adopting optional arguments) - but this makes a judgement that the text **has to exist**, which might not be ideal. could i check here if this has been cleared by our designers and also, that empty states going forward have been tied down to this

src/components/EmptyArea.tsx
```diff
@@ -0,0 +1,45 @@
...
+  mainText,
+  subText,
+  actionButton,
+  children,
+  isItemEmpty,
+}: EmptyAreaOption): JSX.Element => {
+  return (
+    <>
+      {isItemEmpty ? (
+        <Box as="form" w="full">
```
seaerchin, line 23: why do we need a `form` here?

src/components/EmptyArea.tsx
```diff
@@ -0,0 +1,45 @@
...
+  subText?: string
+}
+
+export const EmptyArea = ({
+  mainText,
+  subText,
+  actionButton,
+  children,
+  isItemEmpty,
+}: EmptyAreaOption): JSX.Element => {
```
seaerchin, line 19: we've given this element a more powerful api by extending chakra props but the extra flexibility isn't being used - should we consider having a `...rest` argument which is spread over the `VStack`?

src/layouts/ResourceRoom/ResourceRoom.stories.tsx
```diff
@@ -53,16 +55,19 @@ Empty.parameters = {
...
     handlers: {
       resourceRoomData: buildResourceRoomData([]),
+      resourceRoomName: buildResourceRoomName({ resourceRoomName: "" }),
     },
   },
 }
 
-export const NoResourceRoom = Template.bind({})
-NoResourceRoom.parameters = {
+export const EmptyCategory = Template.bind({})
```
seaerchin, line 63: supernit: could we change this to `EmptyResourceCategory`

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -332,6 +362,7 @@ const ResourceRoomContent = ({
                 <FormLabel>Resource room title</FormLabel>
                 <Input
                   placeholder="New resource room name"
+                  // eslint-disable-next-line react/jsx-props-no-spreading
```
seaerchin, line 365: can just ignore cos it's a warning hahahahahah

src/layouts/layouts/SiteViewLayout/SiteViewLayout.tsx
```diff
@@ -30,6 +30,7 @@ export const SiteViewLayout = ({
 }: StackProps): JSX.Element => {
   return (
     <>
+      {/* eslint-disable-next-line react/jsx-props-no-spreading */}
```
seaerchin, line 33: can remove if u wanna but it's fine to leave too

src/layouts/ResourceRoom/ResourceRoom.stories.tsx
```diff
@@ -60,8 +60,8 @@ Empty.parameters = {
   },
 }
 
-export const EmptyCategory = Template.bind({})
-EmptyCategory.parameters = {
+export const EmptyResoruceCategory = Template.bind({})
+EmptyResoruceCategory.parameters = {
```
seaerchin, line 64: nit: rename

src/components/EmptyArea.tsx
```diff
@@ -1,11 +1,10 @@
 import { Box, Center, VStack, Text, HTMLChakraProps } from "@chakra-ui/react"
-import { ReactFragment } from "react"
 
 import { EmptyBoxImage } from "assets"
 
-export interface EmptyAreaOption extends HTMLChakraProps<"div"> {
+export interface EmptyAreaProps extends HTMLChakraProps<"div"> {
   isItemEmpty: boolean
```
seaerchin, line 6: could we make this prop optional?

src/layouts/ResourceRoom/ResourceRoom.tsx
seaerchin: i think this^ looks better instead of having `divider = {undefined}` - wdyt?

src/layouts/ResourceRoom/ResourceRoom.tsx
seaerchin: could we do this^

src/components/EmptyArea.tsx
seaerchin: If they don't pass it, won't it default to `undefined` and hence, we can still do a type conversion under the hood?

src/components/EmptyArea.tsx
seaerchin: Sure! I think this is just something to be aware of next time. Wrt this component, it's general enough to be placed here and the given API is fine (if text is optional).

src/layouts/ResourceRoom/ResourceRoom.tsx
seaerchin: Actually this is because the component has crashed before due to the `useEffect` ._.

src/layouts/ResourceRoom/ResourceRoom.tsx
seaerchin: I think we should be using design system's `Button`, and the equivalent variant is `clear`; sorry if this was unclear (pun intended)

src/components/EmptyArea.tsx
kishore03109: Hmm when would you have a case where I wont pass it in `isItemEmpty` first place ah? This JSX element is designed such that `isItemEmpty` is `true`, it would show the empty state, else it would render its `children`. I cant think of any case that we would need only show the empty state at **all** times, am I mistaken here?

src/components/EmptyArea.tsx
kishore03109: to be discussed under here [instead](https://github.com/isomerpages/isomercms-frontend/pull/993#discussion_r959316420)

src/layouts/ResourceRoom/ResourceRoom.stories.tsx
kishore03109: ee2d6e

src/components/EmptyArea.tsx
kishore03109: (repost because Idk why the github is duplicating shit ._.)
hmm I disagree here: isItemEmpty is used [here](https://github.com/isomerpages/isomercms-frontend/blob/add-empty-state-for-empty-resource/src/components/EmptyArea.tsx#L22) If I were to make this optional, then developers might not pass this parameter when calling EmptyArea, which then renders this whole component useless.

src/layouts/ResourceRoom/ResourceRoom.tsx
kishore03109: +1 for comment -1 for lame pun

src/layouts/ResourceRoom/ResourceRoom.tsx
kishore03109: eh I think got misunderstanding here bah - this components  for this empty state will be shown below the entire page leh, unless we redefine what `SiteViewLayout` does which I dont feel should be the case here

src/layouts/ResourceRoom/ResourceRoom.tsx
kishore03109: 03555cc

src/layouts/ResourceRoom/ResourceRoom.tsx
seaerchin: hmm i can't repro this! feel free to ping me to pair on it

src/layouts/ResourceRoom/ResourceRoom.tsx
seaerchin: hmm from the [spec](https://www.w3.org/TR/html4/interact/forms.html#buttons) itself, they say that 
> controls outside a form cannot be [successful controls](https://www.w3.org/TR/html4/interact/forms.html#successful-controls).

which seems to imply that a `form` element is required .-.

seaerchin (commented):
overall looks fine! just 2 minor comments wrt dom structure. will approve once addressed!

src/layouts/ResourceRoom/ResourceRoom.tsx
seaerchin: this actually seems fine now!

kishore03109:
> not about the errors but more of setting the type = "submit", which requires that there's a DOM node higher up. if you've verified that this works, i don't have an issue with it (apart from figuring out how it works AHHA)

@seaerchin correct me if I am wrong ya, but the reason this works is because the `handleSubmit` is from the `useForm` hook (This still works when I completely remove the `type` attribute as a whole)

src/layouts/ResourceRoom/ResourceRoom.tsx
seaerchin: resolved offline - think the `onClick` is the one that allows this to work

src/layouts/ResourceRoom/ResourceRoom.tsx
seaerchin: fixed

src/components/EmptyArea.tsx
```diff
@@ -0,0 +1,45 @@
+import { Box, Center, VStack, Text, HTMLChakraProps } from "@chakra-ui/react"
+
+import { EmptyBoxImage } from "assets"
+
+export interface EmptyAreaProps extends HTMLChakraProps<"div"> {
```
seaerchin, line 5: nit: change htis to `StackProps`

seaerchin (approved):
lgtm with minor nit

PR #998, opened 2022-08-01
fix(context-menu): update context menu styling to fix UI breakage

seaerchin:
## Problem
Upgrading from cypress v6 to v10 also involved upgrading certain other packages. in [this](https://github.com/isomerpages/isomercms-frontend/pull/991/commits/197527712b2cab71f54a1f8867a36633c70ccfd8) commit on github, chakra was upgraded from v1.8.3 to v1.8.8 and this (was suspected to have) caused a visual bug on the frontend, where the context menu is not bound due to the `LinkBox` applying css styles on top of the button. 

## Solution
This problem came about because the `LinkBox` applied special styling onto its child components to ensure a11y compat. hence, the fix was to define a parent component for both `LinkBox` and `ContextMenu` in situations where this bug occurred.


alexanderleegs (changes requested):
The `clickContextMenuItem` command might need to be updated - the refactored context menu button is causing tests involving the context menu to fail
Screenshot 2022-08-03 at 10 33 49 AM

seaerchin:
> Reference in ne

good catch, this slipped my mind when pushing this PR. i think this could have been caught with automated e2e actually - how fast does e2e run on your local machine now? 

i think if it's a reasonable amount of time (10-20 mins), we could consider running e2e per PR (v10 seems to be a lil faster)

alexanderleegs (approved):
lgtm

Issue #992, opened 2022-07-25
[FEATURE] Empty state for Workspace + Folders 


kishore03109:
### Problem 

Currently, Isomer does not really explicitly tell the user if there are empty folders or pages, as shown in the image below. 
<img width="1243" alt="Screenshot 2022-07-25 at 9 19 47 AM" src="https://user-images.githubusercontent.com/42832651/180675728-dc6906c6-8760-4708-831c-b1717a78fc24.png">

### Suggested solution 
Implement UI in the this workflow [shown in this figma](https://www.figma.com/file/7OtNYkc4s9BkFk05ConEbd/%E2%AD%90-V1.5?node-id=1539%3A57304).

PR #1000, opened 2022-08-03
feat(workspace): empty state for empty folders and pages

kishore03109:
## Problem
Currently, Isomer does not really explicitly tell the user if there are empty folders or pages, as shown in the image below.

Addresses #992 

## Solution

**Breaking Changes**

**Features**:

Empty states for 

## Before & After Screenshots

**BEFORE**:

Screenshot 2022-08-08 at 1 51 20 PM

**AFTER**:

*No page and folder*
Screenshot 2022-08-08 at 1 52 12 PM

*No folder _only_*
Screenshot 2022-08-08 at 1 53 27 PM

*No page _only_*
Screenshot 2022-08-08 at 1 55 26 PM


seaerchin:
hmm, if this is ready for review, should we request for reviewers?

kishore03109:
> hmm, if this is ready for review, should we request for reviewers?

@seaerchin hmm like other engineers or design ah?

seaerchin:
> > hmm, if this is ready for review, should we request for reviewers?
> 
> @seaerchin hmm like other engineers or design ah?

If we have design changes, then both! Otherwise, engineers only

seaerchin:
needs rebase; feel free to ping me on slack once done

src/layouts/Workspace/WorkspacePagesAndFoldersComponents.tsx
```diff
@@ -0,0 +1,233 @@
+import {
```
seaerchin, line 1: side note - we should probably shift all the components here into `/Workspace/compnents` as their own file rather than putting them in this file

src/layouts/Workspace/WorkspacePagesAndFoldersComponents.tsx
```diff
@@ -0,0 +1,233 @@
...
+import { PageCard, FolderCard } from "./components"
+
+export const Folders = (props: {
+  siteName: string
+  pagesData: PageData[]
+  url: string
+  dirsData: (PageData | DirectoryData)[]
+}): JSX.Element => {
+  const { setRedirectToPage } = useRedirectHook()
+  const { siteName, pagesData, url, dirsData } = props
```
seaerchin, line 45: we should probably inline this above

src/layouts/Workspace/WorkspacePagesAndFoldersComponents.tsx
```diff
@@ -0,0 +1,233 @@
...
+  CreateButton,
+} from "../components"
+
+import { PageCard, FolderCard } from "./components"
+
+export const Folders = (props: {
+  siteName: string
+  pagesData: PageData[]
+  url: string
+  dirsData: (PageData | DirectoryData)[]
```
seaerchin, line 42: hmm this is getting abit out of hand - we should probably create a type alias for readability.

also why is `dirsData` typed as `(PageData | DirectoryData)[]`? wouldn't `DirectoryData[]` suffice?

src/layouts/Workspace/FoldersAndPagesController.tsx
```diff
@@ -0,0 +1,89 @@
...
+    return (
+      <GridItem
+        area="content"
+        as={VStack}
+        spacing="2rem"
+        bgColor="gray.50"
+        w="100%"
+        h="100%"
+        divider={<StackDivider borderColor="border.divider.alt" />}
+      >
```
seaerchin, line 41: we seem to be redefining this for everything as a workaround for having the controller. this seems abit awkward and maybe we should have a context which the pages/folders both read from and then render accordingly

src/layouts/Workspace/Workspace.tsx
```diff
@@ -19,170 +18,103 @@ import {
...
-    params: { siteName },
-  } = useRouteMatch<{ siteName: string }>()
-
-  const { setRedirectToPage } = useRedirectHook()
-  const { url } = useRouteMatch()
-  const { data: _dirsData, isLoading: isDirLoading } = useGetFolders({
+// refers to the homepage, navigation bar and contact us pages.
+export const MainPages = (props: {
+  siteName: string
+  pagesData: PageData[]
```
seaerchin, line 34: we don't seem to be using this except as an indication of whether the page is loading. perhaps we could change this prop to `isLoading`?

src/layouts/Workspace/FoldersAndPagesController.tsx
```diff
@@ -0,0 +1,89 @@
...
+  Folders,
+  EmptyPage,
+} from "./WorkspacePagesAndFoldersComponents"
+
+export const FoldersAndPagesController = (props: {
+  siteName: string
+  url: string
+  pagesData: PageData[]
+}): JSX.Element => {
+  const { siteName, url, pagesData } = props
```
seaerchin, line 20: we should inline this directly above

src/layouts/Workspace/WorkspacePagesAndFoldersComponents.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+export const UngroupedPages = (props: {
+  pagesData: PageData[]
+  url: string
+}): JSX.Element => {
+  const { pagesData, url } = props
+  const { setRedirectToPage } = useRedirectHook()
+  return (
+    <>
+      <Section>
```
seaerchin, line 85: same comment

src/layouts/Workspace/WorkspacePagesAndFoldersComponents.tsx
```diff
@@ -0,0 +1,233 @@
...
+        <Box w="100%">
+          <SectionHeader label="Ungrouped Pages">
+            <CreateButton
+              onClick={() => setRedirectToPage(`${url}/createPage`)}
```
seaerchin, line 89: same comment

src/layouts/Workspace/WorkspacePagesAndFoldersComponents.tsx
```diff
@@ -0,0 +1,233 @@
...
+      </Section>
+    </>
+  )
+}
+
+export const EmptyFolder = (props: { url: string }): JSX.Element => {
+  const { url } = props
+  const { setRedirectToPage } = useRedirectHook()
+  return (
+    <>
```
seaerchin, line 116: same comment

seaerchin (changes requested):
i think using the controller and then redefining what `SiteViewLayout` already does seems pretty awkward here. 

the controller is deciding what to render for each subcomponent of `Folders/Pages` but couldn't we just do this already at the top level and bypass having the vstack? i might be missing something here so do lmk if this is so

src/layouts/Workspace/Workspace.tsx
```diff
@@ -19,170 +18,103 @@ import {
...
-  const dirsData = _dirsData || []
-  const pagesData = _pagesData || []
-
+  useEffect(() => {
+    if (contactUsPage)
+      setContactUsCard(
+        contactUsPage.content?.frontMatter?.layout ===
+          CONTACT_US_TEMPLATE_LAYOUT
+      )
+  }, [contactUsPage])
```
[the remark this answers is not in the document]
kishore03109: I am not too sure about this tbh, [this few lines of code were already in the codebase](https://github.com/isomerpages/isomercms-frontend/blob/develop/src/layouts/Workspace/Workspace.tsx#L61)..
My hypothesis is that the contactUsPage requires a API call to our backend, and there is a chance that the page has already rendered _before_ the round trip, thus the need to use `useEffect`, wdyt?

src/layouts/Workspace/Workspace.tsx
```diff
@@ -19,170 +18,103 @@ import {
...
-              pagesData.length > 0 &&
-              pagesData
-                .filter((page) => page.name !== "contact-us.md")
-                .map(({ name }) => <PageCard title={name} />)}
-          </SimpleGrid>
-        </Skeleton>
-      </Section>
-    </SiteViewLayout>
+    <Section>
+      <Text as="h2" textStyle="h2">
```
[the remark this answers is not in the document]
kishore03109: Hmm I did take this from somewhere else in the codebase haha 
Out of curiosity, why is this ideal ah? Wouldnt setting the `textStyle` or `as` suffice?

src/layouts/Workspace/Workspace.tsx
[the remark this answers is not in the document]
seaerchin: i actually removed it HAHA there's no difference and it's alot more complicated to read (as `useEffect` + `useState`). you might've accidentally added it back!

src/layouts/Workspace/Workspace.tsx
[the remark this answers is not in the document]
seaerchin: i'm not too sure whether setting the `as` prop will hook into the `textStyles` but this is mainly for `textStyles` that don't translate directly to a dom node type (eg: `textStyle = legal`).

this is ideal b/c html is still a markup language and setting the type of the node properly has semantic meaning to the rendering engine in your browser. (and as an aside, even though we don't care about this, it helps screen readers tremendously)

you could check the MDN docs [here](https://developer.mozilla.org/en-US/docs/Glossary/Semantics) for more info and/or google why you shouldn't style a `div` like a `button`

src/layouts/Workspace/Workspace.tsx
kishore03109: hmm `MainPages` here refer to the top three main pages (homepage, navigation, and contact us pages) that exist in all pages by default, not as a placeholder while the pages are loading!

kishore03109:
> i think using the controller and then redefining what `SiteViewLayout` already does seems pretty awkward here.
> 
> the controller is deciding what to render for each subcomponent of `Folders/Pages` but couldn't we just do this already at the top level and bypass having the vstack? i might be missing something here so do lmk if this is so

@seaerchin Yes, I do feel uncomfortable with this solution as well, we can huddle to see if there is a better way to overcome this. 
To summarise, the issue stems from the fact that the VStack does not resolve empty fragments. Eg. for the code below: 

```
<VStack
  divider={<StackDivider borderColor='gray.200' />}
  spacing={4}
  align='stretch'
>
  <Box h='40px' bg='yellow.200'>
    1
  </Box>
  <Box h='40px' bg='yellow.200'>
    2
  </Box>
  <>
      <Box h='40px' bg='tomato'>
        3
      </Box>
      <Box h='40px' bg='pink.100'>
        4
      </Box>
  </>
</VStack>
```

A divider will not be rendered between boxes 3 and 4, which is why I have to manually return wrap a `GridItem` around every single `return` statement for `FoldersAndPagesController`

src/layouts/Workspace/Workspace.tsx
seaerchin: i think you might've gotten confused - i'm not worried about the naming, more about the prop `pagesData`, which is only used as an indicator of whether it's loading. this should just be a boolean and called `isLoading` so that it's clearer and more indicative of what this component requires

src/layouts/Workspace/Workspace.tsx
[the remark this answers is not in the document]
kishore03109: As in this from [develop](https://github.com/isomerpages/isomercms-frontend/blob/develop/src/layouts/Workspace/Workspace.tsx#L61) HAHAA havent merge yet ah? 
If that is the case, I will resolve and use your updated code during merge!

src/layouts/Workspace/WorkspacePagesAndFoldersComponents.tsx
```diff
@@ -0,0 +1,233 @@
...
+        {/* Resource Room does not exist */}
+        <VStack spacing={5}>
+          <EmptyBoxImage />
+          <Center>
+            <VStack spacing={0}>
+              <Text textStyle="subhead-1">There&apos;s nothing here yet.</Text>
+              <Text textStyle="body-2">Create a new item to get started.</Text>
+            </VStack>
+          </Center>
+          <ContextMenu placement="bottom-end" offset={[0, 0]}>
```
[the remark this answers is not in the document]
kishore03109: Added in a comment for this in 63f0c70!

src/layouts/Workspace/FoldersAndPagesController.tsx
kishore03109: I understand that you wanted to remove the `FoldersAndPagesController` altogether. I have added an [experimental branch here](https://github.com/isomerpages/isomercms-frontend/blob/temp/experiment-removing-workspace-controller/src/layouts/Workspace/Workspace.tsx) (to be removed) 
I originally preferred against it as nested inline conditions might be hard to read (if you are ok with this, I will just cherry-pick this commit over to this branch). Please do let me know if you have other way of refactoring this to remove the controller entirely!

src/components/EmptyArea.tsx
```diff
@@ -0,0 +1,45 @@
+import { Box, Center, VStack, Text, HTMLChakraProps } from "@chakra-ui/react"
```
seaerchin, line 1: same comments as on prior pr!

src/layouts/ResourceRoom/ResourceRoom.tsx
```diff
@@ -116,10 +116,8 @@ const EmptyResourceRoom = () => {
           >
             <Input
               maxW="50%"
-              {...register("newDirectoryName", {
-                required:
-                  "Please ensure that you have entered a resource room name!",
-              })}
+              // eslint-disable-next-line react/jsx-props-no-spreading
+              {...register("newDirectoryName", { required: true })}
```
seaerchin, line 120: i think this has different functionality - the previous one has the error message (the string) but this does not. was this intended?

src/layouts/Folders/components/MenuDropdownButton.tsx
```diff
@@ -32,7 +32,10 @@ export const MenuDropdownButton = (): JSX.Element => {
           >
             Create page
           </Button>
-          <ContextMenu>
+          <ContextMenu
+            placement="bottom-end"
+            offset={[0, 0]} // to prevent weird gap between the dropdown menu and button>
```
seaerchin, line 37: could we give a brief overview of what `offset` does (eg: this positions it (x, y) units away relative from the base position)?

src/layouts/Workspace/Workspace.tsx
```diff
@@ -19,123 +22,75 @@ import {
...
+  const { data: _pagesData } = useGetWorkspacePages(siteName)
 
-  const hasContactUsCard =
-    contactUsPage?.content?.frontMatter?.layout === CONTACT_US_TEMPLATE_LAYOUT
-
-  const dirsData = _dirsData?.filter(isDirData) || []
   const pagesData = _pagesData || []
+  const dirsData = (_dirsAndPagesData || []).filter((data) => {
+    return data.type === "dir" // only get directories and not pages
+  }) as DirectoryData[]
```
seaerchin, line 52: we should not need the cast here - there should be a `isDirData` typeguard in `types/utils`. using that should prevent the type casting (through the return type of the function being asserted as `x is y`, where y is the type)

src/layouts/Workspace/components/WorkspacePagesAndFoldersComponents.tsx
```diff
@@ -0,0 +1,133 @@
+import { SimpleGrid, Box, Skeleton, Text } from "@chakra-ui/react"
```
seaerchin, line 1: could we shift the components here to `./components`

src/layouts/layouts/SiteViewLayout/SiteViewLayout.tsx
```diff
@@ -50,20 +51,30 @@ export const SiteViewLayout = ({
...
-        >
-          {children}
-        </GridItem>
+        <ContentGridLayout>{children}</ContentGridLayout>
       </Grid>
     </>
   )
 }
+
+export const ContentGridLayout = ({
```
seaerchin, line 60: nit: could we change the name to `SiteViewContent` so that the grouping w/ `SiteViewLayout` is clearer

src/components/EmptyArea.tsx
kishore03109: Hmm, gotcha, we shall address this is that other then!

src/layouts/ResourceRoom/ResourceRoom.tsx
kishore03109: good catch - have re-included this error message! adc5ecf

src/layouts/Folders/components/MenuDropdownButton.tsx
kishore03109: added in a comment adc5ecf

src/layouts/Workspace/Workspace.tsx
kishore03109: cce8135

src/layouts/layouts/SiteViewLayout/SiteViewLayout.tsx
kishore03109: ac4b18f

src/layouts/Workspace/components/MainPagesComponent.tsx
```diff
@@ -0,0 +1,46 @@
...
+        </SimpleGrid>
+      </Skeleton>
+    </Section>
+  )
+}
+
+export interface MainPagesProps {
+  siteName: string
+  isLoading: boolean
+}
```
seaerchin, line 46: supernit: place above component

seaerchin (approved):
lgtm; we could also remove the `component` from the filename so that it's `MainPages` vs `MainPagesComponent` but this is just a matter of preference

PR #1003, opened 2022-08-08
Feat: add acknowledgement on old login screen

alexanderleegs:
## Problem

This PR introduces the login acknowledgement on the existing login screen.

### DIFF
--- a/src/layouts/Home.tsx
+++ b/src/layouts/Home.tsx
@@ -26,8 +26,7 @@ const Home = (): JSX.Element => {
         By clicking “Log in”, you are acknowledging and agreeing to Isomer’s{" "}
         <Link
           href="https://guide.isomer.gov.sg/terms-and-privacy/terms-of-use"
-          rel="noopener noreferrer"
-          target="_blank"
+          isExternal
           fontSize="0.625rem"
         >
           Terms of Use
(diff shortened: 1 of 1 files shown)


src/layouts/Home.tsx
```diff
@@ -22,6 +22,26 @@ const Home = (): JSX.Element => {
...
         Login with GitHub
       </Button>
+      <Text color="text.helper" fontSize="0.625rem" pt="1rem">
+        By clicking “Log in”, you are acknowledging and agreeing to Isomer’s{" "}
+        <Link
+          href="https://guide.isomer.gov.sg/terms-and-privacy/terms-of-use"
+          rel="noopener noreferrer"
+          target="_blank"
+          fontSize="0.625rem"
+        >
```
seaerchin, line 32: actually i think chakra's `Link` has the `isExternal` property, which does this!

seaerchin (changes requested):

src/layouts/Home.tsx
alexanderleegs: Right, fixed in https://github.com/isomerpages/isomercms-frontend/pull/1003/commits/bd7a9f4fae5776ff8793241d67edcf99a76197a6

seaerchin (approved):

PR #1011, opened 2022-08-12
Fix/hitting character limit when adding <iframe/>

### LINKED ISSUE #994: [Bug] Unable to add iFrame videos due to character limit
### Problem

Since HTML code regarding iframe is part of the 400 character limit (see image below), this does not allow certain agencies to add in iframe. Currently, the workaround is to manually adding the code directly to the back end. 

Screenshot 2022-07-27 at 10 24 59 AM

### Proposed Solution 

Using a regex pattern to get instances of `<iframe></iframe>`, and then not including this subset of the sting as part of the total character count allowed.

kishore03109:
## Problem 

Since HTML code regarding iframe is part of the 400 character limit (see image below), this does not allow certain agencies to add in iframe. Currently, the workaround is to manually adding the code directly to the back end.

Closes #994

## Solution

Using a regex pattern to get instances of `<iframe></iframe>` and `<div></div>` (what our current users are using), and then not including this subset of the sting as part of the total character count allowed.

Note: This is meant to be a temporary workaround, once we move to block editing, this should not be an issue

### DIFF
--- a/src/utils/allowedHTML.ts
+++ b/src/utils/allowedHTML.ts
@@ -1,12 +1,8 @@
 /**
- * Defines the list of allowed HTML tags that do not count towards the total
+ * HTML tag that do not count towards the total
  * character limit.
- *
  */
-const regrexTests = [
-  new RegExp("(<iframe.*/iframe>)", "gm"),
-  new RegExp("(<div.*/div>)", "gm"),
-]
+const IFRAME_TAG_REGEX = new RegExp("(<iframe.*/iframe>)", "gm")
 
 /**
  * @param text raw text input by user
@@ -14,8 +10,8 @@ const regrexTests = [
  */
 export const getLengthWithoutTags = (text: string): number => {
   let finalText = text
-  for (const regrexTest of regrexTests) {
-    finalText = finalText.replace(regrexTest, "")
-  }
+
+  finalText = finalText.replace(IFRAME_TAG_REGEX, "")
+
   return finalText.length
 }
--- a/src/utils/allowedHTML.ts
+++ b/src/utils/allowedHTML.ts
@@ -9,5 +9,5 @@ const IFRAME_TAG_REGEX = new RegExp("(<iframe.*/iframe>)", "gm")
  * @returns the character length after removing specifc HTML tags
  */
 export const getLengthWithoutTags = (text: string): number => {
- return text.replace(IFRAME_TAG_REGEX, "").length
+  return text.replace(IFRAME_TAG_REGEX, "").length
 }
--- a/src/utils/allowedHTML.ts
+++ b/src/utils/allowedHTML.ts
@@ -3,14 +3,19 @@
  * character limit.
  *
  */
-
-const iFrameRegrexTest = new RegExp("(<iframe.*/iframe>)", "gm")
+const regrexTests = [
+  new RegExp("(<iframe.*/iframe>)", "gm"),
+  new RegExp("(<div.*/div>)", "gm"),
+]
 
 /**
  * @param text raw text input by user
  * @returns the character length after removing specifc HTML tags
  */
 export const getLengthWithoutTags = (text: string): number => {
-  const removeIFrameTags = text.replace(iFrameRegrexTest, "")
-  return removeIFrameTags.length
+  let finalText = text
+  for (const regrexTest of regrexTests) {
+    finalText = finalText.replace(regrexTest, "")
+  }
+  return finalText.length
 }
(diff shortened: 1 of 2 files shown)


src/utils/validators.js
```diff
@@ -560,7 +562,7 @@ const validateContactType = (contactType, value) => {
       break
     }
     case "other": {
-      if (value.length > CONTACT_DESCRIPTION_MAX_LENGTH) {
+      if (getLengthWithoutTags(value) > CONTACT_DESCRIPTION_MAX_LENGTH) {
         errorMessage = `Description should be shorter than ${CONTACT_DESCRIPTION_MAX_LENGTH} characters.`
```
alexanderleegs, line 566: I think we shouldn't be allowing iframe inside the contact us page - we can leave these as the original restriction

src/utils/validators.js
```diff
@@ -640,7 +642,9 @@ const validateLocationType = (locationType, value) => {
       break
     }
     case "description": {
-      if (value.length > LOCATION_OPERATING_DESCRIPTION_MAX_LENGTH) {
+      if (
+        getLengthWithoutTags(value) > LOCATION_OPERATING_DESCRIPTION_MAX_LENGTH
+      ) {
```
alexanderleegs, line 647: See above!

src/utils/allowedHTML.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Defines the list of allowed HTML tags that do not count towards the total
+ * character limit.
+ *
+ */
+const regrexTests = [
+  new RegExp("(<iframe.*/iframe>)", "gm"),
+  new RegExp("(<div.*/div>)", "gm"),
+]
```
alexanderleegs, line 9: Hmm why do we also allow div tags to bypass the check? This feels like it might be easily abusable to get abnormally long lines

alexanderleegs: Also nit: `regexTests`!

src/utils/allowedHTML.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * @param text raw text input by user
+ * @returns the character length after removing specifc HTML tags
+ */
+export const getLengthWithoutTags = (text: string): number => {
+  let finalText = text
+  for (const regrexTest of regrexTests) {
+    finalText = finalText.replace(regrexTest, "")
+  }
+  return finalText.length
```
seaerchin, line 20: hmm, we might want to express this as a `reduce` operation if it's cleaner but i'll leave it up to you!

src/utils/allowedHTML.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Defines the list of allowed HTML tags that do not count towards the total
+ * character limit.
+ *
+ */
+const regrexTests = [
+  new RegExp("(<iframe.*/iframe>)", "gm"),
+  new RegExp("(<div.*/div>)", "gm"),
```
seaerchin, line 8: could i ask why the regexp is not `<iframe>.*</iframe>` (technically still the same but a lil clearer) and why we are including the parenthesis in the regexp?

kishore03109: Sure, it was to allow for attributes for the <iframe> element. Eg. <iframe src="abc.xyz"></iframe> will not be matched!

src/utils/validators.js
kishore03109: makes sense - have made the change!

src/utils/allowedHTML.ts
kishore03109: hmm I only added it since our users used a `div` to make align their `iframe` - but I agree that this is easily abusable, made the change likewise

src/utils/allowedHTML.ts
seaerchin: would `<iframe .*>.*</iframe>` be clearer here?

src/utils/allowedHTML.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * @param text raw text input by user
+ * @returns the character length after removing specifc HTML tags
+ */
+export const getLengthWithoutTags = (text: string): number => {
+  let finalText = text
+
+  finalText = finalText.replace(IFRAME_TAG_REGEX, "")
+
+  return finalText.length
```
seaerchin, line 16:
```suggestion
 return text.replace(IFRAME_TAG_REGEX, "").length
```

seaerchin (changes requested):
minor change for clarity

seaerchin (approved):

PR #1012, opened 2022-08-15
refactor(media): shifts files/images to v1.5 UI

seaerchin:
stories are in #1013 

## Problem
`Media` (`Files/Images`) is still stuck at v1 UI, which is visually inconsistent with the rest of the workspace tabs. This PR migrates them to v1.5

## Solution
1. shift `Media` to ts and refactor some minor portions for ts.
2. shift cards to internal components of `Media`
  - for `ImagePreviewCard`, this is an **entirely new** component as our `Card` api sadly isnt flexible enough to support this use case. we should refactor `Card` sometime in the future to better support upcoming use cases shown in the figma


src/layouts/Media/Media.tsx
```diff
@@ -0,0 +1,184 @@
...
+            </Text>
+            <MediaBreadcrumbs />
+          </Box>
+        </Section>
+        <Section>
+          <SectionHeader label="Albums">
+            <CreateButton as={Link} to={`${url}/createMedia`}>
+              {`Create ${directoryLabel}`}
+            </CreateButton>
+          </SectionHeader>
```
alexanderleegs, line 90: This link should be to `${url}/createDirectory`!

src/types/utils.ts
```diff
@@ -1,5 +1,13 @@
...
   return (data as DirectoryData).type === "dir"
 }
+
+// NOTE: This assumes that mediaPath is proof that this is a media file.
+// The corresponding backend code (mediaFiles.js) does not appear to guarantee the existence of
+// this property on media files (or indeed, ANY property) and this is a best effort attempt,
+// which might fail in the future if new media files are discovered to be returned w/o mediaPath.
+export const isMediaData = (data: unknown): data is MediaData => {
+  return (data as MediaData).type === "file" && !!(data as MediaData).mediaPath
+}
```
alexanderleegs, line 13: Just checking, this is only for file data and not media directory data right?

src/layouts/Media/Media.tsx
```diff
@@ -0,0 +1,184 @@
...
+      directoryLabel: "directory",
+    }
+  }
+
+  return {
+    singularMediaLabel: "image",
+    pluralMediaLabel: "images",
+    directoryLabel: "album",
+  }
+}
```
alexanderleegs, line 57: This looks super useful, thanks for doing up a helper method! Is this the only place this is used? Otherwise we could probably shift it to a utils file

alexanderleegs (changes requested):
Left some comments!

Also, tests currently appear to be failing - this is probably due to the swap from js to ts, so the mock in `RouteSelector.spec` needs to be changed accordingly

src/layouts/Media/components/ImagePreviewCard.tsx
```diff
@@ -0,0 +1,128 @@
...
+          >
+            <Text>Edit details</Text>
+          </ContextMenu.Item>
+          <ContextMenu.Item
+            icon={<BiFolder />}
+            as={RouterLink}
+            to={`${url}/moveMedia/${encodedName}`}
+          >
+            --
+            <HStack spacing="4rem" alignItems="center">
```
alexanderleegs, line 109: Is this `--` meant to be here?

src/layouts/Media/components/MediaBreadcrumb.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }>()
+  const directories = mediaDirectoryName.split("%2F")
+  const firstItem = directories[0]
+
+  return (
+    <Breadcrumb spacing="2px" separator={<BiChevronRight color="text.body" />}>
+      {directories
+        .slice(1)
+        .reduce(
+          (acc, cur) => {
```
kishore03109, line 26: Hey could we rename `acc` and `cur`? Am not too sure what acc and cur means :(

src/types/directory.ts
```diff
@@ -38,6 +42,14 @@ export interface MediaData {
   type: "file"
 }
 
+// NOTE: This assumes that mediaPath is proof that this is a media file.
+// The corresponding backend code (mediaFiles.js) does not appear to guarantee the existence of
```
kishore03109: Hmm do you happen to have an example where the backed code returns a mediaPath without having a media file? While it would be beyond the scope of this PR - I feel that we can minimally document this bug as an issue as this front-end fix seems to me like a temporary workaround

src/types/utils.ts
kishore03109: Hmm do you happen to have an example where the backed code returns a mediaPath without having a media file? While it would be beyond the scope of this PR - I feel that we can minimally document this bug as an issue as this front-end fix seems to me like a temporary workaround

kishore03109: Hmm do you happen to have an example where the backed code returns a mediaPath without having a media file? While it would be beyond the scope of this PR - I feel that we can minimally document this bug as an issue as this front-end fix seems to me like a temporary workaround

src/layouts/Media/components/MediaDirectoryCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+          </CardBody>
+        </LinkOverlay>
+      </LinkBox>
+      <ContextMenu>
+        <ContextMenu.Button pos="absolute" />
+        <ContextMenu.List>
+          <ContextMenu.Item
+            icon={<BiEditAlt />}
+            as={RouterLink}
+            to={`/sites/${siteName}/folders/${title}`}
```
kishore03109, line 49: This link seems to be wrong: 
Expected: `/sites/${siteName}/media/images/mediaDirectory/images%2F${title}`
Actual: `/sites/${siteName}/folders/${title}`

Currently, this `Edit folder` button leads to a not found page!

kishore03109 (changes requested):

src/layouts/Media/Media.tsx
seaerchin: good catch, thanks!

src/types/utils.ts
seaerchin: @alexanderleegs - yep! i did a `filter` to ensure that the types are correct

@kishore03109 - actually i **don't**. it would be a bug if we expect the property to exist but it doesn't. for this case, my confusion stems from `MediaFileService.js`, where we do a `find` operation across an array of objects. this doesn't guarantee existence of the object (leading to `sha` to be `undefined`). 

the BE does guarantee that `mediaPath` exists but this might not be true going forward and hence, my comment is more of a warning than proof of a bug ._.

src/layouts/Media/components/ImagePreviewCard.tsx
seaerchin: good catch - removed!

src/layouts/Media/components/MediaBreadcrumb.tsx
seaerchin: yep oops this is on me - forgot to rename variables for clarity...

seaerchin:
> Just to confirm, is the empty state for files/images (the empty box image that Kishore is working on) to be added in later?

@alexanderleegs i think kishore only did empty states for workspace + resources; unfortunately, this wasn't covered ._.

seaerchin:
also, @alexanderleegs @kishore03109 - this PR was p sloppy from me... thanks for reviewing and apologies!!!

kishore03109 (approved):

alexanderleegs (approved):
lgtm

PR #1013, opened 2022-08-15
ref(media): add stories for media tab

seaerchin:
## Problem
follow up to #1012 - this PR adds stories for `Media`.


alexanderleegs (approved):

kishore03109 (approved):
+1 for cat images

PR #1015, opened 2022-08-16
fix: add support for site url config parameter

dcshzj:
## Problem

Closes https://github.com/isomerpages/isomercms-backend/issues/454.

## Solution

**Breaking Changes**

**Bug Fixes**:

- The `url` configuration parameter in `_config.yml` can now be configured on the frontend with a validation of the domain name if given.

## Before & After Screenshots

**BEFORE**:
Screenshot 2022-08-16 at 16 25 49

**AFTER**:
Screenshot 2022-08-16 at 16 25 04

## Tests

Frontend PR for https://github.com/isomerpages/isomercms-backend/pull/490.

## Deploy Notes

*None*

### DIFF
--- a/src/hooks/settingsHooks/useGetSettings.ts
+++ b/src/hooks/settingsHooks/useGetSettings.ts
@@ -29,6 +29,7 @@ const DEFAULT_BE_STATE = {
   faq: "",
   show_reach: false,
   logo: "/images/isomer-logo.svg",
+  url: "",
 }
 
 const DEFAULT_SOCIAL_MEDIA_SETTINGS: SiteSocialMediaSettings = {
@@ -89,14 +90,7 @@ const convertfromBe = (backendSettings: BackendSiteSettings): SiteSettings => {
   }
 
   const rest = _(backendSettings)
-    .omit([
-      ...TOGGLED_VALUES,
-      "socialMediaContent",
-      "colors",
-      // These properties are extra and will lead to errors in validation
-      "resources_name",
-      "url",
-    ])
+    .omit([...TOGGLED_VALUES, "socialMediaContent", "colors"])
     .pickBy()
     .value()
 
--- a/src/layouts/Settings/GeneralSettings.tsx
+++ b/src/layouts/Settings/GeneralSettings.tsx
@@ -53,6 +53,20 @@ export const GeneralSettings = ({
             {...register("description")}
           />
         </FormControl>
+        <FormControl isDisabled={isError} isRequired>
+          <Box mb="0.75rem">
+            <FormLabel mb={0}>Site URL</FormLabel>
+            <FormLabel.Description color="text.description">
+              This is the URL that the Isomer site will be published to.
+            </FormLabel.Description>
+          </Box>
+          <Input
+            w="100%"
+            id="url"
+            placeholder="https://"
+            {...register("url", { required: true })}
+          />
+        </FormControl>
         <FormControl isDisabled={isError}>
           <Flex justifyContent="space-between" w="100%">
             <FormLabel>Display government masthead</FormLabel>
--- a/src/layouts/Settings/GeneralSettings.tsx
+++ b/src/layouts/Settings/GeneralSettings.tsx
@@ -1,9 +1,23 @@
-import { VStack, FormControl, Flex, Box } from "@chakra-ui/react"
-import { FormLabel, Input, Textarea } from "@opengovsg/design-system-react"
+import {
+  VStack,
+  FormControl,
+  Flex,
+  Box,
+  InputLeftAddon,
+  InputGroup,
+} from "@chakra-ui/react"
+import {
+  FormErrorMessage,
+  FormLabel,
+  Input,
+  Textarea,
+} from "@opengovsg/design-system-react"
 import { FormTitle } from "components/Form"
-import { useFormContext } from "react-hook-form"
+import { useFormContext, useFormState } from "react-hook-form"
 
 import { Section, SectionHeader } from "layouts/components"
+
+import { DOMAIN_NAME_REGEX } from "utils/validators"
 
 import { FormToggle } from "./components/FormToggle"
 import { SettingsFormFieldMedia } from "./components/SettingsFormFieldMedia"
@@ -16,6 +30,7 @@ export const GeneralSettings = ({
   isError,
 }: GeneralSettingsProp): JSX.Element => {
   const { register } = useFormContext()
+  const { errors } = useFormState()
 
   return (
     <Section id="general-fields">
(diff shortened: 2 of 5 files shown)


NatMaeTan:
This solution looks good! Is this component pulled from the base design system?

dcshzj:
> This solution looks good! Is this component pulled from the base design system?

@NatMaeTan The "https://" part is not from the OGP design system, it was taken from Chakra UI directly.

src/utils/validators.js
```diff
@@ -16,6 +16,8 @@ export const URL_REGEX_PREFIX = "^(https://)?(www.)?("
 export const URL_REGEX_SUFFIX = ".com/)([a-zA-Z0-9_-]+([/.])?)+$"
 export const TELEGRAM_REGEX = "telegram|t).me/([a-zA-Z0-9_-]+([/.])?)+$"
 export const TIKTOK_REGEX = ".com/@)([a-zA-Z0-9_-]+([/.])?)+$"
+export const DOMAIN_NAME_REGEX =
+  "(?:[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?\\.)+[a-z0-9][a-z0-9-]{0,61}[a-z0-9]"
```
kishore03109, line 20: Hmm out of curiosity, why do we need the `{0,61}`

seaerchin, line 20: we should add a lengthy comment here explaining what this regex does - this is p long and i'm not sure i understand it fully even w/ the help of regex101 (my lifesaver for regex ;--;)

also, does the initial `?` do anything? i was under the impression that it captures the previous group (of which there is none here)

could you also help me to understand teh `{0, 61}` - why do we have an upper bound here?

kishore03109 (commented):
Just to clarify ya - is this functionality already working - or is this currently purely a front-end component? Asking as I cant seem to save the changes whenever I put in a web domain!

src/hooks/settingsHooks/useGetSettings.ts
```diff
@@ -89,14 +90,7 @@ const convertfromBe = (backendSettings: BackendSiteSettings): SiteSettings => {
...
   const rest = _(backendSettings)
-    .omit([
-      ...TOGGLED_VALUES,
-      "socialMediaContent",
-      "colors",
-      // These properties are extra and will lead to errors in validation
-      "resources_name",
-      "url",
-    ])
+    .omit([...TOGGLED_VALUES, "socialMediaContent", "colors"])
```
seaerchin, line 93: double checking here, from the issue linked in the PR description, it seems like this PR only deals w/ `url`. could you help me understand why `resources_name` is being removed?

src/layouts/Settings/GeneralSettings.tsx
```diff
@@ -53,6 +68,30 @@ export const GeneralSettings = ({
...
+            <Input
+              w="100%"
+              id="url"
+              placeholder="www.open.gov.sg"
+              {...register("url", {
+                required: true,
+                pattern: RegExp(DOMAIN_NAME_REGEX),
+              })}
+            />
+          </InputGroup>
```
seaerchin, line 90: double checking here - the intent is for them to be secure right (hence the `https://` prefix)? however, when we capture this, we **do not** validate that their site is actually `https` - could we just check this by entering a website prefixed with `http://` like `http://info.cern.ch/`?

seaerchin (changes requested):
no complains on the code structure/design side - however, we need to check if this functions as we expect and actually enforces a `https://` structure. also, if we do decide to forcefully append `https://`, then we need to either check that it's valid (some sites **do not** have a ssl cert ._.) or warn the users that it'll be a deadlink if we don't check

src/utils/validators.js
dcshzj: [RFC1035](https://www.rfc-editor.org/rfc/rfc1035) specifies that each label is limited to 63 characters, then there is a `[a-z0-9]` before and after the `{0,61}` part, so 63 - 2 = 61.

dcshzj: I think that's a good idea, thanks! I have added a comment to explain each part of the regex plus some context behind some of the values. I also removed the `?:` part as it only differentiates between a capturing and non-capturing group, which is not relevant in this context.

For the `{0,61}`, it is part of RFC1035.

src/hooks/settingsHooks/useGetSettings.ts
dcshzj: The backend PR https://github.com/isomerpages/isomercms-backend/pull/490 removes the `resources_name` parameter from the settings API endpoint, so that would no longer be a need for this omission to be done on the frontend.

src/layouts/Settings/GeneralSettings.tsx
dcshzj: I added the `^` and `$` part in the regex to enforce that the user would not be able to input `https://` at the front. However, we do not have the ability to check or enforce that the destination site has https enabled unfortunately.

dcshzj:
> Just to clarify ya - is this functionality already working - or is this currently purely a front-end component? Asking as I cant seem to save the changes whenever I put in a web domain!

@kishore03109 The functionality needs the `fix/settings-parameters` branch on the backend in order for the url setting to be properly saved.

> no complains on the code structure/design side - however, we need to check if this functions as we expect and actually enforces a `https://` structure. also, if we do decide to forcefully append `https://`, then we need to either check that it's valid (some sites **do not** have a ssl cert ._.) or warn the users that it'll be a deadlink if we don't check

@seaerchin My understanding from @alexanderleegs is that all our supported sites have https enabled (either Let's Encrypt or a custom cert provided by them). Are there any sites that currently do not have https enabled?

src/hooks/settingsHooks/useGetSettings.ts
seaerchin: ah gotcha

src/utils/validators.js
seaerchin: thanks for doing this!

src/layouts/Settings/GeneralSettings.tsx
seaerchin: yeap, we also don't forcefully append the `https://` prefix here - should we do that then? wrt whether **all** sites have `https` by default - i want to say yes but i'm not sure. i'll just wait for @alexanderleegs to shed some light here and if this is so, then it's ok

alexanderleegs: Yeah they should have - government sites should be using `https://` anyway, and if the cert isn't working, that should probably be fixed before SEO is a concern

seaerchin: ok so just to double check - sites will always have `https://` and if their cert expires, how will they know on their side? (i think out of scope here but just piggy backing HAHAHA i think the redir server will resolve hosted sites to `https://www` iirc so i'm fine w/ approving this)

seaerchin (approved):
lgtm - feel free to merge once you clear the copywriting w/ design/product

NatMaeTan:
Yup copy has been cleared, and I see that we don't yet have this component in the base design system – but we should add it in! Will let the team know. But otherwise lgtm

NatMaeTan (approved):

seaerchin:
> Yup copy has been cleared, and I see that we don't yet have this component in the base design system – but we should add it in! Will let the team know. But otherwise lgtm

it's in chakra alr which we can use fine as-is. i suspect that we're not adding it in because the chakra one is alr fine AHHAHA

src/layouts/Settings/GeneralSettings.tsx
dcshzj: I'm guessing the uptime robot checks for any SSL errors? I randomly picked a few sites and they all have the following checked:

Screenshot 2022-08-18 at 15 05 21

PR #1016, opened 2022-08-17
fix(folders): fixed floating context menu for cards in folders

seaerchin:
**note: updates to e2e tests will come in a subsequent PR**
## Problem
See #998 for the issue - this problem also exists in folders; however, this problem only showed up now because of 2 reasons
1. `Folders` didn't have its card component updated to fit figma at point of merge for the context menu PR. 
2. This problem exists only when using an updated chakra version after merging the cypress update PR

## Solution
Refer to #998 for more deets but the only concrete change is to add a full width `Box` as the parent of `Card` and `ContextMenu`


alexanderleegs (approved):
lgtm

PR #1017, opened 2022-08-17
fix(folders.spec): change selector to account for altered dom structure

seaerchin:
## Problem
Altered dom structure in the prior PR #1016 meant that the existing tests that relied on clicking the context menu might fail here. 

## Solution
1. change selector so that context menu clicking works

### DIFF
--- a/cypress/e2e/folders.spec.ts
+++ b/cypress/e2e/folders.spec.ts
@@ -174,7 +174,7 @@ describe("Folders flow", () => {
 
     // Rename
     it("Should be able to rename a sub-folder", () => {
-      cy.contains("button", PRETTIFIED_SUBFOLDER_WITH_PAGES_TITLE)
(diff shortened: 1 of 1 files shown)


cypress/e2e/folders.spec.ts
```diff
@@ -353,7 +353,7 @@ describe("Folders flow", () => {
     })
   })
 
-  describe("Create page, delete page, edit page settings in subfolder", () => {
+  describe.only("Create page, delete page, edit page settings in subfolder", () => {
```
alexanderleegs, line 356: Can we remove the only here?

alexanderleegs (approved):
Minor change, but otherwise lgtm

cypress/e2e/folders.spec.ts
seaerchin: good catch! missed this out when pushing to origin ._.

PR #1027, opened 2022-08-24
fix(hooks): update query keys for page and directory deletion hooks

seaerchin:
## Problem
in #973, some deletion hooks did not have their keys updated and this resulted in pages and folders in `resources` not automatically updating even when the pages/folders within were deleted.  

## Solution
1. update query keys
2. pass `params` and omit extra keys

**note**: this is a bandaid solution and a more permanent fix would be to refactor these deletion hooks akin to what has been done in #973 instead


kishore03109 (approved):
lgtm, thanks for the quick fix!

PR #1035, opened 2022-08-26
fix(e2e-tests): update e2e tests 

### LINKED ISSUE #1033: [BUGS] QA bugs found during release
### Problem
this is verbatim from slack 
- [x] most (if not all) skipped tests are due to our new field in Settings - because it’s set as required, it is not backward compatible. hence, all sites that exist now will need to fill this in to enable saving.
- [x] The failing [bug](https://dashboard.cypress.io/projects/nxbty1/runs/104/test-results/baf5d2ef-f5bc-475a-b3fd-374fa793402a)  here (moving resources across resource categories) is a true positive, prob an issue with the the wrong keys being invalidated
- [x] resource pages (resourceCategories.spec) bug: creating resource page of type file does not cause new file to show up until refresh (react-query issue i think), neither does editing page settings
- [x] reordering items do not work due to the `chakra.button` wrapping
- [ ] #1038
- [ ] #1039

seaerchin:
## Problem
see #1033 for an overview of the bugs uncovered during release ;--; 

Partially resolves #1033

## Solution
1. add query key invalidation for `move`/`createPage` hooks
2. update cypress test command to fill in the SEO by default
3. add success/failure e2e test for SEO Field

### DIFF
--- a/cypress/e2e/settings.spec.ts
+++ b/cypress/e2e/settings.spec.ts
@@ -1,4 +1,8 @@
-import { TEST_PRIMARY_COLOR, TEST_REPO_NAME } from "../fixtures/constants"
+import {
+  TEST_PRIMARY_COLOR,
+  TEST_REPO_NAME,
+  BASE_SEO_LINK,
+} from "../fixtures/constants"
 
 describe("Settings page", () => {
   const BASE_TITLE = "TEST"
@@ -58,6 +62,11 @@ describe("Settings page", () => {
       .parent()
       .find("input")
       .each((elem) => cy.wrap(elem).clear())
+    cy.contains("label", "SEO")
+      .parent()
+      .parent()
+      .find("input")
+      .type(BASE_SEO_LINK)
     cy.contains("label", "Primary")
       .parent()
       .find('button[aria-label="Select colour"]')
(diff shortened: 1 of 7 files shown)


cypress/e2e/settings.spec.ts
```diff
@@ -124,6 +133,40 @@ describe("Settings page", () => {
...
+
+    // Act
+    cy.get("@seoInput").clear().type(expected)
+
+    // Assert
+    cy.saveSettings()
+    cy.get("@seoInput").should("have.value", expected)
+  })
+
+  it("should not be able to input the protocol at the beginning of the url", () => {
```
dcshzj, line 149: I think we can add at least 2 more invalid inputs:
1. A domain with a subdirectory (like `open.gov.sg/space/`)
2. An invalid domain (some random characters)

src/hooks/moveHooks/useMoveHook.jsx
```diff
@@ -29,6 +32,7 @@ export function useMoveHook(params, queryParams) {
       if (queryParams && queryParams.onError) queryParams.onError()
     },
     onSuccess: (resp) => {
+      console.log(params)
```
dcshzj, line 35: I think this can be removed.

dcshzj (changes requested):

src/hooks/moveHooks/useMoveHook.jsx
seaerchin: Oops, good catch!

cypress/e2e/settings.spec.ts
seaerchin: Hmm, what properties does the regex ensure? I think we should write invalid test cases given the invariants and not the other way round

dcshzj: The regex mainly ensures that the user provides a proper domain name, so it will prevent cases like specifying a subdirectory. I think maybe just having at least this invalid input should be sufficient, so that we don't give users the idea that it is possible to host an Isomer site within a subdirectory?

kishore03109 (approved):
lgtm

cypress/e2e/settings.spec.ts
seaerchin: fair enough, added extra test cases!

cypress/e2e/settings.spec.ts
```diff
@@ -124,6 +133,68 @@ describe("Settings page", () => {
...
+      // Assert
+      cy.contains("The web domain you have entered is not valid").should(
+        "exist"
+      )
+    })
+  })
+
+  it("should not allow URLs with subdomains for the SEO URL", () => {
+    // Arrange
+    const urlWithSubdomain = "www.open.gov.sg/sub"
```
dcshzj, line 172: Sorry, to clarify the terminology:
- Domain (or primary domain) is something like `open.gov.sg`
- Subdomain would be something like `space.open.gov.sg`
- Subdirectories would be this case, which is `open.gov.sg/directory`

I think we can rename this to be clear that it is a subdirectory rather than a subdomain (the proper terminology is url-path).

dcshzj (approved):
A small nit, otherwise LGTM.

cypress/e2e/settings.spec.ts
seaerchin: sure thing, thanks for catching!

PR #1036, opened 2022-08-29
fix(reorderingmodal): removed chakra.button from folderItem

seaerchin:
## Problem
Previously, drag and drop did not work on `develop`. this is because having the`chakra.button` interfered with the draggable as they both respond to `onClick` events. 

Related to #1033 

## Solution
1. remove extra stuff from `FolderItem` as it's only used in re-ordering modal now.

### DIFF
--- a/src/components/folders/FolderItem.jsx
+++ b/src/components/folders/FolderItem.jsx
@@ -6,7 +6,7 @@ import contentStyles from "styles/isomer-cms/pages/Content.module.scss"
 import { pageFileNameToTitle } from "utils"
 
 const FolderItem = ({ item }) => {
-  const { name } = item
+  const { name, type } = item
 
   return (
     <Box
@@ -16,7 +16,13 @@ const FolderItem = ({ item }) => {
     >
       <Box className={`${elementStyles.card} ${elementStyles.folderItem}`}>
         <div className={contentStyles.contentContainerFolderRow}>
-          <i className={`bx bxs-file-blank ${elementStyles.folderItemIcon}`} />
+          {type === "file" ? (
+            <i
+              className={`bx bxs-file-blank ${elementStyles.folderItemIcon}`}
+            />
+          ) : (
+            <i className={`bx bxs-folder ${elementStyles.folderItemIcon}`} />
+          )}
           <span className={`${elementStyles.folderItemText} mr-auto`}>
             {pageFileNameToTitle(name)}
           </span>
--- a/src/components/folders/FolderItem.jsx
+++ b/src/components/folders/FolderItem.jsx
@@ -1,40 +1,12 @@
-import {
-  chakra,
-  LinkOverlay,
-  LinkBox,
-  Divider,
-  HStack,
-  Icon,
-  Text,
-  Box,
-} from "@chakra-ui/react"
-import { ContextMenu } from "components/ContextMenu"
-import { useMemo } from "react"
-import {
-  BiChevronRight,
-  BiEditAlt,
-  BiFolder,
-  BiTrash,
-  BiWrench,
-} from "react-icons/bi"
-import { Link as RouterLink, useRouteMatch } from "react-router-dom"
+import { Box } from "@chakra-ui/react"
 
 import elementStyles from "styles/isomer-cms/Elements.module.scss"
 import contentStyles from "styles/isomer-cms/pages/Content.module.scss"
 
 import { pageFileNameToTitle } from "utils"
 
-const FolderItem = ({ item, isDisabled }) => {
-  const { url } = useRouteMatch()
-
-  const { name, type, children } = item
-  const encodedName = encodeURIComponent(name)
-
-  const generatedLink = useMemo(() => {
-    return type === "dir"
-      ? `${url}/subfolders/${encodedName}`
-      : `${url}/editPage/${encodedName}`
-  }, [encodedName, type, url])
+const FolderItem = ({ item }) => {
+  const { name } = item
 
   return (
     <Box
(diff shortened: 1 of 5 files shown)


src/components/folders/FolderItem.jsx
```diff
@@ -1,125 +1,27 @@
...
-                Delete
-              </ContextMenu.Item>
-            </>
-          </ContextMenu.List>
-        </ContextMenu>
-      )}
+      <Box className={`${elementStyles.card} ${elementStyles.folderItem}`}>
+        <div className={contentStyles.contentContainerFolderRow}>
+          <i className={`bx bxs-file-blank ${elementStyles.folderItemIcon}`} />
+          <span className={`${elementStyles.folderItemText} mr-auto`}>
```
alexanderleegs, line 20: The icon specified causes all items in the reordering modal to show up as pages, but we should have a separate icon for folders (`bx bxs-folder ${elementStyles.folderItemIcon}`)

src/components/folders/ReorderingModal.jsx
```diff
@@ -17,15 +17,7 @@ import { deslugifyDirectory } from "utils"
...
   const { collectionName, subCollectionName } = params
-  const [dirOrder, setDirOrder] = useState([])
-
-  /** ******************************** */
-  /*     useEffects to load data     */
-  /** ******************************** */
-
-  useEffect(() => {
-    if (dirData) setDirOrder(dirData)
-  }, [dirData])
```
alexanderleegs, line 28: Just checking - why this this useEffect removed? Does the ReorderingModal rerender without the useEffect if dirData changes?

alexanderleegs (changes requested):
Does this PR also handle the react query cache invalidation? If so, the `useCreatePageHook` and `useUpdatePageHook` both also need to invalidate the resource specific query key.

Also as a minor nit: we've actually modified the icons on the folders pages to use the light version in one of our previous PRs (I think because the solid one doesn't exist on the `react-icons`?), but the reorder modal is still using the solid ones - I don't think this is an important fix and doesn't need to be tackled now, but we should refactor that component in future to make it consistent

src/components/folders/FolderItem.jsx
seaerchin: oof i derped - only saw pages so i thought we only reorder pages... good catch!

src/components/folders/ReorderingModal.jsx
seaerchin: hmm this has subtly different behaviour actually but i didn't see a noticeable difference. i'll change it back just to be safe ._.

seaerchin:
> Does this PR also handle the react query cache invalidation? If so, the `useCreatePageHook` and `useUpdatePageHook` both also need to invalidate the resource specific query key.
> 
> Also as a minor nit: we've actually modified the icons on the folders pages to use the light version in one of our previous PRs (I think because the solid one doesn't exist on the `react-icons`?), but the reorder modal is still using the solid ones - I don't think this is an important fix and doesn't need to be tackled now, but we should refactor that component in future to make it consistent

i think i updated the `useCreatePage` hook in another PR (#1035)! for `useUpdatePage`, it appears to work fine already as there's an `if` condition to invalidate if `params.resourceRoomName` exists (which is the part causing issues)

src/components/folders/FolderItem.jsx
```diff
@@ -1,125 +1,27 @@
...
 
 import { pageFileNameToTitle } from "utils"
 
-const FolderItem = ({ item, isDisabled }) => {
-  const { url } = useRouteMatch()
-
-  const { name, type, children } = item
-  const encodedName = encodeURIComponent(name)
-
-  const generatedLink = useMemo(() => {
```
kishore03109, line 33: Just to clarify here - are we aware why there was a `useMemo` here is the first place?

seaerchin: the `useMemo` is there to avoid recomputation of the `generatedLink` when stuff doesn't change. however, this component is only used in 1 place now (`ReorderingModal`) and will never be used as a `link` so this was removed

src/components/folders/FolderItem.jsx
```diff
@@ -1,125 +1,33 @@
...
-    return type === "dir"
-      ? `${url}/subfolders/${encodedName}`
-      : `${url}/editPage/${encodedName}`
-  }, [encodedName, type, url])
+const FolderItem = ({ item }) => {
```
kishore03109, line 8: If we are removing `isDisabled` as an argument for this function, I believe we should also remove the corresponding parameter `isDisabled` [here](https://github.com/isomerpages/isomercms-frontend/blob/fix/reorering/src/components/folders/ReorderingModal.jsx#L115)

kishore03109 (changes requested):

alexanderleegs:
> 

Ah I missed the create fix, thanks for pointing it out!

The update page one doesn't work correctly currently though - we currently use
`queryClient.invalidateQueries([
            DIR_CONTENT_KEY,
            { siteName: params.siteName, isUnlinked: true },
          ])`
instead of `RESOURCE_CATEGORY_CONTENT_KEY`, so it doesn't get updated when modifying resource pages

seaerchin:
> > 
> 
> Ah I missed the create fix, thanks for pointing it out!
> 
> The update page one doesn't work correctly currently though, I think it might be because we're using `queryClient.invalidateQueries([RESOURCE_CATEGORY_CONTENT_KEY, params])` instead of `queryClient.invalidateQueries([RESOURCE_CATEGORY_CONTENT_KEY, _.omit(params, "fileName")])`

good catch!!! didn't notice this cos i thought `params` is sufficient; fixed!

src/components/folders/FolderItem.jsx
seaerchin: good catch! ts would've caught this... sadly this file is in jsx so it's hard to tell what's omitted @__@

kishore03109 (approved):
lgtm

alexanderleegs (approved):

PR #1045, opened 2022-09-01
fix: allow trailing slash for site URL but remove on save

dcshzj:
## Problem

The site URL field in the site settings page does not allow for trailing slashes.

## Solution

**Breaking Changes**

**Bug Fixes**:

- The domain name regex has been updated to allow for a single trailing slash in the site URL input field.
- Any trailing slash will be removed before sending the domain name to the backend.

## Tests

**Unit tests**:
- Run `npm run tests`

**Smoke tests**:
- Run `npm run dev`.
- Use any site and navigate to the site settings page.
- Add a valid domain to the SEO field (site URL) and add a single trailing slash.
- Click anywhere outside the field, no error message will be shown.
- Click on "Save", the updated settings can be saved and the trailing slash will disappear.

## Deploy Notes

*None*

### DIFF
--- a/src/hooks/settingsHooks/useUpdateSettings.ts
+++ b/src/hooks/settingsHooks/useUpdateSettings.ts
@@ -49,10 +49,12 @@ const convertfromFe = ({
 
   // Remove trailing slash in site URL if present
   if (
-    "url" in renamedBasicSettings &&
-    (<string>renamedBasicSettings.url).endsWith("/")
+    !!renamedBasicSettings.url &&
+    // This is a safe cast as the `url` key gets mapped into a string
+    // if it is present and the property is also a string
+    (renamedBasicSettings.url as string).endsWith("/")
   ) {
-    renamedBasicSettings.url = (<string>renamedBasicSettings.url).slice(0, -1)
+    renamedBasicSettings.url = (renamedBasicSettings.url as string).slice(0, -1)
   }
 
   return {
--- a/src/hooks/settingsHooks/useUpdateSettings.ts
+++ b/src/hooks/settingsHooks/useUpdateSettings.ts
@@ -47,6 +47,14 @@ const convertfromFe = ({
     (_value, key) => FE_TO_BE[key]
   )
 
+  // Remove trailing slash in site URL if present
+  if (
+    "url" in renamedBasicSettings &&
+    (<string>renamedBasicSettings.url).endsWith("/")
+  ) {
+    renamedBasicSettings.url = (<string>renamedBasicSettings.url).slice(0, -1)
+  }
+
   return {
     ...renamedBasicSettings,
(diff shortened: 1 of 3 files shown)


src/hooks/settingsHooks/useUpdateSettings.ts
```diff
@@ -47,6 +47,14 @@ const convertfromFe = ({
     (_value, key) => FE_TO_BE[key]
   )
 
+  // Remove trailing slash in site URL if present
+  if (
+    "url" in renamedBasicSettings &&
+    (<string>renamedBasicSettings.url).endsWith("/")
+  ) {
+    renamedBasicSettings.url = (<string>renamedBasicSettings.url).slice(0, -1)
+  }
```
seaerchin, line 56: ts casting uses the `as` keyword; we should also add comments to prove that the type casting is safe as it might not be clear at a glance

```suggestion
  if (
    !!renamedBasicSettings.url &&
    // This is a safe cast as the `url` key gets mapped
    // into a string if it is present and the property
    // is also a string
    (renamedBasicSettings.url as string).endsWith("/")
  ) {
    renamedBasicSettings.url = (renamedBasicSettings.url as string).slice(0, -1)
  }```

File: src/utils/validators.js [hunk 2]
```diff
@@ -27,7 +27,7 @@ export const TIKTOK_REGEX = ".com/@)([a-zA-Z0-9_-]+([/.])?)+$"
 // 3. The last section after "+" is to match the TLD segment of the domain
 // 4. The "^" and "$" ensures that we are matching the whole string (to prevent the user from adding "https://")
 export const DOMAIN_NAME_REGEX =
-  "^([a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?\\.)+[a-z0-9][a-z0-9-]{0,61}[a-z0-9]$"
+  "^([a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?\\.)+[a-z0-9][a-z0-9-]{0,61}[a-z0-9]/?$"
```
seaerchin, line 30: regex101 seems to complain about the `/?`, saying that 
> / An unescaped delimiter must be escaped; in most languages with a backslash (\)
> ? The preceding token is not quantifiable

i think you might be missing a `\` as follows: `\/?`

src/hooks/settingsHooks/useUpdateSettings.ts
dcshzj: Fixed! Thanks for pointing this out!

src/utils/validators.js
dcshzj: I realised it was being automatically removed by prettier, so I had to add a double escape character. Thanks for the catch!

seaerchin (approved):
lgtm

PR #1047, opened 2022-09-06
fix: improve copy of SEO settings field per design

dcshzj:
## Problem

The copy used for the SEO settings field can be confusing for users.

## Solution

**Breaking Changes**

**Improvements**:

- The copy used for the SEO settings field has been improved as provided by design.
- The error message when the field is empty has also been improved.

## Tests

**Smoke tests**:

- Run `npm run dev`.
- Use any site and navigate to the site settings page.
- The updated description will be shown.
- Add any text to the SEO field (site URL) and then remove it.
- Click anywhere outside the field, the updated error message will be shown.

## Deploy Notes

*None*

### DIFF
--- a/src/layouts/Settings/GeneralSettings.tsx
+++ b/src/layouts/Settings/GeneralSettings.tsx
@@ -72,7 +72,7 @@ export const GeneralSettings = ({
           <Box mb="0.75rem">
             <FormLabel mb={0}>Search Engine Optimisation (SEO)</FormLabel>
             <FormLabel.Description color="text.description">
-              Enter the web domain of your live site, to improve its position on
+              Enter the web domain of your site, to improve its position on
               search result pages.
             </FormLabel.Description>
           </Box>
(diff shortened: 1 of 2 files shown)


src/layouts/Settings/GeneralSettings.tsx
```diff
@@ -89,13 +89,16 @@ export const GeneralSettings = ({
             />
           </InputGroup>
           <FormErrorMessage>
-            The web domain you have entered is not valid.
+            {errors.url?.type === "required" &&
+              "This field cannot be left blank"}
+            {errors.url?.type === "pattern" &&
+              "The web domain you have entered is not valid"}
```
seaerchin, line 95: i think we'd have to specify the type on `useFormState` like so: 

```ts
const { errors } = useFormState<SiteInfo>()
```

this allows the type of errors to be specified - if we hover over it rgiht now, it's typed as `any`

seaerchin (commented):
minor comment

seaerchin (approved):
lgtm

PR #1048, opened 2022-09-08
Feat/add external permalink to resource

### LINKED ISSUE #1046: [Feat] allow external link for resource pages
We should allow the use of resource pages to external links, as this is something frequently requested by users.
image

alexanderleegs:
ON HOLD: Updating resource pages to allow for external permalinks as a third type of resource

## Problem

This PR introduces the ability to use external permalinks for resource pages, to allow users to use resource cards to link to external pages. Resolves #1046.

This PR also separates the validation for resource page permalinks and other pages permalinks, as external permalinks for unlinked/collection pages do not work in the same way. While resource pages can be used to link out to external pages, the other types of pages will continue to interpret it as an internal link - we should therefore be using a separate set of validation specifically for resource page permalinks. 

TODO:
Nat raised the issue that being able to have both internal and external permalinks is confusing - we're looking into making this a third type of resource page instead

### DIFF
--- a/src/components/PageSettingsModal/PageSettingsModal.jsx
+++ b/src/components/PageSettingsModal/PageSettingsModal.jsx
@@ -147,6 +147,13 @@ export const PageSettingsModal = ({
                   >
                     <Box mb="0.75rem">
                       <FormLabel mb={0}>Page URL</FormLabel>
+                      {resourceRoomName && (
+                        <FormLabel.Description color="text.description">
+                          Enter what you want to link to (e.g. /page-url for
+                          pages on your site, or https://www.url.com for
+                          external websites).
+                        </FormLabel.Description>
+                      )}
                       <FormLabel.Description color="text.description">
                         {siteUrl}
                       </FormLabel.Description>
--- a/src/components/PageSettingsModal/PageSettingsSchema.jsx
+++ b/src/components/PageSettingsModal/PageSettingsSchema.jsx
@@ -4,6 +4,7 @@ import * as Yup from "yup"
 import {
   permalinkRegexTest,
   specialCharactersRegexTest,
+  resourcePermalinkRegexText,
   PAGE_SETTINGS_PERMALINK_MIN_LENGTH,
   PAGE_SETTINGS_PERMALINK_MAX_LENGTH,
   PAGE_SETTINGS_TITLE_MIN_LENGTH,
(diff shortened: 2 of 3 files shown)


src/components/PageSettingsModal/PageSettingsModal.jsx
```diff
@@ -147,6 +147,13 @@ export const PageSettingsModal = ({
                   >
                     <Box mb="0.75rem">
                       <FormLabel mb={0}>Page URL</FormLabel>
```
dcshzj, line 149: When a user inputs an external link and clicks on "Save", it brings the user to the edit page which may be confusing (plus the content will not be shown to the site visitor). Should we disable the edit page if an external link is put into the page URL?

src/components/PageSettingsModal/PageSettingsSchema.jsx
```diff
@@ -32,9 +33,27 @@ export const PageSettingsSchema = (existingTitlesArray = []) =>
...
+          return schema
+            .required("Permalink is required")
+            .min(
+              PAGE_SETTINGS_PERMALINK_MIN_LENGTH,
+              `Permalink must be longer than ${PAGE_SETTINGS_PERMALINK_MIN_LENGTH} characters`
+            )
+            .max(
+              PAGE_SETTINGS_PERMALINK_MAX_LENGTH,
+              `Permalink must be shorter than ${PAGE_SETTINGS_PERMALINK_MAX_LENGTH} characters`
+            )
```
dcshzj, line 50: I think this might be quite a constraint for external URLs. For instance, one of the MND's press releases would not be able to be linked because it is 224 characters:

https://www.mnd.gov.sg/newsroom/press-releases/view/15th-meeting-of-the-malaysia-singapore-joint-ministerial-committee-for-iskandar-malaysia-focuses-on-enhancing-investment-and-innovation-ecosystem-to-drive-economic-recovery

Should we lift this limit for external links?

dcshzj (changes requested):

src/components/PageSettingsModal/PageSettingsModal.jsx
```diff
@@ -147,6 +147,13 @@ export const PageSettingsModal = ({
...
+                      {resourceRoomName && (
+                        <FormLabel.Description color="text.description">
+                          Enter what you want to link to (e.g. /page-url for
+                          pages on your site, or https://www.url.com for
+                          external websites).
+                        </FormLabel.Description>
+                      )}
                       <FormLabel.Description color="text.description">
                         {siteUrl}
                       </FormLabel.Description>
```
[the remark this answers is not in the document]
alexanderleegs: Had a chat with Nat, we'll be modifying this to use {siteUrl}{permalink} and separating out the external links into a different type of resource page probably

src/components/PageSettingsModal/PageSettingsModal.jsx
alexanderleegs: Thinking of changing the type of these to something else instead! Which should fix the issue

alexanderleegs:
Superceded by https://github.com/isomerpages/isomercms-frontend/pull/1064

PR #1055, opened 2022-09-13
feat: allow e2e tests to be triggered from a PR comment

### LINKED ISSUE #1037: [TESTS] Implement comment-based triggering of e2e tests
As resolved in the engineering sync on 29 August 2022, the e2e tests should also be run for each pull request. This can be done through a comment on the PR to trigger the e2e tests to run.

Implement a mechanism to detect a specific comment for triggering e2e tests. Additionally:
1. Ensure that only one instance of the e2e tests is running
2. Update the pull request template to prompt the PR creator to run the e2e tests (along with instructions on how to do so)

Note: e2e tests should not be running on every push, as it takes a long time and can be very costly with little added benefit.

dcshzj:
## Problem

Our e2e tests are painful to run and we choose not to run it on our local machines. This can result in potential bugs that could have been caught in our e2e tests to be missed, and to only surface during release day. This is dangerous and can result in significant loss in developer productivity to identify the root cause of e2e test failures.

Fixes #1037.

## Solution

**Breaking Changes**

**Features**:

- A new workflow is introduced to allow e2e tests to be triggered via a special comment. This workflow will ensure that:
    - The user is authorised to perform the command (being part of [isomerpages/core team](https://github.com/orgs/isomerpages/teams/core))
    - There are no existing e2e tests being run before triggering one
- The existing e2e test workflow can now be triggered directly through a `workflow_dispatch` event
- The existing e2e test workflow retains its current capabilities of being run on a PR to the master branch, and that has been configured to be of higher priority, where it is able to cancel existing workflows in order to run the e2e test for that particular PR (when performing a release)
- The pull request template has been updated to prompt the PR creator to run the e2e tests

<details>
<summary><b>Sequence diagram</b></summary>
<br>

image

</details>

## Before & After Screenshots

Screenshot 2022-09-13 at 16 57 45

## Tests

The changes cannot be tested directly on this repository. Instead, you can use my [test repository zhongjun-test](https://github.com/isomerpages/zhongjun-test) to try out this feature. The e2e test workflow is simulated to be a workflow that takes 5 minutes to run.

An example PR with the different scenarios can be observed here: https://github.com/isomerpages/zhongjun-test/pull/2.

### DIFF
--- a/.github/workflows/ci-e2e.yml
+++ b/.github/workflows/ci-e2e.yml
@@ -3,6 +3,12 @@ name: ci-e2e
 on:
   pull_request:
     branches: [master]
+  workflow_dispatch:
+
+# Required for allowing PRs to master branch to be of highest priority
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
 
 jobs:
   install:
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -42,6 +42,10 @@ Closes [insert issue #]
 
 <!-- What tests should be run to confirm functionality? -->
 
+- [ ] Unit tests (using `npm run tests`)
+- [ ] e2e tests (comment on this PR with the text `!run e2e`)
+- [ ] Smoke tests
+
 ## Deploy Notes
 
 <!-- Notes regarding deployment of the contained body of work.  -->
--- a/.github/workflows/trigger-e2e.yml
+++ b/.github/workflows/trigger-e2e.yml
@@ -1,0 +1,163 @@
+name: Trigger e2e tests
+
+on:
+  issue_comment:
+    types: [created, edited]
+
+  workflow_dispatch:
+    inputs:
+      issueNumber:
+        description: "The pull request number to build e2e tests on"
+        required: true
+        default: 1
+        type: number
+      attemptNumber:
+        description: "Attempt number"
+        required: false
+        default: 1
+        type: number
+
+env:
+  # The full comment text to match to trigger this workflow
+  ISOMER_TRIGGER_COMMENT: "!run e2e"
+  # The file name of the e2e test workflow
+  ISOMER_E2E_WORKFLOW_NAME: ci-e2e.yml
+  # The slug for the Isomer core team
+  ISOMER_CORE_TEAM_SLUG: core
+  # How long to wait before queuing again (roughly half the time taken for e2e tests to run)
+  ISOMER_SLEEP_SECONDS: 900 # 15 min
+  # How many times to retry before failing
+  ISOMER_MAX_RETRIES: 5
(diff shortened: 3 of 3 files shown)


.github/workflows/ci-e2e.yml
```diff
@@ -3,6 +3,12 @@ name: ci-e2e
 on:
   pull_request:
     branches: [master]
+  workflow_dispatch:
+
+# Required for allowing PRs to master branch to be of highest priority
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
```
alexanderleegs, line 11: Very minor nit: could the indentation be adjusted for easier reading here?

alexanderleegs: Also just for clarification, how do we indicate that this is the same workflow as `trigger-e2e`?

alexanderleegs (commented):
This looks super useful, thanks for working on this! Left a minor nit/clarification

.github/workflows/ci-e2e.yml
dcshzj: Hmm I think the indentation cannot be adjusted further as the `group` and `cancel-in-progress` keys are siblings. This is similar to the [example from the docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-concurrency-to-cancel-any-in-progress-job-or-run).

Also, this ci-e2e workflow will run differently from the trigger-e2e workflow. To summarise the points:
1. Multiple trigger-e2e workflows can run at the same time, because it functions somewhat like a queue system where they will regularly poll to see if there are any pending or in progress runs of the ci-e2e workflow. It can happen if multiple people were to trigger the e2e tests from different pull requests.
2. If a ci-e2e workflow is currently running, the few trigger-e2e workflows that are running will keep waiting and polling until they find a window where there are no ci-e2e workflow running.
3. If a ci-e2e workflow is currently running and we were to perform a release (defined as creating a PR to the master branch), then any existing ci-e2e runs will be terminated to make way for the e2e tests to be performed on the release.

There is a very edge case where 2 trigger-e2e workflows manage to trigger a ci-e2e workflow run. In that case, GitHub Actions will take the latest one and the earlier one will be terminated. If that happens, the user would not be notified, but the terminated e2e test can run again by sending the command to trigger e2e tests again.

alexanderleegs (approved):
Just wondering, is there a reason why we don't replace the original GITHUB_TOKEN?

Other than that, lgtm!

dcshzj:
> Just wondering, is there a reason why we don't replace the original GITHUB_TOKEN?

I kept it for security reasons, so only steps that require a little more permissions will have access to the more powerful personal access token.

dcshzj:
Made some additional changes to make sure that unrelated comments are ignored.

PR #1064, opened 2022-09-16
Refactor/page settings modal

alexanderleegs:
## Problem

This PR refactors the PageSettingsModal and splits the ResourcePageSettingsModal into a separate component, as the content of the modals has diverged sufficiently. Screens can be found [here](https://www.figma.com/file/7OtNYkc4s9BkFk05ConEbd/%E2%AD%90-V1.5?node-id=131%3A3620), though not fully implemented (calendar field requires chakra 2.0). To be reviewed in conjunction with PR #[500](https://github.com/isomerpages/isomercms-backend/pull/500) on the isomercms-backend repo and PR #[276](https://github.com/isomerpages/isomerpages-template/pull/276) on the isomer-template repo.

This PR also adds in an additional type of resource page, to be used for external links. Resolves https://github.com/isomerpages/isomercms-frontend/issues/1046.


src/components/FormFieldMedia/FormFieldMedia.jsx
```diff
@@ -14,7 +14,14 @@ const FormFieldMedia = ({
   placeholder = "",
   type = "images",
   inlineButtonText = "Choose Item",
-  register = () => {},
+  register = (_args) => {},
+}: {
+  value?: string,
```
kishore03109, line 19: +1 for adding types! This file is meant to be a .tsx file rather than a .jsx right?

src/components/PageSettingsModal/ResourcePageSettingsModal.tsx
```diff
@@ -0,0 +1,335 @@
...
+  resourceRoomName: string
+  resourceCategoryName: string
+}
+
+interface ResourcePageFrontMatter {
+  layout: string
+  title: string
+  permalink: string
+  date: string
+  // eslint-disable-next-line camelcase
```
kishore03109, line 56: Hmm is it possible for us to use camelCase here? (unless its the cypress issue uh, then feel free to dismiss this comment)

src/components/PageSettingsModal/ResourcePageSettingsModal.tsx
```diff
@@ -185,6 +177,13 @@ export const ResourcePageSettingsModal = ({
         })
         setValue("permalink", "")
       }
+      if (pageData.content.frontMatter.layout === "external") {
+        // remove https:// from resource pages with external permalinks
+        setValue(
+          "permalink",
+          pageData.content.frontMatter.permalink.replace("https://", "")
```
kishore03109, line 179: Just to clarify here - why do we need to remove the `https://` ah?

src/components/PageSettingsModal/ResourcePageSettingsModal.tsx
```diff
@@ -206,7 +212,9 @@ export const ResourcePageSettingsModal = ({
       <ModalContent>
         <ModalCloseButton id="settings-CLOSE" />
         <ModalHeader>
-          <Text as="h1">{!fileName ? "Create new page" : "Page settings"}</Text>
+          <Text as="h1">
+            {!fileName ? "Create new resource page" : "Resource page settings"}
```
kishore03109, line 210: supernit - can we change this to
`{fileName ? "Resource page settings" : "Create new resource page"}` for clarity?

src/layouts/ResourceCategory/components/ResourceCard.tsx
```diff
@@ -30,6 +31,17 @@ interface ResourceCardProps {
   resourceType: ResourcePageData["resourceType"]
 }
 
+const getIcon = (resourceType: string): JSX.Element => {
+  switch (resourceType) {
+    case "post":
+      return <Icon as={BiFileBlank} fontSize="1.5rem" fill="icon.alt" />
+    case "link":
+      return <Icon as={BiLink} fontSize="1.5rem" fill="icon.alt" />
+    default:
```
kishore03109, line 40: nit: could we change `default:` to `case "file":` for clarity?

src/components/PageSettingsModal/PageSettingsSchema.jsx
```diff
@@ -56,12 +57,13 @@ export const PageSettingsSchema = (existingTitlesArray = []) =>
       }
     }),
     layout: Yup.string().oneOf(["file", "post", "link"]),
-    date: Yup.date()
+    date: Yup.string()
```
kishore03109, line 60: Hmm am confused here... wouldn't using the `Yup.date()` be a better solution than using a string and then trying to manually match the regex pattern ourself?

kishore03109 (changes requested):
Screenshot 2022-09-19 at 2 31 46 PM
Hey, there is a minor ui bug! Steps to replicate: 
1) create an external link
2) click on the external link's page settings
3) modal shown has the wrong error: `Permalink should start with a slash, and contain alphanumeric characters separated by hyphens and slashes only`

src/components/FormFieldMedia/FormFieldMedia.jsx
alexanderleegs: The types were defined here to specify the types for the typescript file which imports this component - we'd ideally like to migrate all of these files to typescript, but I think there's a whole dependency chain of files which would need to be updated here - I've opted to leave it out for now

src/components/PageSettingsModal/ResourcePageSettingsModal.tsx
alexanderleegs: This is actually from the actual page front matter!

src/components/PageSettingsModal/ResourcePageSettingsModal.tsx
alexanderleegs: This is for the display of the field - we're autopopulating the `https://` for users, so it's stored as `https://sitename.com`. When we retrieve the permalink, we should also only display `sitename.com`, hence the removal

src/components/PageSettingsModal/PageSettingsSchema.jsx
alexanderleegs: Hmm we're expecting a specific date string here(`yyyy-mm-dd`) to be saved in the frontmatter of the file - using `Yup.date()` specifically specifies it as a Date object, which we need to transform back into a string for the purposes of validation and to save - I opted to change it back to string because of this!

src/layouts/ResourceCategory/components/ResourceCard.tsx
alexanderleegs: Changed in https://github.com/isomerpages/isomercms-frontend/pull/1064/commits/36a32368a50a92cf4a0da92b1b2f795d49f0a0aa!

src/components/PageSettingsModal/ResourcePageSettingsModal.tsx
alexanderleegs: Changed in https://github.com/isomerpages/isomercms-frontend/commit/36a32368a50a92cf4a0da92b1b2f795d49f0a0aa!

alexanderleegs:
> Screenshot 2022-09-19 at 2 31 46 PM Hey, there is a minor ui bug! Steps to replicate:
> 
> 1. create an external link
> 2. click on the external link's page settings
> 3. modal shown has the wrong error: `Permalink should start with a slash, and contain alphanumeric characters separated by hyphens and slashes only`

Whoops good catch, this was caused by not retriggering the permalink validation when changing fields! Fixed in https://github.com/isomerpages/isomercms-frontend/pull/1064/commits/18c62e18c387a0b498ec0a66727009ce3fb059c2

kishore03109 (approved):

kishore03109:
@alexanderleegs Locally, my TS Server complains that: 
`For the file Type annotations can only be used in TypeScript files` for `FormFieldMedia.jsx`. Locally to test this branch, I also had to manually restart my frontend server. 
I understand that there will be cascading stuff to fix if this was converted into a `.tsx` which is beyond the scope of the PR + this fix is critical, but I am concerned if this bug will cause issues when deploying to prod. Is there a way we can test if prod will still work, essentially ignoring this compilation error?

alexanderleegs:
> @alexanderleegs Locally, my TS Server complains that: `For the file Type annotations can only be used in TypeScript files` for `FormFieldMedia.jsx`. Locally to test this branch, I also had to manually restart my frontend server. I understand that there will be cascading stuff to fix if this was converted into a `.tsx` which is beyond the scope of the PR + this fix is critical, but I am concerned if this bug will cause issues when deploying to prod. Is there a way we can test if prod will still work, essentially ignoring this compilation error?

As discussed offline, we pushed this to staging and there doesn't appear to be any issues - merging should be fine!

PR #1066, opened 2022-09-19
fix(file not uploaded when having 2 dots): Improve user experience by throwing an error

### LINKED ISSUE #1021: Inaccurate error message when uploading PDF file
Description

When uploading PDF, file name cannot have two full stops. **_The issue will not be resolved by trying again or checking internet connection._** 

How to resolve:

Remove extra full stops in name of file.

image

kishore03109:
## Problem

Currently, when uploading files with dots, IsomerCMS fails with an ambigous message stating "A new media file could not be created. Please try again or check your internet connection" This is a bandage on the front end rather than a permanent fix

Closes #1021. 

## Solution

Throw an error when the user attempts to upload a file with dots within the file name, excluding the extension. 

**Breaking Changes**

- [] Yes - this PR contains breaking changes

### DIFF
--- a/src/components/MediaSettingsModal/MediaSettingsSchema.jsx
+++ b/src/components/MediaSettingsModal/MediaSettingsSchema.jsx
@@ -20,7 +20,7 @@ export const MediaSettingsSchema = (existingTitlesArray = []) =>
       )
       .test(
         "File not supported",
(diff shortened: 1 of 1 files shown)


seaerchin (approved):
lgtm but could i trouble you to file an issue on BE for this so we can come up with a more permanent fix in the future

PR #1068, opened 2022-09-19
fix(edit page): avoid using textContent to display XSS modal

dcshzj:
## Problem

The Isomer CMS will throw an error if the `textContent` value is an empty string, which is possible in the following few examples:
1. `<script></script>` - empty content for a whole tag that will be removed
2. `<p evil>Text</p>` - empty attribute content for the `evil` attribute that will be removed

## Solution

**Breaking Changes**

**Bug Fixes**:

- The use of `textContent` has been replaced with either `outerHTML` for removed elements, or `nodeName` for removed attributes.

## Before & After Screenshots

**AFTER**:

Screenshot 2022-09-19 at 18 05 35

## Tests

- [x] Smoke tests

1. Go to any site and edit a page
2. Try to save content matching any of the above cases
3. Verify that the XSS warning modal appears (instead of throwing an error)

## Deploy Notes

*None*

### DIFF
--- a/src/layouts/EditPage.jsx
+++ b/src/layouts/EditPage.jsx
@@ -162,12 +162,9 @@ const EditPage = ({ match }) => {
                   <br />
                   <Code>{i + 1}</Code>:
                   <Code>
-                    {elem.attribute?.textContent || elem.element?.textContent
-                      ? (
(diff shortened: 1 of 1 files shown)


dcshzj:
!run e2e

PR #1069, opened 2022-09-20
fix(gh-actions): handle failures from bots and refine PR comments

dcshzj:
## Problem

The e2e tests GitHub Actions workflow does not handle `workflow_dispatch` events well, causing the PR comments to be given to incorrect pull requests. Also, the step to check for user's team membership does not work well for bot accounts.

## Solution

**Breaking Changes**

**Bug Fixes**:

- The original e2e test workflow has been updated to allow specifying the pull request number. This number will be used to determine the name of the original branch, the SHA of the HEAD commit for that branch and the title of the pull request. This fix was performed on [advice from the Cypress documentation](https://docs.cypress.io/guides/continuous-integration/github-actions#Pull-requests-commit-message-is-merge-SHA-into-SHA).
- The step to check the user's team membership has been updated to continue on errors, so that the workflow does not fail unexpectedly.

## Tests

- [ ] Smoke tests

No tests were prepared as they are refinements to an earlier PR #1055.

### DIFF
--- a/.github/workflows/ci-e2e.yml
+++ b/.github/workflows/ci-e2e.yml
@@ -90,8 +90,6 @@ jobs:
       USERNAME: ${{secrets.USERNAME}}
       PERSONAL_ACCESS_TOKEN: ${{secrets.PERSONAL_ACCESS_TOKEN}}
       E2E_COMMIT_HASH: ${{secrets.E2E_COMMIT_HASH}}
-      COMMIT_INFO_MESSAGE: ${{github.event.pull_request.title}}
-      COMMIT_INFO_SHA: ${{github.event.pull_request.head.sha}}
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci-e2e.yml
+++ b/.github/workflows/ci-e2e.yml
@@ -4,6 +4,12 @@ on:
   pull_request:
     branches: [master]
   workflow_dispatch:
+    inputs:
+      issueNumber:
+        description: "The pull request number to build e2e tests on"
+        required: true
+        default: 1
+        type: number
 
 # Required for allowing PRs to master branch to be of highest priority
 concurrency:
@@ -84,9 +90,7 @@ jobs:
       USERNAME: ${{secrets.USERNAME}}
       PERSONAL_ACCESS_TOKEN: ${{secrets.PERSONAL_ACCESS_TOKEN}}
       E2E_COMMIT_HASH: ${{secrets.E2E_COMMIT_HASH}}
-      # overwrite commit message sent to Dashboard
       COMMIT_INFO_MESSAGE: ${{github.event.pull_request.title}}
-      # re-enable PR comment bot
       COMMIT_INFO_SHA: ${{github.event.pull_request.head.sha}}
 
     steps:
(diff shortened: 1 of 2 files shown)


seaerchin:
!run e2e

seaerchin:
would it be possible to also have a comment (either by the cypress bot or otherwise) on the test results? it would make it easier for us to see the results of our e2e, whereas now it's obscured

dcshzj:
> would it be possible to also have a comment (either by the cypress bot or otherwise) on the test results? it would make it easier for us to see the results of our e2e, whereas now it's obscured

Yup, this PR aims to fix that by passing the correct commit SHA for workflow_dispatch events to the cypress dashboard.

.github/workflows/trigger-e2e.yml
```diff
@@ -44,6 +44,7 @@ jobs:
       - name: Check if user is part of Isomer core team
         uses: tspascoal/get-user-teams-membership@v1
         id: checkUserMember
+        continue-on-error: true
```
kishore03109, line 47: Hey! Am a bit hesitant to allow continue for **all** errors, is there a way to state continue on error for only `workflow_dispatch` errors?

kishore03109 (changes requested):
>The original e2e test workflow has been updated to allow specifying the pull request number. This number will be used to determine the name of the original branch, the SHA of the HEAD commit for that branch and the title of the pull request. This fix was performed on [advice from the Cypress documentation](https://docs.cypress.io/guides/continuous-integration/github-actions#Pull-requests-commit-message-is-merge-SHA-into-SHA).

Hmm, does github actions not allow you to comment based on the PR number instead? Am asking about this since we do tend to rebase often which changes the SHA!

.github/workflows/trigger-e2e.yml
dcshzj: If there is an error, the workflow will continue processing but the `ISOMER_RUN_E2E` variable will not be set to `true`, thus causing the rest of the workflow to be skipped unless we know who triggered the workflow. Basically, the rest of the workflow will run only if the teams membership check is successful, and will not run if otherwise.

dcshzj:
> Hmm, does github actions not allow you to comment based on the PR number instead? Am asking about this since we do tend to rebase often which changes the SHA!

I will be passing in the pull request number, which will then be used to query the latest commit hash. If a rebase happens then the new hash will be used instead.

.github/workflows/ci-e2e.yml
```diff
@@ -84,9 +90,7 @@ jobs:
       USERNAME: ${{secrets.USERNAME}}
       PERSONAL_ACCESS_TOKEN: ${{secrets.PERSONAL_ACCESS_TOKEN}}
       E2E_COMMIT_HASH: ${{secrets.E2E_COMMIT_HASH}}
-      # overwrite commit message sent to Dashboard
       COMMIT_INFO_MESSAGE: ${{github.event.pull_request.title}}
-      # re-enable PR comment bot
       COMMIT_INFO_SHA: ${{github.event.pull_request.head.sha}}
```
seaerchin, line 94: need remove

seaerchin (approved):

Issue #1070, opened 2022-09-22
[FEAT] Review request


seaerchin:
- [ ] add review request modal
- [ ] add screens for different roles and states 
- [ ] add diff viewer for github changes

PR #1071, opened 2022-09-22
feat(rr): resource room modals

seaerchin:
## Problem
This PR adds the initial 2 resource room modals. These can be seen [here](https://www.figma.com/file/7OtNYkc4s9BkFk05ConEbd/%E2%AD%90-V1.5?node-id=2687%3A72843). This PR also adds the `react-table` package to manage tabular data for us and the `react-virtuoso` package to handle virtualization for large diffs. (usually on first PR, so that the load time isn't very long).

**do note that this is purely the frontend, so there is no logic implemented (will come in later PRs)**

Part of #1070 

## Solution
1. make review request modal + components
2. use `react-virtuoso` to handle large file diffs. this was chosen over `react-window` as it doesn't handle dynamic sizing gracefully (as long folder names will take 2 lines - @NatMaeTan is agreeable w/ this). we could've used `react-virtual` but it was slightly verbose (might change it later if required)


src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -0,0 +1,335 @@
...
+const LastEditedMeta = ({
+  lastEditedBy,
+  lastEditedTime,
+}: LastEditedMetaProps): JSX.Element => {
+  const { date, time } = getDateTimeFromUnixTime(lastEditedTime)
+  return (
+    <HStack>
+      <Avatar
+        bg="primary.100"
+        name={lastEditedBy.slice(0, 2).split("").join(" ")}
```
dcshzj, line 163: I think this logic should be placed as some common function as I think it will also be used for the new navbar. Also to ensure that the way we get the initials is consistent throughout the CMS.

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -0,0 +1,335 @@
...
+  } = useSearchbar({})
+  const columnHelper = createColumnHelper<EditedItemProps>()
+
+  const columns = [
+    columnHelper.display({
+      id: "filter",
+      header: () => (
+        <Th w="3rem">
+          <IconButton
+            aria-label="sort by file type"
```
dcshzj, line 201: I think should change sort to filter here?

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -0,0 +1,335 @@
...
+        id: "lastEdited",
+        header: () => (
+          <Th textTransform="capitalize" w="12.5rem">
+            <HStack spacing="0.25rem">
+              <Text textStyle="subhead-2" textColor="text.label">
+                Last edited
+              </Text>
+              <IconButton
+                icon={<BiSort fontSize="1rem" />}
+                aria-label="view file changes"
```
dcshzj, line 234: I think this label does not match the intention of the button.

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -0,0 +1,335 @@
...
+              aria-label="view file on staging"
+              variant="link"
+            />
+          </Link>
+          <IconButton
+            icon={<BiShow />}
+            aria-label="view file changes"
+            variant="link"
+          />
+        </HStack>
```
dcshzj, line 265: Should we also add a tooltip for these two buttons?

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -0,0 +1,335 @@
...
+
+  return (
+    <VStack spacing="1rem" mt="1rem">
+      <Flex w="100%">
+        <Box>
+          <Text textStyle="subhead-1" textColor="text.body">
+            <b>{items.length}</b> item(s) for review
+          </Text>
+          <Text textStyle="caption-2" textColor="text.placeholder">
+            Items below have been identified to have changes
```
dcshzj, line 284: I think "Items" can be changed to "Item(s)" to match the title.

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -0,0 +1,335 @@
...
+          onSearchIconClick={handleExpansion}
+          isExpanded={isExpanded}
+          onSearch={() => {
+            console.log("hi")
+          }}
+          onBlur={handleCollapse}
+        />
+      </Flex>
+      <Box w="100%" borderWidth="1px" borderRadius="8px" borderColor="gray.100">
+        <TableVirtuoso
```
dcshzj, line 299: I'm not sure why the scrollbar goes all the way to the top. I think should stop before the header row.

Screenshot 2022-09-27 at 10 20 09

src/layouts/Dashboard/components/ReviewRequestModal/ReviewRequestForm/ReviewRequestForm.tsx
```diff
@@ -0,0 +1,46 @@
...
+          // NOTE: This is done to ensure that the width of the MultiSelect
+          // takes the full width of its container.
+          // Using a VStack that has full width constrains the width somehow
+          // and the MultiSelect itself does not expose a width prop.
+          mb="0.75rem"
+        >
+          Your changes can be published when at least **one** reviewer approves
+          your request
+        </FormLabel.Description>
+        <Select options={admins} isMulti />
```
dcshzj, line 42: Placeholder text should be updated.

src/mocks/constants.ts
```diff
@@ -186,3 +188,69 @@ export const MOCK_SUBFOLDER_NAME = "mock-subfolder"
...
+    lastEditedBy: "asdf",
+    lastEditedTime: 129823104823094,
+  },
+  {
+    type: ["nav"],
+    name: "some file",
+    path: ["some", "thing"],
+    url: "www.google.com",
+    lastEditedBy: "asdf",
+    lastEditedTime: 129823104823094,
```
dcshzj, line 207: Small nit: The timestamps should all be different so we can see if the sorting function works as intended.

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -0,0 +1,335 @@
...
+  BiChevronRight,
+} from "react-icons/bi"
+import { TableVirtuoso } from "react-virtuoso"
+
+import { BxFileArchiveSolid } from "assets"
+
+interface DisplayedDateTime {
+  date: string
+  time: string
+}
```
dcshzj, line 51: I think there are lots of types scattered around this file, would it be better if we moved them to their own file in the `types/` folder?

dcshzj (changes requested):

kishore03109 (commented):
> (usually on first PR, so that the load time isn't very long) 
Hey sorry - am abit confused about this statement, why would only the first PR have large diffs ah?

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
kishore03109: Opps a heads up to this - design no longer wants tooltips in isomer haha. I will be removing them soon!

NatMaeTan: Wait what when did I say that, I want tool tips!!!!

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.stories.tsx
```diff
@@ -0,0 +1,114 @@
+import {
+  Modal,
+  ModalOverlay,
+  ModalContent,
+  ModalHeader,
+  VStack,
+  ModalCloseButton,
```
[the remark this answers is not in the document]
seaerchin: good catch! the tragedy of using auto import AHHAHA

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: sounds good! do you think that this will be used outside of the dashboard? if so, i'll probably place this in the root `utils` folder otherwise, jus this folder

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: yep, good catch!

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: yep, will change!

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: can, good catch!

src/mocks/constants.ts
seaerchin: yep, this PR doesn't include sorting - the follow up PR does so this isn't an issue

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: can! what types do you foresee yourself using so that i can shift just those types?

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview.tsx
```diff
@@ -0,0 +1,263 @@
...
+        />
+      )
+    }
+    case "image": {
+      return <Icon as={BiImage} />
+    }
+    default: {
+      // NOTE: This is done to ensure exhaustive type matching
+      const error: never = iconType
+      throw new Error(`Unmatched fileType: ${error}`)
```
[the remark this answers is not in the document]
seaerchin: this isn't casting! what i'm doing here is that i'm saying that the **type of `error` is `never`**. this is similar to eg: 
```c
int x = 1;  
```

in c, where the `int` specifies the type of the variable x. whether `iconType` will be `null` is dependent upon the runtime - at compile time, ts' type system verifies the correctness of this type. at runtime, if there is `js` interop, then it really depends but because the API BE isn't specified yet so as of now, no. in the future, the API fetch will be in `ts` so barring implicit casts/conversions, there shouldn't be

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -24,6 +23,13 @@ import {
   IconButton,
   useSearchbar,
 } from "@opengovsg/design-system-react"
+import {
+  createColumnHelper,
+  flexRender,
+  getCoreRowModel,
+  Row,
+  useReactTable,
+} from "@tanstack/react-table"
```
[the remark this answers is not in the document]
seaerchin: `react-table` is the library helping us to manage the _state_ of the table. the library helping us to manage the load time is `react-virtuoso` through virtualisation

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: i think this is from the library as this is `sticky` - i can check w/ @NatMaeTan to see if this is an issue but this should be relatively minor

seaerchin: @NatMaeTan nodded her head in approval so i think this is ok.

seaerchin:
> > (usually on first PR, so that the load time isn't very long)
> > Hey sorry - am abit confused about this statement, why would only the first PR have large diffs ah?

cos the first PR will have all the changes since we spun up the site sadge

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
dcshzj: I know at least the new navbar will use this, the site dashboard will not be using this component.

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
dcshzj: I won't be using these types! It was more of a matter of preference (separation of concerns) where we keep the declaration of types separate from the actual code logic.

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: placed in `utils/text`!

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: changed to `sort by file type`

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: wrt this - will double check w/ design (@NatMaeTan)

src/layouts/Dashboard/components/ReviewRequestModal/ReviewRequestForm/ReviewRequestForm.tsx
seaerchin: good catch, missed this!

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: hmm in that case i **won't** export these - my reasoning is that i'd rather not place these in the shared folders as that implies that these would be used across the codebase and makes the codebase more difficult to navigate (encapsulation - private knowledge about this component shouldn't escape fi there's no need).  

having said that, if you do foresee yourself using this, please feel free to shift them to the `types/` folder!

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -0,0 +1,336 @@
...
+    date: date.toLocaleDateString("en-GB", {
+      month: "short",
+      year: "numeric",
+      day: "numeric",
+    }),
+    time: date.toLocaleTimeString("en-GB", {
+      timeStyle: "short",
+    }),
+  }
+}
```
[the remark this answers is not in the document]
seaerchin: shure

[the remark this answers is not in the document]
seaerchin: shifted to `utils/date`!

kishore03109 (approved):

dcshzj (approved):
I think just left with the tooltips, but that can be done in a separate PR.

PR #1075, opened 2022-09-26
feat(rr): sorting the modal

seaerchin:
## Problem
This PR adds sorting and filtering to the modal. This is done through adding `react-table` to the frontend. `react-table` presents a headless view of the data, which allows us to plug it into the respective UI library.  

## Solution
- this uses the inbuilt functionality of `react-table` of `filter` + `sort` to achieve this. 
- the docs are abit unclear on how to accomplish this so i'll try to note this down below.

`react-table` uses its `columnHelper` to define its data model. through using the `columnHelper`, we are able to define the following forms of columns: 
1. accessor - these read data from our data model 
2. display - these do **not** read any data but instead, display arbitrary content 
3. grouping - no data model, only used to group other columns together.

accessor columns access our data model through the use of an `accessor` key. **This affects filtering/sorting!** when we sort and filter the column, we pass the **whole** row to the column and let the column decide what it's interested in (through the `accessor` key). 

this allows `react-table` to decouple the data model from the view layer.

### DIFF
--- a/src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
+++ b/src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
@@ -17,6 +17,10 @@ import {
   Breadcrumb,
   BreadcrumbItem,
   useTheme,
+  Menu,
+  MenuButton,
+  MenuList,
+  MenuItem,
 } from "@chakra-ui/react"
 import {
   Searchbar,
@@ -25,9 +29,11 @@ import {
   Button,
 } from "@opengovsg/design-system-react"
 import {
+  ColumnFiltersState,
   createColumnHelper,
   flexRender,
   getCoreRowModel,
+  getFilteredRowModel,
   getSortedRowModel,
   Row,
   SortingState,
@@ -45,6 +51,7 @@ import {
   BiFilterAlt,
   BiChevronRight,
   BiChevronDown,
+  BiCompass,
 } from "react-icons/bi"
 import { TableVirtuoso } from "react-virtuoso"
 
@@ -64,14 +71,7 @@ const getIcon = (iconTypes: EditedItemProps["type"]): JSX.Element => {
       return <Icon {...ICON_STYLE_PROPS} as={BiWorld} />
     }
     case "file": {
-      return (
-        <Icon
-          ml="0.75rem"
-          fontSize="1.25rem"
-          fill="icon.alt"
-          as={BxFileArchiveSolid}
-        />
-      )
+      return <Icon {...ICON_STYLE_PROPS} as={BxFileArchiveSolid} />
     }
     case "setting": {
       return <Icon {...ICON_STYLE_PROPS} as={BiCog} />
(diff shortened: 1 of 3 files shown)


src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -192,63 +192,131 @@ export const RequestOverview = ({
...
+      // like: /some/folder, the correct search term would be some,folder.
+      filterFn: "includesString",
+      header: ({ column }) => (
+        <Button
+          ml="-0.25rem"
+          variant="link"
+          _hover={{
+            textDecoration: "none",
+          }}
+          onClick={column.getToggleSortingHandler()}
```
dcshzj, line 274: Can I check for this sorting function, it can toggle between 3 states:
1. Unsorted
2. Sort ascending
3. Sort descending

Should we only allow it to toggle between states 2 and 3, and only state 1 happens when the user first loads the page?

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -290,37 +366,43 @@ export const RequestOverview = ({
...
             TableRow: Tr,
             TableHead: (props) => <Thead {...props} bg="neutral.100" />,
           }}
-          fixedHeaderContent={() => (
-            <Tr>
-              {table
-                .getFlatHeaders()
-                .map((header) =>
-                  flexRender(
+          fixedHeaderContent={() =>
```
dcshzj, line 390: I think the table header row is missing a bottom border.

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -290,37 +366,43 @@ export const RequestOverview = ({
...
+          onSearch={table.getColumn("itemName").setFilterValue}
           onBlur={handleCollapse}
         />
       </Flex>
       <Box w="100%" borderWidth="1px" borderRadius="8px" borderColor="gray.100">
         <TableVirtuoso
-          style={{ height: 400, borderRadius: "8px" }}
-          data={items}
+          style={{
+            height: 400,
```
dcshzj, line 376: Should we specify this in rem?

dcshzj (changes requested):

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -192,63 +192,131 @@ export const RequestOverview = ({
...
+                <BiFileBlank fontSize="1rem" fill={theme.colors.icon.alt} />
+              }
+              iconSpacing="0.5rem"
+              onClick={() => {
+                column.setFilterValue(["page", "image", "file"])
+              }}
+            >
+              <Text textStyle="subhead-2" textColor="text.body">
+                Pages
+              </Text>
```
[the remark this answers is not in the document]
seaerchin: good catch, i missed this! will add!

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
```diff
@@ -192,63 +192,131 @@ export const RequestOverview = ({
...
+      header: ({ column }) => (
+        <Button
+          ml="-0.25rem"
+          variant="link"
+          _hover={{
+            textDecoration: "none",
+          }}
+          onClick={column.getToggleSortingHandler()}
+        >
+          <HStack w="28rem" spacing="0.5rem">
```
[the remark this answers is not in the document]
seaerchin: yes, we should. however, this was done because using a percentage width would cause the table to freely resize between content. this is because our rows are virtualised so the table doesn't know, at render time, what the max width of the _global_ state is - it only knows the max width of the _local_ state (displayed rows). 

This is most evident when we have a short title and a long title that don't appear together. when the long title appears, the header will look fine. however, when the short title appears, the table will resize so that the cell will take width of item name, which looked weirder to me than just having a fixed width throughout. this was acceptable to me due to horizontal scrolling existing on table + we don't guarantee support for smaller screen sizes (kekw) 

will defer to @NatMaeTan on this tho

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: no strong opinion here, will defer to @NatMaeTan

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: yea... i tried adding it in by setting a `borderBottom` on both `Th` and `Tr` but neither of them worked. the table content actually overlays the header so it hides teh border, annoyingly enough. checked w/ @NatMaeTan on this and she's fine w/ this but if you want, we could file an issue and i could work on this after!

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: yes we should!! good catch!

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
dcshzj: If design is ok I'm fine too!

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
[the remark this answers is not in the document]
NatMaeTan: What are you referring to when you say 'Modal'? Is this on the tables @seaerchin

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
NatMaeTan: Is this for the last updated column? Or which column are you referring to? @dcshzj

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
NatMaeTan: Idk what this is but if I told chin it's fine it should be ok

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
dcshzj: Both the item name and the last edited columns, should we allow the user to be able to go back to being unsorted?

Also, @NatMaeTan should we also change the icon to indicate the sorting order?

kishore03109 (changes requested):
Mostly lgtm - with a lingering bug: clicking on an already selected filter (eg. `Pages`) does not deselect the filter!
Screenshot 2022-09-30 at 1 26 33 PM

src/layouts/Dashboard/components/ReviewRequestModal/RequestOverview/RequestOverview.tsx
seaerchin: ^ wrt sorting order, nat say leave till later :/

seaerchin:
> Mostly lgtm - with a lingering bug: clicking on an already selected filter (eg. `Pages`) does not deselect the filter! Screenshot 2022-09-30 at 1 26 33 PM

good catch, this was an erroneous assumption on my end - i thought there could only be a single filter active ._. will fix!

dcshzj (approved):

PR #1076, opened 2022-09-27
ref(collaborators): refactor to remove shared props from context

seaerchin:
## Problem
This PR refactors the collaborators PR to have less shared state in context. There was an extremely thick context layer in the original PR, but different components used varying small sections of that context, implying that it didn't need to be so huge. 

## Solution
This PR accomplishes this through the following patterns: 
1. using `react-hook-form` to simplify state management in forms 
  - this allows us to defer the form's state management into the dom itself rather than having to repeatedly define `get/set` accessors for the various states. this allows us to simplify our context by removing the props from the 
2. using `react-query` to extract state from backend and use it. 
  - originally, this was done through passing in several `setState` functions but this is not required - we can instead extract out the error message from the error returned to us from the backend and use this to display the inline messages for our form.

## Other problems
1. `string` roles instead of `enum`
2. usage of `any` for types. (we should instead fallback to `unknown` or give a minimal working type)


src/components/CollaboratorModal/submodals/MainSubmodal.tsx
```diff
@@ -119,20 +141,17 @@ const MainSubmodal = () => {
...
           <FormLabel>Only admins can add or remove collaborators</FormLabel>
-          <Input
-            onChange={(event) => {
-              setNewCollaboratorEmail(event.target.value)
-              setAddCollaboratorError("")
-            }}
-          />
-          <FormErrorMessage>{addCollaboratorError}</FormErrorMessage>
+          <Input {...register("newCollaboratorEmail")} />
+          <FormErrorMessage>{errorMessage}</FormErrorMessage>
```
dcshzj, line 145: Are we able to reset the form state when the modal is closed, and after any change? Here's the steps to reproduce:
1. Open the collaborators modal (main modal)
2. Enter an invalid email address into the input box
3. Click on "Add collaborator", the error message will prompt the user that the email is invalid.
4. Change the input to a valid email address
5. The "Add collaborator" button is still disabled and the error message is still shown. (expected: error message disappears on change in the email, and the button is re-enabled)

The same applies even when the modal is closed and re-opened. The difference is the input box is cleared, but the error message still shows and the "Add collaborators" button is still disabled.

src/components/CollaboratorModal/submodals/MainSubmodal.tsx
```diff
@@ -119,20 +141,17 @@ const MainSubmodal = () => {
...
         >
           <FormLabel>Only admins can add or remove collaborators</FormLabel>
-          <Input
-            onChange={(event) => {
-              setNewCollaboratorEmail(event.target.value)
-              setAddCollaboratorError("")
-            }}
-          />
-          <FormErrorMessage>{addCollaboratorError}</FormErrorMessage>
+          <Input {...register("newCollaboratorEmail")} />
```
dcshzj, line 144: Should we also add a validation here for getting a valid email address? Right now this functionality is entirely offloaded to the backend but was thinking it might be better to get immediate feedback if the user entered an invalid email address.

src/components/CollaboratorModal/submodals/AcknowledgementSubmodal.tsx
```diff
@@ -84,25 +83,17 @@ const AcknowledgementSubmodalContent = () => {
...
-            setModalState(CollaboratorModalState.Default)
-            setIsAcknowledged(false)
-            closeModal()
-          }}
-        >
+        <Button variant="clear" color="secondary" onClick={onClose}>
           Cancel
         </Button>
-        <Button isDisabled={!isAcknowledged} onClick={handleAddCollaborator}>
+        <Button isDisabled={!isAcknowledged} type="submit">
```
dcshzj, line 96: When the user clicks submit, the success toast appears but the acknowledgement modal is not closed. I think ideally the user should be brought back to the main submodal with the new collaborator shown in the collaborators list.

dcshzj (changes requested):

src/components/CollaboratorModal/submodals/MainSubmodal.tsx
```diff
@@ -19,32 +18,41 @@ import {
...
+
+import { CollaboratorError } from "types/collaborators"
+import { DEFAULT_RETRY_MSG } from "utils"
+
+import {
+  CollaboratorModalState,
+  ACK_REQUIRED_ERROR_MESSAGE,
+} from "../constants"
 
 const LAST_LOGGED_IN_THRESHOLD_IN_DAYS = 60
```
kishore03109, line 34: nit: shall we shift this `LAST_LOGGED_IN_THRESHOLD_IN_DAYS = 60` into `CollaboratorModal/constants.ts`?

src/components/CollaboratorModal/submodals/RemoveCollaboratorSubmodal.tsx
```diff
@@ -1,35 +1,50 @@
...
 import { useCollaboratorModalContext } from "components/CollaboratorModal/CollaboratorModalContext"
-import { CollaboratorModalState } from "components/CollaboratorModal/constants"
 
+import { useLoginContext } from "contexts/LoginContext"
+
+import { useDeleteCollaboratorHook } from "hooks/collaboratorHooks"
 import useRedirectHook from "hooks/useRedirectHook"
 
-const RemoveCollaboratorSubmodal = () => {
+const TEXT_FONT_SIZE = "14px"
```
kishore03109, line 16: hmm can we just remove this constant? feels like its only being used at one place only, seems trivial enough to be just used directly in the code? 
Nit: let's shift this to rem instead!

src/components/CollaboratorModal/submodals/RemoveCollaboratorSubmodal.tsx
```diff
@@ -51,25 +66,17 @@ const RemoveCollaboratorSubmodalContent = () => {
...
-        >
+        <Button variant="clear" color="secondary" onClick={onClose}>
           Cancel
         </Button>
         <Button
           colorScheme="danger"
-          onClick={() => {
+          onClick={async () => {
             if (userIsDeletingThemselves) {
               setRedirectToPage(`http://localhost:3000/sites`)
```
kishore03109, line 76: nit: oof can we use `RouterLink` instead rather than hard coding the page link out? (trivial comment, feel free to dismiss this)

kishore03109 (changes requested):
Thanks for the refactor @seaerchin 🌮 🌮 ! 
I have left some comments that are beyond the scope of the pr, if you feel that they are high effort, feel free to dismiss them since we are in a time crunch either ways :/

src/components/CollaboratorModal/submodals/RemoveCollaboratorSubmodal.tsx
```diff
@@ -1,35 +1,50 @@
...
+const RemoveCollaboratorSubmodalContent = ({
+  siteName,
+}: {
+  siteName: string
+}) => {
+  const { deleteCollaboratorTarget } = useCollaboratorModalContext()
   const { setRedirectToPage } = useRedirectHook()
-  const TEXT_FONT_SIZE = "14px"
+  const { email } = useLoginContext()
+  const userIsDeletingThemselves = email === deleteCollaboratorTarget?.email
```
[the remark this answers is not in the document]
seaerchin: good point - fixed!

src/hooks/collaboratorHooks/index.ts
```diff
@@ -1,6 +1,12 @@
+import { AxiosError } from "axios"
 import { CollaboratorModalState } from "components/CollaboratorModal/constants"
 import { Dispatch, SetStateAction } from "react"
```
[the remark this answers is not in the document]
seaerchin: removed!

src/components/CollaboratorModal/submodals/MainSubmodal.tsx
seaerchin: this is a good point and i didn't have any good solution to this; i need to check with @NatMaeTan on this flow as i'm of the opinion that we should validate for email here and check for existence later (and feedback using a toast)

src/components/CollaboratorModal/submodals/MainSubmodal.tsx
seaerchin: yea, agree - see above!

src/components/CollaboratorModal/CollaboratorModal.tsx
```diff
@@ -46,90 +63,82 @@ export const CollaboratorModal = ({ siteName }: { siteName: string }) => {
...
+    if (errMessage === ACK_REQUIRED_ERROR_MESSAGE) {
+      setModalState(CollaboratorModalState.Acknowledgement)
+    }
+  }, [addCollaboratorError])
+
+  useEffect(() => {
+    if (addCollaboratorSuccess) {
+      successToast({ description: "Collaborator added successfully" })
+    }
+  }, [addCollaboratorSuccess, successToast])
```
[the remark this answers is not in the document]
seaerchin: because it's used inside `useEffect`!

src/components/CollaboratorModal/CollaboratorModal.tsx
```diff
@@ -8,36 +8,53 @@ import {
...
-  const [addCollaboratorError, setAddCollaboratorError] = useState("")
   const [deleteCollaboratorTarget, setDeleteCollaboratorTarget] = useState(
     undefined
   )
   const [modalState, setModalState] = useState<CollaboratorModalState>(
     CollaboratorModalState.Default
   )
-  const [isAcknowledged, setIsAcknowledged] = useState(false)
+
+  const methods = useForm({
```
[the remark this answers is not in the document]
seaerchin: sure! this is a collection of methods though so i'll rename this to `collaboratorFormMethods`

src/components/CollaboratorModal/submodals/AcknowledgementSubmodal.tsx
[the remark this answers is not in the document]
seaerchin: you're right - the `form` is in `CollaboratorModal`

src/components/CollaboratorModal/submodals/RemoveCollaboratorSubmodal.tsx
seaerchin: think i'lkl shift this to `constants` and change it to rem as it's being used in `AcknowledgementSubmodal` and `RemoveCollaboratorSubmodal`

src/components/CollaboratorModal/submodals/RemoveCollaboratorSubmodal.tsx
seaerchin: `RouterLink` is meant to be used in a declarative way - if this was `RouterLink`, we would have to either always redirect them to `/sites` or to `""` @__@ i think this is a lil weird (we should probably have a context doing this auto-magically) but it's ok for now

src/components/CollaboratorModal/submodals/MainSubmodal.tsx
dcshzj: Hmm the point is a bit separate from the one below. This one applies generally to any errors that the user encounters, not just having an invalid email address. As long as the user encounters an error, even if the user corrects the input to something valid, they are not able to add collaborator unless they refresh the page (the button is always disabled and the error message persists).

kishore03109 (approved):

seaerchin:
@dcshzj for the 2 comments here 

1. closing modal and transitioning back into default state should be done
2. unable to add user - could you verify that this is not the case on the base PR?

dcshzj (changes requested):

seaerchin:
tested flow on story + localdev, seems fine; would still advise reviewers to test also :kekw:

src/components/CollaboratorModal/components/MainSubmodal.tsx
```diff
@@ -0,0 +1,251 @@
...
+  const { email } = useLoginContext()
+  const { siteName } = useParams<{ siteName: string }>()
+  const {
+    data: collaborators,
+    isError,
+  } = CollaboratorHooks.useListCollaboratorsHook(siteName)
+  const { isDisabled } = useFormControlContext()
+
+  return (
+    <Box mt="40px">
```
dcshzj, line 68: Should this be in `rem` instead? Same for the height/padding below.

src/components/CollaboratorModal/components/MainSubmodal.tsx
```diff
@@ -0,0 +1,251 @@
...
+                <Text pl="4px" color="gray.500">
+                  {email === collaborator.email ? "(You)" : null}
+                </Text>
+                {numDaysAgo(collaborator.lastLoggedIn) >=
+                  LAST_LOGGED_IN_THRESHOLD_IN_DAYS && (
+                  <Text pl="4px" color="danger.500">
+                    {`(Last logged in ${numDaysAgo(
+                      collaborator.lastLoggedIn
+                    )} days ago)`}
+                  </Text>
```
dcshzj, line 88: I think the text here should use 0.875rem instead. It's currently too large and causing an overflow:

Screenshot 2022-10-20 at 11 04 04

Also, minor nit: Can we call `numDaysAgo()` once and store the result in a variable, then just use that variable here?

src/components/CollaboratorModal/components/MainSubmodal.tsx
```diff
@@ -0,0 +1,251 @@
...
+                )}
+              </Box>
+            </GridItem>
+            <GridItem colSpan={2}>
+              <Box display="flex" alignItems="center" h="100%">
+                <Text textTransform="capitalize">
+                  {_.capitalize(collaborator.role)}
+                </Text>
```
dcshzj, line 96: Hmm I think we can skip the capitalisation from lodash and only rely on the CSS transformation?

Also, the role does not appear (see above screenshot). I think the data is in `collaborator.SiteMember.role` instead (similar to your `CollaboratorDto`)

src/components/CollaboratorModal/components/MainSubmodal.tsx
```diff
@@ -0,0 +1,251 @@
...
+  }, [addCollaboratorError])
+
+  useEffect(() => {
+    if (addCollaboratorSuccess) {
+      successToast({ description: "Collaborator added successfully" })
+      collaboratorFormMethods.reset()
+    }
+  }, [addCollaboratorSuccess, collaboratorFormMethods, successToast])
+
+  const isDisabled = role !== "ADMIN"
```
dcshzj, line 186: I think we should also disable the button if the input field is empty or invalid.

src/components/CollaboratorModal/components/MainSubmodal.tsx
```diff
@@ -0,0 +1,251 @@
...
+                  isDisabled={isDisabled}
+                >
+                  <FormLabel>
+                    Only admins can add or remove collaborators
+                  </FormLabel>
+                  <Input
+                    {...collaboratorFormMethods.register(
+                      "newCollaboratorEmail"
+                    )}
+                  />
```
dcshzj, line 232: Needs a placeholder text: `name@email.com`.

Also, I think this input should be cleared onClose, so that when the user reopens the modal it starts from a fresh state.

src/components/CollaboratorModal/components/RemoveCollaboratorSubmodal.tsx
```diff
@@ -0,0 +1,91 @@
...
+                {userToDelete?.email}{" "}
+              </Text>
+              <Text as="span">
+                from this site, they will no longer be able to make any changes.
+              </Text>
+            </Text>
+          )}
+
+          <Stack spacing={4} direction="row" justify="right" mt="36px">
+            <Button variant="clear" color="secondary" onClick={onClose}>
```
dcshzj, line 68: I think this `onClose` is insufficient, as it can cause the user to be stuck. Steps to reproduce:
1. Ensure you are the only collaborator left on the site
2. Attempt to remove yourself as the collaborator
3. Error message correctly shows that the site needs at least one admin
4. However, the modal still stays on the RemoveCollaboratorSubmodal. Cancelling/Closing the modal and reopening the collaborators modal again still brings the user to the RemoveCollaboratorSubmodal.

In addition, if the user accidentally clicks on the remove collaborator trash bin icon, there is no way for the user to cancel to go back to the original modal. The user is forced to remove the collaborator.

We should bring the user back to the MainSubmodal on close so that the user is not stuck.

src/components/CollaboratorModal/components/AcknowledgementSubmodal.tsx
```diff
@@ -93,23 +78,17 @@ const AcknowledgementSubmodalContent = () => {
...
-        <Button
-          variant="clear"
-          color="secondary"
-          onClick={() => {
-            setModalState(CollaboratorModalState.Default)
-            setIsAcknowledged(false)
-            closeModal()
-          }}
-        >
+        <Button variant="clear" color="secondary" onClick={onClose}>
```
dcshzj, line 81: I think this is going to be quite nitty and annoying: The `isAcknowledged` should reset to false on cancel. Otherwise, it is possible for the user to "bypass" the acknowledgement. Steps to reproduce:

1. Add a collaborator that requires acknowledgement
2. The acknowledgement submodal appears, click on the "I agree to Isomer terms of use" checkbox
3. Click on "Cancel" or the close button (this indicates that the user has decided to disagree with the terms of use)
4. Open the collaborators modal again, the previous collaborator is prefilled in the input box (separate issue which I will address in another comment)
5. Click on "Add collaborator"
6. Notice that the collaborator is added without the acknowledgement submodal appearing. It should appear to prompt the user to agree to the terms of use.

I think this is being super nitty already so if no time can just skip this. No loss in functionality anyway if we do not fix this.

dcshzj (changes requested):
I'm thinking this rewrites quite a bit of the original PR, should we change the base to directly merge into `feat/identity-phase-2`? Not sure if it might help in the reviewing so we see only the additions.

src/components/CollaboratorModal/components/MainSubmodal.tsx
seaerchin: we can't use `text-transform` sadly - see [here](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform#:~:text=The%20text%2Dtransform%20CSS%20property,help%20improve%20legibility%20for%20ruby.)

the spec states 
>`capitalize`
Is a keyword that converts the first letter of each word to uppercase. Other characters remain unchanged (they retain their original case as written in the element's text).

src/components/CollaboratorModal/components/AcknowledgementSubmodal.tsx
seaerchin: clarified offline - happens when 
1. fill in collaborator which requires ACK
2. go to ack modal and tick
3. click on the **closebutton**
4. reopen modal 
5. fill in same collaborator and click add
6. toast pops up

src/components/CollaboratorModal/components/RemoveCollaboratorSubmodal.tsx
seaerchin: yea, thanks for the catch. 2 fold fix, disable if only 1 admin left + reset state on close completion to show main modal

src/components/CollaboratorModal/components/MainSubmodal.tsx
seaerchin: fixed!

src/components/CollaboratorModal/components/MainSubmodal.tsx
seaerchin: for data - added the data transformation in the hook. didn't expect this because ts wasn't throwing an error... structural typing do be like that

src/components/CollaboratorModal/components/MainSubmodal.tsx
seaerchin: good point - added on the button; this is also used for the form control, which will disable the input as well.

src/components/CollaboratorModal/components/MainSubmodal.tsx
seaerchin: good point wrt placeholder; wrt clearing - i clear only if the submission is successful. my thinking was that if the user mistyped, they'd rather have the old value there and just edit a few chars rather than having to re-enter. wdyt about this?

src/components/CollaboratorModal/components/AcknowledgementSubmodal.tsx
seaerchin: ok turns out `reset` was from the `useMutation` hook not `react-hook-forms` so the state wasn't reset after modal closes .-. 

fixed

src/components/CollaboratorModal/components/MainSubmodal.tsx
seaerchin: this text area is prone to overflowing - if we have an extremely long email, it will overflow its boundary too :/

more inclined to file as issue cos this is cosmetic and has no functional impact (atm) 

will fix the multiple calls

dcshzj: Yeah I think we don't have to fix the overflowing issue, only the text size should still be 0.875rem as defined in Figma.

src/components/CollaboratorModal/components/MainSubmodal.tsx
dcshzj: Yeah that makes sense too, let's go with the original behaviour.

dcshzj: The placeholder text still needs to be added hmm

seaerchin: the placeholder disappeared in the rebase lmao

dcshzj (changes requested):
The rest LGTM, just 2 more changes:
1. Decreasing the text size to 0.875rem to match Figma
2. Adding email placeholder text

seaerchin:
> The rest LGTM, just 2 more changes:
> 
> 1. Decreasing the text size to 0.875rem to match Figma
> 2. Adding email placeholder text

fixed

dcshzj (approved):
LGTM!

PR #1078, opened 2022-09-28
Fix: remove permalink for file type resources

alexanderleegs:
This PR fixes a bug introduced in https://github.com/isomerpages/isomercms-frontend/pull/1064 - a permalink was being generated for all resource pages, even for file type resources, which caused the file specified to not show up. This PR fixes the behaviour by removing this param for file type resources.


alexanderleegs:
Making a hotfix instead

PR #1085, opened 2022-09-29
feat(review-requests): review overview on dashboard

seaerchin:
## Problem
This PR adds in the review request dashboard, which presents an overview of the review request and allows the user to take certain actions depending on their role.

Part of #1070 

## Solution
1. Add in a top component to display top level information about the review request.
     - not exporting as it probably won't be widely used. 
2. extend `MenuDropdownButton` to both take a `colorScheme` prop and compute the icon fill. This might not be the best solution (eg: `colorScheme` is `success` w/ a outline button will result in blue icon but the theme will use red elsewhere. this should also be defined in `theme` but we can do both of this later as it's good enough "for now" 
3. rejig some components so review requests is better structured


src/layouts/ReviewRequest/Dashboard.tsx
```diff
@@ -0,0 +1,167 @@
...
+                border="1px solid white"
+                mr="-0.25rem"
+                bg="primary.100"
+                name={initials}
+                textStyle="caption-1"
+                textColor="primary.400"
+                size="sm"
+              />
+            )
+          })}
```
dcshzj, line 162: Very minor nit, depends on what design says: The overlapping of avatars is different from the Figma design.
* Figma: Left avatar overlaps over the right avatar
Screenshot 2022-09-30 at 11 05 05

* Here: Right avatar overlaps over the left avatar
Screenshot 2022-09-30 at 11 05 32

src/layouts/ReviewRequest/Dashboard.tsx
```diff
@@ -0,0 +1,167 @@
...
+              </PopoverContent>
+            </Popover>
+          </HStack>
+          <Spacer />
+          <ApprovalButton />
+        </Flex>
+        <SecondaryDetails
+          requestor={requestor}
+          reviewers={reviewers}
+          reviewRequestedTime={new Date()}
```
dcshzj, line 92: Should we get the requested time from the props instead?

src/layouts/ReviewRequest/Dashboard.tsx
```diff
@@ -0,0 +1,167 @@
...
+  changedItems,
+}: ReviewRequestDashboardProps): JSX.Element => {
+  const { onCopy, hasCopied } = useClipboard(reviewUrl)
+
+  return (
+    <Box bg="white" w="100%" h="100%">
+      <VStack
+        bg="blue.50"
+        pl="9.25rem"
+        pr="2rem"
```
dcshzj, line 51: Would it work to have these 2 values moved to the `<Box>` above? So then the same `pl` and `pr` values for the RequestOverview Box can be removed.

dcshzj (changes requested):

src/layouts/ReviewRequest/Dashboard.tsx
```diff
@@ -0,0 +1,167 @@
...
+        pt="2rem"
+        pb="1.5rem"
+        spacing="0.75rem"
+        align="flex-start"
+      >
+        <Flex w="100%">
+          <HStack spacing="0.25rem">
+            <Text textStyle="h5">{title}</Text>
```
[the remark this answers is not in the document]
seaerchin: for the heading one: yea, this is a known issue... this is cos the base design system has wrong `textStyles` (fact-check me @NatMaeTan)

src/layouts/ReviewRequest/Dashboard.tsx
seaerchin: ooh oops, think i saw wrongly HAHAHAH

src/layouts/ReviewRequest/Dashboard.tsx
seaerchin: good catch!

src/layouts/ReviewRequest/Dashboard.tsx
```diff
@@ -0,0 +1,167 @@
...
+  reviewers: string[]
+  reviewRequestedTime: Date
+}
+const SecondaryDetails = ({
+  requestor,
+  reviewers,
+  reviewRequestedTime,
+}: SecondaryDetailsProps) => {
+  const { date, time } = getDateTimeFromUnixTime(
+    reviewRequestedTime.getTime() / 1000
```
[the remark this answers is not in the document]
seaerchin: good ctach, fixed!

src/layouts/ReviewRequest/Dashboard.tsx
seaerchin: sadly we can't cos this results in the top having white bg... an alternatiev would be to have the `Box` have the blue bg then we set the `Box` wrapping the `RequestOverview` to have white bg - would you prefer this? (then we shift the `pl + pr` to above)

src/layouts/ReviewRequest/Dashboard.tsx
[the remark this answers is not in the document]
seaerchin: no truncation - will spill over; no strong opinion, will check with @NatMaeTan

src/layouts/ReviewRequest/Dashboard.tsx
dcshzj: Oh I think it's fine, let's just stick to the current way of doing it!

PR #1091, opened 2022-09-30
feat(rr): misc modals

seaerchin:
**note: i have 700 lines from svgs and more from package.json so this PR is not that big**
## Problem
This PR adds the remaining misc modals. Most of the modals here don't have much logic with the exception of `ManageReviewerModal`, which prevents deletion when there is a single reviewer left.

Part of #1070 

## Solution
1. use design system + chakra UI to make the modals 
2. for the `ManageReviewerModal`, used a hook to manage the internal state of the modal and used said hook to conditionally display the cross


src/layouts/ReviewRequest/components/ApprovedModal/ApprovedModal.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+  component: ApprovedModal,
+} as ComponentMeta<typeof ApprovedModal>
+
+const Template: Story<never> = () => {
+  const { isOpen, onOpen, onClose } = useDisclosure({ defaultIsOpen: true })
+  return (
+    <>
+      <Button onClick={onOpen}>Open Modal</Button>
+      <ApprovedModal isOpen={isOpen} onClose={onClose}>
+        <Button colorScheme="danger">Click me</Button>
```
dcshzj, line 18: Minor nit but can this be removed? It was a bit confusing why this was here (same applies for the other stories).

src/layouts/ReviewRequest/components/CancelRequestModal/CancelRequestModal.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+        <ModalFooter>
+          <Button
+            textColor="text.body"
+            variant="clear"
+            mr="1rem"
+            onClick={onClose}
+          >
+            Go back
+          </Button>
```
dcshzj, line 42: Should the colour scheme be secondary for this?

Also need to check with @NatMaeTan when hovering over this button, should it give a blue or purple background?

src/layouts/ReviewRequest/components/ManageReviewerModal/ManageReviewerModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+                  ...base,
+                  width: "100%",
+                  minHeight: "2.75rem",
+                }),
+                control: (base) => ({ ...base, height: "100%" }),
+                // NOTE: Don't allow removal if there is only 1 selected admin
+                multiValueRemove: (base) => {
+                  const canRemove = selectedAdminsList.length > 1
+                  return { ...base, display: canRemove ? "inherit" : "none" }
+                },
```
dcshzj, line 96: Just curious, would we get to a state where the error toast is triggered like how it is shown on Figma?

dcshzj (changes requested):

src/layouts/ReviewRequest/components/PublishedModal/PublishedModal.tsx
```diff
@@ -0,0 +1,62 @@
...
+          <Button
+            variant="clear"
+            mr="1rem"
+            onClick={onClose}
+            colorScheme="primary"
+            textColor="text.title.brandSecondary"
+          >
+            Go to live site
+          </Button>
+          <Button>Publish now</Button>
```
alexanderleegs, line 57:
```suggestion
          <Button
            variant="clear"
            mr="1rem"
            onClick={onClose}
            colorScheme="primary"
            textColor="text.title.brandSecondary"
          >
            Go to live site
          </Button>
          <Button>Back to Site Dashboard</Button>
```
I think the onClose part might belong to the second button also?

alexanderleegs (commented):
Left some comments! Also, the package-lock changes are 58k lines because it uses lockfileversion 2 (node 16) - as discussed offline, we should keep it at node 14 for now and do the switch to node 16 on its own in a separate PR at some point

src/layouts/ReviewRequest/components/ApprovedModal/ApprovedModal.stories.tsx
seaerchin: to clarify, you meant the entire button right?

src/layouts/ReviewRequest/components/CancelRequestModal/CancelRequestModal.tsx
seaerchin: i double checked with nat before, our `primary.500` might be wrong ._.

src/layouts/ReviewRequest/components/ManageReviewerModal/ManageReviewerModal.tsx
seaerchin: could you point me to the error toast section?

seaerchin: but initial thoughts are that it shouldn't be possible, at least from the UI, to remove the last admin.

src/layouts/ReviewRequest/components/ManageReviewerModal/ManageReviewerModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+              components={{
+                Input: () =>
+                  remainingAdmins.length > 0 ? (
+                    <Text
+                      ml="0.25rem"
+                      textDecoration="underline"
+                      color="text.link.default"
+                      textStyle="body-2"
+                    >{`+${remainingAdmins.length} more`}</Text>
+                  ) : null,
```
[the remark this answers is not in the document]
seaerchin: clarified with nat on this - we were initially supposed to use `MultiSelect` (but cos our deps upgrade failed so now we're using `react-select` :pepehands:), which actually **does not** do the above HAHAHAH it displays up to 4 (see [here](https://github.com/opengovsg/design-system/blob/main/react/src/Dropdown/components/MultiSelectCombobox/SelectedItems.tsx#L46)) then thereafter does the `xxx more`. 

we just sort then display shortest 2 so we don't have to resize (hopefully) to 2 lines

src/layouts/ReviewRequest/components/PublishedModal/PublishedModal.tsx
seaerchin: oops, good catch! i think the button should take them to `/sites` on FE - haven't done it yet cos it's not wired to be

src/layouts/ReviewRequest/components/ApprovedModal/ApprovedModal.stories.tsx
dcshzj: Yup the entire danger button with "Click me".

src/layouts/ReviewRequest/components/ManageReviewerModal/ManageReviewerModal.tsx
dcshzj: This is the Figma link: https://www.figma.com/file/7OtNYkc4s9BkFk05ConEbd/%E2%AD%90-V1.5?node-id=2795%3A80957

Screenshot:
Screenshot 2022-10-04 at 09 36 02

src/layouts/ReviewRequest/components/CancelRequestModal/CancelRequestModal.tsx
dcshzj: Hmm I think we will have to adjust this in a separate PR.

src/layouts/ReviewRequest/components/ManageReviewerModal/ManageReviewerModal.tsx
seaerchin: When we try to uncheck last admin but that's not possible under the new ui (no X, drop downs don't have checkbox) so idt this shows heh

dcshzj: If @NatMaeTan is ok then I think this is ok hahaa

seaerchin: as discussed during sync - got impl error!!! should show **all** admins; will fix

src/layouts/ReviewRequest/components/ApprovedModal/ApprovedModal.stories.tsx
seaerchin: actually cos this is the playground so i wanted to show the open/close functionality - i could add a `Open Modal` story?

src/layouts/ReviewRequest/components/ManageReviewerModal/ManageReviewerModal.tsx
seaerchin: tracked by #1108. will raise a separate PR for this, won't fix here.

src/layouts/ReviewRequest/components/ApprovedModal/ApprovedModal.stories.tsx
seaerchin: clarified offline - modal wasn't showing

dcshzj (approved):
LGTM! Just the package-lock.json file needs to be updated.

PR #1093, opened 2022-10-03
Feat/notifications storybook

alexanderleegs:
## Problem

This PR adds a storybook to the header. Builds on PR #1092.

This PR also fixes the issues introduced in #1092 which broke our existing stories - the get notifications endpoint was not handled properly prior to this change


src/mocks/utils.ts
```diff
@@ -15,34 +16,50 @@ import {
...
+        return res(delay ? ctx.delay(delay) : ctx.delay(0), ctx.json(mockData))
+      })
+    case "get":
+      return rest.get(endpoint, (req, res, ctx) => {
+        return res(delay ? ctx.delay(delay) : ctx.delay(0), ctx.json(mockData))
+      })
+    default:
+      return rest.get(endpoint, (req, res, ctx) => {
+        return res(delay ? ctx.delay(delay) : ctx.delay(0), ctx.json(mockData))
+      })
```
seaerchin, line 37: i think we might have more http verbs in the future and we could instead extend this to be `rest[verb](...)`, if we can restrict the verb to `"get" | "post" | "delete"` etc and set a default to `get`. wdyt about this?

seaerchin (approved):
minor nit

src/mocks/utils.ts
alexanderleegs: That makes sense - changed in https://github.com/isomerpages/isomercms-frontend/pull/1093/commits/2cc56329ff030946f3f252bb1133a8d07992987b!

PR #1105, opened 2022-10-12
feat(rr): integration with site dashboard

seaerchin:
apologies in advance for the big pull request, i should have structured this better ;--;

## Problem
This PR adds CRUD hooks for review requests together with integration of the site dashboard together with review requests.

Related to #1070 

## Solution
1. add hooks + base API calls for data fetching and updating
2. add linkages between components so that the flow is cohesive
3. create `RoleContext` to display different views to different users

**notes**
the frontend queries for a **full** view of the RR to determine role - this might be slow with big RRs and if required, we should add another end point to retrieve just the properties on the backend


src/layouts/ReviewRequest/Dashboard.tsx
```diff
@@ -13,127 +13,271 @@ import {
...
+          pl="9.25rem"
+          pr="2rem"
+          pt="2rem"
+          pb="1.5rem"
+          spacing="0.75rem"
+          align="flex-start"
+        >
+          <Flex w="100%">
+            <HStack spacing="0.25rem">
+              <Text textStyle="h5">{data?.title || ""}</Text>
```
dcshzj, line 90: Minor nit: Can we switch to using a skeleton instead? Also display something like "Unable to retrieve data" if the review request could not be loaded.

src/layouts/ReviewRequest/Dashboard.tsx
```diff
@@ -13,127 +13,271 @@ import {
...
+                    <Text textStyle="body-2" color="text.inverse">
+                      Link copied!
+                    </Text>
+                  </PopoverBody>
+                </PopoverContent>
+              </Popover>
+            </HStack>
+            <Spacer />
+            <CancelRequestButton isApproved={isApproved} />
+            <ApprovalButton isApproved={isApproved} />
```
dcshzj, line 119: For some reason these two buttons appear the same on my side:

Screenshot 2022-10-18 at 10 11 10

I think maybe need to check the reviewer role to display the correct button, instead of disabling?

src/layouts/ReviewRequest/components/PublishedModal/PublishedModal.tsx
```diff
@@ -48,15 +69,17 @@ export const PublishedModal = (
 
         <ModalFooter>
           <Button
-            variant="clear"
+            variant="link"
```
dcshzj, line 72: Can we change this to use the ButtonLink component?

src/services/ReviewService.ts
```diff
@@ -0,0 +1,67 @@
...
+  siteName: string,
+  prNumber: number
+): Promise<void> => {
+  const endpoint = `/sites/${siteName}/review/${prNumber}/approve`
+  return apiService.post(endpoint)
+}
+
+export const cancelReviewRequest = async (
+  siteName: string,
+  prNumber: number
```
dcshzj, line 63: I think maybe we should rename this to `reviewId` to be consistent with how it's named for the other endpoints.

dcshzj (changes requested):

src/hooks/reviewHooks/useMergeReviewRequest.ts
```diff
@@ -0,0 +1,43 @@
...
+} from "constants/queryKeys"
+
+import { ErrorDto } from "types/error"
+
+import * as ReviewService from "../../services/ReviewService"
+
+export const useMergeReviewRequest = (
+  siteName: string,
+  prNumber: number,
+  shouldInvalidate = true
```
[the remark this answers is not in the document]
seaerchin: yeap, it's just used there. i exposed this so that if we do use it somewhere else in the future, we can get invalidation by default.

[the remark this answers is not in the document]
seaerchin: if we do remove the invalidation prop, we haev to decide whether we want to manually invalidate or invalidate in the hook.

invalidation in the hook would lead to a desync with the desired flow (on publish, we will trigger a refetch -> go back to dashboard rather than showing the modal). manual desync leads to breaking of established code convention + usages in the future have to be aware of this (burden of knowledge).

i felt that this was a compromise between the 2 but lmk what are your thoughts!

src/hooks/settingsHooks/useGetSiteUrl.ts
```diff
@@ -0,0 +1,22 @@
...
+import { SITE_URL_KEY } from "constants/queryKeys"
+
+import { apiService } from "services/ApiService"
+
+const getSiteUrl = async (siteName: string) => {
+  const settingsResp = await apiService.get(
+    `${process.env.REACT_APP_BACKEND_URL_V2}/sites/${siteName}/settings`
+  )
+  return settingsResp.data.url
+}
```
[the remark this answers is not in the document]
seaerchin: yeap, this is tracked by #1109

src/layouts/ReviewRequest/Dashboard.tsx
seaerchin: oh yes - i'm aware of this and i haven't created a tracking issue yet AHHA

src/layouts/ReviewRequest/Dashboard.tsx
seaerchin: good catch, this change was made for manual testing and i missed out on reverting

src/services/ReviewService.ts
seaerchin: actually wdyt about changing this to be `prNumber` as we are now using `(siteName, prNumber)` to identify each RR

src/contexts/RoleContext.tsx
```diff
@@ -0,0 +1,60 @@
+import { createContext, useContext, PropsWithChildren } from "react"
```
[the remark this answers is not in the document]
seaerchin: sounds good, will do

src/hooks/settingsHooks/useGetSiteUrl.ts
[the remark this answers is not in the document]
dcshzj: Right oops I forgot, I will add a separate endpoint called `/sites/:siteName/siteUrl` which will return the live site URL. This is cos the existing `/sites/:siteName/info` endpoint is a little more expensive due to the calls to the GitHub API to get the commits information.

src/hooks/reviewHooks/useMergeReviewRequest.ts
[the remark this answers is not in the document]
dcshzj: Oh yeah I'm ok with having validation/invalidation but it was more of whether there would be a chance that we will use this hook again somewhere else. It's a very small point so we can just leave it as it is!

src/services/ReviewService.ts
dcshzj: Yup sounds good also, let's go ahead with `prNumber`

src/layouts/ReviewRequest/components/PublishedModal/PublishedModal.tsx
seaerchin: unfortunately, no because the hover styling is hard-coded and leads to the bg obscuring the text on hover.

src/hooks/settingsHooks/useGetSiteUrl.ts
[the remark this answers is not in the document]
dcshzj: Weird but there's also an existing hook in `src/hooks/settingsHooks/useUrlHook.jsx`. Perhaps we can use that directly?

src/layouts/ReviewRequest/Dashboard.tsx
seaerchin: tracked by #1126

src/services/ReviewService.ts
seaerchin: could we file this as a separate issue? i think there might be downstream PRs that would also require this change, making it tedious to do here.

src/hooks/settingsHooks/useGetSiteUrl.ts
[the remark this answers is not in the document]
seaerchin: didn't use it because it was in js, and i wanted type safety. i didn't migrate the hook over as i'd have to change the call-sites, which is tedious + time-consuming.

could we file an issue for this and let me tackle it down the line?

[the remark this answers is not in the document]
dcshzj: Tracked in #1129.

src/services/ReviewService.ts
dcshzj: Tracked in #1130.

dcshzj (approved):
LGTM!

PR #1120, opened 2022-10-17
fix(rr): disallow creation if no admin selected

### LINKED ISSUE #1116: [BUG] Review request creation when no reviewer selected
## Problem
Currently, our submit button is not disabled when there are no admins selected as reviewers. This allows the users to go through with creation of rr 

## Suggested solution
Disable submit until >= 1 reviewer and title is there. IF they bypass this by pinging our backend, it should also be disallowed with an error message.

## Definition of done
unable to create rr on FE w/o title + reviewer; BE should return message

seaerchin:
## Problem
See #1116 for a brief description of the issue

Closes #1116 

## Solution
add a `watch` conditional onto the `reviewers` field so that we disallow creation when `reviewers.length === 0`


dcshzj (approved):

alexanderleegs (approved):
lgtm

PR #1121, opened 2022-10-17
feat(rr): lock editing when active rr is approved

### LINKED ISSUE #1110: [FEAT] RR editing should be locked when a review is approved
## Problem
Editing of a users' site should be prevented when there is an **approved** review request. 

## Suggested solution
(pending with design but probably will redirect to the review request dashboard page w/ a toast) 

## Definition of done
when a rr is approved, users are unable to edit **ANY** files. this is inclusive of 
    - settings
    - pages
    - images 
    - files
when a user accesses the edit page directly from the url, they should also be prevented from editing

seaerchin:
## Problem
See issue for overview of problem

Closes #1110 

## Solution
1. add new `Greyscale` component that displays components in grey + disables pointer events
2. add a `Redirect` component that checks our BE to see if there are any approved RRs.
    - this uses the new component for a loading state (to avoid ui being unreactive) + handles toasting on error/approved rr states.
    - the toast message for **error** is taken from BE; for approved RR, a fixed message is shown together with a redirect to `/dashboard`
    - add a HOC to wrap our legacy components...
3. refactor `ProtectedRoute` to TS


alexanderleegs (approved):
lgtm

PR #1123, opened 2022-10-18
fix: touch up site dashboard

### LINKED ISSUE #1115: [BUG] Unparsed datetime on dashboard
## Problem
currently, our frontend retrieves datetime from the backend, which is given as unix time (integers). this isn't converted into a human readable format on the frontend and is shown as the raw unix time.

## Suggested solution
use util functino (alr up) to retrieve

## Definition of done
human readable datetime on FE

dcshzj:
## Problem

The site dashboard displays the Unix timestamp directly instead of a human-readable date format, which is hard to interpret for users. Also, the site dashboard is designed for email login users, but both GitHub login users and email login users are able to access it.

Fixes #1115.

## Solution

**Breaking Changes**

**Improvements**:

- The dates on the site dashboard is now displayed in a human-readable format.
- The site dashboard has been restricted to only email login users, GitHub login users will be redirected to the workspace.
- Time format has been changed to use 12-hour time instead of 24-hour time.

## Tests

- [x] Smoke tests

1. Run with `npm run dev`.
2. Navigate to the site dashboard of any site (as an email login user)
3. Verify that the dates shown are in the human-readable format
4. Log out and log in through GitHub
5. Navigate to the site dashboard of any site (by changing the URL)
6. Verify that the user will be redirected to the workspace.

## Deploy Notes

*None*

### DIFF
--- a/src/layouts/SiteDashboard/SiteDashboard.tsx
+++ b/src/layouts/SiteDashboard/SiteDashboard.tsx
@@ -60,7 +60,8 @@ export const SiteDashboard = (): JSX.Element => {
     isLoading: isCollaboratorsStatisticsLoading,
   } = useGetCollaboratorsStatistics(siteName)
 
-  const { savedAt, savedBy, publishedAt, publishedBy } = siteInfo || {}
+  const savedAt = getDateTimeFromUnixTime(siteInfo?.savedAt || 0)
+  const publishedAt = getDateTimeFromUnixTime(siteInfo?.publishedAt || 0)
 
   return (
     <SiteViewLayout overflow="hidden">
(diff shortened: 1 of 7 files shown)


src/layouts/SiteDashboard/SiteDashboard.tsx
```diff
@@ -58,6 +67,16 @@ export const SiteDashboard = (): JSX.Element => {
...
 
+  const savedAt = getDateTimeFromUnixTime(siteInfo?.savedAt || 0)
+  const publishedAt = getDateTimeFromUnixTime(siteInfo?.publishedAt || 0)
+
+  useEffect(() => {
+    // GitHub users should not be able to access this page
+    if (userId !== "Unknown user" && !!userId) {
+      setRedirectToPage(`/sites/${siteName}/workspace`)
+    }
+  })
```
alexanderleegs, line 78: I'm wondering how many places we need to do this redirection - is it just here and collaborators? If there are more i'm wondering if it's better to have a different type of route which automatically restricts access and redirects instead, but regardless i think this can be settled in a future pr if necessary

alexanderleegs (approved):
Minor comment for discussion, but otherwise lgtm!

src/layouts/SiteDashboard/SiteDashboard.tsx
dcshzj: Not collaborators but will be needed for review requests. I think this can be handled in the future together with the review requests side of things? @seaerchin

PR #1124, opened 2022-10-18
fix: integrate collaborators modal into the site dashboard

dcshzj:
## Problem

The site collaborators modal cannot be accessed from the site dashboard.

## Solution

**Breaking Changes**

**Improvements**:

- The collaborators modal is now integrated into the main site dashboard, with adjustments to the CollaboratorsModal component to allow for opening and closing of the modal.
- The temporary dashboard page has been removed.

## Tests

- [x] Smoke tests

1. Run the frontend with `npm run dev`
2. Navigate to any site's dashboard
3. Open the collaborators modal using the "Manage" button on the top right corner of the site collaborators card.

## Deploy Notes

*None*


src/styles/isomer-cms/elements/buttons.scss
```diff
@@ -91,8 +91,4 @@ button {
       box-shadow: unset;
     }
   }
-
-  &:last-child {
-    margin-right: 0px;
-  }
```
alexanderleegs, line 97: Actually, is it possible to do this outside of the file? We've been avoiding modification of these files because they tend to have a bunch of unexpected changes haha

alexanderleegs (commented):
Minor comment on css!

src/styles/isomer-cms/elements/buttons.scss
dcshzj: Hmm I was struggling quite a bit with the CSS specificity because it seems like the `last-child` selector has a higher specificity than the `marginRight` prop in the Button component. I also did a cursory look at other parts of the CMS and it doesn't seem like any component was affected.

seaerchin: tbh, we might wanna err on the side of caution here (avoiding visual breakage) but fwiw, this is increasingly become legacy so we might wanna prune too...

i'm fine with both either way but if deciding to stay safe, passing `styles` directly could work to override specificity?

src/components/CollaboratorModal/CollaboratorModal.tsx
```diff
@@ -24,13 +22,16 @@ import * as CollaboratorHooks from "hooks/collaboratorHooks"
 
 import { useSuccessToast } from "utils"
 
+interface CollaboratorModalProps extends Omit<ModalProps, "children"> {
+  siteName: string
+}
+
```
seaerchin, line 28: fyi interface changed a lil - behaves like a fully fledged modal w/o children now (should be minor change)

seaerchin (commented):
in principle approval; am fine w/ removal of global css anyways - only blocker is type change to modal

src/styles/isomer-cms/elements/buttons.scss
dcshzj: I'm inclined to remove this as I think by default, our buttons have the `margin-right` property set to 0px. Any usages of the `marginRight` or `mr` props in our codebase do not apply to the last child button element, so I think it would be safe to remove this.

src/components/CollaboratorModal/CollaboratorModal.tsx
dcshzj: Updated via a rebase in d63380d.

seaerchin (approved):
lgtm, please hold off on merging!!!

PR #1127, opened 2022-10-20
feat(rr): allow update of admins in rr 

### LINKED ISSUE #1108: [BUG] Unable to update admins
## Problem description
reviewers are permanent at the time of review request creation and requestors are unable to update it. The backend for updating rr details is already done and includes admin so the work is purely frontend and involves
1. getting state from RHF
2. use a hook to ping our BE

## Definition of done
1. requestors can change admins **after** review is created and this change is propagated throughout (ie, newly added admins can approve; removed admins cannot)

seaerchin:
## Problem
See #1108 for an overview of the problem

Closes #1108 

## Solution
1. remove `stateManager` functionality and use `react-hook-forms` to manage state instead.
2. create new hooks + service api to ping the backend on update admin change.

**Breaking Changes**
- Requires the corresponding backend PR to be merged to work.

### DIFF
--- a/src/layouts/ReviewRequest/components/ManageReviewerModal/ManageReviewerModal.tsx
+++ b/src/layouts/ReviewRequest/components/ManageReviewerModal/ManageReviewerModal.tsx
@@ -11,26 +11,21 @@ import {
 } from "@chakra-ui/react"
 import { Button, ModalCloseButton } from "@opengovsg/design-system-react"
 import _ from "lodash"
-import Select, { PropsValue, useStateManager } from "react-select"
+import { useEffect } from "react"
+import { Controller, FormProvider, useForm } from "react-hook-form"
+import { useParams } from "react-router-dom"
+import Select from "react-select"
+
+import { useUpdateReviewRequest } from "hooks/reviewHooks/useUpdateReviewRequest"
+
+import { getAxiosErrorMessage } from "utils/axios"
 
 import { User } from "types/reviewRequest"
+import { useErrorToast, useSuccessToast } from "utils"
 
 export interface ManageReviewerModalProps extends Omit<ModalProps, "children"> {
   selectedAdmins: User[]
   admins: User[]
-}
-
-// NOTE: This is with reference to
-// https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/types.ts
-// The initial type is PropsValue, which is either a single value or a multi-value.
-// We need to narrow the type ourself to prove to the TS compiler that this is a list.
-const isList = (admins: PropsValue<unknown>): admins is unknown[] => {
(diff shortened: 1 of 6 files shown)


src/layouts/ReviewRequest/components/ManageReviewerModal/ManageReviewerModal.tsx
```diff
@@ -11,34 +11,65 @@ import {
...
+    }
+  }, [isUpdateReviewRequestSuccess, successToast])
+
+  useEffect(() => {
+    if (isUpdateReviewRequestError) {
+      errorToast({
+        description: getAxiosErrorMessage(updateReviewRequestError),
+      })
+    }
+  }, [errorToast, isUpdateReviewRequestError, updateReviewRequestError])
```
alexanderleegs, line 72: Nit: should this be placed in the hook itself instead, via `onError`/`onSuccess`? It doesn't seem to require anything in particular from the component, and the error/success toast should be the same regardless of where it's called

alexanderleegs (commented):
Small nit, but otherwise lgtm

src/layouts/ReviewRequest/components/ManageReviewerModal/ManageReviewerModal.tsx
seaerchin: no real opinion here - i've just always been placing it inside the component cos i think that's where devs expect to see the toast + gives the component/page the ability to give a better error message in the event that it's a 500.

not too attached though; wdyt?

alexanderleegs: No particularly strong feelings either way also - I'm fine with leaving it as is!

seaerchin: will take path of least resistance

src/types/reviewRequest.ts
```diff
@@ -32,10 +34,16 @@ export interface ReviewRequestInfo {
   reviewers: User[]
 }
 
-export interface ReviewRequestDto extends Omit<ReviewRequestInfo, "reviewers"> {
+export interface CreateReviewRequestDto
+  extends Omit<ReviewRequestInfo, "reviewers"> {
   reviewers: string[]
 }
 
+export type UpdateReviewRequestDto = SetOptional<
```
kishore03109, line 42: hmm quick clarification: why is `reviewers` field set to `optional`?

seaerchin: made `reviewers` a required property! thanks for pointing this out!

dcshzj (approved):
LGTM!

PR #1132, opened 2022-10-21
chore(routeselector): add approval redirect on all sub components

seaerchin:
## Problem
when the site has an approved review request, the user should be redirected to `/dashboard` on login. this was only done for `Workspace` and legacy components (in `.js`) and wasn't done for the new pages.

## Solution
add redirect


alexanderleegs (approved):
lgtm!

PR #1133, opened 2022-10-25
Fix/notifications display

alexanderleegs:
This PR fixes an issue with hardcoded values for the comments drawer, as well as updating the correct parameter name to be used. It also refetches data when a comment is submitted.

### DIFF
--- a/src/layouts/ReviewRequest/components/Comments/CommentsDrawer.tsx
+++ b/src/layouts/ReviewRequest/components/Comments/CommentsDrawer.tsx
@@ -79,11 +79,7 @@ export const CommentsDrawer = ({
     onClose: onCommentsClose,
   } = useDisclosure()
 
-  const {
-    data: commentsData,
-    isLoading: isCommentsLoading,
-    refetch: refetchCommentsData,
-  } = useGetComments({
+  const { data: commentsData, isLoading: isCommentsLoading } = useGetComments({
     siteName,
     requestId,
   })
@@ -160,7 +156,6 @@ export const CommentsDrawer = ({
             <SendCommentForm siteName={siteName} requestId={requestId} />
           </Box>
         </DrawerContent>
-        refetchData={refetchCommentsData}
       </Drawer>
     </>
   )
--- a/src/layouts/ReviewRequest/components/Comments/CommentsDrawer.tsx
+++ b/src/layouts/ReviewRequest/components/Comments/CommentsDrawer.tsx
@@ -79,7 +79,11 @@ export const CommentsDrawer = ({
     onClose: onCommentsClose,
   } = useDisclosure()
 
-  const { data: commentsData, isLoading: isCommentsLoading } = useGetComments({
+  const {
+    data: commentsData,
+    isLoading: isCommentsLoading,
+    refetch: refetchCommentsData,
+  } = useGetComments({
     siteName,
     requestId,
   })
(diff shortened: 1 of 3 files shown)


src/layouts/ReviewRequest/components/Comments/CommentsDrawer.tsx
```diff
@@ -156,6 +160,7 @@ export const CommentsDrawer = ({
             <SendCommentForm siteName={siteName} requestId={requestId} />
           </Box>
         </DrawerContent>
+        refetchData={refetchCommentsData}
```
seaerchin, line 163: this seems to have been misplaced instead of on the `SendCommentForm`

src/layouts/ReviewRequest/components/Comments/SendCommentForm.tsx
```diff
@@ -21,13 +22,17 @@ export interface CommentFormProps {
 export const SendCommentForm = ({
   siteName,
   requestId,
-}: CommentProps): JSX.Element => {
+  refetchData,
+}: CommentProps & {
+  refetchData: () => Promise<QueryObserverResult<CommentData[]>>
```
seaerchin, line 27: could we rename this to `getComments`? 

sidenote (not v impt to fix): does using the original type of the service function not work? ie, `() => Promise<CommentData[]>`

src/layouts/ReviewRequest/components/Comments/SendCommentForm.tsx
```diff
@@ -39,6 +44,8 @@ export const SendCommentForm = ({
 
   const handleUpdateNotifications = async ({ comment }: CommentFormProps) => {
     await updateNotifications({ siteName, requestId, message: comment })
+    await refetchData()
```
seaerchin, line 47: actually if we just want the data to be refetched on data update, we could just pass `CommentData` and invalidate its key on data refetch. this might be slightly clearer also

src/layouts/ReviewRequest/components/Comments/CommentsDrawer.tsx
alexanderleegs: Good catch - removed in https://github.com/isomerpages/isomercms-frontend/pull/1133/commits/38bd4010d56b9540ee4799e69569899935692c63!

src/layouts/ReviewRequest/components/Comments/SendCommentForm.tsx
alexanderleegs: Removed in https://github.com/isomerpages/isomercms-frontend/pull/1133/commits/38bd4010d56b9540ee4799e69569899935692c63!

On the sidenote, the original type doesn't actually work for some reason - possibly because the refetch method is from react-query?

src/layouts/ReviewRequest/components/Comments/SendCommentForm.tsx
alexanderleegs: You're right, invalidating the query is much simpler, I've implemented the suggestion in https://github.com/isomerpages/isomercms-frontend/pull/1133/commits/38bd4010d56b9540ee4799e69569899935692c63

seaerchin: should be done in hook but it's fine for now!

PR #1134, opened 2022-10-27
feat(rr): adds request unapproval ability

### LINKED ISSUE #1117: [BUG] Request unapproval
## Problem
Request unapproval does not update state. This means that even if a request was approved and later unapproved, the view will still be approved

## Suggested solution
Ping backend to unapprove it

## DoD
when user unapproves, the PR becomes `In review` and new reviewers see it as unapproved too

seaerchin:
## Problem
See #1117 

Closes #1117

## Solution
1. add hook + api call to backend endpoint
2. minor style fixes so the rr dashboard has a header


prestonlimlianjie (approved):
lgtm

PR #1136, opened 2022-10-27
fix(siteviewheader): conditional url for back button

### LINKED ISSUE #1113: [BUG] View partitioning by role
## Problem
Currently, the cms does not make a distinction between github users and email users. this results in, for example, github users having the ability to see the dashboard, which is something that should not happen.

## Suggested solution
have a context that store whether they are github or email users.

## Definition of done
Github users are unable to see the dashboard
Back button for **email users** should take them to dashboard rather than the sites view

seaerchin:
## Problem
Part of #1113 - this PR makes the `SiteViewHeader` conditionally go back to either the dashboard or sites, depending on whether they are at the dashboard itself.

As `SiteViewHeader` is **only** used at review request related stuff, this header should only be seen by **email** users.

Closes #1113 

## Solution
add conditional for href on back button

### DIFF
--- a/src/layouts/layouts/SiteViewLayout/SiteViewHeader.tsx
+++ b/src/layouts/layouts/SiteViewLayout/SiteViewHeader.tsx
@@ -11,12 +11,14 @@ import { IconButton } from "@opengovsg/design-system-react"
 import { AvatarMenu } from "components/Header/AvatarMenu"
 import { NotificationMenu } from "components/Header/NotificationMenu"
 import { BiArrowBack } from "react-icons/bi"
-import { Link as RouterLink, useParams } from "react-router-dom"
+import { Link as RouterLink, useLocation, useParams } from "react-router-dom"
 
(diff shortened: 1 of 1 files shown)


src/layouts/layouts/SiteViewLayout/SiteViewHeader.tsx
```diff
@@ -29,16 +31,16 @@ export const SiteViewHeader = (): JSX.Element => {
     >
       <HStack spacing="1.25rem">
         <IconButton
-          aria-label="Back to My sites"
+          aria-label="Back to site dashboard"
```
dcshzj, line 34: Should this label also use the same conditional logic in line 43?

seaerchin: yes, good catch!

prestonlimlianjie (approved):
lgtm

PR #1137, opened 2022-10-31
feat: record view status of review requests

dcshzj:
## Problem

Currently, there is no way for the CMS backend to know if a user has viewed a review request or not, as the frontend does not make an additional call to the backend to record the view status after the page is loaded.

## Solution

**Breaking Changes**

**Features**:

- The site dashboard has been enhanced to send a request to the backend to mark all review requests as viewed after the page has been loaded.
- The review request page has been enhanced to send a request to the backend to mark that specific review request as viewed after the page has been loaded.

## Tests

- [x] Smoke tests

1. Run `npm run dev` on this branch.
2. Navigate to the site dashboard of any site.
3. Look for a call made by the browser to the backend similar to `/sites/:siteName/review/viewed`. This verifies that the site dashboard does send a secondary call to the backend after the page is loaded.
4. Navigate to any review request page.
5. Look for a call made by the browser to the backend similar to `/sites/:siteName/review/:requestId/viewed`. This verifies that the review request page does send a secondary call to the backend after the page is loaded.

## Deploy Notes

*None*


src/hooks/siteDashboardHooks/useUpdateViewedReviewRequests.ts
```diff
@@ -0,0 +1,17 @@
...
+export const useUpdateViewedReviewRequests = (): UseMutationResult<
+  void,
+  AxiosError<{ message: string }>,
+  { siteName: string }
+> => {
+  return useMutation<
+    void,
+    AxiosError<{ message: string }>,
+    { siteName: string }
+  >(({ siteName }) => SiteDashboardService.updateViewedReviewRequests(siteName))
```
seaerchin, line 16: hmm why don't we invalidate the cache for the related getter?

src/hooks/reviewHooks/useUpdateReviewRequestViewed.ts
```diff
@@ -0,0 +1,19 @@
+import type { AxiosError } from "axios"
+import { useMutation } from "react-query"
+import type { UseMutationResult } from "react-query"
+
+import * as ReviewService from "services/ReviewService"
+
+export const useUpdateReviewRequestViewed = (): UseMutationResult<
+  void,
+  AxiosError<{ message: string }>,
```
seaerchin, line 9: fyi, not important to fix here, but the error is either `ResponseErrorBody` or, if there are **exceptions**, then `MiddlewareError`.

src/hooks/reviewHooks/useUpdateReviewRequestViewed.ts
```diff
@@ -0,0 +1,19 @@
...
+  { siteName: string; prNumber: number }
+> => {
+  return useMutation<
+    void,
+    AxiosError<{ message: string }>,
+    { siteName: string; prNumber: number }
+  >(({ siteName, prNumber }) =>
+    ReviewService.updateReviewRequestViewed(siteName, prNumber)
```
seaerchin, line 17: could i clarify why we don't need to invalidate the cache and force a refetch of the related getter?

src/hooks/siteDashboardHooks/useUpdateViewedReviewRequests.ts
seaerchin: could i clarify the difference between this and the above hook? is it just a matter of scale? (one above is single rr, this is for all rr)

dcshzj: > hmm why don't we invalidate the cache for the related getter?

I am thinking that the review request should always be shown as new to the user, until the user explicitly refreshes the page or navigates to another page and comes back. I'm not sure if my understanding of the react query cache is correct here, but I think if I were to invalidate the cache here, it would cause the review request to no longer show as new even though the user did not perform any page refreshes, which may be too fast for the user to notice.

> could i clarify the difference between this and the above hook? is it just a matter of scale? (one above is single rr, this is for all rr)

Yup that interpretation is correct.

seaerchin: ah ok that explanation makes sense! thanks for clarifying!

seaerchin (approved):

PR #1139, opened 2022-11-02
fix(header): conditional url for site edit header

seaerchin:
## Problem
See #1136 for a brief overview but for `SiteViewHeader`. In here, we distinguish between github/identity users and we conditionally alter the text/url based on that. 

This is done through checking the `userId` variable, which is set when the user logs in through github. 

## Solution
1. retrieve `userId` and check


dcshzj (approved):
LGTM!

PR #1141, opened 2022-11-03
fix: increase z-index for Menu.List in header items

dcshzj:
## Problem

The `z-index` value is unset for the notifications and avatar menu dropdown, which would cause other items that have a `z-index` value defined greater than 0 to block parts of those dropdown menus.

## Solution

**Breaking Changes**

**Bug Fixes**:

- The `z-index` value for the notification and avatar menu has been set to 10, which is a sufficiently high value to not be covered by any other components.

## Before & After Screenshots

**BEFORE**:

Screenshot 2022-11-03 at 15 36 11

**AFTER**:

Screenshot 2022-11-03 at 15 36 11

## Tests

- [x] Smoke tests

1. Run `npm run dev` on this branch
2. Open any review request page
3. Verify that the review button no longer blocks the notifications menu and the avatar menu

## Deploy Notes

*None*

### DIFF
--- a/src/components/Header/AvatarMenu.tsx
+++ b/src/components/Header/AvatarMenu.tsx
@@ -75,7 +75,7 @@ export const AvatarMenu = ({
           <Menu.List
             role="menu"
             marginTop="0.375rem"
-            zIndex={10}
+            zIndex="sticky"
             {...menuListProps}
           >
             <Box
--- a/src/components/Header/AvatarMenu.tsx
+++ b/src/components/Header/AvatarMenu.tsx
@@ -72,7 +72,12 @@ export const AvatarMenu = ({
               }
             />
           </AvatarMenuButton>
-          <Menu.List role="menu" marginTop="0.375rem" {...menuListProps}>
+          <Menu.List
+            role="menu"
+            marginTop="0.375rem"
+            zIndex={10}
+            {...menuListProps}
+          >
             <Box
               display="flex"
               alignItems="center"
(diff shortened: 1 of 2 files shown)


src/components/Header/AvatarMenu.tsx
```diff
@@ -72,7 +72,12 @@ export const AvatarMenu = ({
               }
             />
           </AvatarMenuButton>
-          <Menu.List role="menu" marginTop="0.375rem" {...menuListProps}>
+          <Menu.List
+            role="menu"
+            marginTop="0.375rem"
+            zIndex={10}
```
kishore03109, line 78: nit - shall we use a constant for this? I was thinking having a seperate file for this where we can define stuff like 
Z-INDEX__FOR_NOTIFICATIONS = 10
Z-INDEX_FOR_ANNOUNCEMENTS = 50 
^ was thinking this would be neater as we know exactly which `level` each components are - rather than us guessing that 10 is sufficiently enough! wdyt?

kishore03109 (approved):
lgtm, but wondering if we can make this clearer as commented!

src/components/Header/AvatarMenu.tsx
seaerchin: Chakra has default hints - eg: `dropdown`, we should use those

dcshzj: Good point! I missed out on Chakra's one. Updated in 870d1ef6.

I am opting to use Chakra's one instead of defining our own set of constants as this is already done for the SiteViewHeader/SiteEditHeader, only changing the values to match the ones defined for those.

PR #1151, opened 2022-11-11
fix(stories): updated stories + handlers

### LINKED ISSUE #1119: [FIX] Review request stories
## Problem 
Some stories are broken as a result of new hooks being aded into the RR PR. We should create handlers for these and link them to `msw` so that the stories will work properly again.

## Suggested solution
add `msw` handlers for hooks

## Definition of Done
stories work properly

seaerchin:
## Problem
see #1119 for overview. 

additionally, the endpoint for `useListCollaboratorsHook` and `useGetCollaborators` was identical and hence, `useGetCollaborators` has been removed. There will be a corresponding backend PR to update the types but this doesn't affect our FE (no validation done -> our FE accepts the runtime type as-is and as the run-time type is a superset (contains more properties), this is safe for now)

Closes #1119 

## Solution
1. update builders + handlers for new endpoints
2. stories that use APIs now have corresponding handlers


seaerchin:
note: will have another PR up to simplify `useListCollaboratorsHook` as the initial implementation on BE simply converts the model to JSON, which means that the type is very complicated

PR #1152, opened 2022-11-11
chore(collaborators): error messages + parsing

seaerchin:
to be reviewed alongside BE PR [here](https://github.com/isomerpages/isomercms-backend/pull/568)

## Problem
Previously, list collaborators uses a default error message which is uninformative to the user. this PR changes the error message to be extracted from the error body first, prior to falling back to a default. 

this also removes unused properties from our `CollaboratorsDto` type and shifts the parsing of `SiteMember.role` -> `role` to the backend

## Solution
1. remove unused properties
2. use axios util and update it to allow default msg


alexanderleegs (approved):
lgtm

seaerchin:
not merging till be approved fyi

PR #1153, opened 2022-11-11
chore(chromatic.yml): update to trigger when files are changed

seaerchin:
## Problem
previously, our frontend workflow for chromatic triggered alot, even when there are unrelated file changes. this led to multiple rebuilds, which led to us busting our limit and designers being unable to review code.

## Solution
1. use [github action](https://github.com/dorny/paths-filter) that detects if there are file changes
    - we only care about changes that affect chromatic and hence, the list of folders are components + styles/theme.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -9,30 +9,47 @@ on:
 
 # List of jobs
 jobs:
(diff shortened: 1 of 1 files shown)


alexanderleegs (approved):
lgtm

PR #1157, opened 2022-11-17
Feat/restrict identity routes

alexanderleegs:
## Problem

Currently, github login users are still able to access email login features due to the lack of restrictions on the backend. This PR introduces the relevant restrictions to prevent github login users from receiving notifications, as well as handling the new behaviour of review request status to prevent github users from being locked out from editing their repo. To be reviewed in conjunction with PR #[573](https://github.com/isomerpages/isomercms-backend/pull/573) on the isomercms-backend repo.

### DIFF
--- a/src/__tests__/RouteSelector.spec.jsx
+++ b/src/__tests__/RouteSelector.spec.jsx
@@ -1,5 +1,6 @@
 import "../tests/TestDecoder.mock"
 import { render, screen } from "@testing-library/react"
+import { QueryClient, QueryClientProvider } from "react-query"
 import { MemoryRouter } from "react-router-dom"
 import "@testing-library/jest-dom/extend-expect"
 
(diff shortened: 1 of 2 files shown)


seaerchin:
oops, i actually have a different commit on `vapt` [here](https://github.com/isomerpages/isomercms-frontend/commit/7c68220fd1738aea1952dd45a41f8cc71d5eb334) that does something similar - i just rendered `children` as is if it's a github user - wdyt about doing that

alexanderleegs:
> oops, i actually have a different commit on `vapt` [here](https://github.com/isomerpages/isomercms-frontend/commit/7c68220fd1738aea1952dd45a41f8cc71d5eb334) that does something similar - i just rendered `children` as is if it's a github user - wdyt about doing that

oops yeah that does make sense! let me just verify that it works with the backend pr and i'll close this one

kishore03109 (approved):

PR #1174, opened 2023-01-04
Fix/frontmatter for link resources

alexanderleegs:
This PR adds the external param to the front matter of link resources. Previously, as permalink is used by jekyll to determine the files to build, any permalinks with # or ? would cause a build error with jekyll, which disallows those files. As such, this PR modifies the external link url to use a different unrelated param. To be reviewed in conjunction with 
https://github.com/isomerpages/isomercms-backend/pull/598 on the backend and https://github.com/isomerpages/isomerpages-template/pull/285 on the isomerpages-template repos.

### DIFF
--- a/src/components/PageSettingsModal/PageSettingsSchema.jsx
+++ b/src/components/PageSettingsModal/PageSettingsSchema.jsx
@@ -36,9 +36,8 @@ export const PageSettingsSchema = (existingTitlesArray = []) =>
     permalink: Yup.string().when("layout", (layout, schema) => {
       switch (layout) {
         case "file":
-          return schema
         case "link":
-          return schema.required("Permalink is required")
+          return schema
         default:
           return schema
             .required("Permalink is required")
@@ -73,5 +72,8 @@ export const PageSettingsSchema = (existingTitlesArray = []) =>
       ),
     file_url: Yup.string().when("layout", (layout, schema) =>
       layout === "file" ? schema.required("File url is required") : schema
+    ),
+    external: Yup.string().when("layout", (layout, schema) =>
+      layout === "link" ? schema.required("Link is required") : schema
     ),
   })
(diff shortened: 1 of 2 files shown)


src/components/PageSettingsModal/ResourcePageSettingsModal.tsx
```diff
@@ -172,16 +173,21 @@ export const ResourcePageSettingsModal = ({
...
-          "permalink",
-          pageData.content.frontMatter.permalink.replace("https://", "")
-        )
+      if (pageData.content.frontMatter.layout === "link") {
+        if (pageData.content.frontMatter.permalink) {
+          // backwards compatible with previous link files
+          setValue(
+            "external",
+            pageData.content.frontMatter.permalink.replace("https://", "")
+          )
```
kishore03109, line 182: after we `setValue`, should we do a `delete processedData.permalink` rather than during the `onSubmit`?

kishore03109 (changes requested):

src/components/PageSettingsModal/PageSettingsSchema.jsx
```diff
@@ -36,9 +36,8 @@ export const PageSettingsSchema = (existingTitlesArray = []) =>
     permalink: Yup.string().when("layout", (layout, schema) => {
       switch (layout) {
         case "file":
-          return schema
         case "link":
-          return schema.required("Permalink is required")
+          return schema
```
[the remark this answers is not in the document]
alexanderleegs: It was removed because we no longer use `Permalink` for `link` type resources - the param that stores the url being linked to is now `external`! So there's no longer any specific schema to check against for this case

src/components/PageSettingsModal/ResourcePageSettingsModal.tsx
alexanderleegs: Hmm `processedData` is only retrieved on submission! The alternative here would be to modify `pageData.content.frontMatter.permalink` directly, but that causes issues because the value is retrieved via react-query - I think it might be better to leave it on the `onSubmit`, since we'll still need to it handle legacy `file` type resources as well. What are your thoughts?

kishore03109 (approved):

PR #1184, opened 2023-02-15
Fix: remove . when no file extension

alexanderleegs:
This PR fixes an issue where newly created files were unable to be uploaded. This issue was due to the shared create/update media modal - we recently disallowed the changing of extension on renaming of existing images, by automatically appending  the extension to the filename of the existing file. However, for newly added items, they had no such extension, and this caused the uploaded file to have `.` appended to it. This PR fixes this behaviour for the creation flow.

This PR also restricts the renaming of newly uploaded images to prevent users from modifying their extension, and changes our duplicate filename check to be case insensitive.

### DIFF
--- a/src/components/MediaSettingsModal/MediaSettingsModal.jsx
+++ b/src/components/MediaSettingsModal/MediaSettingsModal.jsx
@@ -64,6 +64,7 @@ export const MediaSettingsModal = ({
       context: { mediaRoom, isCreate },
     })
 
+  // fileExt is blank for newly created files - mediaData is undefined for the create flow
   const fileExt = getFileExt(mediaData?.name || "")
 
   /** ******************************** */
@@ -86,6 +87,7 @@ export const MediaSettingsModal = ({
     return onProceed({
       data: {
         ...rest,
+        // Period is appended only if fileExt exists, otherwise MediaCreationModal handles the period and extension appending
         name: `${name}${fileExt ? `.${fileExt}` : ""}`,
       },
     })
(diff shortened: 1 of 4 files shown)


harishv7 (dismissed):

src/components/MediaSettingsModal/MediaSettingsModal.jsx
```diff
@@ -86,7 +86,7 @@ export const MediaSettingsModal = ({
     return onProceed({
       data: {
         ...rest,
-        name: `${name}.${fileExt}`,
+        name: `${name}${fileExt ? `.${fileExt}` : ""}`,
```
QiluXie, line 89: nit: Can we add a brief comment to explain why this operation is needed for a better readability of the code? 
After reading this line `const fileExt = getFileExt(mediaData?.name || "")`, I feel `fileExt` would be `""` if the `name` of `mediaData` is missing. What edge case does this line address by manually checking `fileExt` and assign `""`?

QiluXie (commented):
Left a minor comment @alexanderleegs

seaerchin:
lgtm for now but it seems like the root problem is sharing the modal for create/update, based off what you've described

src/components/MediaSettingsModal/MediaSettingsModal.jsx
alexanderleegs: Oh this actually adds a period when fileExt exists - i've added comments in https://github.com/isomerpages/isomercms-frontend/pull/1184/commits/b8b38e5e54164d5519c0cc22bd552f5ed3d23138 to clarify

We'll might want to split up the create and update modals at some point in future, since this behaviour is a little unintuitive

PR #1186, opened 2023-02-16
feat(identity): announcement modal

seaerchin:
## Problem
This PR adds the announcement modal from [figma](https://www.figma.com/file/7OtNYkc4s9BkFk05ConEbd/%E2%AD%90-V1.5-(original-w-comments)?node-id=6440%3A119091&t=pDjVX8I209vWP0GL-1) into 
our codebase.

## Solution
The announcement modal is used to feedback **major**, breaking changes to a user. Hence, the modal needs to be able to tell what to show to the user. 

There were 2 ways of going about this, either backend via a db table or frontend via `localStorage`. The backend necessitated creating a database table to track user views (annoying) + link an actual URL to show the image + provide an endpoint to retrieve the notes but did have the added advantage of having x-browser support, as well as not showing the modal again on `localStorage` being cleared/incognito. 

For the frontend, doing it through `localStorage` was easier and i think the downside of reshowing the modal is negligible, especially as users can easily escape it. Hence, the list of announcements is stored as an array (see `Announcements`) on the frontend. Note that **we have to roll out a frontend update to update the modal.**

To prevent repeated views for the user and to allow for different users on the same browser, the approach taken was to store the last index seen + email as a `Record<email, lastSeenIndex>`. The flow will be briefly laid out below: 
    1. retrieve last seen index using email
    2. if no last seen index, display latest update
    3. on modal close, update last seen index
  
because we return `undefined` when index >> length of announcements, the caller has to not render the modal in that case so step 3 is not triggered.

some components were taken from the pre-existing announcement modal from forms. (`MotionBox` + `ProgressIndicator`)

**New dependencies**:

- `framer-motion` for animation


src/features/AnnouncementModal/AnnouncementModal.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  const handleNextClick = useCallback(() => {
+    if (isLastAnnouncement) {
+      onClose()
+      setLastSeenAnnouncement()
+    }
+
+    setCurrActiveIdx(Math.min(currActiveIdx + 1, announcements.length - 1))
+  }, [
+    announcements.length,
```
kishore03109, line 54: Hey, `announcements.length` is redundant in this dependency array since `isLastAnnouncement`'s hook also conatains `announcements.length` as part of its dependency array

src/types/announcements.ts
```diff
@@ -0,0 +1,14 @@
...
+export type TagVariant = "New Feature"
+
+export type Announcement = {
+  title: string
+  description: string
+  image: (props: React.SVGProps<SVGSVGElement>) => JSX.Element
+  tags: TagVariant[]
+}
+
+export type Announcements = {
```
kishore03109, line 11: Hmm am abit confused here. So each announcement `batch` is a type of `Announcements` and each announcement `batch` has list of possible tabs called `announcements`? 
If this the case, shouldnt `link` be an attribute of `Announcement` rather than `Announcements`? since it could be possible for 2 announcements in a particular announcement batch to have 2 seperate links?

kishore03109 (changes requested):
Hey, I cant see the announcements when I run this locally apart from the storybook. As a quick sanity check, was there any changes you had to make in your code base to view the announcements in your localhost directly rather than just the storybook?

src/features/AnnouncementModal/AnnouncementModal.tsx
```diff
@@ -0,0 +1,122 @@
...
+  link,
+}: AnnouncementModalProps): JSX.Element => {
+  const [currActiveIdx, setCurrActiveIdx] = useState<number>(0)
+  const { setLastSeenAnnouncement } = useAnnouncements()
+  const isLastAnnouncement = useMemo(
+    () => currActiveIdx === announcements.length - 1,
+    [announcements.length, currActiveIdx]
+  )
+
+  const handleNextClick = useCallback(() => {
```
harishv7, line 46: Why do we need useCallback here?

src/features/AnnouncementModal/AnnouncementModal.tsx
```diff
@@ -0,0 +1,122 @@
...
+    setCurrActiveIdx(Math.min(currActiveIdx + 1, announcements.length - 1))
+  }, [
+    announcements.length,
+    currActiveIdx,
+    isLastAnnouncement,
+    onClose,
+    setLastSeenAnnouncement,
+  ])
+
+  const { title, description, image, tags } = announcements[currActiveIdx]
```
harishv7, line 61: might want to defensive check if announcements.length > 0

src/features/AnnouncementModal/AnnouncementModal.tsx
```diff
@@ -0,0 +1,122 @@
...
+        <Image bg="base.canvas.brandLight" pt="4.5rem" as={image} />
+        <ModalHeader>
+          {tags.map((tagVariant) => {
+            switch (tagVariant) {
+              case "New Feature":
+                return <NewFeatureTag />
+              default: {
+                const unimplVariant: never = tagVariant
+                throw new Error(
+                  `Unimplemented tag variant found: ${unimplVariant}`
```
harishv7, line 77: will this break the page?

src/features/AnnouncementModal/Announcements.ts
```diff
@@ -0,0 +1,26 @@
+import { IsomerThumbsUp } from "assets"
+import { IsomerWaitingLine } from "assets/images/IsomerWaitingLine"
+import { Announcements } from "types/announcements"
+
+// TODO: Add link
+export const ANNOUNCEMENTS: Announcements[] = [
```
harishv7, line 6: Should this be in some kind of config/constants folder rather than under features/components

src/features/AnnouncementModal/Announcements.ts
```diff
@@ -0,0 +1,26 @@
...
+    link: "",
+    announcements: [
+      {
+        title: "Control who can edit your website",
+        description:
+          "Isomer now lets you manage your site’s Collaborators, which includes Admins and Contributors. Only Admins can add or remove these.",
+        image: IsomerWaitingLine,
+        tags: ["New Feature"],
+      },
+      {
```
harishv7, line 17: Are we supporting like multiple announcements? Or only 1 announcement at a time?

src/types/announcements.ts
```diff
@@ -0,0 +1,14 @@
+// NOTE: Only `New Feature` for now but we can extend this for eg: `Bugfix` or `Rework`
+export type TagVariant = "New Feature"
+
+export type Announcement = {
```
harishv7, line 4: are all of the following mandatory or can there be optional keys as well

src/features/AnnouncementModal/AnnouncementModal.tsx
seaerchin: this is true but it's auto added in as part of a rule (i think it's `useEffect exhaustive deps` or something to that effect) and it's because react isn't smart enough to figure out transitive deps (A -> B and B -> C therefore A -> C). 

you could try removing and saving, it should be auto added back in

src/types/announcements.ts
seaerchin: to give this in context of figma, 
Screenshot 2023-02-20 at 2 31 23 PM

each `Announcements` is a modal, and each slide on that modal is an `Announcement`; wrt `link`, i clarified w/ @NatMaeTan prior and the link should be tied to each release (ie, each `Announcements`). but yea i'm open to this and in fact i think a better flow would be for each slide to have a link to the relevant section in the patch note and having an ending slide for the release overall that links to the main patch notes

src/features/AnnouncementModal/AnnouncementModal.tsx
seaerchin: so that we avoid excessive re-rendering if any of the deps change.

src/features/AnnouncementModal/AnnouncementModal.tsx
seaerchin: yes but this should be of type `never` and the intention is to catch this at compile time (since you can never assign a concrete value to `nev`r)`

src/features/AnnouncementModal/Announcements.ts
seaerchin: actually i'm open to this also - no strong opinion but i saw the original form one keep it here so that it avoids conflation w/ actual announcement constants (eg: max ann, static isomer guide url etc)

i can change this if you want to!

src/features/AnnouncementModal/Announcements.ts
seaerchin: supposedly multiple which is why it's like this

src/types/announcements.ts
seaerchin: checked w/ nat prior, all the types are mandatory @__@

src/features/AnnouncementModal/AnnouncementModal.tsx
seaerchin: idt we need cos we do `Math.min(currActiveIdx + 1, announcements.length - 1)` on line 52

seaerchin:
> Hey, I cant see the announcements when I run this locally apart from the storybook. As a quick sanity check, was there any changes you had to make in your code base to view the announcements in your localhost directly rather than just the storybook?

it's not used yet so it's not present in localhost as it's also pending some design changes @__@

src/features/AnnouncementModal/AnnouncementModal.tsx
QiluXie: @seaerchin did you mean `Math.max` on line 52?  
If `announcements.length = 0`, `Math.min(currActiveIdx + 1, announcements.length - 1)` is picking the smallest number, which would be -1 in this case. Then on line 61 `announcements[currActiveIdx]` would have index out of range error.

seaerchin: actually i didn't cos the active index should be incremented up until the last item (`arr.length - 1`). i didn't account for this because the `AnnouncementModal` should not even render if the array length is 0 and this is also statically determined by the `ANNOUNCEMENTS` constants. 

wdyt about changing `isOpen = {isOpen && !!announcements.length}` so that we only show the modal if there is somethign to show? 

note that: 
```ts
const empty = []
const res = empty[-1] // res === undefined
```
so it won't crash

src/types/announcements.ts
kishore03109: right, got it. Let's go with design then, and revisit this if we have a use case of `link per slide`

kishore03109: Nit: could we rename `Announcements` to `AnnouncementBatch` ?
just feels `Announcements.announcements` could be confusing :(

src/features/AnnouncementModal/AnnouncementModal.tsx
kishore03109: just to confirm, we dont need to implement the case where there are no tags correct?

kishore03109 (approved):
LGTM with nits.
once design is up, and this is tackled in a separate pr, some documentation of deployment process on notion would be great!

src/features/AnnouncementModal/AnnouncementModal.tsx
QiluXie: @seaerchin Thanks for the explanation. My point was line 52 doesn't prevent `currActiveIdx` to be less than 0 on line 61. With line 52, `currActiveIdx` could still be -1 on line 61. In this case, like you mentioned, line 61 will produce `undefined` and it won't crash, I think it's still better to account for the negative index situation for a better readability of the code.
And yeah, it's good that we only show the modal if there is something to show. It could be `isOpen = {isOpen && announcements.length > 0}` which is a bit better to read, but the effect is the same, so up to you. Thanks!

src/features/AnnouncementModal/AnnouncementModal.tsx
harishv7: is there value add cos of useCallback here though. is a simple handleNextClick.

If there are deep child component dependencies this would help

src/features/AnnouncementModal/AnnouncementModal.tsx
harishv7: wouldn't a console.error suffice?

src/features/AnnouncementModal/Announcements.ts
harishv7: Am okay with either. Personally, I will keep it in a config file separately

src/features/AnnouncementModal/Announcements.ts
harishv7: okay cool

src/types/announcements.ts
harishv7: okay cool

src/features/AnnouncementModal/Announcements.ts
seaerchin: Alright, I'll resolve this!

src/features/AnnouncementModal/AnnouncementModal.tsx
seaerchin: @kishore03109 nope, cos nat say cfm will have tags 

@harishv7 yes but better to break in local dev than for it to be silently ignored @_@

src/features/AnnouncementModal/AnnouncementModal.tsx
seaerchin: Sure, I'll remove it!

src/types/announcements.ts
seaerchin: Yea that's a good idea, will do!

src/features/AnnouncementModal/AnnouncementModal.tsx
QiluXie: BTW, this is from chatGPT (it's about to replace stack overflow lol)
Screenshot 2023-02-24 at 10 46 04 AM

harishv7:
@seaerchin got some minor conflicts to resolve for local storage

alexanderleegs (approved):

PR #1190, opened 2023-03-02
feat: adding gitguardian precommit hook

harishv7:
## Problem

While we do have checks post-commit to identify any secrets are committed, we should also attempt to prevent this even before a commit can occur.

After considering some solutions - https://github.com/awslabs/git-secrets and Git Guardian, we decided to go with GitGuardian since we are already using it.

## Solution

We add Git Guardian as a pre-commit hook which scans commits for any leaked secrets before actually performing the commit.

**Breaking Changes**

There is a breaking change for local development since this commit makes Git Guardian setup somewhat mandatory.

  - Details ...

**New dev dependencies**:

No new dependencies are added to our package. Git Guardian will be installed on the OS-level instead as a brew package.

### DIFF
--- a/.env-example
+++ b/.env-example
@@ -14,3 +14,6 @@ export CYPRESS_TEST_REPO_NAME=''
 export E2E_COMMIT_HASH=bcfe46da1288b3302c5bb5f72c5c58b50574f26c
 export PERSONAL_ACCESS_TOKEN=''
 export USERNAME=''
+
+# GitGuardian
+export GITGUARDIAN_API_KEY=""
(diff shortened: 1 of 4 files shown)


seaerchin (approved):
lgtm; can i just check if we've added the env var into our 1pw?

kishore03109 (changes requested):
~~Chore: Opps I just realised this - we have a 'prepare' command in package.json. I think we should include `brew install gitguardian/tap/ggshield` inside that! (note this change would also be replicated in our backend code)~~

aye opps I thought the commands in `prepare` were done at the os level, but its not!

kishore03109 (approved):

harishv7:
> lgtm; can i just check if we've added the env var into our 1pw?

Yea, @kishore03109 added it

PR #1191, opened 2023-03-02
chore(permalinks): change permalink to lowercase

### LINKED ISSUE #1188: [FIX] Convert URLs to lower case
### Problem
at present, our cms respects casing and saves the URL of a page with its casing intact. Previously, on Netlify, this was all case insensitive, so a URL of `/UPPER` would become `/upper`, which is fine.

however, amplify respects casing, which means that entering `/upper` for a url of `/UPPER` would not work (and be a 404).

### Solution
1. CMS converts all URL to lower case
2. Investigate if Amplify allow redirects from upper -> lower

kishore03109:
## Problem

Currently, netlify links are lower casing. [To move towards Amplify infra](https://www.notion.so/opengov/Netlify-to-Amplify-Migration-01b9baff55ef4aebbe9f472fadf5a096?pvs=4), and ensure backward compatibility, all permalinks are going to be lower cased. 

Closes #1188 

## Solution

Change permalink to lower casing. 

**Breaking Changes**

  - Details ...
- [ X ] No - this PR is backwards compatible

## Tests

Permalinks for pages are now only lower casing. Typing capital casing throws an error.
Previously, creating new pages within folders with CAPS has a default link in CAPS. This is no longer the case. 
Similar behaviour is observed when creating resource pages of type POST Content.

### DIFF
--- a/src/__tests__/utils.jsx
+++ b/src/__tests__/utils.jsx
@@ -63,7 +63,7 @@ describe("Utils test", () => {
         resourceRoomName,
         resourceCategoryName,
       }
-      const examplePermalink = `/${resourceRoomName}/${resourceCategoryName}/permalink`
+      const examplePermalink = `/${resourceRoomName}/${resourceCategoryName}/permalink`.toLowerCase()
 
       expect(getDefaultFrontMatter(params, [])).toMatchObject({
         title: exampleTitle,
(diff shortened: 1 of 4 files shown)


alexanderleegs (approved):
lgtm

PR #1198, opened 2023-03-15
Fix/remove sensitive data from local storage

alexanderleegs:
## Problem

This PR removes the sensitive data (email and phone number) stored in our local storage on the frontend. Instead of retrieving this information from local storage, we now query our backend directly, as this is not an expensive operation.

### DIFF
--- a/src/contexts/LoginContext.tsx
+++ b/src/contexts/LoginContext.tsx
@@ -5,6 +5,7 @@ import {
   useContext,
   PropsWithChildren,
   useCallback,
+  useState,
 } from "react"
 
 import { LOCAL_STORAGE_KEYS } from "constants/localStorage"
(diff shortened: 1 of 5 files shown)


src/contexts/LoginContext.tsx
```diff
@@ -33,45 +34,29 @@ const useLoginContext = (): LoginContextProps => {
...
-  ] = useLocalStorage(LOCAL_STORAGE_KEYS.Email, "Unknown email")
-
   const [, , removeSites] = useLocalStorage(
     LOCAL_STORAGE_KEYS.SitesIsPrivate,
     false
   )
+  const [storedUserId, setStoredUserId] = useState("")
+  const [storedUserContact, setStoredUserContact] = useState("")
+  const [storedUserEmail, setStoredUserEmail] = useState("")
   const verifyLoginAndSetLocalStorage = useCallback(async () => {
```
harishv7, line 44: does this still set local storage? else we should update the function name

src/contexts/LoginContext.tsx
```diff
@@ -33,45 +34,29 @@ const useLoginContext = (): LoginContextProps => {
...
     const { data: loggedInUser } = await axios.get<LoggedInUser>(
```
harishv7, line 45: since we removed the local storage keys, for example in the phone verification flow where the number was pre-filled, will it re-fetch user details from API ?

src/contexts/LoginContext.tsx
```diff
@@ -33,45 +34,29 @@ const useLoginContext = (): LoginContextProps => {
...
-    removeStoredUserEmail,
-  ] = useLocalStorage(LOCAL_STORAGE_KEYS.Email, "Unknown email")
-
   const [, , removeSites] = useLocalStorage(
     LOCAL_STORAGE_KEYS.SitesIsPrivate,
     false
   )
+  const [storedUserId, setStoredUserId] = useState("")
+  const [storedUserContact, setStoredUserContact] = useState("")
+  const [storedUserEmail, setStoredUserEmail] = useState("")
```
harishv7, line 43: do we need to update e2e where we set session defaults etc to come from local storage?

src/contexts/LoginContext.tsx
alexanderleegs: Ah yeah that's a good point, changed in https://github.com/isomerpages/isomercms-frontend/pull/1198/commits/c0295d7bbe88a08417dc0a8078120363b6b0e9d1

src/contexts/LoginContext.tsx
alexanderleegs: Yeah it refetches - the number will still be prefilled for the user

src/contexts/LoginContext.tsx
alexanderleegs: Our e2e test settings to bypass the normal checks are on the backend, and localStorage isn't used for anything on the frontend I believe!

harishv7 (approved):

QiluXie (approved):
lgtm

PR #1199, opened 2023-03-16
Fix/misc identity cleanup

alexanderleegs:
This PR performs miscellaneous cleanup of minor identity-phase-2 issues. It also cherry-picks the changes added in PR #1185 as the width fix for the sites component is used.

- The Pull Request button for email login has been replaced with a button which brings up the review request modal
- The empty sites page has been modified with the appropriate error message instead of a generic contact us message:
Screenshot 2023-03-16 at 10 25 54 AM
Screenshot 2023-03-16 at 10 25 20 AM
- The site preview icon in `My Sites` has been updated:
Screenshot 2023-03-16 at 1 29 03 PM


src/layouts/Sites.jsx
```diff
@@ -23,24 +34,54 @@ const Sites = ({ siteNames }) => {
...
+              <OGPLogoNoText height="30%" />
+            </Flex>
+            <VStack
+              w="full"
+              h="30%"
+              justifyContent="center"
+              alignItems="start"
+              backgroundColor="interaction.mainLight.default"
+              pl="1rem"
+              borderRadius="0px 0px 4px 4px"
```
seaerchin, line 47: wait why is the border radius like this

seaerchin (approved):
lgtm with small comment

seaerchin:
no story for nat to review?

alexanderleegs:
> no story for nat to review?

Oops I just noticed this file was still in js - I've refactored to ts + added storybook!

src/layouts/Sites.tsx
```diff
@@ -0,0 +1,139 @@
...
+                  </Text>
+                </VStack>
+              </Link>
+            </div>
+          </div>
+        ))}
+      </>
+    )
+
+  if (siteNames && siteNames.length === 0)
```
kishore03109, line 70: nit: rather than the pattern
```
if (siteNames && siteNames.length ===0) return showEmptySites`
 if (siteNames && siteNames.length > 0)` return showSites
 return LoadingState
 ```
 
 could we 
```
if (!siteNames) return LoadingState
if (siteNames.length === 0) return return showEmptySites
return showSites
```

Comment (alexanderleegs) [reply src/layouts/Sites.tsx]: Yeah that makes sense, https://github.com/isomerpages/isomercms-frontend/pull/1199/commits/540deb4475e306d6ed172b2b6c6fdfae355c8a38!

File: src/assets/images/IsomerLogoNoText.tsx [hunk 3]
```diff
@@ -0,0 +1,106 @@
+export const IsomerLogoNoText = (
+  props: React.SVGProps<SVGSVGElement>
+): JSX.Element => {
```
harishv7, line 3: can't we have this as `.svg` under assets folder instead of as code?

alexanderleegs: I think the main reason we do this is so that we can pass props to the svg if necessary!

PR #1202, opened 2023-03-17
Feat: Upgrade major deps and to Webpack 5

harishv7:
# Problem

This is a major PR that has some breaking changes. Currently, most of our dependencies are still using old versions. We are using outdated versions of key dependencies like Chakra UI, React, Design System and dev dependencies like Cypress.

# Solution

This PR addresses the issue (https://github.com/isomerpages/isomercms-frontend/issues/1196) by upgrading the dependencies. Along with this we will be upgrading to Webpack 5 which is required by the dependencies. Some of the key changes:

- Chakra from 1.x to 2.5.1
- React from 16 to 18 (required by Chakra 2 as well)
- Design system to 0.0.12
- React-scripts from version 4 to 5
- Update sentry to 7.x
- Update storybook to 6.5.x
- Remove craco from storybook config
- Update of CI workflow files to use latest packages/yarn
Node version used for testing & dev: 18.15.0

# Breaking Changes

- [x] Yes - this PR contains breaking changes
  - Details ...
- [ ] No - this PR is backwards compatible

# Features

Following on from the older PR,

## Chakra Updates

- updating to chakra 2.x removes the isFullWidth prop - this has been update to use w = 100%.
- `Button, Icon Button, Input, NumberInput, Radio, Checkbox, Select#
We removed the isFullWidth prop from these components. Use width="full" or width="100%" instead.`
- Please see [Chakra docs](https://chakra-ui.com/getting-started/migration) for more info.

## React Upgrades

- ReactDOM.render is no longer supported in React 18. We use `createRoot` instead
- No more package-lock.json. We are moving to `yarn.lock` with the use of yarn as package manager
- Reduced reliance on craco. We still need craco for running the `test` command. Otherwise, all other commands will use `react-scripts`
- Hot module reloading, which was previously broken, is now fixed in this PR

## Storybook Updates

When running `npm run build-storybook` it outputs a weird error:

ERR! ModuleNotFoundError: Module not found: Error: Can't resolve '@mdx-js/react' in '/.storybook/foundations/Colours'

```
❯ npm run build-storybook

> isomercms-frontend@0.16.1 build-storybook
> build-storybook -s public

info @storybook/react v6.5.16
info
info => Cleaning outputDir: isomercms-frontend/storybook-static
(node:20070) DeprecationWarning: --static-dir CLI flag is deprecated, see:

https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated---static-dir-cli-flag
[...]

PR #1203, opened 2023-03-17
Chore/run eslint prettier

harishv7:
## Problem

Run eslint and prettier

## Solution

Run `yarn lint-fix` and `yarn format-fix`. Fixes https://github.com/isomerpages/isomercms-frontend/issues/1205

Note: This PR fixes CI issues arising from lint/formatting for https://github.com/isomerpages/isomercms-frontend/pull/1202


.eslintrc.json
```diff
@@ -44,6 +44,9 @@
     "@typescript-eslint/ban-ts-comment": "off",
     "react/prop-types": "warn",
     "no-param-reassign": "warn",
+    "no-use-before-define": "warn",
+    "no-unused-expressions": "warn",
+    "camelcase": "warn",
```
seaerchin, line 49: can i check why these were added

kishore03109 (approved):

.eslintrc.json
harishv7: we have a lot of cases which causes the lint to fail, i had to make them as warns for ci to pass

we should of course fix these issues in a separate PR, as there has been a lot of accumulated lint/format issues

PR #1222, opened 2023-04-03
fix(navBar): API change

kishore03109:
## Problem

Nav Bar wasnt allowing for adding folders/submenus etc.

Closes IS-61

## Solution

`selectInputRef.current.select` was undefined. Not too sure when the API changed, but looking at the [docs](https://react-select.com/props), suggest that `clearValue` should be used, rather than `select.clearValue`

## Tests 
Should be able to create Folders , submenus and single pages properly now.


alexanderleegs (approved):
lgtm

kishore03109:
closing to merge into hotfix branch

PR #1231, opened 2023-04-12
Is 59/fix/disallow files and subfolders starting with restricted characters

alexanderleegs:
## Problem

This PR fixes an issue with missing pages, which were showing up on the CMS but not on the actual site. On further investigation, we found that [jekyll has a list of restricted first characters for files/folders which will not be built](https://jekyllrb.com/docs/structure/) unless explicitly specified via the config file in the include directive. The special characters are (`.`, `~`, `#`, and `_`), all of which are already restricted for various reasons except for `#`. We have opted to restrict the use of `#` as the first character for subcollections and page titles, as this is the quickest way of resolving this issue - other methods we could require significant testing/engineering effort (either writing to the config file if we detect this special character, or decoupling file name and file front matter title).

To be reviewed in conjunction with PR [#699](https://github.com/isomerpages/isomercms-backend/pull/699) on the isomercms backend repo.


seaerchin (approved):
lgtm

kishore03109 (approved):

PR #1237, opened 2023-04-17
fix: images tests and add additional image tests

alexanderleegs:
## Problem

Previously, our images tests were failing due to the new uploading functionalities on the cms - as we moved towards automatically determining the extension of images instead of having them be user determined, the titles our tests were looking for had to change to use the new image format instead.

This PR also adds additional album level tests and image moving tests to our original suite, which was fairly barebones.

### DIFF
--- a/cypress/e2e/images.spec.ts
+++ b/cypress/e2e/images.spec.ts
@@ -3,20 +3,121 @@ import "cypress-file-upload"
 import { TEST_REPO_NAME, Interceptors } from "../fixtures/constants"
 
 describe("Images", () => {
-  const IMAGE_TITLE = "singapore.jpg"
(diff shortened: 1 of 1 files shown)


cypress/e2e/images.spec.ts
```diff
@@ -3,20 +3,121 @@ import "cypress-file-upload"
...
+        .contains(/^Next$/)
+        .click()
+
+      cy.get("button")
+        .contains(/^Skip$/)
+        .click()
+        .should("not.exist")
+        .wait(Interceptors.POST)
+
+      // ASSERTS
```
seaerchin, line 60: y u angry

seaerchin (approved):
lgtm

PR #1238, opened 2023-04-19
[IS-71] chore(sites.spec): fix e2e

seaerchin:
## Problem
e2e were failing cos we were issuing requests to `v2` instead of `v1` now. 

Closes IS-71

## Solution
- Change to use `setupDefaultInterceptors`, which listens on `v2` endpoints. 
- add an assertion that the `Loading sites` text should not be visible prior to attempting to click


alexanderleegs (approved):
lgtm

PR #1258, opened 2023-05-04
Feat/optimise directory hooks

kishore03109:
## Problem
We are hitting the API rate limit.

## Solution
It doesnt make sense to call this expensive call site on window refresh. As a low hanging fruit, not doing that.

## Breaking Changes
User needs to refresh the page in order to get the updated directory. This inaccuracy hit seems reasonable to improve perf.

## Tests
Open network tab
notice `/directory` calls are not called on window focus

## Other testing [to-do]
ensure functionality for collaborators is not broken


kishore03109:
!e2e

PR #1263, opened 2023-05-04
Fix: publish button delay

alexanderleegs:
This PR fixes an issue with the publish button not immediately reflecting the state of the reviewer's action. See video for current behaviour:

https://user-images.githubusercontent.com/22111124/236171845-5196365d-850a-4ea8-881c-e930184bbca4.mov

The issue is that the publish now button is determined by our backend review request state instead of the user's direct action - the behaviour of the button state has been replicated to the publish now button as well.

As a safeguard, to prevent users from being able to publish even if the approval state has failed, an additional guard has been added to the backend in PR #[752](https://github.com/isomerpages/isomercms-backend/pull/752), to be reviewed in conjunction with this PR.

### DIFF
--- a/src/layouts/ReviewRequest/Dashboard.tsx
+++ b/src/layouts/ReviewRequest/Dashboard.tsx
@@ -76,11 +76,11 @@ export const ReviewRequestDashboard = (): JSX.Element => {
     // TODO!
     isError: isMergeError,
   } = useMergeReviewRequest(siteName, prNumber, false)
+  const [isApproved, setIsApproved] = useState<boolean | null>(null)
 
   const { onCopy, hasCopied } = useClipboard(data?.reviewUrl || "")
-
   const reviewStatus = data?.status
-  const isApproved = reviewStatus === ReviewRequestStatus.APPROVED
+
   useEffect(() => {
     if (
       reviewStatus === ReviewRequestStatus.CLOSED ||
@@ -88,7 +88,10 @@ export const ReviewRequestDashboard = (): JSX.Element => {
     ) {
       setRedirectToPage(`/sites/${siteName}/dashboard`)
     }
-  }, [reviewStatus, setRedirectToPage, siteName])
+    if (reviewStatus && isApproved === null) {
+      setIsApproved(reviewStatus === ReviewRequestStatus.APPROVED)
+    }
+  }, [isApproved, reviewStatus, setRedirectToPage, siteName])
 
   useEffect(() => {
     updateReviewRequestViewed({ siteName, prNumber })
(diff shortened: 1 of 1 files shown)


src/layouts/ReviewRequest/Dashboard.tsx
```diff
@@ -238,20 +249,22 @@ const CancelRequestButton = ({
...
+        onClose={() => {
+          setIsApproved(false)
+          onClose()
+        }}
+      />
     </>
   )
 }
 
 // NOTE: Utility component exists to soothe over state management
```
kishore03109, line 263: is this comment still relevant?

src/layouts/ReviewRequest/Dashboard.tsx
```diff
@@ -76,19 +76,22 @@ export const ReviewRequestDashboard = (): JSX.Element => {
...
+
   useEffect(() => {
     if (
       reviewStatus === ReviewRequestStatus.CLOSED ||
       reviewStatus === ReviewRequestStatus.MERGED
     ) {
       setRedirectToPage(`/sites/${siteName}/dashboard`)
     }
-  }, [reviewStatus, setRedirectToPage, siteName])
+    if (reviewStatus && isApproved === null) {
```
kishore03109, line 91: hey I am confused with this code here :( 
why are we setting it to approved when `isApproved` is null? if the idea to set them when we first run this code, this wouldn't play well well with a flow like `approve` -> `disapprove` -> `approve`? shouldnt this then just be ```if(reviewStatus) setIsApproved(reviewStatus === ReviewRequestStatus.APPROVED)```?

alexanderleegs: The idea for this code is to set `isApproved` only on initial load - `isApproved` is set to null originally, and if `reviewStatus` has finished loading, we set `setIsApproved(reviewStatus === ReviewRequestStatus.APPROVED)`. From this point onwards, isApproved is only affected by the user's changes, which are immediate (as opposed to the previous behaviour of having to wait for the query from the backend)

Without the check for the first load, it gets stuck on the initial state of the page and has to rely on new information from the backend in order to change state!

src/layouts/ReviewRequest/Dashboard.tsx
alexanderleegs: Good catch, removed in https://github.com/isomerpages/isomercms-frontend/pull/1263/commits/414b75d87d6d4472ae4fc30ba514356cbe90c3fb

kishore03109 (approved):

PR #1269, opened 2023-05-10
fix: media name regex check

alexanderleegs:
This PR fixes a bug where our regex was not working properly for `.` in media file names. This was due to the change in behaviour where the user can no longer specify the extension of their file when uploading a new file, so the previous check to only run the regex on the portion before the last `.` no longer holds.

### Testing

- [ ] Attempt to create a new image with any name with a `.` (e.g. `blah.picture`), should disable the save button


seaerchin (approved):
lgtm; as part of deployment practice, could you include a simple manual test so that the on-call for the week knows how to verify that this PR is working properly on `staging`?

harishv7 (approved):

PR #1270, opened 2023-05-10
fix(resourceCategorySpec): fix e2e tests

kishore03109:
## Problem

Resource Category Spec e2e are failing

## Solution

Add a trigger for 'title' to re-run validation for the title field. 
Currently blocked by IS-180 due to current bug in production.


seaerchin:
fyi @kishore03109 IS-180 is done alr! can mark this as ready for review if there's nth else to be done

harishv7 (approved):

src/components/PageSettingsModal/ResourcePageSettingsModal.tsx
```diff
@@ -242,6 +243,7 @@ export const ResourcePageSettingsModal = ({
                 onChange={() => {
                   trigger("permalink")
                   trigger("external")
+                  trigger("title")
```
seaerchin, line 246: could i get more context on why a code change is required for e2e to pass?

kishore03109: Since we dont refresh on focus anymore, the title displayed was stale, so we effectively need to trigger a refetch here

seaerchin (approved):

PR #1272, opened 2023-05-15
Refactor/is 75 use v2 endpoints for special pages

alexanderleegs:
This PR refactors a portion of our special page (homepage, nav bar, contact us) editing views - specifically, the `GET` methods of homepage and contact us, and the `UPDATE` method of all special pages have been moved into hooks and modified to use the updated (`/v2`) endpoints.

The end result of this PR is that our update endpoints now all utilise the sanitisation provided by our `/v2` endpoints, and that the homepage and contact us pages no longer need a page reload to reflect new changes.

Note that the `GET` method for nav bar has not been updated yet - the reason for this is that one of the required calls (`/sites/${siteName}/folders/all`) provides functionality (returning all folders/files under all folders) that has not been replicated sufficiently by any of our existing v2 endpoints - the only way to retrieve this information is by querying each collection individually currently. I've opted to leave this as is for the moment - the main objective of this PR is to sanitise our data going into special pages.

### Testing

- [ ] Homepage
  - [ ] Saving after editing works properly, without reloading the page
  - [ ] Input is sanitised
  - [ ] Special exceptions (e.g. iframe embeds of sites like facebook) still work properly
- [ ] Nav bar
  - [ ] Saving after editing works properly
  - [ ] Input is sanitised
- [ ] Contact us
  - [ ] Saving after editing works properly, without reloading the page
  - [ ] Input is sanitised


src/layouts/EditNavBar.jsx
```diff
@@ -738,7 +724,10 @@ const EditNavBar = ({ match }) => {
               )}
               <LoadingButton
                 isDisabled={hasErrors()}
-                onClick={() => saveNavData(siteName, originalNav, links, sha)}
+                onClick={async () => {
+                  await saveNavData({ originalNav, links, sha })
+                  window.location.reload()
+                }}
```
seaerchin, line 730: side note: should we put the `window.location.reload` into the `saveNavData` hook's `onSuccess` property? this makes the invariant (refetch on success b/c of old API) maintained. if this is not done, we might run into saves w/o reloads so the user won't be able to read their own writes. we might also wish to add a `TODO` comment to remove the reload if it's shifted and/or file an issue pertaining to the removal of the `getNavData` and jot down the removal of the `reload` behaviour once the endpoint is migrated.

src/services/ContactUsService.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { apiService } from "./ApiService"
+
+const getContactUsEndpoint = (siteName: string): string => {
+  return `/sites/${siteName}/contactUs`
+}
+
+export const getContactUs = async (siteName: string): Promise<ContactUsDto> => {
+  const endpoint = getContactUsEndpoint(siteName)
+  return apiService.get(endpoint).then((res) => res.data)
```
seaerchin, line 11: just an fyi that for the http methods, we need to hint the return type otherwise it's inferred as `any`:

```suggestion
  return apiService.get<ContactUsDto>(endpoint).then((res) => res.data)
```
```ts
return api
```

seaerchin (commented):
could i just check if all the properties being optional on our types is intended? this seems abit peculiar because it means that we might accept an empty object which has no semantic meaning/information to convey.

also, could we hint the types on our api methods? this is so that those methods are easier to work w/ in the future

src/layouts/EditContactUs.jsx
```diff
@@ -225,43 +225,11 @@ const EditContactUs = ({ match }) => {
         console.log(err)
       }
 
-      try {
-        const contactUsResp = await axios.get(
-          `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/pages/contact-us.md`
```
kishore03109, line 230: Just checking in ya, is this URL no longer needed?

kishore03109 (commented):
Agreed with Chin regarding the optional params for the objects, I am curious as to why this was chosen! 

Also, how are you testing that the inputs are sanitised for the testing plan ya? whats the easiest way one can verify this?

src/layouts/EditContactUs.jsx
```diff
@@ -757,21 +719,11 @@ const EditContactUs = ({ match }) => {
...
-          `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/pages/contact-us.md`,
-          frontMatterParams,
-          {
-            withCredentials: true,
-          }
-        )
+        await saveContactUs({
+          frontMatter: filteredFrontMatter,
+          sha: frontMatterSha,
+        })
```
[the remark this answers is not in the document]
alexanderleegs: This is the format expected by the endpoint currently - we can actually remove it for the homepage since the page body is not useful in this case, but our backend would end up adding it anyway because of the expected format in github. For now I'm in favour of leaving it as is just to get sanitisation working 

Removing `withCredentials` works fine on staging!

alexanderleegs:
> could i just check if all the properties being optional on our types is intended? this seems abit peculiar because it means that we might accept an empty object which has no semantic meaning/information to convey.

Hey yeah for these components, unfortunately it seems like basically every field can actually be left blank - e.g. it's possible to have an infobar section where the user just deletes every field off of it, and we currently remove blank fields entirely, so the entire section is just blank. I've been following our already specified RequestSchema on the backend where possible, or inspecting the output otherwise

src/hooks/directoryHooks/contactUsHooks/useGetContactUsHook.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const useGetContactUsHook = (
+  siteName: string
+): UseQueryResult<ContactUsDto> => {
+  const errorToast = useErrorToast()
+  return useQuery(
+    [GET_CONTACT_US_KEY, siteName],
+    () => ContactUsService.getContactUs(siteName),
+    {
+      onError: () => {
```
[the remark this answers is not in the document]
alexanderleegs: Yeah that makes sense, added the error message in https://github.com/isomerpages/isomercms-frontend/pull/1272/commits/3764cf32685cf53b84c515ee9e951e8471eb1edb

src/hooks/directoryHooks/contactUsHooks/useGetContactUsHook.ts
```diff
@@ -14,9 +15,9 @@ export const useGetContactUsHook = (
     [GET_CONTACT_US_KEY, siteName],
     () => ContactUsService.getContactUs(siteName),
     {
-      onError: () => {
+      onError: (err: AxiosError) => {
         errorToast({
-          description: `Your Contact Us page details could not be retrieved. ${DEFAULT_RETRY_MSG}`,
+          description: `Your Contact Us page details could not be retrieved. ${DEFAULT_RETRY_MSG}. Error: ${err.message}`,
```
seaerchin, line 20: sanity check here - why is this `err.message` but in the other, `err.response?.data.error.message` even when the type is identical?

src/layouts/EditNavBar.jsx
```diff
@@ -724,10 +724,7 @@ const EditNavBar = ({ match }) => {
               )}
               <LoadingButton
                 isDisabled={hasErrors()}
-                onClick={async () => {
-                  await saveNavData({ originalNav, links, sha })
-                  window.location.reload()
-                }}
+                onClick={async () => saveNavData({ originalNav, links, sha })}
```
seaerchin, line 727: should we still `await` here?

seaerchin (commented):
some sanity checks, will approve when done. 

side-note, the axios typings should be done for the rest of the new services introduced also but shouldn't be a blocker, so feel free to raise that as another PR

seaerchin (commented):
some sanity checks, will approve when done. 

side-note, the axios typings should be done for the rest of the new services introduced also but shouldn't be a blocker, so feel free to raise that as another PR

src/services/ContactUsService.ts
seaerchin: unresolving for visibility that if we wanna do this, please do so for the other services as well! no need to do here but please raise another PR for the 3 new services introduced

src/layouts/EditNavBar.jsx
alexanderleegs: There's no need for this because it's effectively 
```
async () => {
  return saveNavData(...)
}
```

src/hooks/directoryHooks/contactUsHooks/useGetContactUsHook.ts
alexanderleegs: Oops i changed the file but didn't save it - fixed in https://github.com/isomerpages/isomercms-frontend/pull/1272/commits/17904cef5b32809997235e5929b468af114b1ce4

seaerchin: actually i realised there's a `getAxiosErrorMessage` util which does this - we can do this in a subsequent PR or this one.

seaerchin (approved):
lgtm, good to haves like adding types on axios calls and using the util method can be done as a follow up

seaerchin (approved):
lgtm, good to haves like adding types on axios calls and using the util method can be done as a follow up

PR #1273, opened 2023-05-15
Is 157 view request link does not bring user to the view request page

harishv7:
## Problem

The "Request Review" button should be disabled if there is an already existing PR in "Open" status

Closes IS-157

## Solution

Fetch and filter the current list of PRs and if there exists a PR in Open state, we disable the button

**Breaking Changes**

  - Details ...

## Test Plan

You should be able to click on Review Request iff there are no open PRs. Else, the button should be disabled

### DIFF
--- a/src/layouts/layouts/SiteEditLayout/SiteEditHeader.tsx
+++ b/src/layouts/layouts/SiteEditLayout/SiteEditHeader.tsx
@@ -49,13 +49,9 @@ export const SiteEditHeader = (): JSX.Element => {
     isLoading: isReviewRequestsLoading,
   } = useGetReviewRequests(siteName)
 
-  console.log(reviewRequests)
-
   const openReviewRequests = reviewRequests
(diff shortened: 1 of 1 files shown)


seaerchin (approved):
lgtm, wondering if we should extend the capability of the endpoint (eg: accept query params). 

also, if this use case is common, we might wanna consider adding a `useHasOpenReviewRequest(siteName)` hook in the future. this would wrap around this `useGetReviewRequests` hook and use the same underlying axios cache

kishore03109 (approved):

harishv7:
> lgtm, wondering if we should extend the capability of the endpoint (eg: accept query params).
> 
> also, if this use case is common, we might wanna consider adding a `useHasOpenReviewRequest(siteName)` hook in the future. this would wrap around this `useGetReviewRequests` hook and use the same underlying axios cache

So far this was the only place. But agreed, if we have more areas for this check, lets have a separate hook

PR #1280, opened 2023-05-19
fix: file name in test

alexanderleegs:
## Problem

This PR fixes our files test spec - it currently works on its own but not after being run sequentially with edit page, due to the partial match of the file being uploaded in the `editpage` spec (`singapore-pages` in `editpage`, `singapore` in `files`). The file being searched for in `files` has been changed to `singapore-file` in order to avoid this overlap.

### DIFF
--- a/cypress/e2e/files.spec.ts
+++ b/cypress/e2e/files.spec.ts
@@ -2,7 +2,7 @@ import "cypress-file-upload"
 import { Interceptors, TEST_REPO_NAME } from "../fixtures/constants"
 
 describe("Files", () => {
-  const FILE_TITLE = "singapore"
+  const FILE_TITLE = "singapore-file"
   const OTHER_FILE_TITLE = "america"
   const EXISTING_FILE_TITLE = "New Uber BrandSystem QuickGuide"
   const TEST_FILE_PATH = "files/singapore.pdf"


cypress/e2e/files.spec.ts
```diff
@@ -2,7 +2,7 @@ import "cypress-file-upload"
 import { Interceptors, TEST_REPO_NAME } from "../fixtures/constants"
 
 describe("Files", () => {
-  const FILE_TITLE = "singapore"
+  const FILE_TITLE = "singapore-file"
```
seaerchin, line 5: seems liek the same file might be reused across `editPage` and `files`, maybe we should consider having both of the specs refer to the same constant by exporting it from our fixtures file? 

this means that a change to filename in `editPage` might cause a downstream failure still

alexanderleegs: We actually want the files referred to be different - the tests were made independently of each other, and the issue is that running them sequentially mistakenly referred to the file created in `editPage` which had a similar title, but was not meant to be the same file

seaerchin: oh ok i see!

seaerchin (approved):

kishore03109 (approved):

PR #1282, opened 2023-05-24
fix: facebook-pixel expected type

alexanderleegs:
This PR fixes an issue where sites with facebook-pixel were unable to save their settings, due to a mismatch of expected type (number vs string). To be reviewed in conjunction with PR #[776](https://github.com/isomerpages/isomercms-backend/pull/776) on the isomercms backend repo, which has a more detailed writeup on the issue.

The change being introduced in the frontend is the facebook-pixel field enforcing that the length is 15.

### Testing

- [ ] Manually edit a site's facebook-pixel value to be missing
  - [ ] Should be able to save any change in settings
- [ ] Manually edit a site's facebook-pixel value to be an empty string
  - [ ] Should be able to save any change in settings
- [ ] Manually edit a site's facebook-pixel value to be a string (e.g. `facebook-pixel: "12345"`)
  - [ ] Should be able to save any change in settings
- [ ] Should be able to modify facebook-pixel value and have it show up as a number
- [ ] Should be able to delete facebook-pixel and have it show up as null

### DIFF
--- a/src/layouts/Settings/AnalyticsSettings.tsx
+++ b/src/layouts/Settings/AnalyticsSettings.tsx
@@ -37,15 +37,13 @@ export const AnalyticsSettings = ({
           <FormLabel>Facebook Pixel</FormLabel>
           <Input
             w="100%"
-            type="number"
             {...register("pixel", {
-              valueAsNumber: true,
-              min: {
-                value: 100000000000000,
+              minLength: {
+                value: 15,
                 message: "Your Facebook Pixel id should be a 15 digit number.",
               },
-              max: {
-                value: 999999999999999,
+              maxLength: {
+                value: 15,
                 message: "Your Facebook Pixel id should be a 15 digit number.",
               },
             })}
--- a/src/layouts/Settings/AnalyticsSettings.tsx
+++ b/src/layouts/Settings/AnalyticsSettings.tsx
@@ -29,7 +29,11 @@ export const AnalyticsSettings = ({
       <VStack spacing="1.5rem" align="flex-start" w="50%">
         <FormControl isDisabled={isError}>
           <FormLabel>Facebook Pixel</FormLabel>
-          <Input w="100%" {...register("pixel")} />
+          <Input
+            w="100%"
+            type="number"
+            {...register("pixel", { valueAsNumber: true })}
+          />
         </FormControl>
 
         <FormControl isDisabled={isError}>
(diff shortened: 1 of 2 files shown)


src/layouts/Settings/AnalyticsSettings.tsx
```diff
@@ -29,7 +29,11 @@ export const AnalyticsSettings = ({
       <VStack spacing="1.5rem" align="flex-start" w="50%">
         <FormControl isDisabled={isError}>
           <FormLabel>Facebook Pixel</FormLabel>
-          <Input w="100%" {...register("pixel")} />
+          <Input
+            w="100%"
+            type="number"
+            {...register("pixel", { valueAsNumber: true })}
+          />
```
seaerchin, line 36: nit: we might want to validate this to be of length 15.

seaerchin (commented):
how are we handling `null` and `NaN` values?

seaerchin (commented):
how are we handling `null` and `NaN` values?

src/layouts/Settings/AnalyticsSettings.tsx
kishore03109: sanity check: what happens to existing values that are of type string ah?

kishore03109 (approved):
Sanity check here: regarding the test, by manually edit, you do mean going to their github page and editing the config.yaml right? If not please dismiss this review!

src/layouts/Settings/AnalyticsSettings.tsx
```diff
@@ -27,9 +33,24 @@ export const AnalyticsSettings = ({
...
       </VStack>
       <VStack spacing="1.5rem" align="flex-start" w="50%">
-        <FormControl isDisabled={isError}>
+        <FormControl isDisabled={isError} isInvalid={isInvalid("pixel")}>
           <FormLabel>Facebook Pixel</FormLabel>
-          <Input w="100%" {...register("pixel")} />
+          <Input
+            w="100%"
+            {...register("pixel", {
+              minLength: {
```
kishore03109, line 41: Could we also add a regex pattern to ensure that this only accepts digits and not just any string?

alexanderleegs: Good catch, changed in https://github.com/isomerpages/isomercms-frontend/pull/1282/commits/e49d24fec203b33140aeb69ed31e605cbe8e3637

kishore03109 (approved):

PR #1290, opened 2023-06-01
feat(featureFlag): Site launch implementation

kishore03109:
## Problem

We want to trial out the site launch features with a subset of users. 

This PR aims to get early feedback on a hacky approach that the upcoming site launch PRs aims to implement this. 

## Solution

We store the repos that we intend to trial this on as an env var in the FE separated by commas, and this state is held only in the FE.  

Benefits: 
+ easy to implement and subsequently remove
+ don't have to redeploy to add agencies that want to trial this out, just need change the env var

Limitations: 
- super hacky 
- technically backend APIs exposed to all users (but not really a concern since its an authenticated route anyways)

Ideally, we would leverage on a platform like Darkly to implement feature flags like this. While that is still a possible in the future, am opting for a more brute-force quicker solution to gather feedback from our users early.

## Reviewer notes
Specifically seeking feedback on
- Cleaner, easy to implement solutions 
- Strong concerns against using this approach in the short term 

- `env var` : SITE_LAUNCH_FEATURE_WHITELISTED_REPOS

### DIFF
--- a/src/services/SiteLaunchService.ts
+++ b/src/services/SiteLaunchService.ts
@@ -1,0 +1,7 @@
+export const isUserUsingSiteLaunchFeature = (siteName: string): boolean => {
+  // whitelisting sites that we want to trial this feature with
+  const whiteListedRepos = process.env.SITE_LAUNCH_FEATURE_WHITELISTED_REPOS?.split(
+    ","
+  )
+  return whiteListedRepos ? whiteListedRepos.includes(siteName) : false
+}


alexanderleegs (commented):
Will the endpoint on the backend have checks as well? It's possible for users to directly query the endpoint instead if the only check is on FE

I don't think this is a huge issue though, since anyone trying to do this still needs to be an isomer user, but just something to consider

kishore03109:
@alexanderleegs 
> Will the endpoint on the backend have checks as well? It's possible for users to directly query the endpoint instead if the only check is on FE

nope, I was thinking of only holding the state purely on the FE because its easier to then just change one env var. (else it will an env var in the BE, and might need to worry about inconsistent env vars)
while your concern is valid, I am assuming the risk here is low since it is auth-ed routes anyways, wdyt?

alexanderleegs:
> @alexanderleegs
> 
> > Will the endpoint on the backend have checks as well? It's possible for users to directly query the endpoint instead if the only check is on FE
> 
> nope, I was thinking of only holding the state purely on the FE because its easier to then just change one env var + the time taken to update FE's env var < time taken to update BE env var. (else it will an env var in the BE, and might need to worry about inconsistent env vars) while your concern is valid, I am assuming the risk here is low since it is auth-ed routes anyways, wdyt?

I'm okay with this, since this is a trial release - can't think of any other simple ways to implement

alexanderleegs (approved):

src/services/SiteLaunchService.ts
```diff
@@ -0,0 +1,7 @@
+export const isUserUsingSiteLaunchFeature = (siteName: string): boolean => {
```
seaerchin, line 1: hmm, i think a better way would be to just expose a `/features` endpoint on our backend and ask that the frontend queries this to see if we should render. 

something of this form could be like (on the fe): 
```ts
<FeatureFlag shouldRender = {...}>
  {our components}
</FeatureFlag>
```

and on the backend, we just expose a k/v mapping of features (`showSiteLaunch: true` or `siteLaunchRepos: [...]`)

this might be too heavy for this use case but 1 benefit is that we control the feature roll out (it's backend driven) whereas here, we have to rely on the user hard-refreshing + the browser refetching the data. 

Also pls do file a ticket for removal/followup for this feature flag!

harishv7: well agreed, we will be moving towards that direction with the introduction of launchdarkly, will be working on that in coming sprints, where we will fetch via sdk.

 i think in the time being, for this use case am fine with the hacky approach. wdyt?

kishore03109: I actually didnt think of that, thanks for the suggestion @seaerchin! Since we will be moving towards that with the introduction of launchdarkly anyways, we can 'unhackify' this approach then!

kishore03109: [Issue filed](https://isomeropengov.atlassian.net/browse/IS-238?atlOrigin=eyJpIjoiZWQ0YjVkNWY1YmJhNGNkYWI5MDBlOGY0YzVkN2ZhYjQiLCJwIjoiaiJ9)

harishv7 (approved):

src/services/SiteLaunchService.ts
seaerchin: i think as long as 

1. we won't need to roll this back (ie, all future rollouts must be back compat with this) 
2. we have a clear deprecation plan (on the issue) 

this is ok

PR #1291, opened 2023-06-01
Fix: trim whitespace

alexanderleegs:
This PR fixes an issue where having whitespace when adding collaborators would cause the adding to fail.


seaerchin (approved):

kishore03109 (approved):
Sanity check here: are there other instances of email input that we need to trim too?

alexanderleegs:
> Sanity check here: are there other instances of email input that we need to trim too?

The other input is for login and is already trimmed!

PR #1295, opened 2023-06-06
test(e2e): add initial collaborators e2e tests

seaerchin:
**note: this requires the corresponding [be](https://github.com/isomerpages/isomercms-backend/pull/782) PR**

## Problem
current e2e tests only cover the github flow. this PR adds in collaborator tests that cover the email login flow.

## Solution
1. add in spec for collab tests

- `CYPRESS_BACKEND_URL` : backend url of our website, required to make the request to create a user

## Tests
- [ ] **first, add the e2e email repo to your db** (required b/c e2e email users only have access to this/gh repo)
- [ ] run `npm run cypress:open` 
- [ ] click on the collaborators spec

### DIFF
--- a/cypress/e2e/collaborators.spec.ts
+++ b/cypress/e2e/collaborators.spec.ts
@@ -3,6 +3,7 @@ import {
   E2E_EMAIL_COLLAB,
   E2E_EMAIL_TEST_SITE,
   Interceptors,
+  TEST_REPO_NAME,
 } from "../fixtures/constants"
 import { USER_TYPES } from "../fixtures/users"
 
@@ -44,7 +45,7 @@ const getCollaboratorsModal = () => {
   return cy.get("form").should("be.visible")
 }
 
-const addCollaboratorsFor = (user: string) => {
+const inputCollaborators = (user: string) => {
   getCollaboratorsModal().get(ADD_COLLABORATOR_INPUT_SELECTOR).type(user).blur()
   // NOTE: need to ignore the 422 w/ specific error message because we haven't ack yet
   cy.contains("Add collaborator").click().wait(Interceptors.POST)
@@ -77,14 +78,14 @@ describe("collaborators flow", () => {
     })
     it("should not be able to add a non-whitelisted collaborator", () => {
       // Act
-      addCollaboratorsFor("some_gibberish@gmail.com")
+      inputCollaborators("some_gibberish@gmail.com")
 
       // Assert
       cy.contains(ADD_COLLABORATOR_ERROR_MESSAGE).should("be.visible")
     })
     it("should not be able to add an existing user", () => {
       // Act
-      addCollaboratorsFor(E2E_EMAIL_ADMIN.email)
+      inputCollaborators(E2E_EMAIL_ADMIN.email)
       ignoreDuplicateError()
 
       // Assert
@@ -93,7 +94,7 @@ describe("collaborators flow", () => {
     it("should not be able to add a collaborator without an isomer account", () => {
       // Act
       // NOTE: Initial admin will always be added manually
-      addCollaboratorsFor("gibberish@nonsense.gov.sg")
+      inputCollaborators("gibberish@nonsense.gov.sg")
       ignoreNotFoundError()
 
       // Assert
@@ -111,7 +112,7 @@ describe("collaborators flow", () => {
 
       // Act
       // NOTE: Initial admin will always be added manually
-      addCollaboratorsFor(collaborator)
+      inputCollaborators(collaborator)
       // NOTE: Cannot proceed without acknowledgement
       cy.contains("Continue").should("be.disabled")
       cy.get("input[name='isAcknowledged']").next().click()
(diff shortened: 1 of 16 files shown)


cypress/e2e/collaborators.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+const DUPLICATE_COLLABORATOR_ERROR_MESSAGE =
+  "User is already a member of the site"
+
+const NON_EXISTING_USER_ERROR_MESSAGE =
+  "This user does not have an Isomer account. Ask them to log in to Isomer and try adding them again."
+
+const ADD_COLLABORATOR_INPUT_SELECTOR = "input[name='newCollaboratorEmail']"
+
+const visitE2eEmailTestRepo = () => {
+  cy.visit(`http://localhost:3000/sites/${E2E_EMAIL_TEST_SITE.repo}/dashboard`)
```
harishv7, line 33: to replace localhost with env

cypress/e2e/collaborators.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+      )
+
+      // Act
+      // NOTE: Initial admin will always be added manually
+      inputCollaborators(collaborator)
+      // NOTE: Cannot proceed without acknowledgement
+      cy.contains("Continue").should("be.disabled")
+      cy.get("input[name='isAcknowledged']").next().click()
+      cy.contains("Continue").click().wait(Interceptors.POST)
+      ignoreAcknowledgementError()
```
harishv7, line 120: why do we need this here btw?

cypress/e2e/collaborators.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+          }
+        })
+
+      // Assert
+      cy.get('button[id^="delete-"]').should("be.disabled")
+    })
+
+    it("should prevent admins of a site from removing collaborators of another site", () => {
+      // Arrange
+      cy.visit(`http://localhost:3000/sites/${TEST_REPO_NAME}/dashboard`)
```
harishv7, line 171: localhost

cypress/support/session.ts
```diff
@@ -4,10 +4,49 @@ import {
...
+    cy.setCookie(E2E_COOKIE.Auth.key, E2E_COOKIE.Auth.value)
+    cy.setCookie(E2E_COOKIE.EmailUserType.key, userType)
+    cy.setCookie(E2E_COOKIE.Email.key, email)
+    cy.request("GET", `${Cypress.env("BACKEND_URL")}/auth/whoami`).wait(
+      Interceptors.GET
+    )
+
+    cy.setEmailSessionDefaults(initialUserType)
+    cy.visit(
+      `http://localhost:3000/sites/${E2E_EMAIL_TEST_SITE.repo}/dashboard`
```
harishv7, line 49: localhost?

cypress/e2e/collaborators.spec.ts
seaerchin: ps didn't write this down in the comments - this is because the error message is gotten from the _backend_. concretely, this means that we receive a network error, fe extracts the error message and shows it to the user. because of that, we need to ignore network errors so that cypress doesn't fail

cypress/e2e/collaborators.spec.ts
seaerchin: good catch, fixed!

cypress/e2e/collaborators.spec.ts
seaerchin: added a comment!

kishore03109 (commented):
While code wise looks ok, am quite confused about what the env var `CYPRESS_BACKEND_URL` and why it is needed for this PR. Is this env var needed for future PRs?

harishv7 (approved):
Lgtm, tests pass successfully on local

Screenshot 2023-06-08 at 4 36 16 PM

seaerchin:
> While code wise looks ok, am quite confused about what the env var `CYPRESS_BACKEND_URL` and why it is needed for this PR. Is this env var needed for future PRs?

good question! the reason why we have this here is to allow for cypress to hit our backend programmatically.

concretely speaking, we want our tests to be: 
1. isolated
2. repeatable (as a result of ^)

one problem that we have at present (as a result of not fulfilling ^) is that we have _cascading_ failures - a prior test failure leads to downstream tests all failing.

one way we could solve this is by cleaning up server state between test suites or by doing the test spec that mutates server state last (then cos all the earlier tests are idempotent, we can rerun them in isolation anyw)

in order to clean up server state, we could either do a user flow fully or issue a request to our backend. while the user flow way is good, we should **only do it for the test spec that requires it** - ie, if i wanted to, for example, rename a media file as part of another test suite, i should **not** do it via the UI if it's part of another test suite, since the correctness checking is already done by the `mediaFiles.spec.ts` (for eg) and this is costly + flaky.

so we let cypress know of our backend url, perform the pre-requisites for our test suite (or clean up server state) via API requests in order to make our tests faster + more reliable (no longer depend on DOM for test pre-req/cleanup)

kishore03109:
@seaerchin 
> good question! the reason why we have this here is to allow for cypress to hit our backend programmatically.

Correct me if I am wrong here, but this is to allow cypress to _mock_ our backend responses, ie it acts as an interceptor to allow for more comprehensive _isolated_ front end testing? Are we going to mock all backend responses? If yes, while I get the rationale for enabling for faster, more reliable testing, but my concern here is that we are losing the benefits of E2E testing as this no longer tests the functionality of the backend. 

I think a more ideal approach would be critical (esp security access related flows like adding/deleting collaborators) flows should still rely on the backend calls while non critical flows like say rename some page could be something that we mock.
Thoughts? 

> (also am open to changing the backend url to be generated in code also, just that our env naming isn't standardised so this is abit annoying lmao)

hahha no opinions on this, you can make the call! But a sanity check here, this var isnt referenced in code as of yet right? My assumption here is that this is going to be used for future PRs?

seaerchin:
> @seaerchin
> 
> > good question! the reason why we have this here is to allow for cypress to hit our backend programmatically.
> 
> Correct me if I am wrong here, but this is to allow cypress to _mock_ our backend responses, ie it acts as an interceptor to allow for more comprehensive _isolated_ front end testing? Are we going to mock all backend responses? If yes, while I get the rationale for enabling for faster, more reliable testing, but my concern here is that we are losing the benefits of E2E testing as this no longer tests the functionality of the backend.
> 
> I think a more ideal approach would be critical (esp security access related flows like adding/deleting collaborators) flows should still rely on the backend calls while non critical flows like say rename some page could be something that we mock. Thoughts?
> 
> > (also am open to changing the backend url to be generated in code also, just that our env naming isn't standardised so this is abit annoying lmao)
> 
> hahha no opinions on this, you can make the call! But a sanity check here, this var isnt referenced in code as of yet right? My assumption here is that this is going to be used for future PRs?

This actually hits the backend, there's no mocking being done. This is basically an `axios` call out to our backend that cypress is explicitly aware of. 

You can verify this yourself also by adding a `cy.request` call in our existing tests and have it hit `/ping`. 

This is faster than doing the user flow because we don't have to wait on dom stabilization + it's only one call

kishore03109 (approved):
@seaerchin 
Ok, got it. Concretely, this would be the local BE url, ie lcoalhost:8081 right? 
[Nit]  is this something that we can already get from the pre-existing env var REACT_APP_BACKEND_URL_V2 instead?

seaerchin:
> @seaerchin Ok, got it. Concretely, this would be the local BE url, ie lcoalhost:8081 right? [Nit] is this something that we can already get from the pre-existing env var REACT_APP_BACKEND_URL_V2 instead?

this would be the local BE url **most of the time**. if, for example, we wanted to verify failing e2e tests on a deploy to master, we could do this via changing the url (assuming that we're testing on the same frontend) - which violates that assumption. 

we **can't** actually get env vars from `process.env` directly. it _looks_ like we can, but that's a trick that the application uses - your frontend is not running on your terminal directly (there's **no process**) but instead, it's a subprocess (might be wrong here) of your chosen browser. hence, how frontend apps get your env vars is through injection at **build** time. 

to verify this, you could simply `console.log(process.env)` and see that it's empty. i've also attached a screenshot showing the cypress runner process: it transpiles from ts -> js, then uses a runner to execute the built js. hence, we need a separate env var with `CYPRESS_` prefixed (see [here](https://docs.cypress.io/guides/guides/environment-variables#Option-3-CYPRESS_)) 
Screenshot 2023-06-09 at 2 07 06 PM

kishore03109:
woah ok makes sense, I learnt something new today, thanks sir @seaerchin

PR #1296, opened 2023-06-08
feat(site launch): site dashboard stories

kishore03109:
## Problem

Figma screens for sitelaunch pad only for the dashboard. 

## Solution

The dashboard storybook will now have 3 more states [[1](https://www.chromatic.com/component?appId=61fcc8e14c3af9003a102ac9&csfId=pages-sitedashboard--site-launch-steps-completed&buildNumber=1667&k=648183b79a212ceb30b514ff-1200-interactive-true&h=9&b=-6), [2](61fcc8e14c3af9003a102ac9), [3](https://www.chromatic.com/component?appId=61fcc8e14c3af9003a102ac9&csfId=pages-sitedashboard--site-launch-partial-steps-completed&buildNumber=1667&k=648183b79a212ceb30b51501-1200-interactive-true&h=11&b=-8)] to represent the various states of site lunch that the user will see/interact with. 

This PR also introduces the use of a context. It might seem like an over kill for this particular PR, but the reason for this is to share state across a different page (`/siteLaunchPad`) in the upcoming PRs. 

Note: merging directly to dev since this feature hides behind a feature flag + no site is currently using the feature. 
## Reviewer Notes 
To allow for storybook to change the context info, I had to expose an initial value for the `siteLaunchStatusProps` [here](https://github.com/isomerpages/isomercms-frontend/pull/1296/files#diff-35c1d0b117128f7b6e5c5689991a7846a4bdaa6fe765374132c9799ed71bf705R31-R32). I am not sure if this is the best way to tackle this. 

  Not breaking per se, just as a word of caution, the padding for the display cards have been adjusted. It seems that this is only being used within the dashboard, so visual breaking changes.

## Limitations
Since we are storing the intermediate states in the context for now, if the user refreshes the page, this would lead to the their checklist progress being lost. This is an intentional engineering decision as a first cut, can we can tackle storing the state properly in the backend in the future as we iterate + find the need to. 

## Screenshots
image

## Deploy Notes 
Realised that the env var is not 'sourced' by react since there is no prefix. Modifying this such that there is a prefix now. NOTE: _This is a rename of the env var, not creation of one_

REACT_APP_SITE_LAUNCH_FEATURE_WHITELISTED_REPOS: list of repos to whitelist for this feature 

## Designer notes
 @NatMaeTan 
The font now abit ugly cuz of the padding and the margins 
Screenshot 2023-06-14 at 5 25 47 PM

need your help to make a decision on
1) change the padding and margin for all the display cards 
2) (dont think this ideal) just for this card Ill change the margin
3) do you want me to change the padding left as well for the display card

### Changes made after discussion with Design
1. Icons have changed to 20 X 20 
2. Display card from 16 X 16 become 24 X 24 left and right then top and bottom 16 X 16 
3. Make words less squished for site launch display cards

Screenshot 2023-06-19 at 8 24 29 AM
Screenshot 2023-06-19 at 8 24 24 AM

Changes in padding:
Screenshot 2023-06-19 at 8 34 19 AM


src/contexts/SiteLaunchContext.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+import {
+  SiteLaunchDto,
+  SiteLaunchFrontEndStatus,
+  SiteLaunchStatusProps,
+  SITE_LAUNCH_TASKS_LENGTH,
+} from "types/siteLaunch"
+
+interface SiteLaunchContextProps {
+  setSiteName: (siteName: string) => void
```
alexanderleegs, line 13: Do we need a hook for this? Will we be unable to get it from `useParams<{ siteName: string }>()`?

alexanderleegs (commented):
> Since we are storing the intermediate states in the context for now, if the user refreshes the page, this would lead to the their checklist progress being lost. This is an intentional engineering decision as a first cut, can we can tackle storing the state properly in the backend in the future as we iterate + find the need to.

Just checking - can we recover from this state in our backend? Or are we expecting users to always go through the flow at one shot?

kishore03109:
@alexanderleegs 
Curent flow is 
1) user comes to launch pad
2) checks all the pre-req
3) generates DNS 
4) puts in the generated DNS in their own nameservers
5) Launches site

For steps 1-2, this is stored purely in the frontend as a first cut
everything from step 3 onwards will be stored in the backend

src/types/siteLaunch.ts
```diff
@@ -0,0 +1,46 @@
+export interface DNSRecord {
+  source: string
+  type: string
+  target: string
+}
+
+export type SiteLaunchFrontEndStatus =
+  | "LAUNCHED"
```
harishv7, line 8: nit: Can expose as object first?

```
const SiteLaunchStatus = {
  Launched: "LAUNCHED",
  NotLaunched: "NOT_LAUNCHED",
  Launching: "LAUNCHING"
ChecklistTasksPending: "CHECKLIST_TASKS_PENDING
} as const;

type SiteLaunchFrontEndStatus = typeof SiteLaunchStatus[keyof typeof SiteLaunchStatus]
```

src/services/SiteLaunchService.ts
```diff
@@ -1,7 +1,10 @@
+const whiteListedRepos = process.env.SITE_LAUNCH_FEATURE_WHITELISTED_REPOS?.split(
+  ","
+)
+
 export const isUserUsingSiteLaunchFeature = (siteName: string): boolean => {
-  // whitelisting sites that we want to trial this feature with
-  const whiteListedRepos = process.env.SITE_LAUNCH_FEATURE_WHITELISTED_REPOS?.split(
-    ","
-  )
+  if (siteName === "storybook") {
```
harishv7, line 6: Can i understand what's this check for? so all users using site launch feature will have sitename that is storybook?

src/contexts/SiteLaunchContext.tsx
```diff
@@ -0,0 +1,87 @@
...
+  siteLaunchStatusProps: SiteLaunchStatusProps
+  setSiteLaunchStatusProps: (
+    siteLaunchStatusProps: SiteLaunchStatusProps
+  ) => void
+}
+
+const SiteLaunchContext = createContext<SiteLaunchContextProps | null>(null)
```
harishv7, line 20: referencing your pr description -
I thought you will be using local storage to persist the checklist progress, is that not the case anymore?

src/contexts/SiteLaunchContext.tsx
kishore03109: Right yes this is true, thanks for this!
dfd54ce12e139b20cd5a0e955ac63b72e513814a

src/contexts/SiteLaunchContext.tsx
kishore03109: Uhm regarding our previous discussion, once I fixed the issue I was facing I left it as this for the first cut, but personally no strong opinions on using local storage instead.
Thoughts?

src/services/SiteLaunchService.ts
kishore03109: Oh this is solely for our storybooks so that our storybooks can see the UI changes. Concretely, when I run `npm run storybook`
I can see the site launch screens:

Screenshot 2023-06-09 at 1 40 54 PM

alternatively, all eng would need to update their FE env var to include storybook, but opting the simpler solution here and checking it inline. added a comment [here](2a0fe7a2fe8859442382332b435ba62a2e1eb00f) for clarity

src/types/siteLaunch.ts
kishore03109: Hey is there a reason why you prefer this? A bit inclined not to do this as there is a semantic difference between the front end site launch status vs the backend ones since some of the state is solely held in the front end. 

am ok to change it to object but do you have suggestions for some-var-name where it is clear it only refers to the states in front end? 
```
const <some-var-name> = {
  Launched: "LAUNCHED",
  NotLaunched: "NOT_LAUNCHED",
  Launching: "LAUNCHING"
ChecklistTasksPending: "CHECKLIST_TASKS_PENDING
} as const;

type SiteLaunchFrontEndStatus = typeof SiteLaunchStatus[keyof typeof some-var-name]
```

src/contexts/SiteLaunchContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export const SiteLaunchProvider = ({
+  children,
+  initialSiteLaunchStatus = "NOT_LAUNCHED",
+  initialStepNumber = 0,
+}: {
+  children: React.ReactNode
+  initialSiteLaunchStatus?: SiteLaunchFrontEndStatus
+  initialStepNumber?: number
```
seaerchin, line 36: nit: should prefer to have these as a separate type

src/contexts/SiteLaunchContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+    siteLaunchStatus: initialSiteLaunchStatus,
+    stepNumber: initialStepNumber,
+  })
+  const { data: siteLaunchDto } = useGetSiteLaunchStatus(siteName)
+  const { siteStatus: siteStatusDto, stepNumber: stepNumberDto } =
+    siteLaunchDto ||
+    ({
+      siteStatus: "NOT_LAUNCHED", // defaulting to not launched
+      stepNumber: 0,
+    } as SiteLaunchDto)
```
seaerchin, line 53:
```suggestion
  const {
    siteStatus: siteStatusDto,
    stepNumber: stepNumberDto,
  } = siteLaunchDto || {
    siteStatus: "NOT_LAUNCHED", // defaulting to not launched
    stepNumber: 0,
  }
```

seems to work w/o the cast

src/contexts/SiteLaunchContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+  useEffect(() => {
+    if (siteStatusDto === "LAUNCHED" || siteStatusDto === "LAUNCHING") {
+      setSiteLaunchStatusProps({
+        ...siteLaunchStatusProps,
```
seaerchin, line 57: i think this gets overridden so it's abit confusing why it's here

src/contexts/SiteLaunchContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+    stepNumber: initialStepNumber,
+  })
+  const { data: siteLaunchDto } = useGetSiteLaunchStatus(siteName)
+  const { siteStatus: siteStatusDto, stepNumber: stepNumberDto } =
+    siteLaunchDto ||
+    ({
+      siteStatus: "NOT_LAUNCHED", // defaulting to not launched
+      stepNumber: 0,
+    } as SiteLaunchDto)
+  useEffect(() => {
```
seaerchin, line 54: `siteLaunchStatusProps` seems to be a mapping from `siteStatusDto` -> `siteLaunchStatusProps`, which would be clearer as a function (and we can shift it out of the component) - the `useEffect` should ideally perform some effect (as the name implies) like data fetching/mutation on our backend.

src/hooks/siteDashboardHooks/useGetSiteLaunchStatus.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const useGetSiteLaunchStatus = (
+  siteName: string
+): UseQueryResult<SiteLaunchDto> => {
+  return useQuery<SiteLaunchDto>(
+    [SITE_DASHBOARD_LAUNCH_STATUS_KEY, siteName],
+    () => SiteDashboardService.getSiteLaunchStatus(siteName),
+    {
+      retry: false,
+      refetchOnWindowFocus: false,
```
seaerchin, line 17: nit: i think both of these can be set on the `queryClient` directly, saving us the trouble of repeating it.

separately, if this fetches data from db (and it should, right?) then `refetch: true` should be preferred as it won't cost us any gh api calls

src/services/SiteDashboardService.ts
```diff
@@ -3,6 +3,7 @@ import {
   SiteDashboardInfo,
   SiteDashboardReviewRequest,
 } from "types/siteDashboard"
+import { SiteLaunchDto, SiteLaunchStatusProps } from "types/siteLaunch"
```
seaerchin, line 6:
```suggestion
import { SiteLaunchDto } from "types/siteLaunch"
```

src/services/SiteDashboardService.ts
```diff
@@ -35,3 +36,10 @@ export const updateViewedReviewRequests = async (
   const endpoint = `/sites/${siteName}/review/viewed`
   return apiService.post(endpoint)
 }
+
+export const getSiteLaunchStatus = async (
+  siteName: string
+): Promise<SiteLaunchDto> => {
+  const endpoint = `/sites/${siteName}/launchInfo`
+  return apiService.get(endpoint)
```
seaerchin, line 44: do we not need a `.then((res) => res.data)`?

src/layouts/SiteDashboard/SiteDashboard.tsx
```diff
@@ -101,6 +105,9 @@ export const SiteDashboard = (): JSX.Element => {
     ({ status }) => status === "OPEN" || status === "APPROVED"
   )
 
+  const { siteLaunchStatus } = siteLaunchStatusProps
+  // todo add loading state for site launch api call, after IS-219
```
seaerchin, line 109: is there a tracking issue or is this near term? if it's near term then omitting the issue is ok i feel

src/services/SiteLaunchService.ts
seaerchin, line 1: could we shift the `whiteListedRepos` to a constants file (and if required, `SCREAMING_SNAKE_CASE` it?) and the `isUserUsingSiteLaunchFeature` to a utils file?

the `services/` folder should contain only files that make raw api calls

src/layouts/SiteDashboard/SiteDashboard.tsx
```diff
@@ -193,9 +200,67 @@ export const SiteDashboard = (): JSX.Element => {
...
+                      <Skeleton isLoaded={!isSiteLaunchLoading}>
+                        <DisplayCardHeader
+                          button={
+                            <Link
+                              variant="link"
+                              textStyle="subhead-1"
+                              color="text.title.brand"
+                              marginRight="0.75rem"
+                              href="./siteLaunchPad"
+                            >
```
seaerchin, line 217: this `Link` isn't actually what you want - what this does is perform a hard fetch (when you click on it in the storybook, it fails for this reason). 

we want to use the `Link` from `react-router-dom` and `as` the ogp `Link` (see [here](https://plainenglish.io/blog/const-vs-function-for-react-functional-components)) for details

src/layouts/SiteDashboard/SiteDashboard.tsx
```diff
@@ -193,9 +200,67 @@ export const SiteDashboard = (): JSX.Element => {
...
+                              {siteLaunchStatus === "NOT_LAUNCHED" && (
+                                <Text>Go to Launchpad</Text>
+                              )}
+                              {siteLaunchStatus ===
+                                "CHECKLIST_TASKS_PENDING" && (
+                                <Text>Visit launchpad</Text>
+                              )}
+                              {siteLaunchStatus === "LAUNCHING" && (
+                                <Text>Check Status</Text>
+                              )}
```
seaerchin, line 227: not an issue on my end, will let nat decide if this should be fixed but the text looks abit cramped

src/layouts/SiteDashboard/SiteDashboard.tsx
```diff
@@ -193,9 +200,67 @@ export const SiteDashboard = (): JSX.Element => {
...
+                          <DisplayCardCaption>
+                            Associate your isomer site to a domain
+                          </DisplayCardCaption>
+                        </DisplayCardHeader>
+                        <DisplayCardContent>
+                          <Skeleton isLoaded={!isSiteLaunchLoading} w="100%">
+                            {siteLaunchStatus === "CHECKLIST_TASKS_PENDING" && (
+                              <Text textStyle="body-2">
+                                {siteLaunchStatusProps?.stepNumber}/
+                                {Object.keys(SITE_LAUNCH_TASKS).length}
```
seaerchin, line 247:
```suggestion
                                {SITE_LAUNCH_TASKS_LENGTH}
```

src/layouts/SiteDashboard/SiteDashboard.tsx
```diff
@@ -193,9 +200,67 @@ export const SiteDashboard = (): JSX.Element => {
           {/* Right column */}
           <Box w="40%">
             <VStack spacing="1.25rem">
+              {/* Site launch display card */}
+              {isUserUsingSiteLaunchFeature(siteName) &&
+                siteLaunchStatus !== "LAUNCHED" && (
```
seaerchin, line 205: Actually i feel that this might be better off as a stand-alone component (in this file) that's used here. 

why this is so is because this chunk relies solely on a flag + context variable that the rest of the file is ignorant of (less the default portion), so having the site launch stuff exposed here might not make the most sense, especially because the text etc is also dependent on the context.

src/layouts/SiteDashboard/SiteDashboard.tsx
```diff
@@ -220,11 +285,33 @@ export const SiteDashboard = (): JSX.Element => {
                             : `${publishedAt.date}, ${siteInfo?.publishedBy}`}
                         </Text>
                       </Skeleton>
+
+                      <Skeleton isLoaded={!isSiteLaunchLoading} w="100%">
+                        {isUserUsingSiteLaunchFeature(siteName) &&
+                          siteLaunchStatus === "LAUNCHED" && (
+                            <Box
+                              style={{ display: "flex", alignItems: "center" }}
+                            >
```
seaerchin, line 294: should prefer using `Flex` here

src/layouts/SiteDashboard/SiteDashboard.tsx
```diff
@@ -193,9 +200,67 @@ export const SiteDashboard = (): JSX.Element => {
           {/* Right column */}
           <Box w="40%">
             <VStack spacing="1.25rem">
+              {/* Site launch display card */}
+              {isUserUsingSiteLaunchFeature(siteName) &&
```
seaerchin, line 204: i think this is fine as-is, but we might consider making a utility component such as `FeatureFlag`, which exposes a `showFeature` prop that renders the child component iff `showFeature` is `true`. 

this is still isolated to only this portion of the codebase so it's fine for now (assuming that we're moving to launchdarkly soon) but otherwise it'd be more important

src/layouts/SiteDashboard/SiteDashboard.stories.tsx
```diff
@@ -118,4 +123,58 @@ RequestApproved.parameters = {
...
+
+export const SiteLaunchPartialStepsCompleted = Template.bind({})
+SiteLaunchPartialStepsCompleted.decorators = [
+  (Story) => {
+    return (
+      <MemoryRouter initialEntries={["/sites/storybook/dashboard"]}>
+        <Route path="/sites/:siteName/dashboard">
+          <SiteLaunchProvider
+            initialSiteLaunchStatus="CHECKLIST_TASKS_PENDING"
+            initialStepNumber={1}
```
seaerchin, line 170: not a problem wtih the story, but if the `stepNumber` is implicitly tied to the status, then we shouldn't expose them individually

src/layouts/SiteDashboard/SiteDashboard.stories.tsx
```diff
@@ -118,4 +123,58 @@ RequestApproved.parameters = {
   },
 }
 
+export const SiteLaunchStepsCompleted = Template.bind({})
+SiteLaunchStepsCompleted.decorators = [
+  (Story) => {
+    return (
+      <MemoryRouter initialEntries={["/sites/storybook/dashboard"]}>
+        <Route path="/sites/:siteName/dashboard">
```
seaerchin, line 131: nit: for the stories here, we can actually omit the `SiteLaunchProvider` defined up above and define them individually. also, idt we need the extra `MemoryRouter` + `Router` on the new stories cos it seems like they are the same

src/contexts/SiteLaunchContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+}): JSX.Element => {
+  const { siteName } = useParams<{ siteName: string }>()
+
+  const [
+    siteLaunchStatusProps,
+    setSiteLaunchStatusProps,
+  ] = useState<SiteLaunchStatusProps>({
+    siteLaunchStatus: initialSiteLaunchStatus,
+    stepNumber: initialStepNumber,
+  })
```
seaerchin, line 46: we don't have to do ti now but do remember that we'd need to feedback network failures to our users somehow

kishore03109: Yup, agreed, this is only for the first cut!

src/layouts/SiteDashboard/SiteDashboard.stories.tsx
kishore03109: Hey, while they are related, but it is not possible to derive the step number from the siteLaunchStatus!
The checklist step number state is solely held within the frontend for now!

src/services/SiteLaunchService.ts
kishore03109: Am ok to shift the file around to a utils folder but I am not too keen on shifting this inline, since that would mean I have to redeploy the code everytime I want to add a new site to trial this out with, which is why I opted for a env var change instead. Thoughts?

src/hooks/siteDashboardHooks/useGetSiteLaunchStatus.ts
kishore03109: hey, yup the `refetechOnWindow` is redundant. However, since this query client is a central client used in multiple parts of this codebase, I dont want to increase the surface area of possible bugs in this PR! 

I am assuming you are referring to `refetchOnWindowFocus` ya. For now, the part of the state of the site launch process (ie, the current checklist number) is currently held in the frontend, so I want to be more careful on when we refetch this! (ie, the frontend should invalidate the query intentioanlly when the user is ready to launch the site and completed all the tasks in the checklist)

src/layouts/SiteDashboard/SiteDashboard.stories.tsx
kishore03109: arh right, removed the redundant codes below as well!

src/services/SiteDashboardService.ts
kishore03109: this tripped me in the later PRs :/ thanks for the catch !
1205cb98b7e60d37b8efce04262cd2e422100151

src/layouts/SiteDashboard/SiteDashboard.tsx
kishore03109: yup agreed!

src/services/SiteLaunchService.ts
seaerchin: no, not inlining it - more of
```ts
export whiteListedRepos = config.get(...)
```

whether you `SCREAMING_SNAKE` is up to you (cos this is technically unchanged over the lifetime of the app)

src/layouts/SiteDashboard/SiteDashboard.tsx
kishore03109: done

src/contexts/SiteLaunchContext.tsx
kishore03109: done

src/contexts/SiteLaunchContext.tsx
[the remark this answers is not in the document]
kishore03109: Have restructured the entire bit for clarity, let me know if it makes sense now

src/contexts/SiteLaunchContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+        dnsRecords: siteLaunchDto?.dnsRecords,
+      })
+      return
+    }
+    if (
+      siteStatusDto !== "NOT_LAUNCHED" &&
+      siteLaunchStatusProps.siteLaunchStatus !== "NOT_LAUNCHED"
+    ) {
+      setSiteLaunchStatusProps({
+        ...siteLaunchStatusProps,
```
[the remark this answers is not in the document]
kishore03109: yup, that was confusing, have shifted out the code instead, let me know if it makes sense now!

src/contexts/SiteLaunchContext.tsx
kishore03109: thanks for this man, I knew that `useEffect` is used to sync external to internal flow, but here I tried using it to "sync" two hooks which the `useEffect` shoudnt be used for. 
Abstracted this out for clarity as suggested, let me know if this makes sense!

src/layouts/SiteDashboard/SiteDashboard.tsx
kishore03109: removed liao

src/services/SiteLaunchService.ts
kishore03109: sure, done

src/layouts/SiteDashboard/SiteDashboard.tsx
kishore03109: sure I have used `Link` but to be clear, I do want to perform a hard fetch to the newer url. Ie, the storybook link _should_ fail as the sitelaunchPad Stuff is not introduced in this PR (its introduced in #1297 instead)!

src/layouts/SiteDashboard/SiteDashboard.tsx
kishore03109: ok refactored this bit

src/utils/siteLaunchUtils.ts
```diff
@@ -0,0 +1,11 @@
+import { WHITE_LISTED_REPOS } from "constants/siteLaunch"
+
+export const isUserUsingSiteLaunchFeature = (siteName: string): boolean => {
```
seaerchin, line 3: nit: `shouldUseSiteLaunchFeature` reads better and less verbish (flags should be nouns)

src/layouts/SiteDashboard/SiteDashboard.tsx
```diff
@@ -306,3 +348,66 @@ export const SiteDashboard = (): JSX.Element => {
     </SiteViewLayout>
   )
 }
+
+const SiteLaunchDisplayCard = ({
+  siteName,
+  siteLaunchStatus,
+  isSiteLaunchLoading,
+  siteLaunchChecklistStepNumber,
+}: SiteLaunchDisplayCardProps): JSX.Element => {
```
seaerchin, line 357: could we remove the `siteName/status/stepnumber` props? this is because of the (implicit) assumption that this component should always be rendered inside a `SiteLaunchContext` + `Route`.

this allows us to simplify this component and, if used without the appropriate context, throw an error

src/types/siteLaunch.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export const SITE_LAUNCH_TASKS_LENGTH = Object.keys(SITE_LAUNCH_TASKS).length
+
+export const SITE_LAUNCH_PAGES = {
+  DISCLAIMER: 1,
+  INFO_GATHERING: 2,
+  RISK_ACCEPTANCE: 3,
+  CHECKLIST: 4,
+}
```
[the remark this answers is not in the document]
kishore03109: Hmm the later PRs does add additional enforcement on the types here, is this [sufficient](https://github.com/isomerpages/isomercms-frontend/blob/IS-218-FE-Add-screens-for-creating-launch/src/types/siteLaunch.ts)?

src/constants/siteLaunch.ts
```diff
@@ -0,0 +1,3 @@
+export const WHITE_LISTED_REPOS = process.env.REACT_APP_SITE_LAUNCH_FEATURE_WHITELISTED_REPOS?.split(
```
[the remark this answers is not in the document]
kishore03109: 17d62b968be42196c3f3a441bd4448a49802eabc

src/contexts/SiteLaunchContext.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+export const useSiteLaunchContext = (): SiteLaunchContextProps => {
+  const SiteLaunchContextData = useContext(SiteLaunchContext)
+  if (!SiteLaunchContextData)
+    throw new Error("useSiteLaunchContext must be used within an RoleProvider")
+  return SiteLaunchContextData
+}
+
+function updateSiteLaunchStatusOnApiCall(
+  siteLaunchDto: SiteLaunchDto | undefined,
```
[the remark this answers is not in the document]
kishore03109: sure, changed the API contract here 6064915636c51c55d28fe047e93b2871f54a7c2d

src/utils/siteLaunchUtils.ts
kishore03109: a355257f56fd66c1d93c015a41fae38ef75594da

src/layouts/SiteDashboard/SiteDashboard.tsx
kishore03109: > could we remove the siteName/status/stepnumber props?
Hmm I am lost here :(
these are variables that I intend to use later in this function, what exactly do you mean by removing the props ah? What does this got to do with being rendered inside a SiteLaunchContext + Route ya?

src/contexts/SiteLaunchContext.tsx
```diff
@@ -0,0 +1,100 @@
...
+    stepNumber: initialStepNumber || 0,
+  })
+
+  const { data: siteLaunchDto } = useGetSiteLaunchStatus(siteName)
+
+  updateSiteLaunchStatusOnApiCall(
+    siteLaunchDto,
+    siteLaunchStatusProps,
+    setSiteLaunchStatusProps
+  )
```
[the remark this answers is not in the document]
kishore03109: Hey I am quite confused here :(
doesnt this line here
setState = (props) => setSiteLaunchStatusProps(getSiteLaunchStatusProps(props))
set the props to the value that props already is set to?
may I get your help to give a concrete example of your suggestion?

src/utils/siteLaunchUtils.ts
kishore03109: a355257f56fd66c1d93c015a41fae38ef75594da

src/layouts/SiteDashboard/SiteDashboard.tsx
seaerchin: these props are exposed on their respective hooks - `useParams`/`useSiteLaunchContext`. This eliminates the need for them to be passed as explicit props because **this component should not be used outside of the context**.

src/types/siteLaunch.ts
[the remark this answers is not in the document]
seaerchin: not really sure how it's being used so i think i'll have to see it on that PR bah

src/contexts/SiteLaunchContext.tsx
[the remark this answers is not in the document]
seaerchin: resolved offline - changing to `useEffect` to sync states between BE + FE

src/contexts/SiteLaunchContext.tsx
```diff
@@ -81,11 +81,13 @@ export const SiteLaunchProvider = ({
...
-    siteLaunchStatusProps,
-    setSiteLaunchStatusProps
-  )
+  if (siteLaunchDto) {
+    updateSiteLaunchStatusOnApiCall(
+      siteLaunchDto,
+      siteLaunchStatusProps,
+      setSiteLaunchStatusProps
+    )
+  }
```
seaerchin, line 90: as discussed offline, please use a `useEffect` here!

seaerchin (commented):
change to `useEffect`, rest lgtm - caveat that the number one is dependent on how it's being used in later PRs

src/contexts/SiteLaunchContext.tsx
seaerchin: kek turns out is a `useEffect` after all

src/hooks/siteDashboardHooks/useGetSiteLaunchStatus.ts
seaerchin: sure

src/contexts/SiteLaunchContext.tsx
kishore03109: donez [3515832](https://github.com/isomerpages/isomercms-frontend/pull/1296/commits/3515832b98db3c081558a1494858273c9382309d)

NatMaeTan:
@kishore03109 from design discussion, spacing btw card title and icon should also have been dropped to 8px

kishore03109:
@NatMaeTan opps ps missed it, this is done [now](4daba5c3192b1cbe44e33263596cc33defe1f0ff)

src/components/DisplayCard/DisplayCard.tsx
```diff
@@ -77,7 +76,7 @@ export const DisplayCardTitle = ({
   return (
     <>
       <HStack paddingBottom="0.25rem">
-        <Text {...(styles.title as TextProps)} {...props}>
+        <Text {...(styles.title as TextProps)} mr="0" {...props}>
```
seaerchin, line 79: i might be lacking context here but why did we make this change? the `props` allow for passing in `mr` already and this is a change that will affect **all** existing display card header components so it might be problematic

src/theme/components/DisplayCard.ts
```diff
@@ -48,10 +48,10 @@ export const DisplayCard: ComponentMultiStyleConfig = {
...
       justifyContent: "flex-start",
       boxSizing: "border-box",
-      paddingInlineEnd: "2rem",
-      paddingInlineStart: "2rem",
-      paddingTop: "2rem",
-      paddingBottom: "1.5rem",
+      paddingInlineEnd: "1.5rem",
+      paddingInlineStart: "1.5rem",
+      paddingTop: "1rem",
+      paddingBottom: "1rem",
```
seaerchin, line 54: can i double check on why this was changed?

seaerchin (commented):
sanity check on why we are making broad style changes rather than just customizing the style on the specific component

nit: for `SiteLaunchDisplayCard`, we should prefer using the hooks to get data rather than passing it in because the context already provides them. this allows us to make the API simpler.

src/types/siteLaunch.ts
harishv7: Sure we can use something like `SiteLaunchFrontEndStatusOptions` and `type SiteLaunchFrontEndStatus` which is one of those options?

src/theme/components/DisplayCard.ts
kishore03109: Design wanted the the display cards  from 16 X 16 X 16 X 16 to become 24 X 24 left and right and the top and bottom 16 X 16. 

This is intentionally made as a broad change to allow all other display cards in the dashboard (Ie all display cards since they are not being used anywhere else, correct me if I am wrong about this) to follow this format instead

src/services/SiteLaunchService.ts
harishv7: Okay got it. But what's this env var change? Is it noted in the PR description?

> update their FE env var

src/components/DisplayCard/DisplayCard.tsx
kishore03109: >i might be lacking context here but why did we make this change?

Design wanted the dist from the header to the icon to be only 8px. Changing the text to have no right margin here. 

> the props allow for passing in mr already and this is a change that will affect all existing display card header components so it might be problematic

This is intentionally made as a broad change to allow all other display cards in the dashboard (Ie all display cards since they are not being used anywhere else, correct me if I am wrong about this) to follow this format instead

kishore03109:
@seaerchin  
> nit: for SiteLaunchDisplayCard, we should prefer using the hooks to get data rather than passing it in because the context already provides them. this allows us to make the API simpler.

good point, but in this case, these props are being used at a component that is not `SiteLaunchDisplayCard`, and while I could simplify the contract, I think duplicating the same code again in the same file might less ideal!

src/services/SiteLaunchService.ts
kishore03109: it is changed with the [correct prefix](https://create-react-app.dev/docs/adding-custom-environment-variables/) to allow react to actually source it. Yes, it's already noted in the PR desc.

src/types/siteLaunch.ts
kishore03109: sure, 38945a2083e6b0c7d7043ce3711449d837da7889

src/theme/components/DisplayCard.ts
seaerchin: ok if design says is for all display cards then is fine

src/contexts/SiteLaunchContext.tsx
kishore03109: agreed, it will either be local storage or ideally will shift state away from FE in the future as a whole. Nat is aware of this, and this limitation will be tackled this in a separate PR!

seaerchin (approved):

PR #1301, opened 2023-06-09
chore: update dependencies

alexanderleegs:
Fixes vulnerability raised by synk by updating `@datadog/datadog-ci`


kishore03109 (approved):
Nit: Moving forward can we name the branches more descriptively? 
eg. chore/upgradeDataDogV2.14.0 just to locally we dont have many of these chore/update-Deps name

alexanderleegs:
> Nit: Moving forward can we name the branches more descriptively? eg. chore/upgradeDataDogV2.14.0 just to locally we dont have many of these chore/update-Deps name

whoops that's a good point, will do this moving forward!

PR #1302, opened 2023-06-09
test(e2e): scaffold dashboard tests 

seaerchin:
## Problem
Isomer doesn't have e2e tests at present. as part of the effort to add more e2e tests, this PR scaffolds dashboard tests. 

Closes IS-251

## Solution
1. Add test cases for a few obvious cases
2. Add remaining test description + skip

**Notes**
1. for existing test cases - just see if there are additional conditions that we should be checking for. using `cypress:open` should allow the tests to be ran successfully
2. for test descriptions - check to see if there's any i've missed out.

### DIFF
--- a/cypress/e2e/dashboard.spec.ts
+++ b/cypress/e2e/dashboard.spec.ts
@@ -1,0 +1,270 @@
+import {
+  addAdmin,
+  approveReviewRequest,
+  closeReviewRequests,
+  createReviewRequest,
+  listReviewRequests,
+} from "../api"
+import { editUnlinkedPage } from "../api/pages.api"
+import {
+  E2E_EMAIL_COLLAB,
+  E2E_EMAIL_REPO_STAGING_LINK,
+  E2E_EMAIL_TEST_SITE,
+  MOCK_REVIEW_DESCRIPTION,
+  MOCK_REVIEW_TITLE,
+} from "../fixtures/constants"
+import { USER_TYPES } from "../fixtures/users"
+import {
+  addCollaborator,
+  getOpenStagingButton,
+  removeOtherCollaborators,
+  visitE2eEmailTestRepo,
+} from "../utils"
+
+const getReviewRequestButton = () => cy.contains("button", "Request a Review")
+const getSubmitReviewButton = () => cy.contains("button", "Submit Review")
+
+const selectReviewer = (email: string) => {
+  cy.get("div[id^='react-select']").contains(E2E_EMAIL_COLLAB.email).click()
+}
+const removeReviewers = () => cy.get('div[aria-label^="Remove"]').click()
+
+const TITLE_INPUT_SELECTOR =
+  "input[placeholder='Title your request'][name='title']"
+const DESCRIPTION_TEXTAREA_SELECTOR =
+  "textarea[name='description'][placeholder='Briefly describe the changes you’ve made, and add a review deadline (if applicable)']"
+
+const getAddReviewerDropdown = () =>
+  cy.contains("Select Admins to add as reviewers").parent()
+
+const goToWorkspace = () => cy.contains("a", "Edit site").click()
+
+const REVIEW_MODAL_SUBTITLE =
+  "An Admin needs to review and approve your changes before they can be published"
+
+describe("dashboard flow", () => {
+  beforeEach(() => {
+    cy.setupDefaultInterceptors()
+    cy.setEmailSessionDefaults("Email admin")
+    visitE2eEmailTestRepo()
+  })
+  describe("no review requests", () => {
+    beforeEach(() => {
+      closeReviewRequests()
+    })
+    it('should open the staging site on click of the "Open staging" button', () => {
+      // Act
+      getOpenStagingButton()
+        .should("have.attr", "href", E2E_EMAIL_REPO_STAGING_LINK)
+        .should("have.attr", "target", "_blank")
+    })
+    it('should open the "Request a Review" modal on click of the "Request a Review" button', () => {
+      // Act
+      getReviewRequestButton().click()
+
+      // Assert
+      cy.contains(REVIEW_MODAL_SUBTITLE).should("be.visible")
+    })
+    it.skip("should be able to navigate to the staging site using the dropdown button", () => {
+      throw new Error("Not implemented")
+    })
+    it.skip("should be able to navigate to the production site using the dropdown button", () => {
+      throw new Error("Not implemented")
+    })
+    it.skip('should navigate to the isomer guide on click of the "Get help" button', () => {
+      throw new Error("Not implemented")
+    })
+    it.skip("should navigate to the settings page when manage site settings is clicked", () => {
+      throw new Error("Not implemented")
+    })
+    it.skip("should navigate to the workspace when edit site is clicked", () => {
+      // Arrange
+      // NOTE: There shouldn't be an alert
+      // Act
+      // Assert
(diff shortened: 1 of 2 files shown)


cypress/e2e/dashboard.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+      // NOTE: We define changes as being
+      // 1. Adding a new page
+      // 2. Editing an existing page
+      // 3. Deleting an existing page
+      // 4. Rename an existing page
+      // 5. Moving an existing page
+      // Act
+      // Assert
+    })
+    it("should be able to have the request approved as a collaborator", () => {
```
kishore03109, line 201: [Non-nit] we should have two tests here to test critical access control flows, ie 
'contributor should not be able to approve the request' 
'admin should be able to approve the request'

cypress/e2e/dashboard.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+    })
+  })
+  describe("create review request", () => {
+    beforeEach(() => {
+      // NOTE: We want to start this test suite on a clean slate
+      // and prevent other tests from interfering
+      removeOtherCollaborators()
+    })
+
+    it("should not be able to create a review request when the site has 1 collaborator", () => {
```
kishore03109, line 95: should we also add a 
"should not be able to add contributor as a reviewer"? This seems like a critical access flow control we should test for

cypress/e2e/dashboard.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+        if (!requests || !requests.length) {
+          createReviewRequest(
+            MOCK_REVIEW_TITLE,
+            [E2E_EMAIL_COLLAB.email],
+            MOCK_REVIEW_DESCRIPTION
+          ).then((id) => approveReviewRequest(id))
+        }
+      })
+    })
+    it("should have the alert on the workspace", () => {
```
kishore03109, line 222: can we rename the phrase 'the alert'? 
something liek 
"should have an alert stating that editing is disabled when accessing workspace"

cypress/e2e/dashboard.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+    })
+    it('should open the "Request a Review" modal on click of the "Request a Review" button', () => {
+      // Act
+      getReviewRequestButton().click()
+
+      // Assert
+      cy.contains(REVIEW_MODAL_SUBTITLE).should("be.visible")
+    })
+    it.skip("should be able to navigate to the staging site using the dropdown button", () => {
+      throw new Error("Not implemented")
```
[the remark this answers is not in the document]
seaerchin: yes

cypress/e2e/dashboard.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+      // Arrange
+      // Act
+      // Assert
+    })
+    it("should allow the requestor to close the review request", () => {
+      // Arrange
+      // Act
+      // Assert
+    })
+    it("should disallow users from viewing a closed review request", () => {
```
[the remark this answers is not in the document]
seaerchin: yeap

cypress/e2e/dashboard.spec.ts
seaerchin: contributors won't be able to be added as a reviewer in the first place! there's a test for this

```ts
    it("should not be able to create a review request without a reviewer", () => {
      // Arrange
      cy.createEmailUser(
        E2E_EMAIL_COLLAB.email,
        USER_TYPES.Email.Collaborator,
        USER_TYPES.Email.Admin
      )
      addCollaborator(E2E_EMAIL_COLLAB.email)

      // Act
      // NOTE: There should be at least 1 other admin to be able to create a review request
      getReviewRequestButton().click()
      cy.contains(REVIEW_MODAL_SUBTITLE).should("be.visible")

      // Assert
      getAddReviewerDropdown().click()
      cy.contains("No options").should("be.visible")
    })
```

cypress/e2e/dashboard.spec.ts
seaerchin: see above comment - maybe i can change the wording of the test to make it clear that we're also trying to add a contributor

cypress/e2e/dashboard.spec.ts
seaerchin: sure - i think `review request alert` might be more descriptive in context. will update to your desc + the phrase

seaerchin: done

cypress/e2e/dashboard.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+    })
+    it.skip('should navigate to the isomer guide on click of the "Get help" button', () => {
+      throw new Error("Not implemented")
+    })
+    it.skip("should navigate to the settings page when manage site settings is clicked", () => {
+      throw new Error("Not implemented")
+    })
+    it.skip("should navigate to the workspace when edit site is clicked", () => {
+      // Arrange
+      // NOTE: There shouldn't be an alert
```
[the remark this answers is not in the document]
seaerchin: yeap!

cypress/e2e/dashboard.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+      // Assert
+      cy.contains("No options").should("be.visible")
+    })
+    it("should not be able to create a review request without a reviewer", () => {
+      // Arrange
+      cy.createEmailUser(
+        E2E_EMAIL_COLLAB.email,
+        USER_TYPES.Email.Collaborator,
+        USER_TYPES.Email.Admin
+      )
```
kishore03109, line 112: correct me if I am wrong here.This code updates the existing collab user from an admin right?

cypress/e2e/dashboard.spec.ts
kishore03109: Wait in that test case, we dont see any reviewers as an option because I cannot self review right? This is not the same as checking that collab cannot be a reviewer? 

seems like to fully test that flow we might need another user email here (which might be out of scope of the pr) but it feels like a critial access flow control we want to add e2e tests for. Wdyt?

seaerchin: this is the creation of a review request - the review request has **not** been created yet. the test case for the creation one is the below one, where there is only 1 admin (the creator of the RR) and hence, the RR cannot be created

```ts
should not be able to create a review request when the site has 1 collaborator
```

in this one, we add a collaborator (now there's 2 site members, admin + collab) and the admin tries to create a RR but fails because there's only 1 other collab that cannot be a reviewer

cypress/e2e/dashboard.spec.ts
seaerchin: no, it creates a collaborator, the last argument is to set the session type. the ideal flow would be to use `getCookies -> setCookie` + `cy.session` but this results in cypress skipping over the test body for some odd reason :/

seaerchin:
@isomerpages/iso-engineers realised that this test spec has alot of overlap with review requests - will probably cherry pick the commits into a new branch once approved to avoid dropping the comments here. this will allow this spec to solely be focused on the dashboard

kishore03109 (approved):
as discussed, could you add a comment on why we cannot use the ideal API contract for `cy.createEmailUser()`? 
I don't even think it needs to be solved, as long as it is clear why we need the `initialUserType` is think its fine for our use case

apart from this LGTM

PR #1305, opened 2023-06-12
test(e2e): scaffold notif tests

seaerchin:
## Problem
Currently, isomer does not have e2e tests for identity flows. this PR adds scaffolding and util functions for notifs together with some tests.

## Solution
1. add notfis.api for api requests out to backend
2. add utils + selectors for ease of development
3. write specs + test cases


seaerchin:
@alexanderleegs help to take a look n see what other test suites i missed out! (you don't have to be the one writing the remaining specs)

cypress/e2e/notifications.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  E2E_EMAIL_COLLAB,
+  E2E_EMAIL_TEST_SITE,
+} from "../fixtures/constants"
+import {
+  genRandomString,
+  getNotificationsButton,
+  visitE2eEmailTestRepo,
+} from "../utils"
+
+const NEW_COMMENTS_BG_COLOR = "rgb(235, 248, 255)"
```
kishore03109, line 13: is _unread_ comments rather than new comments a better phrasing here?

cypress/e2e/notifications.spec.ts
alexanderleegs, line 1: There should also be a test case for notifications being created (sent to all reviewers, except the person making the edit) if a rr is open and someone has made an edit!

cypress/utils/pages.ts
```diff
@@ -0,0 +1,4 @@
+// NOTE: generates a random string of length 7
+export const genRandomString = (): string => {
+  return Math.random().toString(36).substring(7)
```
kishore03109, line 3: Nit: What do you feel about including special allowed characters as well ah? 
my thought process here is that we dont really care about the randomness of the string, but we do care if any allowed string is rendered to the user in the notification panel correctly. 

Concretely speaking rather than `genRandomString()`, I am suggesting to use
```
export const generateNotificationString = (): string => {
    const specialCharacters = "!@#$%^&*()"
    const otherLang = "文கி"
  return specialCharacters + otherLang + Math.random().toString(36).substring(100) // longer string to test concatenation 
```

then check if the DOM has rendered a similar string

cypress/api/notifications.api.ts
```diff
@@ -0,0 +1,9 @@
+import { E2E_EMAIL_SITE_API_URL } from "./constants"
+
+const BASE_URL = `${E2E_EMAIL_SITE_API_URL}/notifications`
+/**
+ * NOTE: This uses the **cookie's** site and email to determine which notifications to mark as read.
```
kishore03109, line 5: hey am confused about this comment here :( 
`cy.request("POST", BASE_URL)` works as a network inceptor at the network layer, how are we using a cookies's info for this ah

cypress/e2e/notifications.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+const EMPTY_NOTIFICATIONS_TEXT = "There are no notifications for display."
+const getReviewRequestedNotifText = (email: string) =>
```
kishore03109, line 15: nit: Feels weird that we are defining these constants here.. 
should we extract this out in our non-test codes, and then export those strings here? 
this way if we change the underlying string our tests will still pass

cypress/e2e/notifications.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+    // NOTE: Mark all notifications as read as email collab.
+    // This is because previous tests may have created notifications.
+    api.markNotificationsAsRead()
+
+    cy.setEmailSessionDefaults("Email admin")
+    cy.setupDefaultInterceptors()
+    api.closeReviewRequests()
+    // NOTE: Mark all notifications as read as email collab.
+    // This is because previous tests may have created notifications.
+    api.markNotificationsAsRead()
```
kishore03109, line 50: sanity check here, is the duplication of the lines 
```
 // NOTE: Mark all notifications as read as email collab.
    // This is because previous tests may have created notifications.
    api.markNotificationsAsRead()
```
intended? There are no additional tests that occur within the duplicated 
`api.markNotificationsAsRead()` right?

cypress/api/notifications.api.ts
seaerchin: `cy.request` **does not work as an interceptor** - gonna copy paste the [docs](https://docs.cypress.io/api/commands/request#Cookies) here for clarity and if there's still unclear stuff, i'm more than happy to help!

tl;dr: `request` makes it perform similarly (dk if it's identically) to a user request from a browser

Screenshot 2023-06-14 at 12 02 20 PM

cypress/e2e/notifications.spec.ts
seaerchin: wait my comment is wrong but the duplication **is intended**. in short, we're acting as 2 users (teh collab and the admin) so we clear notifs for both. 

the backend clears notifs through the session data, which is generated from the request's cookies. because we set cookies on the `setUpXxxDefaults` function, we need to do it twice.

seaerchin: updated comment for clarity

seaerchin: updated comment for clarity

cypress/e2e/notifications.spec.ts
seaerchin: good point, changed

cypress/utils/pages.ts
seaerchin: sure - do take note that this function isn't actually used by `editPage.spec` which is where those invariants are asserted (that special characters + other languages are shown properly)

cypress/e2e/notifications.spec.ts
seaerchin: sure - putting it in `NotificationsMenu.tsx` cos that's the only place where it's used.

cypress/e2e/notifications.spec.ts
seaerchin: double checking this: 
Reviewers (both admin): Editor_A, Reviewer_B 
Contributors (admin of site), **not reviewers:** Admin_C
Contributors (collab of site): Contrib_D

for this case, 
1. Editor_A should **not** see a notif
2. Reviewer_B should
3. will Admin_C/Contrib_D see any notifs?

seaerchin: does the role matter for the editor actually?

cypress/api/notifications.api.ts
kishore03109: got it, thanks for the detailed explanation kind sir

kishore03109 (approved):
LGTM assuming, 
fixing chromatic CICD (which is the only thing failing) is out of scope of this PR + we have another PR upcoming to fix that issue!

PR #1307, opened 2023-06-14
fix(constants): update diff values to be constant

seaerchin:
## Problem
chromatic cannot render our diff view cos too big ._. also, because the diff view is randomly generated atm, this means that misc changes also incur a deploy to chromatic (cos the diff view value changed)

## Solution
1. hard code `const` values for display
2. change # lines to 10 instead of 100 ._.

## Tests
- [ ] check if chromatic deploy works

### DIFF
--- a/src/mocks/constants.ts
+++ b/src/mocks/constants.ts
@@ -489,12 +489,7 @@ export const MOCK_SITES_DATA: SiteDataRequest = {
 }
 
 const generateDiffValues = () =>
-  Array(10)
-    .fill([])
-    .map(() => {
-      return "the quick brown fox jumps over the lazy dog"
-    })
-    .join("\n")
+  Array(10).join("the quick brown fox jumps over the lazy dog\n")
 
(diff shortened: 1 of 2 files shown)


src/mocks/constants.ts
```diff
@@ -480,22 +487,14 @@ export const MOCK_SITES_DATA: SiteDataRequest = {
...
-      return Array(length)
-        .fill("")
-        .map(() => {
-          const char = _.random(48, 122)
-          return String.fromCharCode(char)
-        })
-        .join("")
+      return "the quick brown fox jumps over the lazy dog"
     })
     .join("\n")
```
kishore03109, line 497: can just `Array(10).join("the quick brown fox jumps over the lazy dog\n")` bah

kishore03109 (approved):

harishv7 (approved):
lgtm, apart from the 1 comment from @kishore03109

PR #1313, opened 2023-06-19
fix(tokens): change tokens to sync with figma

kishore03109:
## Problem

Our tokens and the ones in figma deviates quite a bit due to legacy reasons. We know that this has been an issue for us for a while, and this might not be the best way to go about doing this, but am hoping to speed this up by using a more brute force approach + then change the other component if there are other issues. (ie breaking fast then fixing fast)

## Solution

Change the tokens 

**Breaking Changes**

  Since this is referenced in multiple parts of the code base, it might make visual changes that might defer from our original intention. 

## Tests
Nat agreed to spend a bit of time to check the build of this (both at storybook level and at the staging site level) as a quick sanity check before merging into prod. 
Note: this PR is currently blocked by #1202 since the testing will involve push to staging and #1202 is currently using staging. Will merge this in once Nat approves this PR. 

## Reviewer Notes 
This PR is more to start the ball rolling with regards to our differing tokens. Open to feedback about less-brute force approach that we could use to fix this!

Also seeking feedback for the var name `baseIsomerColour`. Not sure what a better name is that we can use to represent the colour scale that we are trying to use!

To allow us to use the entire scale, we have in code the entire scale to be used in the future, although we dont have a specific token name for this yet.


src/theme/foundations/colours.ts
```diff
@@ -30,16 +31,16 @@ export const colours: { [k in IsomerColorScheme]: NestedRecord } = {
...
-    100: "#EFE7FF",
-    200: "#C9B3FF",
-    300: "#AA90FA",
-    400: "#8C73DB",
-    500: "#6D58BB",
-    600: "#5543A4",
-    700: "#493897",
-    800: "#3B2E8A",
-    900: "#271E78",
+    50: "#F8F9FA",
```
harishv7, line 34: have we checked for visual inconsistencies?

src/theme/foundations/colours.ts
```diff
@@ -104,6 +105,20 @@ export const colours: { [k in IsomerColorScheme]: NestedRecord } = {
...
+      200: "#C9B3FF",
+      300: "#AA90FA",
+      400: "#8C73DB",
+      500: "#6D58BB",
+      600: "#5543A4",
+      700: "#493897",
+      800: "#3B2E8A",
+      900: "#271E78",
+    },
+  },
```
seaerchin, line 121: had some comments for @NatMaeTan - will leave it to her to follow up + jot notes TIA nat

NatMaeTan:
Discussion notes with @seaerchin 

Concerns:
Mapping a color scale to a non-color token may cause difficulty in remapping in the future, i.e.

defining hex value directly at `brand.illustration.50 `
vs
defining hex value at `purple.50` and mapping it to `brand.illustration.50`

With the second option, if we choose to change the brand.illustration colors in the future we can still remap it to a different color scale e.g. red

Actionables:
1. To redefine the hex values as purple.50 etc.
2. To check with design system team what would be the appropriate approach to map purple to in this case. There are 2 options:
- Map to alt tokens, but am not sure currently what else they map to (There may be surprises)
- Define a new token

@NatMaeTan I'm checking in with the design system team for some advice on this ah, but good point raised by chin

seaerchin:
> defining hex value at purple.50 and mapping it to brand.illustration.50

we gonna use a token eg `brand.illustration.light`, not `brand.illustration.50`

src/theme/foundations/colours.ts
kishore03109: locally yes, but I am not the expert on this. 
Once we clarify this implementation with design system, can move forward to test on staging + get OK from design

harishv7: Okay sounds good

kishore03109:
@seaerchin just a heads up, design opted not to map each scale to a non-colour token because we don't have enough names for now to make non-color token to a scale.

As such, rather than a `brand.illustration.light` we will use  `brand.illustration.50`, and this still allows us in the future to remap it to a different colour scale.

seaerchin:
> @seaerchin just a heads up, design opted not to map each scale to a non-colour token because we don't have enough names for now to make non-color token to a scale.
> 
> As such, rather than a `brand.illustration.light` we will use `brand.illustration.50`, and this still allows us in the future to remap it to a different colour scale.

writing here for doc - just add a name rather than using a number for the value that's being used

src/theme/foundations/colours.ts
```diff
@@ -1,4 +1,26 @@
 // Colour schemes available for isomer
+
+export type IsomerBaseColorScheme = "purple"
+
+// We are not using all the shades of the base colour scheme,
+// but this is placed here if we need them in the future.
+export const baseColourMapping: {
```
seaerchin, line 7: could we rename this to `palettes`? more indicative of what it's supposed to be

src/theme/foundations/colours.ts
```diff
@@ -1,4 +1,26 @@
 // Colour schemes available for isomer
+
+export type IsomerBaseColorScheme = "purple"
```
seaerchin, line 3: maybe out of scope but can we shift `primary + secondary` here and rename to `IsomerColourPalettes`?

kishore03109: This change will incur alot of file changes right? In separate PR uh

src/theme/foundations/colours.ts
kishore03109: 646f8aaf4d994084f96998cd864ef963623aea02

kishore03109:
!run-chromatic

src/theme/foundations/colours.ts
seaerchin: Sorry, I'm not clear why it would incur alot of changes - could you go into more detail?

The theme that chakra reads from is passed by us as an object then chakra just reads the key we pass in, so as long as the final object contains the `primary.<palette>` key, nothing should change!

kishore03109:
Closing this PR in favour of #1329 as that would be a better way to go about fixing this issue

PR #1314, opened 2023-06-19
docs(setup): add docs for e2e setup

seaerchin:
## Problem
No docs previously for how to setup e2e test, which was confusing. This PR adds doc for setting up email + github e2e tests.

## Solution
add doc

### DIFF
--- a/cypress/SETUP.md
+++ b/cypress/SETUP.md
@@ -1,0 +1,35 @@
+### Pre-requisites
+
+This document assumes that you have valid read-only access to the **production** database, in order to extract the data of the e2e test sites.
+
+### Setup
+
+#### Github e2e tests
+
+In order to run the github e2e tests successfully, we need to perform the following one-off setup:
+
+1. go to the backend folder (assuming that both FE + BE are in the same folder, you can `cd ../isomercms-backend`)
+2. set `E2E_ISOMER_ID` to your desired value
+3. ensure that the `E2E_TEST_GH_TOKEN` has sufficient permissions (try running any test)
+
+#### Email e2e tests
+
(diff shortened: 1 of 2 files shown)


cypress/SETUP.md
```diff
@@ -0,0 +1,35 @@
...
+In order to run the github e2e tests successfully, we need to perform the following one-off setup:
+
+1. go to the backend folder (assuming that both FE + BE are in the same folder, you can `cd ../isomercms-backend`)
+2. set `E2E_ISOMER_ID` to your desired value
+3. ensure that the `E2E_TEST_GH_TOKEN` has sufficient permissions (try running any test)
+
+#### Email e2e tests
+
+In order to run the email e2e tests successfully, we need to perform the following one-off setup:
+
```
kishore03109, line 18: hmm could we use a script to automate this in the backend? the repos has a mapping to the `sites` so there will be a manual change of the site id to match local db, which isnt ideal

kishore03109 (approved):
The content of this PR LGTM, but I think we can have a quick win here by automating this in the backend. 
a naive opinion here is that if we automate this via a script in the backend, we dont even need this `SETUP.md` and new devs have lesser overhead when ramping up

cypress/SETUP.md
seaerchin: good suggestion but probably not going to use a script - i'll be making this a migration with `upsert` so that it's idempotent as `staging + production` already have this

seaerchin:
fyi @kishore03109 - gonna be merging this cos it's an improvement over what we have at present. will be writing out a script for this on the backend and file a corresponding PR on FE later.

PR #1315, opened 2023-06-20
fix(ci): chromatic

seaerchin:
## Problem
high chromatic usage because we listen on push event for every branch

## Solution
remove `push` event for feature branches only do so for `develop`

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -4,8 +4,10 @@ name: "Chromatic"
 # Event for the workflow
 on:
   push:
+    branches:
+      - develop
   pull_request:
-    types: [opened, reopened]
+    types: [opened, reopened, synchronize]
 
 # List of jobs
 jobs:


.github/workflows/chromatic.yml
```diff
@@ -4,8 +4,10 @@ name: "Chromatic"
 # Event for the workflow
 on:
   push:
+    branches:
+      - develop
   pull_request:
-    types: [opened, reopened]
+    types: [opened, reopened, synchronize]
```
kishore03109, line 10: am confused by this change :( 
I thought we were opting for a comment based trigger instead?

kishore03109: I dont think this fits in our mental model for a PR leh, eg when we have stacked PRs, design should still be able to comment on that changes even if they are not directly pushed to develop!

seaerchin: we are, can close this first if you want. 

> when we have stacked PRs, design should still be able to comment on that changes even if they are not directly pushed to develop

this will work still cos when the PR is synced, it will trigger the workflow

seaerchin:
will close this first bah since we going iwth comment based fo rnow

PR #1317, opened 2023-06-21
test(spec): shard out tests for rr

seaerchin:
## Problem
Previously, rr tests were in `dashboard.spec.ts` - this PR shifts them out to be in their own spec file.

Closes IS-254

## Solution
shift tests involving rr into another file, add a `before` + `beforeEach` hook to ensure pre-conditions satisfied for rr tests, nothing was changed in spec itself


alexanderleegs (approved):
lgtm

PR #1320, opened 2023-06-23
feat: set cookies with domain

harishv7:
## Problem

With Cypress upgrade, setting cookies needs an explicit domain set. In the absence of a domain, the cookies are not forwarded from FE to BE which causes the auth checks to fail and hence log out of the E2E tests.

Closes Is-270

## Solution

Create an util function that injects domains into the cookies from the env var (`CYPRESS_BACKEND_URL`)

**Breaking Changes**

  - Details ...

## Tests

Ensure E2E tests run properly both against local and staging

### DIFF
--- a/cypress/utils/cookies.ts
+++ b/cypress/utils/cookies.ts
@@ -1,0 +1,6 @@
+import { BACKEND_URL } from "../fixtures/constants"
+
+export const setCookieWithDomain = (name: string, value: string) => {
+  const baseUrl = new URL(BACKEND_URL)
+  cy.setCookie(name, value, { domain: baseUrl.host })
+}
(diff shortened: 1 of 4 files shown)


cypress/utils/cookies.ts
```diff
@@ -0,0 +1,6 @@
+import { BACKEND_URL } from "../fixtures/constants"
+
+export const setCookieWithDomain = (name: string, value: string) => {
+  const baseUrl = new URL(BACKEND_URL)
+  cy.setCookie(name, value, { domain: baseUrl.host })
+}
```
alexanderleegs, line 6: Nit: could we also have an optional param for options, to keep with the inbuilt functionality that setCookie has?

harishv7: added

alexanderleegs (approved):
lgtm

QiluXie:
lgtm

QiluXie (approved):

PR #1322, opened 2023-06-23
fix(ci-e2e): inject env var from workflow

seaerchin:
## Problem
Cypress env var was mistakenly put into netlify rather than workflow file -> e2e tests couldn't run 

## Solution
put env var into workflow


alexanderleegs (approved):
lgtm

PR #1323, opened 2023-06-26
IS-271: fix bugs on simple mde editor

harishv7:
## Problem

With the upgrade of mde editor to v5 in core-deps PR (https://github.com/isomerpages/isomercms-frontend/pull/1202/), the editor’s insert image and hyperlink does not seem to work.

Closes IS-271

## Solution

I removed the use of `mdeRef` which no long contained the `codemirror` object we were looking for inside the ref. I suppose this is due to breaking changes from v4 to v5. See https://www.npmjs.com/package/react-simplemde-editor?activeTab=readme#autosaving for more information.

From docs, the new suggested method to retrieve the editor or codemirror is to use the hooks & callbacks provided.

Have tested this locally via E2E on editPage. Results show all tests are passing:
Screenshot 2023-06-26 at 11 30 48 AM

Note: Due to React's Strict Mode, the file selection dialog (`MediaCreationModal`) gets re-rendered twice in development. However, this should not occur on production-ready builds.

**Breaking Changes**

  - Details ...

## Tests

- Ensure E2E tests, specifically `editPage.spec.ts` runs successfully
- Perform visual inspection by:
    - adding/removing images, 
    - inserting hyperlinks via editor and ensure it comes up fine on preview. 
    - Also test image names with spaces in between.
    - Test inserting files

### DIFF
--- a/src/components/pages/EditorModals.jsx
+++ b/src/components/pages/EditorModals.jsx
@@ -19,7 +19,10 @@ const EditorModals = ({
       setLinePos(lineAndCursor)
       simpleMde.codemirror.setSelection(lineAndCursor)
     }
-    if (lineAndCursor && lineAndCursor.line !== 0 && lineAndCursor.ch !== 0) {
+    if (
+      lineAndCursor &&
+      !(lineAndCursor.line === 0 && lineAndCursor.ch === 0)
+    ) {
       setLinePos(lineAndCursor)
     }
   }, [lineAndCursor])
@@ -80,6 +83,20 @@ EditorModals.propTypes = {
   onSave: PropTypes.func.isRequired,
   modalType: PropTypes.oneOf(["hyperlink", "media"]).isRequired,
   onClose: PropTypes.func.isRequired,
+  simpleMde: PropTypes.shape({
+    codemirror: PropTypes.shape({
+      getSelection: PropTypes.func,
+      getValue: PropTypes.func,
+      replaceSelection: PropTypes.func,
+      setCursor: PropTypes.func,
+    }),
+  }),
+  lineAndCursor: PropTypes.shape({
+    line: PropTypes.number,
+    ch: PropTypes.number,
+    sticky: PropTypes.string,
+    xRel: PropTypes.number,
+  }),
 }
 
 export default EditorModals
--- a/src/components/pages/EditorModals.jsx
+++ b/src/components/pages/EditorModals.jsx
@@ -1,18 +1,36 @@
 import HyperlinkModal from "components/HyperlinkModal"
 import MediaModal from "components/media/MediaModal"
 import PropTypes from "prop-types"
+import { useEffect, useState } from "react"
 
 const EditorModals = ({
-  mdeRef,
+  // mdeRef,
   onSave,
   modalType,
   onClose,
   mediaType,
   simpleMde,
+  lineAndCursor,
 }) => {
+  const [linePos, setLinePos] = useState({ line: 0, ch: 0, sticky: null })
+
+  console.log(`Current linepos`, linePos)
+
+  useEffect(() => {
+    if (!lineAndCursor && simpleMde) {
+      setLinePos(lineAndCursor)
+      simpleMde.codemirror.setSelection(lineAndCursor)
+    }
+    if (lineAndCursor && lineAndCursor.line !== 0 && lineAndCursor.ch !== 0) {
+      setLinePos(lineAndCursor)
+    }
+  }, [lineAndCursor])
+
   const onHyperlinkSave = (text, link) => {
     // const cm = mdeRef.current.simpleMde.codemirror
     const cm = simpleMde.codemirror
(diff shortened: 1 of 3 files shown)


alexanderleegs (commented):
Thanks for the quick fix, adding an existing image/file/link looks good now! I'm encountering a strange newly introduced bug for adding a new image/file now though, whether from this screen or from the actual files/images page, where the image/file to be uploaded needs to be selected twice - this issue doesn't seem to be present on staging currently, I'm wondering if the MediaModal is being rerendered for some reason and losing the initial data?

https://github.com/isomerpages/isomercms-frontend/assets/22111124/9013608e-82cc-4a7d-8c3f-c640bcd68516

harishv7:
> Thanks for the quick fix, adding an existing image/file/link looks good now! I'm encountering a strange newly introduced bug for adding a new image/file now though, whether from this screen or from the actual files/images page, where the image/file to be uploaded needs to be selected twice - this issue doesn't seem to be present on staging currently, I'm wondering if the MediaModal is being rerendered for some reason and losing the initial data?
> 
>  Screen.Recording.2023-06-26.at.11.56.40.AM.mov

@alexanderleegs thanks for pointing this out. After some debugging, this is happening because of React's strict mode which causes component re-render. Since our useeffect hook triggers the file dialog on component render, this causes the file dialog to be "clicked" twice

As such this should not be reproducible on prod-builds. Will be pushing this branch to staging to confirm this shortly

kishore03109:
@harishv7 
> @alexanderleegs thanks for pointing this out. After some debugging, this is happening because of React's strict mode which causes component re-render. Since our useeffect hook triggers the file dialog on component render, this causes the file dialog to be "clicked" twice
See https://stackoverflow.com/questions/58603209/react-hooks-render-twice for more info

hey, this is concerning right? the [usefulness of using strict mode ](https://react.dev/reference/react/StrictMode#reference)is so that we are aware of bugs caused by impure rendering or missing Effect cleanup. Is there a bug here that we should be aware of?

harishv7:
> @harishv7
> 
> > @alexanderleegs thanks for pointing this out. After some debugging, this is happening because of React's strict mode which causes component re-render. Since our useeffect hook triggers the file dialog on component render, this causes the file dialog to be "clicked" twice
> > See https://stackoverflow.com/questions/58603209/react-hooks-render-twice for more info
> 
> hey, this is concerning right? the [usefulness of using strict mode ](https://react.dev/reference/react/StrictMode#reference)is so that we are aware of bugs caused by impure rendering or missing Effect cleanup. Is there a bug here that we should be aware of?

hey @kishore03109 , yes it's a reflection that somewhere there is unnecessary re-render happening. Separately, I am also looking at where this could be happening along the tree

harishv7:
> > Thanks for the quick fix, adding an existing image/file/link looks good now! I'm encountering a strange newly introduced bug for adding a new image/file now though, whether from this screen or from the actual files/images page, where the image/file to be uploaded needs to be selected twice - this issue doesn't seem to be present on staging currently, I'm wondering if the MediaModal is being rerendered for some reason and losing the initial data?
> > Screen.Recording.2023-06-26.at.11.56.40.AM.mov
> 
> @alexanderleegs thanks for pointing this out. After some debugging, this is happening because of React's strict mode which causes component re-render. Since our useeffect hook triggers the file dialog on component render, this causes the file dialog to be "clicked" twice
> 
> As such this should not be reproducible on prod-builds. Will be pushing this branch to staging to confirm this shortly

See

> > @harishv7
> > > @alexanderleegs thanks for pointing this out. After some debugging, this is happening because of React's strict mode which causes component re-render. Since our useeffect hook triggers the file dialog on component render, this causes the file dialog to be "clicked" twice
> > > See https://stackoverflow.com/questions/58603209/react-hooks-render-twice for more info
> > 
> > 
> > hey, this is concerning right? the [usefulness of using strict mode ](https://react.dev/reference/react/StrictMode#reference)is so that we are aware of bugs caused by impure rendering or missing Effect cleanup. Is there a bug here that we should be aware of?
> 
> hey @kishore03109 , yes it's a reflection that somewhere there is unnecessary re-render happening. Separately, I am also looking at where this could be happening along the tree

@kishore03109 @alexanderleegs Confirmed on staging that this issue is NOT present. As for the root cause, in strict mode, React tries to re-render effects to ensure they have predictable mount + cleanup code. The issue in this component is that we are opening a file dialog on mount. On strict mode, this will happen twice as React tries to execute the useEffect twice. By right, we should have a cleanup returned in the useEffect. So the effect on mount opens the file dialog, and on cleanup closes the file dialog. However, there seems to be no programmatic way to close the file dialog with JS. See https://stackoverflow.com/questions/54919762/is-there-a-way-to-close-file-upload-dialog-using-javascript for more info.

Criticality of this issue seems to be low as there is no impact on prod.

Let me know if you have any other ideas!

Relevant docs: https://react.dev/reference/react/StrictMode#fixing-bugs-found-by-re-running-effects-in-development

PR #1324, opened 2023-06-28
feat(siteLaunchPad): prevent users from accessing site launchpad

kishore03109:
## Problem

Right now site launch feature hides behind a feature flag, but errant users could still access the url directly, which is not protected by this. 
[link to comment that sparked this pr ](https://github.com/isomerpages/isomercms-frontend/pull/1308#discussion_r1237182313)

## Solution

Redirect to dashboard + add toast  

https://github.com/isomerpages/isomercms-frontend/assets/42832651/eccf57e8-1940-4c59-b492-1549156d78c0

NOTE: this one got no screen so i just use __**creativity**__. @NatMaeTan let me know if this ok or not ya


src/layouts/SiteLaunchPad/components/SiteLaunchPadBlockedModel.tsx
seaerchin, line 1: Since we have a site launch context, why not prefer rerouting to the dashboard and display a toast?

Ie, we hide this behind a route so that users who see this don't even get a modal at all and are transparently redirected

NatMaeTan:
What happens if a logg-ed out user tries to enter the link? Do they get redirected to login?

Can we also change the copy to 

`This feature is in Beta and only open to a select group of users.`   Dont think we need the support link cos dont want to drive tickets to ops

kishore03109:
@NatMaeTan 
> What happens if a logg-ed out user tries to enter the link? Do they get redirected to login?
ya this is the case, it will be akin to going to the /sites/dashboard without being logged in the first place

alexanderleegs (commented):
Will need rebasing! Otherwise lgtm after addressing chin's point

src/layouts/SiteLaunchPad/components/SiteLaunchPadBlockedModel.tsx
kishore03109: @seaerchin hmm also can, but ill check in with Nat regarding what she prefer.

NatMaeTan: That's would actly be better ah so that users dont even have to click out, we just keep kicking them out haha. The copy can be:

`You do not have access to this page.`

Dont need to tell them much

kishore03109: @NatMaeTan done, can check the pr desc for the newer flow

NatMaeTan: LGTM

src/contexts/SiteLaunchContext.tsx
```diff
@@ -39,7 +41,10 @@ export const SiteLaunchProvider = ({
   initialStepNumber,
 }: SiteLaunchProviderProps): JSX.Element => {
   const { siteName } = useParams<{ siteName: string }>()
-
+  const [
+    isSiteLaunchBlockedToastShown,
+    setIsSiteLaunchBlockedToastShown,
+  ] = useState<boolean>(false)
```
seaerchin, line 47: could i get some context on why we need this? seems like it's to avoid multiple re-renders?

src/layouts/SiteLaunchPad/SiteLaunchPad.tsx
```diff
@@ -189,14 +198,18 @@ export const SiteLaunchPad = (): JSX.Element => {
...
+          {title}
+          {body}
+          <RiskAcceptanceModal
+            isOpen={pageNumber === SITE_LAUNCH_PAGES.RISK_ACCEPTANCE}
+            setPageNumber={setPageNumber}
+          />
+        </VStack>
+      ) : (
+        history.push(`/sites/${siteName}/dashboard`)
+      )}
```
seaerchin, line 212: can we prefer using a `Redirect` here? see [here](https://v5.reactrouter.com/web/api/Redirect) for details - more declarative and intuitive rather than putting an effect in our jsx.

src/contexts/SiteLaunchContext.tsx
kishore03109: yes plus when the redirection originally occurs, that happens from another page (`siteLaunchPad`). This is to store the state between these two pages.

src/layouts/SiteLaunchPad/SiteLaunchPad.tsx
kishore03109: good point, changed e869f3bb9c8c2432ef868054fb2c58599d686ab8!

src/contexts/SiteLaunchContext.tsx
seaerchin: Can we rebase on top of the toast PR to avoid having to do this in the context?

kishore03109: still need this bah to sync state between the two pages! is there a better way to do that without prop drilling?

src/layouts/SiteLaunchPad/components/SiteLaunchInfoGathering.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/jsx-props-no-spreading */
```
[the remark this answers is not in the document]
kishore03109: e50a49f717bed4b55d5326fed7906e56f5c825e4

src/contexts/SiteLaunchContext.tsx
kishore03109: to be clear ya, this is already rebased on top of the error toast pr already

seaerchin: seems like the main usage is 
```ts
if (!shouldUseSiteLaunchFeature(siteName)) {
    setIsSiteLaunchBlockedToastShown(true)
  }
``` in `SiteLaunchPad` - because this set state is given by a function call, it should not matter whether the function call happens in site launch pad or site dashboard. 

would be more in favour of 
```ts
useEffect(() => {
    if (!shouldUseSiteLaunchFeature(siteName)) {
      errorToast({
        id: "no_access_to_launchpad",
        description: "You do not have access to this page.",
      })
    }
  }, [
    errorToast,
    siteName
  ])
```

which makes the flow clearer and allows us to bypass all the set state + prop passing. the key thing here is to decide what's computed data and what's actually state. in this case, the decision to show the toast appears to be computed data so we should just use the computation directly. (also can hide the computation itself in context if you want)

kishore03109: hmm wait ah wouldnt this then re-render everytime the site dashboard is rendered?

seaerchin: why would it re-render? the `errorToast` dep already existed prior + `siteName` is constant for a given site, no?

seaerchin: resolved offline - put the `useEffect` in `SiteLaunchPad` works without having to go through this hassle

seaerchin (approved):

PR #1325, opened 2023-06-28
Feat/site launch integration with be

kishore03109:
## Problem

Add integration with BE 

## Solution

Note: the video is 2 mins + long, just recorded the whole thing to visualise the actual user flow for this (since it will take 2 minutes for them to get the dns records anyways). Can skip to the last portion of the video so see the end state!

https://github.com/isomerpages/isomercms-frontend/assets/42832651/46f411ff-19de-49aa-a302-cc9a0f2a42d0

add routes + invalidate query after 2mins since the call is async (by right the BE would be updated in 90 seconds, but putting 2 mins here as a buffer)

### DIFF
--- a/src/layouts/SiteLaunchPad/SiteLaunchPad.tsx
+++ b/src/layouts/SiteLaunchPad/SiteLaunchPad.tsx
@@ -177,7 +177,6 @@ export const SiteLaunchPad = (): JSX.Element => {
       title = <SiteLaunchChecklistTitle />
       body = (
         <SiteLaunchChecklistBody
-          setPageNumber={setPageNumber}
           handleIncrementStepNumber={handleIncrementStepNumber}
           handleDecrementStepNumber={handleDecrementStepNumber}
         />
--- a/src/layouts/SiteLaunchPad/SiteLaunchPad.tsx
+++ b/src/layouts/SiteLaunchPad/SiteLaunchPad.tsx
@@ -161,29 +161,6 @@ export const SiteLaunchPad = (): JSX.Element => {
     }
   }
 
-  const handleIncrementStepNumber = () => {
-    if (
-      siteLaunchStatusProps &&
-      siteLaunchStatusProps.stepNumber < SITE_LAUNCH_TASKS_LENGTH
-    ) {
-      setSiteLaunchStatusProps({
-        ...siteLaunchStatusProps,
-        stepNumber: (siteLaunchStatusProps.stepNumber +
-          1) as SiteLaunchTaskTypeIndex, // safe to assert since we do a check
-        siteLaunchStatus: "CHECKLIST_TASKS_PENDING",
-      })
-    }
-  }
-  const handleDecrementStepNumber = () => {
-    if (siteLaunchStatusProps && siteLaunchStatusProps.stepNumber > 0) {
-      setSiteLaunchStatusProps({
-        ...siteLaunchStatusProps,
-        stepNumber: (siteLaunchStatusProps?.stepNumber -
-          1) as SiteLaunchTaskTypeIndex, // safe to assert since we do a check
-      })
-    }
-  }
-
   let title: JSX.Element
   let body: JSX.Element
   switch (pageNumber) {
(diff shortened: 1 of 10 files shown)


src/layouts/SiteLaunchPad/SiteLaunchPad.tsx
```diff
@@ -160,6 +161,29 @@ export const SiteLaunchPad = (): JSX.Element => {
...
+  }
+  const handleDecrementStepNumber = () => {
+    if (siteLaunchStatusProps && siteLaunchStatusProps.stepNumber > 0) {
+      setSiteLaunchStatusProps({
+        ...siteLaunchStatusProps,
+        stepNumber: (siteLaunchStatusProps?.stepNumber -
+          1) as SiteLaunchTaskTypeIndex, // safe to assert since we do a check
+      })
+    }
+  }
```
alexanderleegs, line 185: These are duplicate definitions? Will need to remove one of them

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
```diff
@@ -385,14 +381,12 @@ export const SiteLaunchChecklistBody = ({
         </TableContainer>
       </Box>
 
-      <Box display="flex" justifyContent="flex-end">
-        <Button
-          variant="link"
-          ml="auto"
-          onClick={() => setPageNumber(SITE_LAUNCH_PAGES.DISCLAIMER)}
-        >
```
alexanderleegs, line 393: Do we not need the disclaimer anymore?

src/layouts/SiteLaunchPad/SiteLaunchPad.tsx
kishore03109: arh sorry, this was due to rebasing, somehow it duplicated the functions, fixed here 8133a6842261bb2da7ed7c31617fb7bba0e73f01

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
kishore03109: oh hey this is not a disclaimer, but a back page behaviour. Once you have agreed to the disclaimer, user should not be able to go back to disagree with the disclaimer anymore

src/layouts/SiteLaunchPad/components/SiteLaunchInfoGathering.tsx
```diff
@@ -55,16 +72,43 @@ export const SiteLaunchInfoCollectorBody = ({
...
+        Do you want to host the both the www and non-www version of your domain?
+      </Text>
+      <Text textStyle="subhead-2">
+        Select &quot;Yes&quot; if you are not sure
+      </Text>
+      <RadioGroup
+        {...register("useWww")}
+        onChange={setSelectedRadio}
+        mt="0.25rem"
+      >
```
seaerchin, line 86: double checking - i haven't actually tried this but it seems like it won't work as the `Radio` has their `value` set? feel free to resolve if you've tried before/you tried and it works

src/layouts/SiteLaunchPad/components/SiteLaunchInfoGathering.tsx
```diff
@@ -55,16 +72,43 @@ export const SiteLaunchInfoCollectorBody = ({
...
         What is the nature of the domain you are launching with?
       </Text>
-      <RadioGroup {...register("nature")} onChange={setSelectedRadio}>
-        <Radio value="new">
+      <RadioGroup
+        {...register("nature")}
+        onChange={setSelectedRadio}
+        mt="0.25rem"
+      >
+        <Radio value={SiteNature.New}>
```
seaerchin, line 106: are these just `"true" || "false"`?

seaerchin (approved):
lgtm - do we need to also remove the feature flag that was introduced in previous PRs?

src/layouts/SiteLaunchPad/components/SiteLaunchInfoGathering.tsx
kishore03109: they are "new" and "live" respectively

kishore03109:
No feature flag in FE will remain for a while

src/layouts/SiteLaunchPad/components/SiteLaunchInfoGathering.tsx
kishore03109: uhm they resolve to true and false uh. creating an object of 
{
"true":true,
"fasle":false
}

seems weird. Is there a better way I could have structured this?

PR #1328, opened 2023-06-28
Chore/modify e slint rule

kishore03109:
## Problem

I think as a team we have been ok with ignoring this rule in favour of cleaner code (ie we dont have to manually type out the props). This can be reflected by the mere fact that there are many files with this rule being ignored. 

Rather than having multiple ignore warnings, this PR changes that to allow such eslint rules to pass, rather than us needed to ignore this. 

## Solution

change eslint.rc

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -42,7 +42,7 @@
   "rules": {
     "no-shadow": "off",
     "@typescript-eslint/no-shadow": "error",
-    "react/jsx-props-no-spreading": "warn",
+    "react/jsx-props-no-spreading": "off",
     "react/require-default-props": "off",
     "import/prefer-default-export": "off",
     "react/prop-types": "warn",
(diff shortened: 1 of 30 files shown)


CHANGELOG.md
```diff
@@ -4,8 +4,16 @@ All notable changes to this project will be documented in this file. Dates are d
```
harishv7, line 4: can rebase to drop these commits?

.eslintrc.json
```diff
@@ -42,7 +42,7 @@
   "rules": {
     "no-shadow": "off",
     "@typescript-eslint/no-shadow": "error",
-    "react/jsx-props-no-spreading": "warn",
+    "react/jsx-props-no-spreading": "off",
```
harishv7, line 45: thanks for the clean up!

CHANGELOG.md
kishore03109: opps sorry never see properly

harishv7 (approved):

PR #1330, opened 2023-07-03
feat(review): add placeholder filetype and hide them in dashboard

caando:
This is the mirroring PR of https://github.com/isomerpages/isomercms-backend/pull/815 for frontend.

## Problem

Frontend does not have a filetype for the new placeholder `.keep` files. Review request dashboard preview currently shows changes to `.keep` files, which are implementation details that we would want to hide. 

Closes IS-236

## Solution

Add placeholder filetype into frontend. Hide placeholder files in review request dashboard previews. 

## Issues
As Github returns number of files changes with the placeholder file, the number of files change shown in the main dashboard and in the review request page is different (it does not have placeholder files). Possible solution can be a line of text is added to show `and X number of hidden files changed` to account for the difference.

### DIFF
--- a/src/layouts/ReviewRequest/components/RequestOverview/RequestOverview.tsx
+++ b/src/layouts/ReviewRequest/components/RequestOverview/RequestOverview.tsx
@@ -174,9 +174,6 @@ export const RequestOverview = ({
   const [path, setPath] = useState<string[]>([])
   const [title, setTitle] = useState("")
   const [itemStagingUrl, setItemStagingUrl] = useState<string | false>("")
-  const itemsWithoutPlaceholder = items.filter(
-    (item) => item.type !== "placeholder"
-  )
 
   return fileName ? (
     <DiffView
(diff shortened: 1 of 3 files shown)


src/types/reviewRequest.ts
```diff
@@ -12,7 +12,7 @@ export interface User {
   label: string
 }
 
-type FileType = "page" | "nav" | "setting" | "file" | "image"
+type FileType = "page" | "nav" | "setting" | "file" | "image" | "placeholder"
```
seaerchin, line 15: add another type like `DisplayedFileType` that's just `Omit<FileType, "placeholder">` and use it for the display level stuff

src/layouts/ReviewRequest/components/RequestOverview/RequestOverview.tsx
```diff
@@ -88,6 +88,10 @@ const getIcon = (iconType: EditedItemProps["type"]): JSX.Element => {
     case "image": {
       return <Icon {...ICON_STYLE_PROPS} as={BiImage} />
     }
+    case "placeholder": {
```
seaerchin, line 91: this method should take `DisplayedFileType`

caando:
Closing this PR as placeholder file type has been removed from backend, so frontend will never receive changes to placeholder files.

PR #1332, opened 2023-07-04
chore(build): update storybook to v7

seaerchin:
## Problem
Storybook has vulns that is fixed in v7.

## Solution
1. previously, our `bind` function for storybook templates weren't working properly due to it being typed as `never`; this has been changed to `Record<string, never>`, which is satisfied by an empty obj
2. we weren't starting up the `serviceWorker` previously, which led to mocked API calls failing. this was fixed by calling it when it's in development
3. updated `package.json` + `staticDirs` for storybook - weren't serving `/public` previously even though `-s public` was specified (this gives the static dir as an argument). have instead resorted to specifying it as an option

## Tests

**sanity checks**
- [ ] run `npm run storybook` 
- [ ] it should start up successfully
- [ ] go to `PublishedModal` in storybook (or any of the changed stories)
- [ ] should display correctly


package.json
```diff
@@ -88,8 +88,8 @@
     "cypress:open": "source .env && node scripts/run-e2e.js open",
     "prepare": "husky install",
     "version": "auto-changelog -p && git add CHANGELOG.md",
-    "storybook": "source .env && start-storybook -p 6006 -s public",
-    "build-storybook": "build-storybook -s public",
+    "storybook": "source .env && REACT_APP_ENV=test && storybook dev -p 6006",
+    "build-storybook": "REACT_APP_ENV=test && storybook build -s public",
```
kishore03109, line 92: Hmm may I get context why we need this? when devs run `npm run storybook`, do we need `REACT_APP_ENV` to be test? 
also, should we leverage on the existing process.env.NODE_ENV for this?

src/index.jsx
```diff
@@ -21,6 +21,10 @@ if (
   })
 }
 
+if (process.env.REACT_APP_ENV === "test") {
+  import("./mocks/browser").then(({ worker }) => worker.start())
```
kishore03109, line 25: when do _not_ want msws to run ah?

kishore03109 (commented):
Out of curiosity, why were we typing to <never> in the first ah?

package.json
seaerchin: i discovered this by accident **after** the PR so i didn't add it to the description but will explain it here: 

1. running the worker when `REACT_APP_ENV === LOCAL_DEV` means that for some API endpoints, they **will be mocked**. this will show you wrong values on our local env, 1 example wihch i encountered was showing `@mockUser` for the avatar rather than the actual user email
2. as a result of the above, we need `REACT_APP_ENV` to be something other than `LOCAL_DEV` - i chose test because it fits storybook's purpose and we need this env to allow for `msw` to mock our endpoints (for certain stories, we retrieve data from be) 
3. re: leveraging on existing `process.env.NODE_ENV` - this doesn't exist and is instead `REACT_APP_ENV`. we have to use something other than the existing one for the reasons given above

src/index.jsx
seaerchin: we only want msw to run for storybook. shouldn't be run in other environments because it will mock endpoints, returning false data

seaerchin:
> Out of curiosity, why were we typing to <never> in the first ah?

cos some stories take no args

src/index.jsx
kishore03109: oo right sorry I thought this was a `.story` file

package.json
kishore03109: right ok this makes sense to me

kishore03109 (approved):
wah they have night mode for storybook now p0 solid feature

harishv7 (approved):

PR #1333, opened 2023-07-05
Fix/is 276 toast rerender

alexanderleegs:
This PR fixes an issue with multiple rerenders of the toast components - the behaviour of react-hook-form and react-query seem to have changed after our dependencies update, and this caused issues for all of our toasts which were contained within `useEffect`, as the component would get rerendered, which changed the reference to `toast` and triggered the useEffect again.

The solution to this issue is to modify the toasts to use the `id` param - this prevents duplicate toasts from appearing ([link](https://chakra-ui.com/docs/components/toast#preventing-duplicate-toast)). The components which make use of toasts have been modified to support this.

Also resolves IS-284 once the privatisation PR has been rebased onto this one.

### DIFF
--- a/src/utils/dropdownUtils.js
+++ b/src/utils/dropdownUtils.js
@@ -12,7 +12,7 @@ const retrieveThirdNavOptions = async (
   collectionName,
   isExistingCollection
 ) => {
-  const errorToast = useErrorToast
+  const errorToast = useErrorToast()
   try {
     let thirdNavArr = []
     let allCollectionPages = []
@@ -35,6 +35,7 @@ const retrieveThirdNavOptions = async (
   } catch (err) {
     console.log(err)
     return errorToast({
+      id: "third-nav-error",
       description: `There was a problem trying to retrieve data from your repo. ${DEFAULT_RETRY_MSG}`,
     })
   }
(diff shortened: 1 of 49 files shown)


src/utils/dropdownUtils.js
```diff
@@ -12,7 +12,7 @@ const retrieveThirdNavOptions = async (
   collectionName,
   isExistingCollection
 ) => {
-  const errorToast = useErrorToast
+  const errorToast = useErrorToast()
```
kishore03109, line 15: huh so this never worked :sadge:

seaerchin (approved):

src/utils/toasts.ts
```diff
@@ -9,22 +10,40 @@ const DEFAULT_TOAST_PROPS = {
   isClosable: true,
 }
 
-export const useErrorToast = (): UseToastReturn =>
-  useToast({
+// Not specifying return type because type is in the argument
+const useIsomerToast = (args: UseToastProps) => {
```
kishore03109, line 14: wah I think we need to define a return type here, but fell into a rabbit hole for this cuz we got a version mismatch between chakra and design system. 
specifically, `danger` doesnt exist anymore, should be `error` now, but design system is lagging behind, actually not sure if this is already a bug in prod :sadge: 

Rather than putting this, could you create a ticket and assign it to me? is related to the design system changes

kishore03109 (approved):

PR #1338, opened 2023-07-12
fix(comments): disable comment drawer if rr has invalid status

seaerchin:
## Problem
Previously, we relied on redirection of users away from the review request to prevent them from taking any action when the review request is invalid. However, this leads to a delay between content paint and redirection (as we want the user to be able to interact with elements on the page), which results in users being able to access a old merged or closed rr, open the comments and see the comments prior to redirect. 

## Solution
pass an `isDisabled` prop. technically we could ping the server using `getReviewRequest` and check status in `CommentsDrawer` but because the component seems like its state should be controlled by the parent, prop passing is preferred instead (same computed data used for redir + disablign icon button)

next, if a rr fails to fetch from backend, we get the error message and redirect them to the dashboard.

## Manual tests
- [ ] Enter a link where the PR is previously merged/closed directly into your browser
- [ ] the icon button for comments drawer should be disabled

### DIFF
--- a/src/hooks/reviewHooks/useGetReviewRequest.ts
+++ b/src/hooks/reviewHooks/useGetReviewRequest.ts
@@ -1,3 +1,4 @@
+import { AxiosError } from "axios"
 import { useQuery } from "react-query"
 import type { UseQueryResult } from "react-query"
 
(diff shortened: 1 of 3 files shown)


kishore03109 (commented):
Screenshot 2023-07-13 at 10 16 43 AM
hey not sure if i am doing something wrong here, but visiting an invalid rr does not disable + redirect. Instead it just loads like this, do you happen to know why?

kishore03109 (commented):
Screenshot 2023-07-19 at 11 22 25 AM
I still get this loading state, may I check in if this behaviour is only replicable in my local machine?

seaerchin:
> Screenshot 2023-07-19 at 11 22 25 AM I still get this loading state, may I check in if this behaviour is only replicable in my local machine?

idk, i asked @caando to verify this prior cos i wanted to unblock his tests PR. i've updated the behaviour now so it shouldn't occur!

kishore03109 (approved):
lgtm, but please merge after cicd passes ya

PR #1316, opened 2023-06-21
Feat/repo privatisation

alexanderleegs:
**NOTE**
This feature is currently disabled on the end user side while we work on blocking access to the netlify site of privatised sites - will be added in a later pr.

Video for end to end user flow:
https://github.com/isomerpages/isomercms-frontend/assets/22111124/97a1d624-1e8a-479e-81b5-eb7a0d8f292f

## Problem

This PR introduces a repo privatisation feature to the settings page for repos. To be reviewed in conjunction with PR #[806](https://github.com/isomerpages/isomercms-backend/pull/806) on the isomercms-backend repo.

The repo privatisation field should only show up for Amplify sites. This solution automatically generates a password when first enabled by the user, but can be manually changed by the user. 

## Tests

- [ ] Navigate to the settings page of an amplify site (which is recorded in your local db, with entries in `sites` and `deployments`
- [ ] Test password functionality
  - [ ] Can add password
    - [ ] Verify that password works on amplify site and github repo is set to private
  - [ ] Can change password
    - [ ] Verify that new password works on amplify site and github repo is still private
  - [ ] Can remove password
    - [ ] Verify that no password needed for amplify site and github repo is set to public

Before deployment 

- DONE - The IAM role associated to our ec2 instances needs to add additional permissions to be able to modify amplify apps - we will need to modify the prod role as well before merging

PR #1342, opened 2023-07-14
Test/repo privatisation

alexanderleegs:
This PR adds tests for privatisation into our existing settings cypress tests. To facilitate this, the repo being used for the settings tests has been changed to the email login version, and the helper methods have been changed to fit the new changes.

Resolves IS-278

### DIFF
--- a/cypress/e2e/settings.spec.ts
+++ b/cypress/e2e/settings.spec.ts
@@ -1,7 +1,7 @@
 import {
   TEST_PRIMARY_COLOR,
-  TEST_REPO_NAME,
   BASE_SEO_LINK,
+  E2E_EMAIL_TEST_SITE,
   Interceptors,
 } from "../fixtures/constants"
 
@@ -19,6 +19,7 @@ describe("Settings page", () => {
   const BASE_CONTACT_US = "contact/"
   const BASE_FEEDBACK = "www.feedback.com"
   const BASE_FAQ = "/faq/"
+  const TEST_REPO_NAME = E2E_EMAIL_TEST_SITE.repo
 
   const TEST_TITLE = "Test title"
   const TEST_LOGO_IMAGES = [
@@ -50,7 +51,7 @@ describe("Settings page", () => {
     cy.visitLoadSettings(TEST_REPO_NAME, sitePath)
 
   before(() => {
-    cy.setGithubSessionDefaults()
+    cy.setEmailSessionDefaults("Email admin")
     cy.setupDefaultInterceptors()
 
     visitLoadSettings(`/sites/${TEST_REPO_NAME}/settings`).wait(
@@ -116,12 +117,11 @@ describe("Settings page", () => {
   })
 
   beforeEach(() => {
-    cy.setGithubSessionDefaults()
-    visitLoadSettings(`/sites/${TEST_REPO_NAME}/settings`)
-    // Double check that settings are loaded before running tests cos sometimes the tests run too quickly
-    cy.get("#title").should((elem) => {
-      expect(elem.val()).to.have.length.greaterThan(0)
-    })
+    cy.setEmailSessionDefaults("Email admin")
+    cy.setupDefaultInterceptors()
+    visitLoadSettings(`/sites/${TEST_REPO_NAME}/settings`).wait(
+      Interceptors.GET
+    )
     cy.contains("Verify").should("not.exist")
   })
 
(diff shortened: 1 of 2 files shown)


cypress/e2e/settings.spec.ts
```diff
@@ -245,6 +245,130 @@ describe("Settings page", () => {
...
+    cy.get("@privatiseStaging")
+      .check({ force: true, timeout: 0 })
+      .then(() => {
+        // Checking within then because of wonky focus preventing the field from being populated
+        cy.contains("You should change this every 90 days")
+          .parent()
+          .next()
+          .find("input")
+          .as("passwordInput")
+        cy.wait(1000)
```
harishv7, line 270: what is this wait for ?

harishv7 (approved):

cypress/e2e/settings.spec.ts
alexanderleegs: I was getting wonky results when attempting to inspect the contents of the password field immediately for some reason - I think it might have been because of the auto generation + population of the field taking some time. I've opted to add in a wait time here as this made the test less flaky, but open to suggestions on making this better!

harishv7: got it

seaerchin: i can work with you on this - this shouldn't block the PR cos it works fine as-is but it is an issue because static wait times, if propagated, either waits for no reason (when wait time is insufficient) or causes extra delay (resolve early prior to wait). 

the trick to solve this is to know _what_ signals completion and either assert on that or `wait` on the request

cypress/e2e/settings.spec.ts
```diff
@@ -245,6 +245,130 @@ describe("Settings page", () => {
...
+  })
+
+  it("Should regenerate a password if password field is emptied and privatise button toggled again", () => {
+    // Arrange
+    cy.contains("Privatise staging site")
+      .parent()
+      .parent()
+      .find("input")
+      .as("privatiseStaging")
+    cy.get("@privatiseStaging").should("be.checked")
```
seaerchin, line 288: this has implicit dep on prev test so if the prev one fail, this will fail too ._.

cypress/e2e/settings.spec.ts
```diff
@@ -245,6 +245,130 @@ describe("Settings page", () => {
...
+      .check({ force: true })
+      .then(() => {
+        // Assert
+        cy.contains("You should change this every 90 days")
+          .parent()
+          .next()
+          .find("input")
+          .as("passwordInput")
+        cy.wait(1000)
+        cy.get("@passwordInput").should("have.length.gt", 0)
```
seaerchin, line 310: we should aim to make some simple assertions about the nature of this pw - eg, that it is random (might be hard) or that it is regenerated **per toggle** (easy + impt). 

given that this is regarding security, i'm of the mind to make our invariants super clear

src/hooks/settingsHooks/useGetSettings.ts
```diff
@@ -127,17 +127,25 @@ const extractPassword = (
...
+      let passwordSettings
+      const isLaunchDarklyImplemented = false
+      if (shouldGetPrivacyDetails && isLaunchDarklyImplemented) {
+        // TODO: LaunchDarkly to allow specific groups to access this feature first
+        passwordSettings = await SettingsService.getPassword({ siteName })
+      } else {
+        passwordSettings = {
+          isAmplifySite: false,
+          password: "",
+        }
```
seaerchin, line 148: why's this here ._. also, we should remove the toggle as a whole rather than simply changing the request (+ enforce on BE also)

we should also use an env var for this because having it in code requires a full deploy, which defeats the purpose

src/layouts/Settings/Settings.tsx
```diff
@@ -34,11 +36,13 @@ import { SocialMediaSettings } from "./SocialMediaSettings"
 
 export const Settings = (): JSX.Element => {
   const { siteName } = useParams<{ siteName: string }>()
+  const { userId } = useLoginContext()
+  const isGithubUser = userId !== "Unknown user" && !!userId
```
seaerchin, line 40: could we add a note explaining hwy this means that it is a github user?

seaerchin (commented):
non-blockers but good to haves that will help us in future.

src/hooks/settingsHooks/useGetSettings.ts
alexanderleegs: Accidental merge into the wrong branch - will put this in the main branch!

src/layouts/Settings/Settings.tsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/1316/commits/4c6368275c31a6c5cf27bacac104a1ac9207f9bb

src/hooks/settingsHooks/useGetSettings.ts
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/1316/commits/b26ab91a104f550adf0b34beb59e0f6d1654de41

alexanderleegs:
> @alexanderleegs is this ready to be merged?

Will need to address chin's comments first - but currently merging these tests are not particularly useful, because they always result in failure since the privatisation feature is hidden!

alexanderleegs:
Tests are erroring because of the settings being a conditional field for email login repos only - going to close this for now, we can reference later on when creating tests for playwright for settings

PR #1343, opened 2023-07-14
feat(e2e): e2e tests for comments

caando:
Added E2E tests for comments feature of review requests.

Closes IS-258

**Note**

Occasionally `cy.request` fails and return undefined for reasons unknown. This will cause the tests to break.

## Tests

- [ ] Smoke tests

## Deploy Notes

Requires #1338 to be merged first.

### DIFF
--- a/cypress/api/reviewRequest.api.ts
+++ b/cypress/api/reviewRequest.api.ts
@@ -35,24 +35,20 @@ export const closeReviewRequests = (): void => {
   })
 }
 
-function removeResponse(response: any) {
-  // Do nothing
-}
-
 export const approveReviewRequest = (
   id: number
 ): Cypress.Chainable<Cypress.Response<void>> => {
-  return cy.request("POST", `${BASE_URL}/${id}/approve`).then(removeResponse)
+  return cy.request("POST", `${BASE_URL}/${id}/approve`)
 }
 
 export const mergeReviewRequest = (
   id: number
 ): Cypress.Chainable<Cypress.Response<void>> => {
-  return cy.request("POST", `${BASE_URL}/${id}/merge`).then(removeResponse)
+  return cy.request("POST", `${BASE_URL}/${id}/merge`)
 }
 
 export const closeReviewRequest = (
   id: number
 ): Cypress.Chainable<Cypress.Response<void>> => {
-  return cy.request("DELETE", `${BASE_URL}/${id}`).then(removeResponse)
+  return cy.request("DELETE", `${BASE_URL}/${id}`)
 }
--- a/cypress/api/reviewRequest.api.ts
+++ b/cypress/api/reviewRequest.api.ts
@@ -35,20 +35,24 @@ export const closeReviewRequests = (): void => {
   })
 }
 
+function removeResponse(response: any) {
+  // Do nothing
+}
+
 export const approveReviewRequest = (
   id: number
-): Cypress.Chainable<Cypress.Response<any>> => {
-  return cy.request("POST", `${BASE_URL}/${id}/approve`)
+): Cypress.Chainable<Cypress.Response<void>> => {
+  return cy.request("POST", `${BASE_URL}/${id}/approve`).then(removeResponse)
 }
 
 export const mergeReviewRequest = (
   id: number
-): Cypress.Chainable<Cypress.Response<any>> => {
-  return cy.request("POST", `${BASE_URL}/${id}/merge`)
+): Cypress.Chainable<Cypress.Response<void>> => {
+  return cy.request("POST", `${BASE_URL}/${id}/merge`).then(removeResponse)
 }
 
 export const closeReviewRequest = (
   id: number
-): Cypress.Chainable<Cypress.Response<any>> => {
-  return cy.request("DELETE", `${BASE_URL}/${id}`)
+): Cypress.Chainable<Cypress.Response<void>> => {
+  return cy.request("DELETE", `${BASE_URL}/${id}`).then(removeResponse)
 }
--- a/cypress/api/reviewRequest.api.ts
+++ b/cypress/api/reviewRequest.api.ts
@@ -30,11 +30,25 @@ export const listReviewRequests = (): Cypress.Chainable<{ id: number }[]> => {
 export const closeReviewRequests = (): void => {
   listReviewRequests().then((reviewRequests) => {
     reviewRequests.forEach(({ id }) => {
-      cy.request("DELETE", `${BASE_URL}/${id}`)
+      closeReviewRequest(id)
     })
   })
 }
 
-export const approveReviewRequest = (id: number): void => {
-  cy.request("POST", `${BASE_URL}/${id}/approve`)
+export const approveReviewRequest = (
+  id: number
+): Cypress.Chainable<Cypress.Response<any>> => {
+  return cy.request("POST", `${BASE_URL}/${id}/approve`)
+}
+
+export const mergeReviewRequest = (
+  id: number
+): Cypress.Chainable<Cypress.Response<any>> => {
+  return cy.request("POST", `${BASE_URL}/${id}/merge`)
+}
+
+export const closeReviewRequest = (
+  id: number
+): Cypress.Chainable<Cypress.Response<any>> => {
+  return cy.request("DELETE", `${BASE_URL}/${id}`)
 }
--- a/cypress/e2e/comments.spec.ts
+++ b/cypress/e2e/comments.spec.ts
@@ -45,11 +45,6 @@ const checkCommentVisible = (
 const checkCommentDisabled = () =>
   cy.get(COMMENTS_DRAWER_BUTTON_SELECTOR).should("be.disabled")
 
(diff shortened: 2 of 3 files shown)


cypress/api/reviewRequest.api.ts
```diff
@@ -30,11 +30,25 @@ export const listReviewRequests = (): Cypress.Chainable<{ id: number }[]> => {
...
-  cy.request("POST", `${BASE_URL}/${id}/approve`)
+export const approveReviewRequest = (
+  id: number
+): Cypress.Chainable<Cypress.Response<any>> => {
+  return cy.request("POST", `${BASE_URL}/${id}/approve`)
+}
+
+export const mergeReviewRequest = (
+  id: number
+): Cypress.Chainable<Cypress.Response<any>> => {
```
seaerchin, line 46: for the below 2, can we give the type as `void` instead of `any`?

cypress/e2e/comments.spec.ts
```diff
@@ -28,17 +30,33 @@ const MOCK_COMMENT = "mock comment"
...
+    .contains(userEmail)
+    .should("be.visible")
+}
+
+const checkCommentDisabled = () =>
+  cy.get(COMMENTS_DRAWER_BUTTON_SELECTOR).should("be.disabled")
+
 const USER_NOT_AUTHORISED_COMMENT_ERROR_MESSAGE =
   "Only collaborators of a site can view review request comments!"
+const NOTIFICATION_CANNOT_BE_RETRIVED_ERROR_MESSAGE =
```
seaerchin, line 50: these seem to be unused so feel free to remove em

cypress/e2e/comments.spec.ts
```diff
@@ -100,32 +114,73 @@ describe("Comments", () => {
...
     it("should not allow empty comments to be posted", () => {
       // Arrange
+      visitE2eEmailTestRepo()
+
       // Act
+      openReviewRequest()
+      cy.wait(`@${COMMENT_INTERCEPTOR}`)
+      ignoreResizeError()
+      openCommentsDrawer()
+      getCommentInput().type(MOCK_COMMENT).clear().blur()
```
seaerchin, line 129: can we also check that it is disabled on initial load?

cypress/e2e/comments.spec.ts
```diff
@@ -28,17 +30,33 @@ const MOCK_COMMENT = "mock comment"
...
 
-const submitComment = () => cy.get(SUBMIT_COMMENT_BUTTON_SELECTOR).click()
 const COMMENT_INTERCEPTOR = "getComments"
 
 describe("Comments", () => {
   beforeEach(() => {
     cy.setEmailSessionDefaults("Email admin")
+    cy.intercept("GET", "**/comments").as(COMMENT_INTERCEPTOR)
     cy.setupDefaultInterceptors()
-    visitE2eEmailTestRepo()
```
seaerchin, line 41: why was this line deleted?

cypress/e2e/comments.spec.ts
```diff
@@ -175,46 +228,129 @@ describe("Comments", () => {
...
         // before the redirect is done
-        openCommentsDrawer()
-        cy.get(COMMENTS_INPUT_SELECTOR).type(MOCK_COMMENT).blur()
-        submitComment()
 
-        // Assert
-        cy.contains(USER_NOT_AUTHORISED_COMMENT_ERROR_MESSAGE).should(
-          "be.visible"
-        )
+        // // Assert
```
seaerchin, line 232: nit: remove double comment

cypress/e2e/comments.spec.ts
```diff
@@ -175,46 +228,129 @@ describe("Comments", () => {
         // but on `localhost`, cypress is fast enough
         // to execute the commands + assertion
         // before the redirect is done
-        openCommentsDrawer()
-        cy.get(COMMENTS_INPUT_SELECTOR).type(MOCK_COMMENT).blur()
-        submitComment()
```
seaerchin, line 180: this is safe to remove cos it's rebased off the fix branch right?

cypress/e2e/comments.spec.ts
```diff
@@ -153,8 +208,6 @@ describe("Comments", () => {
     describe("not a site member", () => {
       before(() => {
         removeOtherCollaborators()
-      })
-      beforeEach(() => {
```
seaerchin, line 157: why remove this? if we do some permission changes in the test, this change makes it so that the permission changes will propagate to other tests in the same suite

cypress/e2e/comments.spec.ts
```diff
@@ -175,46 +228,129 @@ describe("Comments", () => {
...
+        // as the backend recognises that we lack sufficient permissions
+        // in order to view this site
+        ignoreAuthError()
+        ignoreNotFoundError()
+        // NOTE: There is a redirect being done
+        // but on `localhost`, cypress is fast enough
+        // to execute the commands + assertion
+        // before the redirect is done
+
+        // // Assert
```
seaerchin, line 257: nit: remove double comment

cypress/e2e/comments.spec.ts
```diff
@@ -175,46 +228,129 @@ describe("Comments", () => {
...
       })
-      it("should not be able to create comments for closed review requests", () => {
+      it("should not be able to see/create comments for closed review requests", () => {
         // Arrange
         // Act
+        cy.visit(
+          `${CMS_BASEURL}/sites/${E2E_EMAIL_TEST_SITE.repo}/review/${reviewId}`
+        )
         // Assert
+        checkCommentDisabled()
```
seaerchin, line 335: wrong assumption - this only checks on initial load if the **button** is disabled. we need to also assert that the button is disabled _after_ the user types something into the input. (if the assertion in `checkCommentDisabled` was instead changed to be both button + input then this would be ok)

cypress/e2e/comments.spec.ts
```diff
@@ -175,46 +228,129 @@ describe("Comments", () => {
...
+      it("should not be able to see/create comments for non-existent review requests", () => {
         // Arrange
+        visitE2eEmailTestRepo()
+
         // Act
+
         // Assert
+        cy.contains(
+          `a[href^="/sites/${E2E_EMAIL_TEST_SITE.repo}/review/"]`
+        ).should("not.exist")
```
seaerchin, line 353: at the point of testing this step, are we not still at the e2e test repo (url is `/dashboard`) and not the actual `/review/<id>` url?

src/layouts/ReviewRequest/Dashboard.tsx
seaerchin, line 1: should rebase off

jandonyada:
Brother @caando , pls resolve the PR comments so we can close this chapter and move on with life 👼

cypress/e2e/comments.spec.ts
caando: For this test there is no review requests, so it goes to the site dashboard, then checks if there is a link to the currently opened review request. This exerts that the link does not exist.

cypress/e2e/comments.spec.ts
caando: For this test the user is unable to open the comments drawer to type stuff into the input box. Do you think I should skip this step and force the comment box to open?

cypress/e2e/comments.spec.ts
caando: Doesn't affect tests because most of them visit either this page or other pages on set up. I can add it back

cypress/e2e/comments.spec.ts
caando: The second test for `not a site member` does `setUserAsUnauthorised()` twice. Moving it to test set up instead.

cypress/e2e/comments.spec.ts
caando: This is passing test thou, users are supposed to be able to submit comments

cypress/api/reviewRequest.api.ts
```diff
@@ -30,11 +30,29 @@ export const listReviewRequests = (): Cypress.Chainable<{ id: number }[]> => {
...
     reviewRequests.forEach(({ id }) => {
-      cy.request("DELETE", `${BASE_URL}/${id}`)
+      closeReviewRequest(id)
     })
   })
 }
 
-export const approveReviewRequest = (id: number): void => {
-  cy.request("POST", `${BASE_URL}/${id}/approve`)
+function removeResponse(response: any) {
```
seaerchin, line 38: why do we need this function?

cypress/e2e/comments.spec.ts
seaerchin: oh yea we updated the behaviour - ignore the above comment! (don't force open)

cypress/e2e/comments.spec.ts
seaerchin: ah ok this sounds logical - could we add this description above?

seaerchin (approved):
lgtm. please add the description of the link!

PR #1344, opened 2023-07-18
Fix/restrict repo privatisation to email login

alexanderleegs:
This PR adds an additional restriction to the repo privatisation feature such that only email login users can see the feature - this is due to the prerequisite of restricting github access before being able to convert repos to private, as private repos are unable to set additional branch protection rules, meaning the PR approval requirement can be bypassed. To be reviewed in conjunction with PR #[840](https://github.com/isomerpages/isomercms-backend/pull/840) on the isomercms-backend repo.

### DIFF
--- a/src/hooks/settingsHooks/useGetSettings.ts
+++ b/src/hooks/settingsHooks/useGetSettings.ts
@@ -127,17 +127,26 @@ const extractPassword = (
 }
 
 export const useGetSettings = (
-  siteName: string
+  siteName: string,
+  isEmailLogin: boolean
 ): UseQueryResult<SiteSettings> => {
   return useQuery<SiteSettings>(
     [SETTINGS_CONTENT_KEY, siteName],
     async () => {
       const siteSettings = await SettingsService.get({ siteName })
-      // TODO: reimplement actual functionality once netlify issue resolved
-      // const passwordSettings = await SettingsService.getPassword({ siteName })
-      const passwordSettings = {
-        isAmplifySite: false,
-        password: "",
+      let passwordSettings
+      if (isEmailLogin) {
+        // TODO: reimplement actual functionality once netlify issue resolved
+        // passwordSettings = await SettingsService.getPassword({ siteName })
+        passwordSettings = {
+          isAmplifySite: false,
+          password: "",
+        }
+      } else {
+        passwordSettings = {
+          isAmplifySite: false,
+          password: "",
+        }
       }
       const convertedSettings = convertFromBe(siteSettings)
       const parsedPassword = extractPassword(passwordSettings)
--- a/src/hooks/settingsHooks/useGetSettings.ts
+++ b/src/hooks/settingsHooks/useGetSettings.ts
@@ -128,20 +128,19 @@ const extractPassword = (
 
 export const useGetSettings = (
   siteName: string,
-  isEmailLogin: boolean
+  isEmailLogin?: boolean
 ): UseQueryResult<SiteSettings> => {
+  const shouldGetPrivacyDetails =
+    isEmailLogin === undefined ? false : isEmailLogin
   return useQuery<SiteSettings>(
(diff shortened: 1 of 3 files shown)


src/hooks/settingsHooks/useGetSettings.ts
```diff
@@ -127,17 +127,26 @@ const extractPassword = (
...
+        // passwordSettings = await SettingsService.getPassword({ siteName })
+        passwordSettings = {
+          isAmplifySite: false,
+          password: "",
+        }
+      } else {
+        passwordSettings = {
+          isAmplifySite: false,
+          password: "",
+        }
```
kishore03109, line 149: Hey, in both control flows the password settings get set to the same value, is this intended?

[Style, nit]: Since we have version control, I prefer not to have commented code in production

src/layouts/EditContactUs.jsx
```diff
@@ -205,7 +205,7 @@ const EditContactUs = ({ match }) => {
     onClose: onDeleteModalClose,
   } = useDisclosure()
   const { data: contactUsDetails } = useGetContactUsHook(siteName)
-  const { data: settingsData } = useGetSettings(siteName)
+  const { data: settingsData } = useGetSettings(siteName, false) // Doesn't need the privatisation data, so we can always put this as a github login user
```
kishore03109, line 208: sanity check here: this should be another todo right?

src/layouts/Settings/Settings.tsx
```diff
@@ -34,11 +36,13 @@ import { SocialMediaSettings } from "./SocialMediaSettings"
 
 export const Settings = (): JSX.Element => {
   const { siteName } = useParams<{ siteName: string }>()
+  const { userId } = useLoginContext()
+  const isGithubUser = userId !== "Unknown user" && !!userId
```
kishore03109, line 40: shall we add a todo here as well? Since this would prob have to be modified as this is not a real `isGithubUser`?

kishore03109 (commented):
Hey sorry i am abit confused about this. Shouldn't the control flow differ here for amplify vs non amplify sites? 
correct me if I am wrong here ya, but what this change does is that it effectively blocks this feature from all sites right?

src/layouts/EditContactUs.jsx
alexanderleegs: Nope, this is a comment for clarity - we never need the privacy settings here, and `useGetSettings` does not retrieve privacy settings if the user is from github - this is just a small optimisation

src/layouts/Settings/Settings.tsx
alexanderleegs: What do you mean by this sorry? `userId` is a field that's only present on github login (or people not logged in completely)

alexanderleegs:
> Hey sorry i am abit confused about this. Shouldn't the control flow differ here for amplify vs non amplify sites? correct me if I am wrong here ya, but what this change does is that it effectively blocks this feature from all sites right?

Yep, we're actually releasing this feature completely blocked from all sites - we'll slowly start to roll this out with proper feature flags after releasing LaunchDarkly

src/hooks/settingsHooks/useGetSettings.ts
alexanderleegs: As discussed offline - we're making this clearer that the branch is blocked off until launchDarkly! https://github.com/isomerpages/isomercms-frontend/pull/1344/commits/83e780e3743cf8b573cf942a0ba576ddf03f316d

src/layouts/EditContactUs.jsx
alexanderleegs: As discussed offline, we're changing the second argument to an optional value - https://github.com/isomerpages/isomercms-frontend/pull/1344/commits/83e780e3743cf8b573cf942a0ba576ddf03f316d

kishore03109 (approved):

PR #1350, opened 2023-07-19
Fix: parse env var from string to boolean

alexanderleegs:
This PR fixes an issue with our env vars - netlify passes env vars as strings, which means that interpreting it as a boolean value does not work as expected. This PR modifies the way we parse the `REACT_APP_IS_SITE_PRIVATISATION_ACTIVE` to ensure we parse the boolean value correctly.

### DIFF
--- a/src/hooks/settingsHooks/useGetSettings.ts
+++ b/src/hooks/settingsHooks/useGetSettings.ts
@@ -16,6 +16,10 @@ import {
 
 import { BE_TO_FE } from "./constants"
 
+const IS_SITE_PRIVATISATION_ACTIVE =
+  process.env.REACT_APP_IS_SITE_PRIVATISATION_ACTIVE &&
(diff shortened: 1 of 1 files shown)


kishore03109 (commented):
hey just to check in, does this mean that local and netlify behaviour are the same in terms of parsing the env var as string? Ideally we would want the same behaviour in local and in prod ya. May I also get your help to update 1pw for this please

alexanderleegs:
> hey just to check in, does this mean that local and netlify behaviour are the same in terms of parsing the env var as string? Ideally we would want the same behaviour in local and in prod ya. May I also get your help to update 1pw for this please

Yeah they are - to circumvent this we can put the local env var as `export REACT_APP_IS_SITE_PRIVATISATION_ACTIVE="false"` - this forces it to be interpreted as string. I've updated the 1pw!

kishore03109 (approved):

PR #1351, opened 2023-07-19
feat(form): add NPS modal

seaerchin:
## Problem
right now isomer has no way of collecting feedback from users. this adds a modal that shows an embedded form, with some additional logic to avoid repetitive pings on the user.\

see the modal [here](https://www.chromatic.com/component?appId=61fcc8e14c3af9003a102ac9&csfId=components-feedbackmodal&buildNumber=1868&k=64ba05a458dfcb14aa7fc25f-1200-interactive-true&h=12&b=-3)

Closes IS-324

## Solution
1. add modal, set it on `SiteEditLayout` -> this is because the `SiteEditLayout` is set on almost all editing layouts for our CMS
2. write hook to determine when to trigger the modal
    - currently, this is in intervals of 3 wks since the last seen timing (or if it's the first ever view) + whether the env var is set. the interval of 1 wk was chosen so that there's enough time lapse between toggles and so that we can distinguish between runs of showing our modal. (ie, if the env var is set but you've prev done the survey within a time interval of 3 wks, we don't show again)
    - we will trigger the modal every time the user navigates away from an editing page. (agreed w/ PM -> want to make sure that they have done something) 

- `REACT_APP_SHOW_NPS_FORM` : decides whether to show the NPS form. leave empty if not required.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "prepare": "husky install",
     "version": "auto-changelog -p && git add CHANGELOG.md",
     "storybook": "source .env && REACT_APP_ENV=test && storybook dev -p 6006",
-    "build-storybook": "REACT_APP_ENV=test && storybook build -s public",
+    "build-storybook": "REACT_APP_ENV=test && storybook build",
     "chromatic": "npx chromatic --project-token=$CHROMATIC_PROJECT_TOKEN",
     "sourcemaps:upload": "npx datadog-ci sourcemaps upload ./build --service=isomercms-frontend --release-version=$REACT_APP_VERSION --minified-path-prefix=$REACT_APP_URL"
   },
--- a/src/components/FeedbackModal/FeedbackModal.tsx
+++ b/src/components/FeedbackModal/FeedbackModal.tsx
@@ -1,0 +1,76 @@
+import {
+  Modal,
+  ModalOverlay,
+  ModalContent,
+  ModalHeader,
+  ModalBody,
+  Heading,
+} from "@chakra-ui/react"
+import { ModalCloseButton } from "@opengovsg/design-system-react"
+
+const FeedbackForm = () => {
+  return (
+    <>
+      <div
+        style={{
+          fontFamily: "sans-serif",
+          fontSize: "15px",
+          color: "#000",
+          opacity: "0.9",
+          paddingTop: "5px",
+          paddingBottom: "8px",
+        }}
+      >
+        If the form below is not loaded, you can also fill it in{" "}
+        <a href="https://form.gov.sg/64b76a1cb319a90012e8e8db">here</a>.
+      </div>
+
+      <iframe
+        title="FormSG Feedback form for Isomer"
+        id="iframe"
+        src="https://form.gov.sg/64b76a1cb319a90012e8e8db"
+        width="100%"
+        height="650px"
+      />
+
+      <div
+        style={{
+          fontFamily: "sans-serif",
+          fontSize: "12px",
+          color: "#999",
+          opacity: "0.5",
+          paddingTop: "5px",
+        }}
+      >
+        Powered by{" "}
+        <a href="https://form.gov.sg" color="#999">
+          Form
+        </a>
+      </div>
+    </>
+  )
+}
+
+export interface FeedbackModalProps {
+  isOpen: boolean
+  onClose: () => void
+}
+export const FeedbackModal = ({
+  isOpen,
+  onClose,
+}: FeedbackModalProps): JSX.Element => {
+  return (
+    <Modal isOpen={isOpen} onClose={onClose}>
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader>
+          <Heading as="h4">Help make Isomer better!</Heading>
+        </ModalHeader>
+        <ModalCloseButton />
+        <ModalBody pb={6}>
+          <FeedbackForm />
+        </ModalBody>
+      </ModalContent>
+    </Modal>
+  )
+}
--- a/src/components/FeedbackModal/FeedbackModal.tsx
+++ b/src/components/FeedbackModal/FeedbackModal.tsx
@@ -22,13 +22,13 @@ const FeedbackForm = () => {
         }}
       >
         If the form below is not loaded, you can also fill it in{" "}
-        <a href="https://form.gov.sg/64b76a1cb319a90012e8e8db">here</a>.
+        <a href="https://form.gov.sg/64b7a11823e54700118bad90">here</a>.
       </div>
 
       <iframe
         title="FormSG Feedback form for Isomer"
         id="iframe"
-        src="https://form.gov.sg/64b76a1cb319a90012e8e8db"
+        src="https://form.gov.sg/64b7a11823e54700118bad90"
         width="100%"
         height="650px"
       />
--- a/src/hooks/useFeedbackDisclosure.ts
+++ b/src/hooks/useFeedbackDisclosure.ts
@@ -23,15 +23,15 @@ type UseFeedbackStorageReturn = readonly [number, () => void]
 
 // NOTE: Wrapper to handle get/set of individual keys
 const useFeedbackStorage = (): UseFeedbackStorageReturn => {
-  const { userId } = useLoginContext()
+  const { displayedName } = useLoginContext()
   const [
     userMappings,
     setUserMappings,
   ] = useLocalStorage<FeedbackStorageMappings>(LOCAL_STORAGE_KEYS.Feedback, {})
 
-  const lastSeen = userMappings[userId]
+  const lastSeen = userMappings[displayedName]
   const setLastSeen = () => {
-    setUserMappings({ ...userMappings, [userId]: Date.now() })
+    setUserMappings({ ...userMappings, [displayedName]: Date.now() })
   }
   // NOTE: `const` cast to infer types properly
   // so that it knows that the return value is a tuple
--- a/src/hooks/useFeedbackDisclosure.ts
+++ b/src/hooks/useFeedbackDisclosure.ts
@@ -1,0 +1,66 @@
+import { useDisclosure } from "@chakra-ui/react"
+import { useLocation } from "react-router-dom"
+
+import { LOCAL_STORAGE_KEYS } from "constants/localStorage"
+
+import { useLoginContext } from "contexts/LoginContext"
+
+import { isEditPageUrl, isSpecialPagesUrl } from "utils/pages"
+
+import { useLocalStorage } from "./useLocalStorage"
+
+type LastSeenFeedbackTime = number
+type UserId = string
+
+type FeedbackStorageMappings = Record<UserId, LastSeenFeedbackTime>
+
+const NPS_SURVEY_INTERVAL = 1814400000 as const // 3 weeks
+
+const { REACT_APP_SHOW_NPS_FORM } = process.env
+
+// NOTE: Wrapper to handle get/set of individual keys
+const useFeedbackStorage = () => {
+  const { userId } = useLoginContext()
+  const [
+    userMappings,
+    setUserMappings,
+  ] = useLocalStorage<FeedbackStorageMappings>(LOCAL_STORAGE_KEYS.Feedback, {})
+
+  const lastSeen = userMappings[userId]
+  const setLastSeen = () => {
+    setUserMappings({ ...userMappings, [userId]: Date.now() })
+  }
+  // NOTE: `const` cast to infer types properly
+  // so that it knows that the return value is a tuple
+  return [lastSeen, setLastSeen] as const
+}
+
+export const useFeedbackDisclosure = () => {
+  const location = useLocation<{ from?: string }>()
+  // NOTE: Despite the typing from the library,
+  // state is NOT guaranteed to exist.
+  const from = location.state?.from ?? ""
+  // NOTE: We show the feedback modal to the users iff
+  // they are navigating away from the editor
+  const isLeavingContentPage = isEditPageUrl(from) || isSpecialPagesUrl(from)
+
+  const [lastSeen, setLastSeen] = useFeedbackStorage()
+  // NOTE: Either this is the first time the user has ever seen the survey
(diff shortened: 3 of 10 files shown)


src/hooks/useFeedbackDisclosure.ts
```diff
@@ -0,0 +1,66 @@
...
+  // so that it knows that the return value is a tuple
+  return [lastSeen, setLastSeen] as const
+}
+
+export const useFeedbackDisclosure = () => {
+  const location = useLocation<{ from?: string }>()
+  // NOTE: Despite the typing from the library,
+  // state is NOT guaranteed to exist.
+  const from = location.state?.from ?? ""
+  // NOTE: We show the feedback modal to the users iff
```
QiluXie, line 43: super nit: typo, change `iff` to `if`

seaerchin: iff is intentional, it means if and only if

src/components/FeedbackModal/FeedbackModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+  ModalHeader,
+  ModalBody,
+  Heading,
+} from "@chakra-ui/react"
+import { ModalCloseButton } from "@opengovsg/design-system-react"
+
+const FeedbackForm = () => {
+  return (
+    <>
+      <div
```
kishore03109, line 14: Prefer to use `<Box>` ya

src/components/FeedbackModal/FeedbackModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+        style={{
+          fontFamily: "sans-serif",
+          fontSize: "12px",
+          color: "#999",
+          opacity: "0.5",
+          paddingTop: "5px",
+        }}
+      >
+        Powered by{" "}
+        <a href="https://form.gov.sg" color="#999">
```
kishore03109, line 46: shall we use a existing token for this rather than hardcode

seaerchin:
!run-chromatic

src/hooks/useFeedbackDisclosure.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import { LOCAL_STORAGE_KEYS } from "constants/localStorage"
+
+import { useLoginContext } from "contexts/LoginContext"
+
+import { isEditPageUrl, isSpecialPagesUrl } from "utils/pages"
+
+import { useLocalStorage } from "./useLocalStorage"
+
+type LastSeenFeedbackTime = number
```
kishore03109, line 12: curious here, what type benefit are we getting here ah? this seems to attempt to use the `Brand {__kind: UserId}` flow

src/hooks/useFeedbackDisclosure.ts
```diff
@@ -0,0 +1,66 @@
...
+type UserId = string
+
+type FeedbackStorageMappings = Record<UserId, LastSeenFeedbackTime>
+
+const NPS_SURVEY_INTERVAL = 1814400000 as const // 3 weeks - 3 * 7 * 24 * 60 * 60 * 1000
+
+const { REACT_APP_SHOW_NPS_FORM } = process.env
```
kishore03109, line 19: do remember to parse here, this is not true but "true"

src/utils/pages.ts
```diff
@@ -0,0 +1,12 @@
+import _ from "lodash"
+
+export const SPECIAL_PAGES = ["homepage", "navbar", "contact-us"]
+
+export const isEditPageUrl = (url: string): boolean => {
+  // NOTE: Lowercase here because `/editpage` also works
+  return url.toLowerCase().includes("/editpage") && url.endsWith(".md")
```
kishore03109, line 7:
```suggestion
  return url.toLowerCase().includes("/editpage/") && url.endsWith(".md")
```

kishore03109 (commented):
May I get your help to include a video/screen shot of how the model looks like as a quick sanity check here?

src/components/FeedbackModal/FeedbackModal.tsx
kishore03109: or at least a mapping from chakra?

src/components/FeedbackModal/FeedbackModal.tsx
seaerchin: wrt anything on `FeedbackForm` - this is an `iframe`, so the embed code was taken **directly** from `form.gov.sg`. this isn't intended to be styled (or edited) and it's just extracted out for readability purposes

src/components/FeedbackModal/FeedbackModal.tsx
seaerchin: see above comment - this is what form gave, so idt we should change

src/hooks/useFeedbackDisclosure.ts
seaerchin: for _nominal_ type systems (ts is structural), they wouldn't be mutually assignable. however, the value here is just so that devs know what this is actually intended to be. Tbh i can do the `brand` pattern too, it's not a big deal

src/hooks/useFeedbackDisclosure.ts
QiluXie: @seaerchin I got your point, I still think it's better to control it using string "true" and "false" value, i.e. do a parse here to boolean based on the string value of `process.env.REACT_APP_SHOW_NPS_FORM` because: 

- it's clear when we want to turn on/off the flag, we need to use true/false respectively in env var
- we can identify whether the env var is missing or has unintended string value
- the const here can be guaranteed to be boolean which guarantee `isSurveyRequired` to be boolean on line 52 and saves the cast using `!!` on line 59

src/hooks/useFeedbackDisclosure.ts
```diff
@@ -0,0 +1,66 @@
...
+import { isEditPageUrl, isSpecialPagesUrl } from "utils/pages"
+
+import { useLocalStorage } from "./useLocalStorage"
+
+type LastSeenFeedbackTime = number
+type UserId = string
+
+type FeedbackStorageMappings = Record<UserId, LastSeenFeedbackTime>
+
+const NPS_SURVEY_INTERVAL = 1814400000 as const // 3 weeks - 3 * 7 * 24 * 60 * 60 * 1000
```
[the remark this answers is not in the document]
seaerchin: re: multiplication directly - sure, no worries will do that! did the `as const` to get it to infer exact type but maybe not worth.

re: 3 wks - actually we just need it to be the survey interval. i think i'll change the name + update time (1 wk) for that. the reason why is because we show it on a monthly basis for the duration of a week. if we set to a month, a person could do on the sunday of the prev month, and when we toggle it on again, only has the sunday to do it. hence, the time interval should be duration of survey.

re: constant - not incliend to. this constant isn't meant to be exported cos it's only used in 1 place for now and i want to hide it. if we got more use case for this, then i'll change its location + naming

src/hooks/useFeedbackDisclosure.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const { REACT_APP_SHOW_NPS_FORM } = process.env
+
+// NOTE: Wrapper to handle get/set of individual keys
+const useFeedbackStorage = () => {
```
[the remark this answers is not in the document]
seaerchin: sure

src/hooks/useFeedbackDisclosure.ts
```diff
@@ -0,0 +1,66 @@
...
+  const lastSeen = userMappings[userId]
+  const setLastSeen = () => {
+    setUserMappings({ ...userMappings, [userId]: Date.now() })
+  }
+  // NOTE: `const` cast to infer types properly
+  // so that it knows that the return value is a tuple
+  return [lastSeen, setLastSeen] as const
+}
+
+export const useFeedbackDisclosure = () => {
```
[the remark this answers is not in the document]
seaerchin: sure

src/hooks/useFeedbackDisclosure.ts
seaerchin: sure, not a big deal

seaerchin:
> May I get your help to include a video/screen shot of how the model looks like as a quick sanity check here?

added chromatic link

src/hooks/useFeedbackDisclosure.ts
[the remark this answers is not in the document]
QiluXie: Sounds good, thanks for the explanation.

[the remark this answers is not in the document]
QiluXie: @seaerchin wait, but if the time interval is 1 week, a person can do it every week in the same month, right?
As long as the it's been more than a week since he/she did it last time. 
Is this the intended product requirement?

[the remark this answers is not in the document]
seaerchin: The env var will be off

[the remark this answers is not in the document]
seaerchin: Only if duration + env var - the env var means that ppl need to do survey and the duration means whether they need to do the survey in this current week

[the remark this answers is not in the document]
QiluXie: Got it @seaerchin. Yeah, that makes sense. Thanks for the clarification.

QiluXie (approved):
LGTM, @seaerchin thanks for resolving my comments

package.json
```diff
@@ -89,7 +89,7 @@
     "prepare": "husky install",
     "version": "auto-changelog -p && git add CHANGELOG.md",
     "storybook": "source .env && REACT_APP_ENV=test && storybook dev -p 6006",
-    "build-storybook": "REACT_APP_ENV=test && storybook build -s public",
+    "build-storybook": "REACT_APP_ENV=test && storybook build",
```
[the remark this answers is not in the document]
seaerchin: done using `staticDirs` in `.storybook/main` - without this change, `storybook build` will fail cos got 2 sources for `staticDirs`. it was initially required cos the deps upgrade caused msw to not work.

src/hooks/useFeedbackDisclosure.ts
```diff
@@ -0,0 +1,75 @@
...
+
+import { useLocalStorage } from "./useLocalStorage"
+
+type LastSeenFeedbackTime = number
+type UserId = string
+
+type FeedbackStorageMappings = Record<UserId, LastSeenFeedbackTime>
+
+// 1 week in ms
+const NPS_SURVEY_DURATION = 7 * 24 * 60 * 60 * 1000
```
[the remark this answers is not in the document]
seaerchin: changed for clarity cos i think the initial 3 weeks was abit confusing. as long as the last seen + duration is less than the current time, it means that they haven't done the survey in the current internval

src/hooks/useFeedbackDisclosure.ts
```diff
@@ -0,0 +1,75 @@
...
+
+type UseFeedbackStorageReturn = readonly [number, () => void]
+
+// NOTE: Wrapper to handle get/set of individual keys
+const useFeedbackStorage = (): UseFeedbackStorageReturn => {
+  const { userId } = useLoginContext()
+  const [
+    userMappings,
+    setUserMappings,
+  ] = useLocalStorage<FeedbackStorageMappings>(LOCAL_STORAGE_KEYS.Feedback, {})
```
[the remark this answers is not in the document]
seaerchin: it seems to suggest that your `userId` is `undefined` monkas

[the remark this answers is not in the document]
seaerchin: can you paste the repro steps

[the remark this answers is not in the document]
seaerchin: fixed

src/hooks/useFeedbackDisclosure.ts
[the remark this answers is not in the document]
harishv7: Based on the description,

> this is in intervals of 3 wks since the last seen timing

So this should be last seen + 3 weeks duration > current time right?

src/components/FeedbackModal/FeedbackModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+        If the form below is not loaded, you can also fill it in{" "}
+        <a href="https://form.gov.sg/64b7a11823e54700118bad90">here</a>.
+      </div>
+
+      <iframe
+        title="FormSG Feedback form for Isomer"
+        id="iframe"
+        src="https://form.gov.sg/64b7a11823e54700118bad90"
+        width="100%"
+        height="650px"
```
harishv7, line 33: curious, can this be 100% (or similar) to fill the modal height instead of scrolling? Since it is a short form, can be embedded to fullheight right? might look nicer to user

but if not, am fine with this also

src/components/FeedbackModal/FeedbackModal.tsx
```diff
@@ -0,0 +1,76 @@
+import {
```
harishv7, line 1: have we updated .env-example?

seaerchin: good catch, updated

src/hooks/useFeedbackDisclosure.ts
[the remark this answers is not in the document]
seaerchin: changed PR desc, using 1 week now

src/components/FeedbackModal/FeedbackModal.tsx
seaerchin: i actually didn't want to edit this because this is as-is by forms. i could try but fwiw, i'm rolling out a UI rehash (PM asked) so it might not be of value

src/hooks/useFeedbackDisclosure.ts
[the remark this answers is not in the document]
harishv7: ok!

src/components/FeedbackModal/FeedbackModal.tsx
harishv7: Okay, let's proceed as it is now

harishv7 (approved):

PR #1353, opened 2023-07-20
feat(setup.md): add command

kishore03109:
## Problem

Currently, we are seeding the rows manually for our e2e tests. However, writing directly to DB is generally not good eng practice and can cause problems like [Postgres's primary key sequence going out of sync](https://stackoverflow.com/questions/244243/how-to-reset-postgres-primary-key-sequence-when-it-falls-out-of-sync). 

## Solution

This PR aims to alievate that by just providing 2 commands to run instead.

## Tests 
1. run 
```
INSERT INTO sites (name, site_status, job_status, creator_id)
VALUES ('e2e email test site 2', 'INITIALIZED', 'READY', 1);
INSERT INTO repos (name, url, created_at, updated_at, site_id)
SELECT 'e2e-email-test-repo 2', 'https://github.com/isomerpages/e2e-email-test-repo-2', NOW(), NOW(), id FROM sites WHERE name = 'e2e email test site 2';
```
2. verify that rows are properly added in the `sites` and `repos` tables respectively. 
3. delete added rows from the respective tables.

### DIFF
--- a/cypress/SETUP.md
+++ b/cypress/SETUP.md
@@ -20,8 +20,8 @@ In order to run the email e2e tests successfully, we need to perform the followi
 2. Run these commands:
 
 ```
-INSERT INTO sites (name, api_token_name, site_status, job_status, creator_id)
-VALUES ('e2e email test site', '', 'INITIALIZED', 'READY', '1');
+INSERT INTO sites (name, site_status, job_status, creator_id)
+VALUES ('e2e email test site', 'INITIALIZED', 'READY', '1');
 INSERT INTO repos (name, url, created_at, updated_at, site_id)
 SELECT 'e2e-email-test-repo', 'https://github.com/isomerpages/e2e-email-test-repo', NOW(), NOW(), id FROM sites WHERE name = 'e2e email test site';
 ```
--- a/cypress/SETUP.md
+++ b/cypress/SETUP.md
@@ -21,7 +21,7 @@ In order to run the email e2e tests successfully, we need to perform the followi
 
 ```
 INSERT INTO sites (name, site_status, job_status, creator_id)
-VALUES ('e2e email test site', 'INITIALIZED', 'READY', '1');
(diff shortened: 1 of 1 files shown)


seaerchin (approved):
lgtm if you tested that it works on local

seaerchin:
can hold off on merging until monday? see if i got enough time to do this tmrw

kishore03109:
@seaerchin yes sir

kishore03109:
@seaerchin sir any updates ah

seaerchin:
> @seaerchin sir any updates ah

merge bah didn't work on this

cypress/SETUP.md
```diff
@@ -16,11 +16,15 @@ In order to run the github e2e tests successfully, we need to perform the follow
...
-2. search for the `e2e email test site` in `sites`
-3. copy the value over to your local db in `sites`
-4. search for the `e2e-email-test-repo` in `repos`
-5. copy the value over to your local db in `repos`
+1. go to the local database
+2. Run these commands:
+
+```
+INSERT INTO sites (name, site_status, job_status, creator_id)
+VALUES ('e2e email test site', 'INITIALIZED', 'READY', '1');
```
QiluXie, line 24: the `creator_id` in `sites` table is integer type (`int8`) instead of text/string, we should use `1` without the quote here to match the data type.

QiluXie (changes requested):

cypress/SETUP.md
kishore03109: sure, but just a heads up that [postgres auto converts the type for us](https://www.postgresql.org/docs/current/typeconv.html)

QiluXie: Understand that in this case since text `1` can be parsed to number `1` without any ambiguity, postgres will perform the implicit conversion. But still the best practice is to use the correct type that matches the target column when it's possible.

QiluXie (approved):
lgtm

PR #1360, opened 2023-07-20
feat(preview): added preview images

caando:
## Problem

CMS dashboard page displays Isomer logo for all sites

Closes IS-266

## Solution

This PR queries a new backend endpoint with a list of site names to get url of preview images. It tries to get the favicon from production site as preview image. If it is unsuccessful, Isomer logo will still be displayed. 

**Breaking Changes**

  - Requires the corresponding PR https://github.com/isomerpages/isomercms-backend/pull/848 to be merged in

**Features**:

- Going to the site dashboard now displays the favicon of production site as preview image. 

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

- [x] Smoke tests


src/layouts/Sites.tsx
```diff
@@ -29,8 +30,57 @@ import { EmptySitesImage, IsomerLogoNoText } from "assets"
...
+const SitePreviewImage = ({ imageUrl }: { imageUrl?: string }) => {
+  if (!imageUrl) {
+    return <IsomerLogoNoText height="50%" />
+  }
+  return (
+    <Image
+      src={imageUrl}
+      height="50%"
+      fallback={<IsomerLogoNoText height="50%" />}
+    />
```
QiluXie, line 42: Can we omit the if condition since the `fallback` property is already defined in the `Image` tag? 
If `imageUrl` is not provided, `IsomerLogoNoText` will be displayed as a fallback.

src/hooks/allSitesHooks/useSitePreview.ts
```diff
@@ -0,0 +1,25 @@
+import { useQuery } from "react-query"
+import type { UseQueryResult } from "react-query"
```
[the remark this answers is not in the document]
kishore03109: ^ because `useQuery` is a function while `UseQueryResult` is a type!

src/hooks/allSitesHooks/useSitePreview.ts
```diff
@@ -0,0 +1,25 @@
...
+  sites: string[]
+): UseQueryResult<SitePreviewRequest[]> => {
+  return useQuery<SitePreviewRequest[]>(
+    [`${SITE_PREVIEW_KEY}`, userEmail],
+    () => AllSitesService.getSitePreview(sites),
+    {
+      retry: false,
+      refetchOnWindowFocus: false,
+      onError: (error) => {
+        console.error(error)
```
kishore03109, line 21: Hmm do we want to console log here? this would lead to a stream of console logs for us admins. Even if the agency looks at this, it might be unlikely that this would be useful information here

src/types/sites.ts
```diff
@@ -8,3 +8,7 @@ export interface SiteData {
 export interface SiteDataRequest {
   siteNames: SiteData[]
 }
+
+export interface SitePreviewRequest {
+  imageUrl: string
```
kishore03109, line 13: our backend DTO is `imageUrl?` while this is defined as `imageUrl`. Should we make them consistent?

src/services/AllSitesService.ts
```diff
@@ -1,8 +1,15 @@
...
 export const getAllSites = async (): Promise<SiteDataRequest> => {
   const endpoint = `/sites`
   return apiService.get(endpoint).then((res) => res.data)
 }
+
+export const getSitePreview = async (
+  sites: string[]
+): Promise<SitePreviewRequest[]> => {
+  const endpoint = `/sites/preview`
+  return apiService.post(endpoint, { sites }).then((res) => res.data)
```
kishore03109, line 14: is there a reason why this is a post and not a get ah?

src/layouts/Sites.tsx
```diff
@@ -63,41 +113,30 @@ const SitesContent = ({ siteNames }: { siteNames?: SiteData[] }) => {
         </VStack>
       </VStack>
     )
+  const { data: previews } = useSitePreview(
+    email,
+    siteNames.map((site) => site.repoName)
+  )
+  if (previews && previews.length === siteNames.length) {
```
kishore03109, line 120: Hmm sorry I dont understand this code :(
we only render this when we get the preview images for __**all**__ sites have been loaded? shouldnt we load it eagerly, ie for each site that has a preview, use the url for that site?

src/hooks/allSitesHooks/useSitePreview.ts
```diff
@@ -0,0 +1,25 @@
...
+import * as AllSitesService from "services/AllSitesService"
+
+import { SitePreviewRequest } from "types/sites"
+
+export const useSitePreview = (
+  userEmail: string,
+  sites: string[]
+): UseQueryResult<SitePreviewRequest[]> => {
+  return useQuery<SitePreviewRequest[]>(
+    [`${SITE_PREVIEW_KEY}`, userEmail],
```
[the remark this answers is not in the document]
caando: will change :+1:

src/layouts/Sites.tsx
caando: There are 2 possibilities for failure. The condition checks the backend is able to retrieve a favicon from going to prod homepage and return undefined for imageUrl if its not possible. I think we should not feed a blank url to image src given that it makes a query. The fallback is when there is a imageUrl but does not lead to the image such as if the image is missing. In that case there's no way we would have known and have to rely on the fallback.

src/hooks/allSitesHooks/useSitePreview.ts
caando: okay, the other onError uses `console.log()` too

src/types/sites.ts
caando: okay

src/services/AllSitesService.ts
caando: I didn't find a way to send a list of strings (sites) through a get request, and therefore resorted to a post request. If you know of a way pls lmk

src/layouts/Sites.tsx
caando: The if statement `if (previews && previews.length === siteNames.length) {` only serves as a out of bounds check as the if frontend sends an array of sites to request for previews, it should expect an array of the same length back to frontend. We chose to use a list of sites as 1 request because for admin users that has access to 1k repos, it was making 1k requests to the backend which completely flood the all logs and gets ratelimited after the first 100 or so.

src/hooks/allSitesHooks/useSitePreview.ts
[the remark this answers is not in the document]
caando: 👍 missed it out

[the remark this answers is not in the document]
QiluXie: Sorry @caando & @kishore03109, good lord, how deeply mistaken I was!!
I meant to say importing both of them in one line without `type` 😢

src/layouts/Sites.tsx
QiluXie: @caando so if `imageUrl` is undefined, the `Image` component will still perform a network query?

kishore03109: I think if undefined goes straight to fallback bah. can test this [here](https://chakra-ui.com/docs/components/image/usage) when we put the code 
```
<Box boxSize='sm'>
  <Image src={undefined} alt='Dan Abramov' fallback={<Text>hihi</Text>} />
</Box>
```
it renders the text properly

QiluXie: Thanks @kishore03109 . I think the Image component is smart enough to avoid a network query if the imageUrl is undefined. If so, @caando  I'd suggest to omit the if condition as it doesn't add any value.

src/layouts/Sites.tsx
kishore03109: > it was making 1k requests to the backend which completely flood the all logs and gets ratelimited after the first 100 or so

Hmm why would we get rate limited ah?

src/services/AllSitesService.ts
kishore03109: hmm why cant we do a 

get request on `/sites/preview`, and return only a list of sites that 
1. has an image that is defined 
2. if admin return empty array so this wouldnt slow down our network calls

our cookies already has the user info, so the FE does not need to compute this

src/layouts/Sites.tsx
kishore03109: ya shouldnt be 1k calls bah regardless ah

caando: there's a ratelimiter in `src/services/utilServices/RateLimiter.ts` which limits to around 200 calls per 15mins. Admins completely exhaust this and can no longer perform any actions.

src/services/AllSitesService.ts
caando: Can I check if you're suggesting to get list of permitted sites via DB? I think that would result in Github users not being able to get images thou

src/layouts/Sites.tsx
caando: image
@QiluXie @kishore03109 looks like its not smart enough and will make a request.

kishore03109 (approved):

PR #1374, opened 2023-07-27
test(e2e): add homepage 

seaerchin:
## Problem
Previously, homepage did not have any e2e tests. This PR adds e2e tests for homepage.

## Solution
1. change dnd package - old dnd package had an issue with react strict mode, which causes it to crash on local.
2. add tests for special cases and drag/drop functionality
    - see the base drag and drop spec from react beautiful dnd [here](https://github.com/atlassian/react-beautiful-dnd/blob/master/cypress/integration/reorder.spec.js) 
    - other special cases are given on the ticket and guarntee some baseline behaviour

### DIFF
--- a/cypress/e2e/homepage.spec.ts
+++ b/cypress/e2e/homepage.spec.ts
@@ -124,6 +124,8 @@ describe("Homepage", () => {
       )
     })
 
+    // TODO (IS-377): Disallow resource creation
+    // when a resource section already exists
     it("should limit users to a single resource component", () => {
       // Arrange
       // NOTE: Precondition here is that there are NO existing resources
--- a/cypress/e2e/homepage.spec.ts
+++ b/cypress/e2e/homepage.spec.ts
@@ -183,7 +183,7 @@ describe("Homepage", () => {
 
       // Assert
       // NOTE: Order is now Infopic -> Infobar
-      // note: not using get aliases as they where returning incorrect results
+      // note: not using get aliases as they were returning incorrect results
       cy.get(getHandleSelector()).eq(0).should("contain", "Infopic")
       cy.get(getHandleSelector()).eq(1).should("contain", "Infobar")
       cy.get(infopicPreviewSelector)
--- a/cypress/e2e/homepage.spec.ts
+++ b/cypress/e2e/homepage.spec.ts
@@ -1,4 +1,5 @@
-import { TEST_REPO_NAME } from "../fixtures/constants"
+import { keyCodes, TEST_REPO_NAME, timings } from "../fixtures/constants"
+import { getHandleSelector } from "../utils/dnd"
 
 const visitHomepage = () => {
   cy.visit(`/sites/${TEST_REPO_NAME}/homepage`)
@@ -35,29 +36,6 @@ const SECTION_DROPDOWN_OPTIONS = {
 
 const toggleSectionDropdown = (section: string) => {
   cy.contains(section).next().click()
-}
-
-const timings = {
-  outOfTheWay: 0.2,
-  // greater than the out of the way time
-  // so that when the drop ends everything will
-  // have to be out of the way
-  minDropTime: 0.33,
-  maxDropTime: 0.55,
-}
-
-const keyCodes = {
-  space: 32,
-  arrowDown: 40,
-}
-
-const REACT_DRAG_HANDLE_PROPERTY = "data-rfd-drag-handle-draggable-id" as const
-
-const getHandleSelector = (draggableId?: string) => {
-  if (draggableId) {
-    return `[${REACT_DRAG_HANDLE_PROPERTY}="${draggableId}"]`
-  }
-  return `[${REACT_DRAG_HANDLE_PROPERTY}]`
 }
 
 describe("Homepage", () => {
--- a/cypress/e2e/homepage.spec.ts
+++ b/cypress/e2e/homepage.spec.ts
@@ -1,0 +1,217 @@
+import { TEST_REPO_NAME } from "../fixtures/constants"
+
+const visitHomepage = () => {
+  cy.visit(`/sites/${TEST_REPO_NAME}/homepage`)
+}
+
+const NOTIFICATION_BAR_PREVIEW_SELECTOR = "div[id='notification-bar']"
+const NOTIFICATION_INPUT_SELECTOR = 'input[id="site-notification"]'
+
+const HERO_BUTTON_INPUT_SELECTOR = 'input[id*="hero-button"]'
+const HERO_BUTTON_PREVIEW_SELECTOR = 'div[class*="search-button"]'
+
+const HERO_SECTION_HIGHLIGHTS_SELECTOR = 'input[id="radio-highlights"]'
+const HERO_SECTION_DROPDOWN_SELECTOR = 'input[id="radio-dropdown"]'
+
+const ADD_SECTION_DROPDOWN_SELECTOR = 'select[id="section-new"]'
+
+const getAddSectionButtonSelector = (): "@addSectionButton" => {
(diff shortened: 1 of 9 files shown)


cypress/e2e/homepage.spec.ts
```diff
@@ -0,0 +1,217 @@
...
+}
+
+const REACT_DRAG_HANDLE_PROPERTY = "data-rfd-drag-handle-draggable-id" as const
+
+const getHandleSelector = (draggableId?: string) => {
+  if (draggableId) {
+    return `[${REACT_DRAG_HANDLE_PROPERTY}="${draggableId}"]`
+  }
+  return `[${REACT_DRAG_HANDLE_PROPERTY}]`
+}
```
alexanderleegs, line 61: Could we consider moving this selector (and the associated reordering) into a utils file? Nav bar/Contact us also make use of this and are similarly missing tests

alexanderleegs (commented):
I think it might also be worth testing individual component behaviour (specifically for infopic, which contains uploading of images from the modal), as well as attempting to save the overall page to ensure that changes don't mess up the expected formatting!

cypress/e2e/homepage.spec.ts
seaerchin: can! i just left it here first because the other tests aren't up

kishore03109:
!run-chromatic

cypress/e2e/homepage.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+      // Act
+      // reorder operation
+      cy.get("@first")
+        .focus()
+        .trigger("keydown", { keyCode: keyCodes.space })
+        // need to re-query for a clone
+        .get("@first")
+        .trigger("keydown", { keyCode: keyCodes.arrowDown, force: true })
+        // finishing before the movement time is fine - but this looks nice
+        .wait(timings.outOfTheWay * 1000)
```
QiluXie, line 181: Just curious, how did we come up with the time that we need to wait here (0.2s)?

cypress/fixtures/constants.ts
```diff
@@ -61,3 +61,17 @@ export const MOCK_REVIEW_DESCRIPTION = "Some interesting description"
...
 export const FORBIDDEN_CHARACTERS = "!@#$%^&*()"
 export const NON_ENGLISH_CHARACTERS = "文கி"
+
+export const timings = {
+  outOfTheWay: 0.2,
+  // greater than the out of the way time
+  // so that when the drop ends everything will
+  // have to be out of the way
+  minDropTime: 0.33,
+  maxDropTime: 0.55,
```
QiluXie, line 71: Couldn't find usage of these two constants, will they be used in future test? 
Same question, how did we come up with the values here?

cypress/e2e/homepage.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+      cy.get(ADD_SECTION_DROPDOWN_SELECTOR).select(
+        SECTION_DROPDOWN_OPTIONS.RESOURCES
+      )
+      cy.get("@addSectionButton").click()
+
+      // Assert
+      // NOTE: This is an existing bug on `production`
+      // where adding a resource section doesn't disable.
+      cy.get("@addSectionButton").should("be.disabled")
+    })
```
QiluXie, line 148: It means this assertion will fail if we don't fix the existing bug in production, right?

cypress/e2e/homepage.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+        // need to re-query for a clone
+        .get("@first")
+        .trigger("keydown", { keyCode: keyCodes.arrowDown, force: true })
+        // finishing before the movement time is fine - but this looks nice
+        .wait(timings.outOfTheWay * 1000)
+        .trigger("keydown", { keyCode: keyCodes.space, force: true })
+
+      // Assert
+      // NOTE: Order is now Infopic -> Infobar
+      // note: not using get aliases as they where returning incorrect results
```
QiluXie, line 186: Super nit: typo in the comment? should be `will` instead of `where`?

QiluXie (commented):
Looks good, left some minor comments (mostly questions).

cypress/e2e/homepage.spec.ts
seaerchin: it's the time given by the authors of the library in their tests (see [here](https://github.com/atlassian/react-beautiful-dnd/blob/master/cypress/integration/reorder.spec.js)).

cypress/fixtures/constants.ts
seaerchin: read the test spec for `react-beautiful-dnd` (see [here](https://github.com/atlassian/react-beautiful-dnd/blob/master/cypress/integration/reorder.spec.js)) and took these values

cypress/e2e/homepage.spec.ts
seaerchin: yeap, that's correct.

cypress/e2e/homepage.spec.ts
seaerchin: typo from the lib spec itself AHHAHA will fix

cypress/e2e/homepage.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+
+      // Act
+      cy.get(ADD_SECTION_DROPDOWN_SELECTOR).select(
+        SECTION_DROPDOWN_OPTIONS.RESOURCES
+      )
+      cy.get("@addSectionButton").click()
+
+      // Assert
+      // NOTE: This is an existing bug on `production`
+      // where adding a resource section doesn't disable.
```
harishv7, line 146: are we planning to fix this? is there a ticket?

cypress/e2e/homepage.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+      cy.get("@addSectionButton").should("be.disabled")
+    })
+
+    it.skip("should snap preview to that section when clicking on the section in the editor", () => {
+      // NOTE: Not testing this because cypress doesn't support viewports checking
+    })
+
+    it("should rearrange the blocks correctly and the preview should also reflect the new order", () => {
```
harishv7, line 154: Screenshot 2023-08-02 at 10 37 28 AM

Somehow I can't get the last test to pass, it keeps failing after the click

seaerchin: Oh I think here it's not react beautiful DND so the selector id might be different

cypress/e2e/homepage.spec.ts
seaerchin: Unsure, I think it's good to fix but it's difficult due to how messy this section of the code is...

harishv7: Hmm, then the test will keep failing right? Shall we comment this assert with a TODO on top?

ELse, it becomes like a legacy knowledge that oncalls need to know this failure is expected

cypress/e2e/homepage.spec.ts
harishv7: hmm so do we need to update the selector?

cypress/e2e/homepage.spec.ts
seaerchin: Sure, sounds food

cypress/e2e/homepage.spec.ts
seaerchin: Haven't try yet, will let you know.

harishv7: resolving as this was tested against staging (not local)

harishv7 (approved):

seaerchin:
Screenshot 2023-08-02 at 3 32 42 PM
 
pic for proof

PR #1380, opened 2023-08-01
feat(site launch): add in refetch behvaiour

kishore03109:
## Problem

During the site launch process, users would have to do a hard refresh to get the lastest site launch results. This is not ideal.

## Solution

Refresh behaviour every 5 mins. tbh, the 5 mins is an arbitarary number, we can adjust this param as this feature is more used by our users in the future. This is just to prevent the need for a hard refesh when the launch is success.

### DIFF
--- a/src/hooks/siteDashboardHooks/useGetSiteLaunchStatus.ts
+++ b/src/hooks/siteDashboardHooks/useGetSiteLaunchStatus.ts
@@ -6,6 +6,8 @@ import * as SiteLaunchService from "services/SiteLaunchService"
 
 import { SiteLaunchDto } from "types/siteLaunch"
 
+const FIVE_MINUTES = 1000 * 60 * 5
+
 export const useGetSiteLaunchStatus = (
   siteName: string
 ): UseQueryResult<SiteLaunchDto> => {
@@ -16,6 +18,11 @@ export const useGetSiteLaunchStatus = (
     },
     {
       retry: false,
+      // we want to refetch the data every 5 minutes
+      // during the site launch process. Else, it will
+      // only be updated on a hard refresh.
+      refetchInterval: FIVE_MINUTES,
+      refetchIntervalInBackground: true,
     }
   )
 }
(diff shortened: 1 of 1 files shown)


src/hooks/siteDashboardHooks/useGetSiteLaunchStatus.ts
```diff
@@ -6,6 +6,8 @@ import * as SiteLaunchService from "services/SiteLaunchService"
 
 import { SiteLaunchDto } from "types/siteLaunch"
 
+const FIVE_MINUTES = 1000 * 60 * 5
```
seaerchin, line 9: suggest calling this `SITE_LAUNCH_REFETCH_INTERVAL` instead of tying it to the time. we can comment that this is 5 mins if required.

src/hooks/siteDashboardHooks/useGetSiteLaunchStatus.ts
```diff
@@ -16,6 +18,11 @@ export const useGetSiteLaunchStatus = (
     },
     {
       retry: false,
+      // we want to refetch the data every 5 minutes
+      // during the site launch process. Else, it will
+      // only be updated on a hard refresh.
+      refetchInterval: FIVE_MINUTES,
+      refetchIntervalInBackground: true,
```
seaerchin, line 25: can i check that the component that calls this hook only exists when users attempt to launch their site? (iirc should be hor otherwise is the card alr)

kishore03109: Uhm no not true, it also gets called when the site is in the dashboard. But since it is once every 5 mins I think this is ok, wdyt?

src/hooks/siteDashboardHooks/useGetSiteLaunchStatus.ts
kishore03109: 8ee21805f49692b9254cbd47cba168b39aabaf2c

src/hooks/siteDashboardHooks/useGetSiteLaunchStatus.ts
seaerchin: Isn't that component conditionally shown and if no site launch then it's human image? That's off the top of my head but not v impt cos low freq

seaerchin (approved):

src/hooks/siteDashboardHooks/useGetSiteLaunchStatus.ts
kishore03109: yup this is accurate!

PR #1382, opened 2023-08-01
feat: storybook improvement for site launch pad

kishore03109:
## Problem

Now that there are different flows for the site launch flow, it would make sense to have various flows for the corresponding storybooks for faster reviewal process for upcoming PRs. 

## Solution

Individual pages have their own storybook.

Screenshot 2023-08-01 at 11 35 48 PM
Screenshot 2023-08-01 at 11 35 37 PM
Screenshot 2023-08-01 at 11 35 28 PM
Screenshot 2023-08-01 at 11 35 14 PM


kishore03109:
!run-chromatic

alexanderleegs (approved):
lgtm!

(Non-blocker) The docs page in the `SiteLaunchPad` folder in storybook has an unclosable modal, I think this is fine to leave as is though because we should be going to the individual pages to view the screens

PR #1384, opened 2023-08-02
IS-349-FE-only-admins-can-launch-sites

kishore03109:
## Problem

No access controls for site launch feature. 

## Solution

Only allow site admins to launch sites. Currently, our CMS defines admins as someone whose email ends with a `.gov.sg`. Can be reviewed with [be pr](https://github.com/isomerpages/isomercms-backend/pull/873)

## Screenshot

View of non-admin email user: 
Screenshot 2023-08-02 at 9 30 53 AM

View of contributor access site launch pad page: 

https://github.com/isomerpages/isomercms-frontend/assets/42832651/8f294c51-1510-4cfe-ae2b-5eb0872c01cb

### DIFF
--- a/src/utils/siteLaunchUtils.ts
+++ b/src/utils/siteLaunchUtils.ts
@@ -1,6 +1,6 @@
 import { WHITELISTED_REPOS } from "constants/siteLaunch"
 
-export const shouldUseSiteLaunchFeature = (siteName: string): boolean => {
+const isRepoWhitelistedForSiteLaunch = (siteName: string): boolean => {
   // The name of our sites for our storybooks is "storybook".
   // Therefore, to allow the storybook to use the site launch feature, we need to
   // explicitly check for it.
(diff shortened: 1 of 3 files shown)


src/utils/siteLaunchUtils.ts
```diff
@@ -9,3 +9,14 @@ export const shouldUseSiteLaunchFeature = (siteName: string): boolean => {
   }
   return WHITELISTED_REPOS ? WHITELISTED_REPOS.includes(siteName) : false
 }
+
+const isAdminUser = (email: string): boolean => {
+  return email.endsWith(".gov.sg")
+}
+
```
alexanderleegs, line 16: Are we only allowing `.gov.sg` emails to launch sites? We also have healthcare orgs and other stuff which are considered site owners, but don't have `.gov.sg` emails. Is using the site member information from the backend insufficient?

seaerchin: Suggest exposing an endpoint on the backend tgt with a custom hook so that we can get all non-null expiry values for whitelist and use that for owners. 

For now, this is fine **if** our is only gov us rs.

seaerchin (approved):
Lgtm under assumption that MVP is only for gov.sg. please dismiss if we have plans for launching to healthcare etc.

src/utils/siteLaunchUtils.ts
seaerchin: Other option is we manually sync but v prone to error

kishore03109:
@seaerchin dismissing review, yes, i do intend to put this out for other agencies. There is already an endpoint that does this, which this PR leverages on!

seaerchin (approved):

PR #1386, opened 2023-08-03
fix(header): fix context for storybook

kishore03109:
## Problem

Chromatic builds are failing + keeps building dev. Not 100% sure why, but the facts are:

1. We know that the way Chromatic works is that it compares UI changes against a baseline
2. Chromatic keeps building develop when `!run-chromatic` is commented on the PRs
3. There exists errors in the current develop. 
4. By extension, our baseline (develop) has errors.

## Solution

This PR attempts to test the hypothesis that is the components are fixed, does the chromatic functionality gets back up. However, I cannot test this out until this is merged into develop. 

[Note: this is a current blocker for DS upgrades, UI review of site launch changes that affect the current sprint]

### Before 
Screenshot 2023-08-03 at 9 27 13 AM

### After 
Screenshot 2023-08-03 at 9 26 56 AM

### DIFF
--- a/src/components/Header/Header.stories.tsx
+++ b/src/components/Header/Header.stories.tsx
@@ -1,5 +1,7 @@
-import { ComponentStory, ComponentMeta } from "@storybook/react"
+import { Meta, StoryFn } from "@storybook/react"
 import { MemoryRouter, Route } from "react-router-dom"
+
+import { DirtyFieldContextProvider } from "contexts/DirtyFieldContext"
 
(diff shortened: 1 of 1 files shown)


alexanderleegs (approved):

seaerchin:
i fixed it in the design system PR alr - if required can just copy the stuff from there fyi

kishore03109:
@seaerchin ya looks similar bah, I changed the storybooks components too to use non-deprecated ones

PR #1393, opened 2023-08-04
feat(is-326): refactor dnd

seaerchin:
## Problem
Previously, drag and drop was difficult to decipher, making it difficult to work on the file. This PR tackles the issue of confusing dnd logic by extracting it to a separate file and making the updates return a copy of the state. In addition, previous logic used to perform CRUD operations were extracted out into util functions (kept hidden). 

## Solution
1. Make 3 separate functions for CRUD operations - technically we can combine them, but i haven't figured out a nice way to do so that allows extension.
    - these functions are **pure**, they do not have any side effects beyond updating the state of the homepage. This is to allow for ease of testing/extraction in the future. additionally, their interfaces have been kept similar to allow future work. These functions should follow a similar pattern of decide where to update (from elemType) -> perform update using helper function -> return new state. 
    - `scrollRefs` is explicitly not here because it is not data (ie, it doesn't show in the final markdown) but it is behaviour. as such, i have opted to keep it in the original `EditHomepage` file. 

**How the update cycle works**
1. Get the index of the element to perform the operation on 
2. based on the `elemType`, update the appropriate section accordingly using the helper functions (`updatePositions/deleteElement/createElement`)
3. based on the `elemType`, return a new `homepageState` based on the update above.

## Tests 
- [ ] Go to homepage
- [ ] add **all** of the blocks - should snap to preview
- [ ] update all the added blocks 
- [ ] reorder the blocks using drag and drop
- [ ] delete all of the blocks 
- [ ] clicking on dropdown toggle (the down arrow) should snap to preview
- [ ] repeat the above tests but for hero section -> dropdown
- [ ] toggle between dropdown -> highlights 
- [ ] repeating the above tests should work

## Notes
The following shortcuts were taken: 
1. cast to avoid type check - this was done because the update type is tied to the shape of the resource. ie, if update type === `dropdown`, for example, we are always performing the update on the `HeroSection`, on the `Dropdown` option. This makes it safe.
5. type as `unknown` to avoid giving proper typing - the actual CRUD operations should operate on arrays and be agnostic of the type. This oculd be given as a type parameter but not done in view of time
6. Operating on a whole `HomepageState` rather than each individual section - this is more involved because our errors are dynamically generated along with each sub-section. we should ideally make the update loop agnostic of the type but was not done as a first pass.


src/hooks/useDrag.tsx
```diff
@@ -0,0 +1,418 @@
...
+      )
+    }
+    default:
+      return homepageState
+  }
+}
+
+// NOTE: Handles only placement,
+// let the caller decide what to create
+export const onCreate = <E,>(
```
kishore03109, line 257: quick qn, why do we type this to `E` when we are not referencing the type here ah

src/hooks/useDrag.tsx
```diff
@@ -0,0 +1,418 @@
...
+        newSectionErrors
+      )
+    }
+    // inner dnd for hero
+    // passed in via droppableId
+    case "dropdownelem": {
+      // TODO: type check to avoid casting
+      const draggedElem = (frontMatter.sections[0]
+        .hero as EditorHeroDropdownSection).dropdown.options[source.index]
+      //! Thsis needs to be updated
```
kishore03109, line 178:
```suggestion
      //! This needs to be updated
```

kishore03109 (approved):

src/hooks/useDrag.tsx
```diff
@@ -0,0 +1,418 @@
...
+  })
+}
+
+const deleteElement = <T,>(section: T[], indexToDelete: number): T[] => {
+  return update(section, {
+    $splice: [[indexToDelete, 1]],
+  })
+}
+
+const UPDATE_PATHS = {
```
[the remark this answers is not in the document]
seaerchin: removing, was an experiment but i didn't realise that i left it in

src/hooks/useDrag.tsx
seaerchin: we are referencing the type in the error isn't it

src/hooks/useDrag.tsx
seaerchin: removing, old comment

src/types/homepage.ts
```diff
@@ -77,3 +79,38 @@ export interface HomepageDto {
...
+}
+
+export interface EditorHeroDropdownSection {
+  dropdown: {
+    options: []
+  }
+}
+
+export interface EditorHeroHighlightsSection {
+  key_highlights: []
```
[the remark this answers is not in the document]
seaerchin: this is for compat reasons - the legacy code uses `snake_case` and you can see that, for example, `HighlightHeroSection` (part of homepage dto) uses `key_highlights` also to communicate with the backend. 

not inclined to change - no impact on functionality and have to touch FE (legacy stuff) + BE as well, which makes this annoying.

src/hooks/useDrag.tsx
seaerchin: we are referencing the type in the error isn't it

seaerchin:
waiting on 1 more engineer to review so that at least half the reviewers can agree this makes sense

src/layouts/EditHomepage.jsx
```diff
@@ -525,131 +550,62 @@ const EditHomepage = ({ match }) => {
       const idArray = id.split("-")
       const elemType = idArray[0]
 
-      let newSections = []
-      let newErrors = []
+      // The Isomer site can only have 1 resources section in the homepage
+      // Set hasResources to prevent the creation of more resources sections
+      if (elemType === "section" && value === "resources") {
+        setHasResources(true)
+      }
```
alexanderleegs, line 557: This doesn't actually fix the issue with the section creation - the existing issue is due to the section creation not properly resetting the state of the selected section after a user clicks, which allows the create to continue to be clicked even if the option currently is the generic selector

alexanderleegs (approved):
The abstractions make sense to me, it makes the homepage code easier to work with.

Quick question about the drag hooks though - these are homepage specific operations right, and we'll need to reimplement something similar for the other sections in future (nav, contact us), so would it make sense to specify that this is specifically `useHomepageDrag`?

seaerchin:
> The abstractions make sense to me, it makes the homepage code easier to work with.
> 
> Quick question about the drag hooks though - these are homepage specific operations right, and we'll need to reimplement something similar for the other sections in future (nav, contact us), so would it make sense to specify that this is specifically `useHomepageDrag`?

ideally, this would be abstract over the type of structure being dragged - there's a ticket filed for follow up to do exactly that when we want to operate on dnd for the other 2 (contact-us/nav) pages, which is why i'm **not** gonna call this homepage drag. 

also, do take note that the base function that does the logic is agnostic of said structure as well (update/create/delete element), the structure is wrapped by the topmost exported function

src/layouts/EditHomepage.jsx
seaerchin: yea this doesn't. this was part of the original codebase - i just shifted it out to enable the code to be refactored cleanly AHHA

seaerchin:
merging - for future reviewers, feel free to ask on #isomer or ping me directly if got any qns.

PR #1395, opened 2023-08-07
fix: move removeOtherCollaborators to use api call directly

alexanderleegs:
This PR fixes our review request tests so that it's no longer dependent on the number of collaborators - previously we were removing collaborators via the UI, which caused the tests to break if we did not have the correct number of collaborators. This PR modifies the util method `removeOtherCollaborators` to make an api call to our backend directly in order to bypass this issue.

### DIFF
--- a/cypress/api/collaborators.api.ts
+++ b/cypress/api/collaborators.api.ts
@@ -19,10 +19,10 @@ export const listCollaborators = (): Cypress.Chainable<CollaboratorDto[]> => {
   return cy.request("GET", `${BASE_URL}`).then(({ body }) => body.collaborators)
 }
 
-export const removeOtherCollaborators = (): void => {
+export const removeOtherCollaborators = (remainingEmail: string): void => {
   listCollaborators().then((collaborators) => {
     collaborators.forEach((collaborator) => {
-      if (collaborator.email === E2E_EMAIL_ADMIN.email) return
+      if (collaborator.email === remainingEmail) return
       const { id } = collaborator
       cy.request("DELETE", `${BASE_URL}/${id}`)
     })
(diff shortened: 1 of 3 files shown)


cypress/api/collaborators.api.ts
```diff
@@ -9,3 +14,17 @@ export const addAdmin = (admin: string): void => {
...
 }
+
+export const listCollaborators = (): Cypress.Chainable<CollaboratorDto[]> => {
+  return cy.request("GET", `${BASE_URL}`).then(({ body }) => body.collaborators)
+}
+
+export const removeOtherCollaborators = (): void => {
+  listCollaborators().then((collaborators) => {
+    collaborators.forEach((collaborator) => {
+      if (collaborator.email === E2E_EMAIL_ADMIN.email) return
```
seaerchin, line 25: we should remove this case - it is entirely possible that the writer of the test swapped the email to be a collab email.

we should instead rely on the guarantee that our cms guarantees >=1 collab (canot remove last admin)

cypress/e2e/reviewRequests.spec.ts
seaerchin, line 1: just a sanity check as i didn't open this up in vscode - have all the multiple calls been removed ah? 

separately, can i also check that this has been ran on local (**against staging**) and found to work?

alexanderleegs: I ran this local against local - will also run against staging (once github is ok) to make sure!

cypress/api/collaborators.api.ts
alexanderleegs: Hmm in that case, I think we might be better off allowing the user that should remain as an argument for this method?

seaerchin (approved):
Lgtm

PR #1396, opened 2023-08-07
Is 346 fe UI flow for new domain and old domains

kishore03109:
## Problem

The flows for site launch should be different for new and old domains. This PR changes the storybook to allow for that functionality. 

## Solution

Modify the storybooks to show these two distinct flows.

https://github.com/isomerpages/isomercms-frontend/assets/42832651/3bb7c487-2eeb-454d-bb1c-85f889f90bd1

### DIFF
--- a/src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
+++ b/src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
@@ -138,7 +138,7 @@ const addSubtitlesForChecklist = (
         <>
           {tasksDone === SITE_LAUNCH_TASKS.SET_DNS_TTL - 1 && (
             <Text fontSize="small">
-              You can check you current DNS TTL through 3rd party applications
+              You can check your current DNS TTL through 3rd party applications
               such as <Link href="https://www.nslookup.io/">nslookup.io</Link>
             </Text>
           )}
--- a/src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
+++ b/src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
@@ -39,8 +39,8 @@ export const SiteLaunchChecklistTitle = (): JSX.Element => {
 }
 
 const getTextProps = (index: number, tasksDone: number): TextProps => {
-  const disabled = tasksDone < index - 1
-  if (disabled) {
+  const tasksPending = tasksDone < index - 1
+  if (tasksPending) {
     return {
       color: "interaction.support.disabled-content",
     }
@@ -49,7 +49,7 @@ const getTextProps = (index: number, tasksDone: number): TextProps => {
 
   if (taskDone) {
     return {
-      color: "base.content.strong",
+      color: "base.content.default",
     }
   }
 
--- a/src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
+++ b/src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
@@ -295,7 +295,7 @@ export const SiteLaunchChecklistBody = ({
           <Table variant="simple">
             <Thead>
               <Tr>
-                <Td bg="gray.100">
+                <Td bg="gray.100" width="70%">
                   <Text>
                     Site launch tasks ({tasksDone}/{numberOfTasks})
                   </Text>
--- a/src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
+++ b/src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
@@ -41,13 +41,13 @@ export const SiteLaunchChecklistTitle = (): JSX.Element => {
 }
 
 const getTextProps = (index: number, tasksDone: number): TextProps => {
-  const tasksPending = tasksDone < index - 1
+  const tasksPending = tasksDone < index
   if (tasksPending) {
     return {
       color: "interaction.support.disabled-content",
     }
   }
-  const taskDone = tasksDone >= index + 1
+  const taskDone = tasksDone >= index
 
   if (taskDone) {
     return {
@@ -259,7 +259,7 @@ export const SiteLaunchChecklistBody = ({
           key={i}
           textStyle="subhead-2"
           textColor="base.content.dark"
-          {...getTextProps(i + 1, tasksDone)}
+          {...getTextProps(i, tasksDone)}
         >
           {siteLaunchStatusProps?.isNewDomain
             ? TITLE_TEXTS_NEW_DOMAIN[getNewDomainTaskFrmIdx()]
@@ -335,8 +335,8 @@ export const SiteLaunchChecklistBody = ({
                     textColor="base.content.dark"
                     {...getTextProps(
                       siteLaunchStatusProps?.isNewDomain
-                        ? NEW_DOMAIN_SITE_LAUNCH_TASKS_LENGTH
-                        : SITE_LAUNCH_TASKS_LENGTH,
+                        ? NEW_DOMAIN_SITE_LAUNCH_TASKS_LENGTH - 1
+                        : SITE_LAUNCH_TASKS_LENGTH - 1,
                       tasksDone
                     )}
                   >
--- a/src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
+++ b/src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
@@ -21,6 +21,8 @@ import { useSiteLaunchContext } from "contexts/SiteLaunchContext"
 import { BxCopy, BxLifeBuoy } from "assets"
 import {
   DNSRecord,
+  getNewDomainTaskFrmIdx,
+  getOldDomainTaskFrmIdx,
   NEW_DOMAIN_SITE_LAUNCH_TASKS_LENGTH,
   SITE_LAUNCH_TASKS,
   SITE_LAUNCH_TASKS_LENGTH,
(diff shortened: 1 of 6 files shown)


src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
```diff
@@ -44,13 +38,24 @@ export const SiteLaunchChecklistTitle = (): JSX.Element => {
 }
 
 const getTextProps = (index: number, tasksDone: number): TextProps => {
-  const disabled = tasksDone >= index + 1 || tasksDone < index - 1
+  const disabled = tasksDone < index - 1
```
seaerchin, line 42: could i get more context on this change? previously, the text showed enabled styling if `tasksDone` was in the interval `[index - 1, index]` but this change causes the enabled styling to be `[index, ]`, which is a behavioural change.

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
```diff
@@ -124,38 +130,44 @@ const addSubtitlesForChecklist = (
...
-    <>
-      {tasksDone === SITE_LAUNCH_TASKS.DELETE_EXISTING_DNS_RECORDS - 1 && (
-        <Text fontSize="small">
-          This should be done as soon as domains are dropped
-        </Text>
-      )}
-    </>
-  )
-  return newTableMapping
+  return TABLE_MAPPING.map((task) => {
```
seaerchin, line 134: mainability: this is not a functionality concern, but having this code structure makes it difficult to decipher what this code chunk does. the reason for this is because don't directly map into a new component but return a component as part of an obj as part of an array, which is then consumed somewhere.

would be good if this returned a plain object but because this would qualify as a refactor, am ok with filing as ticket + add `TODO` + comment

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
```diff
@@ -124,38 +130,44 @@ const addSubtitlesForChecklist = (
...
+    const newTask = { ...task }
+    if (task.stepNumber === SITE_LAUNCH_TASKS.SET_DNS_TTL - 1) {
+      newTask.subTitle = (
+        <>
+          {tasksDone === SITE_LAUNCH_TASKS.SET_DNS_TTL - 1 && (
+            <Text fontSize="small">
+              You can check you current DNS TTL through 3rd party applications
```
seaerchin, line 141:
```suggestion
              You can check your current DNS TTL through 3rd party applications
```

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
```diff
@@ -233,9 +248,16 @@ export const SiteLaunchChecklistBody = ({
   let TABLE_MAPPING: TableMappingProps[] = Array.from(
```
seaerchin, line 249: nit: not an issue with this PR, but `UPPER_SNAKE_CASE` should be reserved for `const`, whicsh this isn't

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
```diff
@@ -273,7 +295,7 @@ export const SiteLaunchChecklistBody = ({
                     Site launch tasks ({tasksDone}/{numberOfTasks})
                   </Text>
                 </Td>
-                <Td bg="gray.100" textAlign="center">
+                <Td bg="gray.100" textAlign="center" width="30%">
```
seaerchin, line 303: take note that we need something to fill the missing 70%

src/types/siteLaunch.ts
```diff
@@ -47,9 +53,27 @@ export interface SiteLaunchDto {
...
+  "Drop existing domains on Cloudfront",
+  "Delete existing DNS records from your nameserver",
+  "Wait 1 hour to flush existing records",
+] as const
+
+export const TITLE_TEXTS_NEW_DOMAIN = [
+  TITLE_TEXTS_OLD_DOMAIN[
+    NEW_DOMAIN_SITE_LAUNCH_TASKS.APPROVE_FIRST_REVIEW_REQUEST
+  ],
+] as const
```
seaerchin, line 79: maintability: honestly i'm not a fan of this (using array based `const`) because it's not really descriptive and teh indexes are just raw numbers under the hood. 

if it's not too much effort, would it be possible to change this to const obj

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
kishore03109: be22fb8ab0e4d0062c20bd8e7ceb2af5a7d1460b

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
kishore03109: sure, for clarity 8a0385d39e653c69de3c4b0ee8cfef3def76ef71

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
kishore03109: thanks for the catch c9b0362f84fbcb41e2dd885d53437a313b2d6bf1

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
kishore03109: Yes, this is intended. The figma differentiates the colour of the "current" task, a "task" that is pending and a task that is already "done"

src/types/siteLaunch.ts
kishore03109: sure, 8d5806aad19d3bb374c7982d2ebe358b08b11ed7

I still want to iterate thr the tasks in order, so have introduced util funcs achieve this functionality

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
kishore03109: sure 6f6f7f4f1114bb40d81808192ad7005381a21ded

seaerchin:
i noticed a double render on initial click of the first checkbox, which led to it being selected -> unselected. was this a byproduct of the filming sequence (ie, not an actual bug) or does it impact functionality?

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
```diff
@@ -44,13 +41,24 @@ export const SiteLaunchChecklistTitle = (): JSX.Element => {
...
+    return {
+      color: "interaction.support.disabled-content",
+    }
+  }
+  const taskDone = tasksDone >= index + 1
+
+  if (taskDone) {
     return {
-      color: "interaction.support.disabled",
+      color: "base.content.default",
```
seaerchin, line 54: i'm just assuming that this is correct because it's quite hard to figure out what exactly `index` refers to here fyi

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
seaerchin: actually for additional clarification, what i meant here is that this `maps` and returns an object where `subTitle` is a jsx component and **not** a value.

this mixes in render (the component) together with the value (the text shown by the string). is this a big effort to chaneg?

src/types/siteLaunch.ts
seaerchin: sure. can just do `_.values` but the `switch` is fine too

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
```diff
@@ -230,20 +249,33 @@ export const SiteLaunchChecklistBody = ({
...
+      stepNumber: i,
       title: (
-        <Text fontSize="s" {...getTextProps(i + 1, tasksDone)} key={i}>
-          {TITLE_TEXTS[i]}
+        <Text
+          fontSize="s"
+          key={i}
+          textStyle="subhead-2"
+          textColor="base.content.dark"
+          {...getTextProps(i + 1, tasksDone)}
```
seaerchin, line 262: can i check why we need to pass the `+ 1` here? the `getTextProps` seem to already account for indexing?

seaerchin (approved):
lgtm - we do have some good to haves here like shifting the render logic away from the data. we don't have to do it here as functionality seems to be fine (barring the possible bug on initial render)

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
kishore03109: Correct me if I am wrong here, but yes, because `subtitle` is not a `string`! 
Reason for this is that the subtitles have things like links!

kishore03109:
@seaerchin 
> i noticed a double render on initial click of the first checkbox, which led to it being selected -> unselected. was this a byproduct of the filming sequence (ie, not an actual bug) or does it impact functionality?

hahah i tried again, I think i double clicked in the video :/ 
sanity check here: 
https://github.com/isomerpages/isomercms-frontend/assets/42832651/adcc4e4e-68d7-44d4-868b-1e3e6873e046

kishore03109:
!run-chromatic

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
kishore03109: hey you are right, this can be made clearer a663288b7575ae7d94a4d65d227cab787131dd99

sehyunidaaa:
Please review comments [on Chromatic](https://www.chromatic.com/test?appId=61fcc8e14c3af9003a102ac9&id=64d43eea175ed1920d3eaf3c)

PR #1397, opened 2023-08-07
ci(Mergify): only request reviews for stale open PRs

dcshzj:
## Problem

Mergify keeps going to old PRs and pinging the team for reviews, when they have already been merged or closed.

## Solution

Fix the conditions to only ping the team for stale and open PRs.


seaerchin (approved):
Lgtm, when I configured this I didn't realise mergify so dumb... Then end up I keep being lazy to fix

PR #1399, opened 2023-08-08
fix: selectors

alexanderleegs:
This PR fixes several e2e test selectors which were looking for the wrong string/colour. Tested on local against staging.

### DIFF
--- a/cypress/e2e/editPage.spec.ts
+++ b/cypress/e2e/editPage.spec.ts
@@ -10,12 +10,11 @@ import {
   Interceptors,
   TEST_REPO_NAME,
 } from "../fixtures/constants"
+import { SUCCESSFUL_EDIT_PAGE_TOAST } from "../fixtures/messages"
 
 // Constants
 const PRIMARY_COLOUR = "rgb(255, 0, 0)"
 const SECONDARY_COLOUR = "rgb(0, 255, 0)"
-const SUCCESSFUL_EDIT_PAGE_TOAST =
-  "Changes saved. See a preview on Staging, or request a Review for them to be published."
 
 describe("editPage.spec", () => {
   beforeEach(() => {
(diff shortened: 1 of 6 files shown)


cypress/e2e/editPage.spec.ts
```diff
@@ -14,6 +14,8 @@ import {
 // Constants
 const PRIMARY_COLOUR = "rgb(255, 0, 0)"
 const SECONDARY_COLOUR = "rgb(0, 255, 0)"
+const SUCCESSFUL_EDIT_PAGE_TOAST =
```
seaerchin, line 17: could we check to see if the old text was being referenced anywhere else? (or if this has been done already, mark thi  as closed)

seaerchin (approved):

harishv7 (approved):

cypress/e2e/notifications.spec.ts
```diff
@@ -12,7 +12,7 @@ import {
   visitE2eEmailTestRepo,
 } from "../utils"
 
-const UNREAD_COMMENTS_BG_COLOR = "rgb(235, 248, 255)"
+const UNREAD_COMMENTS_BG_COLOR = "rgb(248, 250, 254)"
```
harishv7, line 15: just to check - was these color changes from design system upgrade?

cypress/e2e/editPage.spec.ts
alexanderleegs: It was, changed in https://github.com/isomerpages/isomercms-frontend/pull/1399/commits/6445e6b05f8b56b2c228ec8fa48e72be84375c65, and also modified other references to all draw from the same source

cypress/e2e/notifications.spec.ts
alexanderleegs: I think so, @seaerchin can confirm

seaerchin: Screenshot 2023-08-08 at 1 57 22 PM

`blue.50` token here got changed to isomer specific token

cypress/e2e/editPage.spec.ts
seaerchin: i don't have anything else alr, feel free to merge

PR #1400, opened 2023-08-08
fix(preview): update styling

seaerchin:
## Problem
There are 2 user-facing problems, post design system upgrade (tbh idk why the upgrade causes it too).

Firstly, the homepage title was extremely small.
Secondly, hovering over the key-highlights section does not display the cursor as a pointer.

Closes IS-413

## Solution
1. For the homepage title, it was observed that the css variables required were missing. to fix this, the actual css file was inspected and the variables defined therein was **copied** over. Do take note that somehow, `sgds` styling is applied **over** our template styling despite the the `isomer-template` import occurring later... (see `.preview-panel`)
**Hero title post removal**
Screenshot 2023-08-08 at 2 07 48 PM

2. Removed `cursor: default` in our **template** styling so that the cursor will be `pointer` when hovering over key-highlights. To verify this, i checked against the actual staging site itself (ss attached below)
Screenshot 2023-08-08 at 1 24 12 PM

**NOTE: This isn't a design system issue** - checked against `master` prior to ds release, also had this ._.

Note that for `#key-highlights > .col`, there is **no** cursor styling, suggesting that our css in the CMS is out of date... 

3. removal of extra import (`buttons.scss`) - previously removed
4. removal of variable import in **template** css. This was checked by searching for css variables in the template css file and none from the variable import was found. 

## Tests 
- [ ] Go to homepage section
- [ ] Hovering over the highlights section on the preview panel should have your cursor change to pointer
- [ ] Add a hero title
- [ ] The hero title should look the same as the one on the staging site

### DIFF
--- a/src/styles/isomer-cms/pages/Editor.module.scss
+++ b/src/styles/isomer-cms/pages/Editor.module.scss
@@ -3,6 +3,43 @@
 @import "../elements/variable";
 
 .preview-styles {
+  // NOTE: These css variables all come from sgds.
+  // This has to be done because the css variables will
+  // otherwise be undefined at runtime as
+  // they are not loaded by sass.
+  --font-size-h1-display: 5.25rem;
+  --font-size-h1-mobile: 2.75rem;
+  --font-size-h1-display-mobile: 3.5rem;
+  --font-size-h1: 3.375rem;
+  --font-size-h1-lh: 3.75rem;
+  --font-size-h2: 2.75rem;
+  --font-size-h2-lh: 3.75rem;
+  --font-size-h2-mobile: 2rem;
+  --font-size-h3: 2rem;
+  --font-size-h3-lh: 2.8125rem;
+  --font-size-h3-mobile: 1.75rem;
+  --font-size-h4: 1.625rem;
+  --font-size-h4-lh: 2.25rem;
+  --font-size-h4-mobile: 1.5rem;
+  --font-size-h5: 1.375rem;
+  --font-size-h5-lh: 1.875rem;
+  --font-size-h6: 1.3rem;
+  --font-size-h6-lh: 1.5rem;
+  --font-size-p: 1.125rem;
+  --font-size-p-lh: 2rem;
+  --font-size-p-lh-content: 2.2rem;
+  --font-size-small: 1rem;
+  --card-variant-highlight-margin-top: 80px;
+  --global-alignment: 1rem;
+  --global-alignment-multiplier-xs: 0.25;
+  --global-alignment-multiplier-sm: 0.5;
+  --global-alignment-multiplier-lg: 2;
+  --global-alignment-multiplier-xl: 4;
+  --timepicker-height: 70px;
+  --radius: 0;
+  --radius-sm: 2px;
+  --radius-lg: 5px;
+
   @include meta.load-css("../../preview-panel.scss");
 }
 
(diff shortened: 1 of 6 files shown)


src/styles/isomer-cms/pages/Editor.module.scss
```diff
@@ -3,6 +3,43 @@
 @import "../elements/variable";
 
 .preview-styles {
+  // NOTE: These css variables all come from sgds.
+  // This has to be done because the css variables will
+  // otherwise be undefined at runtime as
+  // they are not loaded by sass.
```
harishv7, line 9: why are they not loaded?

src/styles/isomer-template.scss
```diff
@@ -11901,7 +11900,6 @@ html.has-navbar-fixed-top-widescreen {
 }
 
 #key-highlights .col {
```
harishv7, line 11902: which part of the Isomer UI do I test to see if it looks good?

src/styles/preview-panel.scss
```diff
@@ -2,6 +2,7 @@
 
 @include meta.load-css("sgds-govtech/css/sgds");
 @include meta.load-css("./isomer-template.scss");
```
harishv7, line 4: can we can before/after ss to the PR description followed by test plan (to test this PR + for release to assert what we need to verify looks fine on staging before deploy)?

src/styles/isomer-cms/pages/Editor.module.scss
seaerchin: honestly i'm not sure too. my initial thinking was that this was because of `load_css` rather than `use` (see [here](https://sass-lang.com/documentation/modules/meta/#load-css)) but using `@use` has the same issue ._.

src/styles/isomer-template.scss
seaerchin: i mentioned in the PR description that 
> cursor will be pointer when hovering over key-highlights

is this unclear ah? will add ss for additional clarity also

src/styles/preview-panel.scss
seaerchin: added test plan, will add ss!

src/styles/isomer-template.scss
seaerchin: ps, taking ss removes the cursor so i'm not sure how to show this

src/styles/isomer-cms/pages/Editor.module.scss
seaerchin: oddly enough, using an old version of `master` does **not** have this issue. however, given that design system doesn't actually impact this, i think this might be caused by deps upgrade (of `sass` possibly) leading to behaviour change ._.

harishv7: Oh, seems like that might be the root issue instead we should be solving for? But, am okay to merge this for now

src/styles/isomer-template.scss
harishv7: Can record as 5 second video haha

harishv7: I think it's fine i will just run through your test plan

harishv7: oh also you can do cmd+shift+3 then crop hahaha (for future use)

harishv7 (approved):
lgtm, verified as per test plan

QiluXie (approved):
lgtm

PR #1404, opened 2023-08-08
ref(homepage): use new components

seaerchin:
## Problem
Previously, the new components were introduced but not used. This PR replaces the existing stuff in `EditHomepage` w/ the new components

## Solution
The steps to replacement will be listed below for clarity: 
1. Replace `Draggable` with `Editable.Draggable` as resp shifted
2. for each component, shift the creation within the `map` loop into 2 segments - first, creating the `Droppable` (this is the actual dnd stuff) and secondly, creating the body (this is the `form` - retained existing code as-is)
3. for each `<Editor_type>Section.jsx`, the code dictating the expand/collapse logic was removed. (shifted into respective `AccordionItem`)

## Tests
- see #1403 for base list
- [ ] for new draggable components, clicking on the accordion button should expand (tbc w design if only the arrow, or the whole section)
- [ ] for new draggable components, **clicking on the card** should drag the component

## Screenshots
Screenshot 2023-08-09 at 2 29 19 AM

no vid cos laze

### DIFF
--- a/src/components/homepage/HeroSection.jsx
+++ b/src/components/homepage/HeroSection.jsx
@@ -1,5 +1,5 @@
 import { Droppable, Draggable } from "@hello-pangea/dnd"
-import { Button, IconButton } from "@opengovsg/design-system-react"
+import { Button } from "@opengovsg/design-system-react"
 import PropTypes from "prop-types"
 
 import { FormContext, FormError, FormTitle } from "components/Form"
@@ -11,8 +11,6 @@ import KeyHighlight from "components/homepage/KeyHighlight"
 
 import styles from "styles/App.module.scss"
 import elementStyles from "styles/isomer-cms/Elements.module.scss"
-
-import { isEmpty } from "utils"
 
 /* eslint
   react/no-array-index-key: 0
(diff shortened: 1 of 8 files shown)


src/components/homepage/HeroSection.jsx
```diff
@@ -33,197 +31,169 @@ const EditorHeroSection = ({
...
-          <FormTitle>Hero background image</FormTitle>
-          <FormFieldMedia
-            value={background}
-            id={`section-${sectionIndex}-hero-background`}
-            inlineButtonText="Select"
+          Dropdown
+        </label>
+      </div>
+      <span className={elementStyles.info}>
+        Note: you can only have either Key Highlights+Hero button or a Hero
```
kishore03109, line 102: are we missing a space here for `Highlights+Hero`?

src/components/homepage/HeroSection.jsx
```diff
@@ -33,197 +31,169 @@ const EditorHeroSection = ({
...
-      <>
-        <FormContext isRequired hasError={!!errors.sections[0].hero.title}>
-          <FormTitle>Hero title</FormTitle>
-          <FormField
-            placeholder="Hero title"
-            id={`section-${sectionIndex}-hero-title`}
-            value={title}
-            onChange={onFieldChange}
+  <>
+    <FormContext isRequired hasError={!!errors.sections[0].hero.title}>
```
kishore03109, line 41: wah, do you know why our shape is `errors.sections[0]` even though the shape only has one element `hero`? While this is out of scope of this pr, if during dev you find no such reason to use an array here, can shift to obj please! 

I am worried here that the original intention was to have an "array of errors" and we are only displaying the "first" error here

kishore03109 (approved):
> for new draggable components, only clicking on the dragHandle should drag the component

Hmm wait shouldnt it be the entire rect? was this the intended design? 
@sehyunidaaa let me know if I am mistaken

alexanderleegs:
> * [ ]  for new draggable components, **only clicking on the `dragHandle`** should drag the component

Not sure if i'm understanding this test correctly, but currently it seems that you're able to drag the component from any point?

https://github.com/isomerpages/isomercms-frontend/assets/22111124/1538fa01-4d99-4939-9b52-5dc466b08aca

src/components/homepage/HeroSection.jsx
alexanderleegs: I think it's because of the implicit assumption that the first section is the `hero` section - how the existing data structure works is that the homepage is represented by multiple sections, and each section is one of `hero`/`infopic`/`infobar`/`resources`, which will have that particular key only, so `sections[0]` always contains `hero`

seaerchin:
> > * [ ]   for new draggable components, **only clicking on the `dragHandle`** should drag the component
> 
> Not sure if i'm understanding this test correctly, but currently it seems that you're able to drag the component from any point?
> 
>  Screen.Recording.2023-08-17.at.5.13.12.PM.mov

My bad, confirmed behaviour with design and the inverse is true (drag at any point); will update test

src/components/homepage/HeroSection.jsx
kishore03109: wait are you trying to say that `sections[1]` could be of type `infopic, infobar, resources`?

src/components/homepage/HeroSection.jsx
seaerchin: not doing this - this whole component gets removed later on and this text doesn't exist anymore in figma (see [here](https://www.figma.com/file/7OtNYkc4s9BkFk05ConEbd/%E2%AD%90-V1.5?type=design&node-id=12732-119415&mode=dev)). 

also note that this missing space is present on prod too, see ss below

src/components/homepage/HeroSection.jsx
seaerchin: ^ yes, this is true. the existing code doesn't do a good job of fleshing this out but the invariant of `sections[0] === hero` is always maintained because we don't allow CRUD operations on it, only on subelements of it

src/components/homepage/HeroSection.jsx
```diff
@@ -33,197 +31,169 @@ const EditorHeroSection = ({
...
+                className={styles.card}
+                ref={droppableProvided.innerRef}
+                {...droppableProvided.droppableProps}
+              >
+                {highlights && highlights.length > 0 ? (
+                  <>
+                    <b>Hero highlights</b>
+                    {highlights.map((highlight, highlightIndex) => (
+                      <Draggable
+                        draggableId={`highlight-${highlightIndex}-draggable`}
```
kishore03109, line 142: we are using an array index as an id here -> could we use a uuid/ anything else unique instead? Worried that this will introduce bugs considering we are reordering the elements, which is specifically where using array as index key _will_ break.

seaerchin: hmm sounds good! this has been using this `draggableId` all this time (even on `production`) so idt it'll break but this is a good change to make regardless

seaerchin: fyi, gonna make this change in `Editable.DraggableAccordionItem` - this particular `draggableId` will be left as is becaues the abstraction we want to present is all draggable stuff goes through the new `Editable`. This particular `HeroSection.jsx` will be entirely removed later on

seaerchin: https://github.com/isomerpages/isomercms-frontend/pull/1404/commits/91ba5cf30a8ca374903d7a6f8ac640773b006955

alexanderleegs (approved):
lgtm

PR #1406, opened 2023-08-10
fix(chromatic): comment based chromatic builds

kishore03109:
## Problem

Currently, while the GH actions pass for comment based trigger, it keep building develop. The issue stems from `actions/checkout@v1` checks out the branch that tiggered the workflow. However, the `issue-comment` seems to be a separate [trigger by itself](https://github.com/actions/checkout/issues/331), and as such this feature does not work as intended.  

## Solution

Use solution as described [here](https://github.com/actions/checkout/issues/331#issuecomment-1438220926). This leverages on the fact that "issue number" === "pr number".

## Tests
I cannot test this unless this code since [this event will only trigger a workflow run if the workflow file is on the default branch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows).

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -27,6 +27,8 @@ jobs:
       frontend: ${{ steps.filter.outputs.frontend }}
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: refs/pull/${{ github.event.issue.number }}/head
       - uses: dorny/paths-filter@v2
         id: filter
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/chromatic.yml
```diff
@@ -65,10 +67,11 @@ jobs:
       - name: Don't run job if user is not part of Isomer core team or trigger words are not present
         if: ${{ steps.checkUserMember.outputs.isTeamMember != 'true' || !(github.event_name == 'issue_comment' && github.event.issue.pull_request && steps.check.outputs.triggered == 'true') }}
         run: exit 0
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
+        with:
+          ref: refs/pull/${{ github.event.issue.number }}/head
```
dcshzj, line 72: For this step, I think we can also do a `fetch-depth: 0` so that checking out will be faster, since we only need the files from the latest commit here.

kishore03109: Hmm wait arent we already getting the `head` no? (effectively latest commit here?)

dcshzj: Ah oops looks like I misunderstood the difference between `fetch-depth: 0` and `fetch-depth: 1`, the latter is what I was referring to (and looks like it is already done by default). Can ignore this!

PR #1414, opened 2023-08-15
Fix/is 422 update csp retrieval

alexanderleegs:
## Problem

This PR fixes an issue where CSP retrieval was failing for pages due to us still using the v1 endpoint. The `useCspHook` has been updated to typescript and now uses the v2 endpoints. In addition, the library being used (`csp-parse`) has been deprecated, so we've swapped the library out to `content-security-policy-parser`. The relevant methods in `cspUtils` have also been modified to account for the new formatting, namely that `csp-parse` used to return a string which had to be parsed, while `content-security-policy-parser` returns an object instead.

## Tests

- [ ] Go to edit page
- [ ] Attempt to embed something from a whitelisted site (e.g. youtube) - embed should show up properly
- [ ] Attempt to embed something from a non-whitelisted site - error message should show up in preview


harishv7 (commented):
lgtm, do we need a test plan for this PR?

alexanderleegs:
> lgtm, do we need a test plan for this PR?

Whoops forgot to add to the description, have added a quick test!

harishv7 (approved):

PR #1415, opened 2023-08-15
feat(homepage): add section button

seaerchin:
## Problem
Previously, the add section button was the old design. This refactors the add section button to take only an `onClick` and checks the `isDisabled` from the current state of the homepage (whether there's resources)

Closes IS-425
Closes IS-377

## Solution
1. use `Menu` from design system 
2. for `Resource` option, check prior to render if there already exists such an option to prevent duplicate creation


src/layouts/components/Editable/AddSectionButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+      {({ isOpen }) => (
+        <>
+          <Menu.Button
+            variant="outline"
+            isStretch
+            isOpen={isOpen}
+            rightIcon={undefined}
+            // NOTE: Need `sx` here as there is styling set on the exported `Menu.Button` component
+            // that overrides the `textAlign` prop
+            sx={{
```
kishore03109, line 17: wait so our figma deviates from ds? is this intended?

src/layouts/components/Editable/AddSectionButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <Text textStyle="body-1">{title}</Text>
+      <Text textColor="base.content.medium" textStyle="body-2">
+        {subtitle}
+      </Text>
+    </Menu.Item>
+  )
+}
+
+AddSectionButton.Option = AddSectionButtonOption
+AddSectionButton.List = Menu.List
```
alexanderleegs, line 54: Nit: is `AddSectionButton.List` being used?

src/types/homepage.ts
```diff
@@ -89,11 +89,16 @@ export type HomepageEditorHeroSection =
   | EditorHeroDropdownSection
   | EditorHeroHighlightsSection
 
+export type HeroFrontmatterSection = { hero: HomepageEditorHeroSection }
+export type ResourcesFrontmatterSection = { resources: Record<string, unknown> }
+
+export type EditorHomepageFrontmatterSection =
+  | HeroFrontmatterSection
+  | ResourcesFrontmatterSection
```
alexanderleegs, line 97: Hmm should this also include the other types of sections (infobar, infopic)?

alexanderleegs (approved):
lgtm, minor nits

src/layouts/EditHomepage.jsx
```diff
@@ -43,6 +49,7 @@ import { DEFAULT_RETRY_MSG } from "utils"
 import { useDrag, onCreate, onDelete } from "../hooks/useDrag"
 
 import { CustomiseSectionHeader, Editable } from "./components/Editable"
+import { AddSectionButton } from "./components/Editable/AddSectionButton"
 
 /* eslint-disable react/no-array-index-key */
```
kishore03109, line 54: [non-nit, but non blocking to this particular pr] we should remove this eslint :(

src/layouts/EditHomepage.jsx
```diff
@@ -904,6 +899,16 @@ const EditHomepage = ({ match }) => {
     return false
   }
 
+  // NOTE: sectionType is one of `resources`, `infopic` or `infobar`
+  const onClick = (sectionType) => {
+    createHandler({
+      target: {
+        value: sectionType,
+        id: "section-new",
```
harishv7, line 907: Shouldn't the id be dynamic based on the sectionType?

src/layouts/EditHomepage.jsx
seaerchin: if we ignore the error, it'll magically go away! jk, i think this can be removed post clean-up PR!

src/layouts/components/Editable/AddSectionButton.tsx
seaerchin: oh i wanted to use it - i remembered thinking that the API given here should be closer to the exported `Menu` api. 

updated and now the api here more closely matches that!

src/types/homepage.ts
seaerchin: yes - not done due to time. there's an existing jira issue filed for this (off the top of my head, IS-408 or IS-409)

src/layouts/components/Editable/AddSectionButton.tsx
seaerchin: good catch! i added this in cos it wasn't aligning (if memory serves, but this was ages ago). i tried removing and nothing seemed to changed so... removed!

src/layouts/EditHomepage.jsx
seaerchin: sadly this was the `id` used in the legacy code too ._. (see [here](https://github.com/isomerpages/isomercms-frontend/pull/1415/commits/0c02ca9b68fefa30a574dae88a5dd9862968e225#diff-a2ed50d7269adce5178a9e394aed981d74165d870a24d31393f4152af4521e68L35)) - the `createHandler` itself actually does all the logic iirc

PR #1416, opened 2023-08-15
chore(edithomepage): shift constants out

seaerchin:
## Problem
There were many constants whose reference wasn't clear in `EditHomepage`. This PR shifts them out for clarity and exposes a helper method to construct the error message.

## Solution
1. shift constants out to `ts` file. 
2. expose util method for getting error object.

### DIFF
--- a/src/layouts/EditHomepage/constants.ts
+++ b/src/layouts/EditHomepage/constants.ts
@@ -1,0 +1,41 @@
+export const RESOURCES_SECTION = {
+  title: "Resources",
+  subtitle: "Add a preview and link to your Resource Room",
+  id: "resources",
+} as const
+
+export const INFOBAR_SECTION = {
+  title: "Infobar",
+  subtitle: "Add informational text",
+  id: "infobar",
+  description: "Infobar description",
+  button: "Button Text",
+  url: "",
+} as const
+
+export const INFOPIC_SECTION = {
+  title: "Infopic",
+  subtitle: "Add an image and text",
+  id: "infopic",
+  description: "Infopic description",
+  button: "Button Text",
+  url: "", // No default value so that no broken link is created
+  image: "", // Always blank since the image modal handles this
+  alt: "Image alt text",
+} as const
+
+export const KEY_HIGHLIGHT_SECTION = {
+  title: "Key Highlight Title",
+  description: "Key Highlight description",
+  url: "", // No default value so that no broken link is created
+} as const
+
+export const DROPDOWN_ELEMENT_SECTION = {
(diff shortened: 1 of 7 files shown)


src/layouts/EditHomepage/constants.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export const DROPDOWN_ELEMENT_SECTION = {
+  title: "Hero Dropdown Element Title",
+  url: "", // No default value so that no broken link is created
+} as const
+
+export const DROPDOWN_SECTION = {
+  title: "Hero Dropdown Title",
+  options: [],
+} as const
```
harishv7, line 41: do we need the as const?

src/layouts/EditHomepage/utils.ts
```diff
@@ -0,0 +1,7 @@
+import _ from "lodash"
+
+export const getErrorValues = (
+  obj: Record<string, string>
+): Record<string, string> => {
+  return _.mapValues(obj, () => "")
```
harishv7, line 6: for err val we just return empty string for all keys?

seaerchin: apparently so - see [here](https://github.com/isomerpages/isomercms-frontend/blob/develop/src/layouts/EditHomepage.jsx#L53)

src/layouts/EditHomepage/constants.ts
seaerchin: yeap - the `as const` is to tell `ts` that this is a `readonly const` value. there's 2 differences with this cast

1. type inference becomes alot stricter - w/ the cast, `DROPDOWN_SECTION.title` is always inferred as `Hero dropdown title` and not `string` 
2. no reassignment of properties - with the cast, we cannot do `DROPDOWN_SECTION.title = "mutated"`; without the cast, we **can**. the `const DROPDOWN_SECTION` applies only to the _identifier_, which means that we can't do `const DROPDOWN_SECTION = ... // something with same shape` but can still do reassignment of properties

harishv7: cool, makes sense

harishv7 (approved):

src/layouts/EditHomepage/constants.ts
QiluXie: @seaerchin good explanation, using `as const` enforces literal types and read-only properties

QiluXie (approved):
lgtm

seaerchin:
**not merging this first because this involves a file rename, which is annoying to rebase**

PR #1417, opened 2023-08-15
feat(homepage): subsume site notif

seaerchin:
## Problem
Currently, our existing homepage has the site notification on top **as a separate block.** However, the figma wants the site notification to be shifted into the hero sectoin

Closes IS-428

## Solution
1. **Temporary measure**: shift notification section into `HeroSection` and add a prop for `notification`
2. For stuff in `HeroSection` where we were using our own home-rolled `Form` components prior to design system, shifts those to design system **with the exception of `FormFieldMedia`**. This is because `FormFieldMedia` is more involved + has modal so deferred fo rnow
3. in `EditHomepage`, removed unused code 

Screenshots: 
Screenshot 2023-08-15 at 3 10 32 PM

### DIFF
--- a/src/layouts/EditHomepage.jsx
+++ b/src/layouts/EditHomepage.jsx
@@ -1,6 +1,12 @@
 /* eslint-disable @typescript-eslint/no-shadow */
-import { useDisclosure, Text, HStack, VStack, Divider } from "@chakra-ui/react"
-import { Box } from "@chakra-ui/react"
+import {
+  Box,
+  useDisclosure,
+  Text,
+  HStack,
+  VStack,
+  Divider,
+} from "@chakra-ui/react"
 import { Button, Tag } from "@opengovsg/design-system-react"
 import update from "immutability-helper"
 import _ from "lodash"
--- a/src/layouts/EditHomepage.jsx
+++ b/src/layouts/EditHomepage.jsx
@@ -1,13 +1,7 @@
 /* eslint-disable @typescript-eslint/no-shadow */
-import {
-  useDisclosure,
-  Text,
-  HStack,
-  VStack,
-  Divider,
-  Box,
-} from "@chakra-ui/react"
-import { Button, Input, Tag } from "@opengovsg/design-system-react"
+import { useDisclosure, Text, HStack, VStack, Divider } from "@chakra-ui/react"
+import { Box } from "@chakra-ui/react"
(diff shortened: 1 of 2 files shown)


src/layouts/EditHomepage.jsx
```diff
@@ -954,192 +954,162 @@ const EditHomepage = ({ match }) => {
...
+                          <>
+                            {/* Hero section */}
+                            {section.hero && (
+                              <>
+                                <Editable.EditableAccordionItem title="Hero section">
+                                  <EditorHeroSection
                                     key={`section-${sectionIndex}`}
-                                    {...section.resources}
+                                    {...section.hero}
+                                    // TODO: Shift to rhf
```
kishore03109, line 987: what is rhf ah

src/layouts/EditHomepage.jsx
```diff
@@ -954,192 +954,162 @@ const EditHomepage = ({ match }) => {
...
-                                >
-                                  <EditorResourcesSection
+                        {frontMatter.sections.map((section, sectionIndex) => (
+                          <>
+                            {/* Hero section */}
+                            {section.hero && (
+                              <>
+                                <Editable.EditableAccordionItem title="Hero section">
+                                  <EditorHeroSection
                                     key={`section-${sectionIndex}`}
```
kishore03109, line 985: could we uuid instead?

kishore03109 (commented):
quick qn
> Temporary measure: shift notification section into HeroSection and add a prop for notification

why is this a temp measure? isnt noti section meant to be optional in the long run?

src/layouts/EditHomepage.jsx
seaerchin: react-hook-forms

src/layouts/EditHomepage.jsx
seaerchin: not gonna do it here - this is going to be deleted in subsequent RPs

seaerchin:
> quick qn
> 
> > Temporary measure: shift notification section into HeroSection and add a prop for notification
> 
> why is this a temp measure? isnt noti section meant to be optional in the long run?

this is a temporary measure because we have a new block design coming out soon! there won't be nested drag and drop anymore so this `notification` being here might change

kishore03109 (approved):

PR #1418, opened 2023-08-16
chore: downgrade webpack-dev-server

alexanderleegs:
This PR fixes an issue where errors on dev were showing up in an overlay -
Screenshot 2023-08-15 at 9 33 49 PM

This prevents us from running tests suites locally. It is likely that this issue was due to the default behaviour of `webpack-dev-server` changing to automatically include the overlay when errors are encountered. This PR explicitly downgrades the package so that we can run tests normally on local.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -138,7 +138,8 @@
         "ts-jest": "^29.0.5",
         "tsconfig-paths": "^4.0.0",
         "type-fest": "^2.19.0",
-        "typescript": "^4.7.4"
+        "typescript": "^4.7.4",
+        "webpack-dev-server": "4.14.0"
       },
       "engines": {
         "node": ">=16.0.0"
@@ -37248,9 +37249,10 @@
       }
     },
     "node_modules/webpack-dev-server": {
-      "version": "4.15.1",
+      "version": "4.14.0",
+      "resolved": "https://registry.npmjs.org/webpack-dev-server/-/webpack-dev-server-4.14.0.tgz",
+      "integrity": "sha512-KUgiUNUZldyx5xz3uK0dnXmvsSz03TAMCLtO1cUOb5oishh9sfP3vaI4XNY3EztrPUu98WKzamNfuaydTedYWQ==",
       "dev": true,
-      "license": "MIT",
       "dependencies": {
         "@types/bonjour": "^3.5.9",
         "@types/connect-history-api-fallback": "^1.3.5",
@@ -37258,7 +37260,7 @@
         "@types/serve-index": "^1.9.1",
         "@types/serve-static": "^1.13.10",
         "@types/sockjs": "^0.3.33",
-        "@types/ws": "^8.5.5",
+        "@types/ws": "^8.5.1",
         "ansi-html-community": "^0.0.8",
         "bonjour-service": "^1.0.11",
         "chokidar": "^3.5.3",
--- a/package.json
+++ b/package.json
@@ -184,7 +184,8 @@
     "ts-jest": "^29.0.5",
     "tsconfig-paths": "^4.0.0",
     "type-fest": "^2.19.0",
-    "typescript": "^4.7.4"
+    "typescript": "^4.7.4",
+    "webpack-dev-server": "4.14.0"
   },
   "config": {
     "commitizen": {


kishore03109 (commented):
hmmm hey, can we please not do this HAHAH 
throwing errors here is a good thing, instead we should be changing our API to return 2xx rather than a 4xx for unauth to prevent errors thrown in the login page

harishv7:
> hmmm hey, can we please not do this HAHAH throwing errors here is a good thing, instead we should be changing our API to return 2xx rather than a 4xx for unauth to prevent errors thrown in the login page

I am not sure I understood it right. But just to clarify right now we have these error overlays popping up everywhere there is a 4xx or 5xx error returned by our BE. And hence we are not able to run E2E locally as it interferes with Cypress. This is not just for the login page but for across the app. Have been encountering this recently after one of the past 1 or 2 releases

alexanderleegs:
> hmmm hey, can we please not do this HAHAH throwing errors here is a good thing, instead we should be changing our API to return 2xx rather than a 4xx for unauth to prevent errors thrown in the login page

Hmm it's not just for whoami that this occurs right? Every time we receive a 4xx/5xx error from the backend, we get this overlay, even though it does make sense for certain endpoints to be returning 4xx errors (e.g. conflict errors for updating an outdated page)

kishore03109:
@harishv7 @alexanderleegs hmm ok, this issue seems bigger than what I thought. What do we then feel about [changing the config just for e2e](https://webpack.js.org/configuration/dev-server/#overlay)? 

Even if we cannot figure that out, feels like just putting (even for local dev)
```
overlay: true,
```

might be a better solution that downgrading webpack and then us having to upgrade in the future again?

alexanderleegs:
> @harishv7 @alexanderleegs hmm ok, this issue seems bigger than what I thought. What do we then feel about [changing the config just for e2e](https://webpack.js.org/configuration/dev-server/#overlay)?
> 
> Even if we cannot figure that out, feels like just putting (even for local dev)
> 
> ```
> overlay: true,
> ```
> 
> might be a better solution that downgrading webpack and then us having to upgrade in the future again?

I didn't opt for this because it would involve moving our build script from `react-scripts` back to `craco` or some other equivalent, since the webpack config is not directly accessible with our existing setup. If we're encountering more cases where we need to modify webpack config, it might make sense to spend some time on this?

harishv7:
> > @harishv7 @alexanderleegs hmm ok, this issue seems bigger than what I thought. What do we then feel about [changing the config just for e2e](https://webpack.js.org/configuration/dev-server/#overlay)?
> > Even if we cannot figure that out, feels like just putting (even for local dev)
> > ```
> > overlay: true,
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > might be a better solution that downgrading webpack and then us having to upgrade in the future again?
> 
> I didn't opt for this because it would involve moving our build script from `react-scripts` back to `craco` or some other equivalent, since the webpack config is not directly accessible with our existing setup. If we're encountering more cases where we need to modify webpack config, it might make sense to spend some time on this?

@alexanderleegs @kishore03109 
I think the long-term solution might be to disable via webpack which is the recommended solution, but I would like to explore on this before we eject from react-scripts and go to craco or something - feels over-engineered for this use case.

For local dev wise, this might actually be useful - might need a quick discussion during retro perhaps. However the overlay seems to occur even for expected returns of 4xx and 5xx like whoami returning 401 and we redirect to login - even this shows up as overlay which is annoying.

However, if we decide to keep the overlay for local dev, then for local E2E we can do something like https://github.com/cypress-io/cypress/discussions/22500 (not yet tested) to prevent it from affecting tests.

for now, I am ok with the suggested package downgrade:
- it is not a major version downgrade
- we can make a decision and come back to this issue while unblocking local E2E for time being

kishore03109:
@harishv7 
> For local dev wise, this might actually be useful - might need a quick discussion during retro perhaps. However the overlay seems to occur even for expected returns of 4xx and 5xx like whoami returning 401 and we redirect to login - even this shows up as overlay which is annoying.

Hmm I think this is annoying by design right? ie we shouldnt be designing for expected 4xx and 5xx errors, and they have had an implication on our alarms. 

ok with this bandaid solution for now

kishore03109 (approved):

harishv7:
> ie we shouldnt be designing for expected 4xx and 5xx errors,

@kishore03109 can you expand on this? My understanding is that 4xx and 5xx are to be expected, but they should be properly handled on FE. Hence, the overlay showing already well-handled errors is distracting and adds noise to me as a developer.

kishore03109:
> @kishore03109 can you expand on this? My understanding is that 4xx and 5xx are to be expected, but they should be properly handled on FE. Hence, the overlay showing already well-handled errors is distracting and adds noise to me as a developer.

If we are testing for errors -> yes then that makes the error expected, and we should preserve that behaviour 

things like `whoami` -> we can treat it like how we used to treat `/siteInfo`. We used to return 4xx if the site did not exist in our repo (pre-migration). now we return a 200, but with an empty object. 

If normal dev work throws 4xx and 5xx errors, would that not be a sign of bad code smell?

harishv7:
> > @kishore03109 can you expand on this? My understanding is that 4xx and 5xx are to be expected, but they should be properly handled on FE. Hence, the overlay showing already well-handled errors is distracting and adds noise to me as a developer.
> 
> [since this convo is already outside scope of pr, this should be non-blocking] @harishv7 If we are testing for errors -> yes then that makes the error expected, and we should preserve that behaviour
> 
> things like `whoami` -> we can treat it like how we used to treat `/siteInfo`. We used to return 4xx if the site did not exist in our repo (pre-migration). now we return a 200, but with an empty object. Similar approach can be used here
> 
> If normal dev work throws a lot of 4xx and 5xx errors, would that not be a sign of bad design?
> 
> This also has a direct implication on alarms, and if we designed for expected 4xx + 5xx, even in happy paths we might get false +ves

> If we are testing for errors -> yes then that makes the error expected, and we should preserve that behaviour

Yes I am talking more along local dev expected test scenarios. I feel the overlay is valuable when it shows an unhandled promise/error which we didn't catch, but either way I think this is not a huge issue - we can disable if required perhaps

kishore03109:
@harishv7 @alexanderleegs hey turns out to just disable it for e2e is quite easy, opened a pr here #1427. Considering this, which solution would yall prefer? 
1. This were we disable overlays as a whole or 
2. only disable overlays for e2e?

harishv7:
> @harishv7 @alexanderleegs hey turns out to just disable it for e2e is quite easy, opened a pr here #1427. Considering this, which solution would yall prefer?
> 
> 1. This were we disable overlays as a whole or
> 2. only disable overlays for e2e?

Nice! I think we can disable for e2e for now then. @alexanderleegs what do you think

kishore03109:
> Nice, thanks @kishore03109 ! I think we can disable for e2e for now then (option 2). @alexanderleegs what do you think

oof when testing I didnt get the overlay by luck. If we need to disable for e2e, we need to disable for localhost:3000 -> which means we need to disable for local dev. 

tldr, #1427 is not functional.

PR #1421, opened 2023-08-16
feat(homepage): styling changes for infobar/pic/resources

seaerchin:
## Problem
We need to update the styling for the various bodies of the card components so that it is in alignment with our envisioned UI. This PR does that through defining a new component that controls the layout (`Editable.Section`) of the body together with replacing the components from the ones from chakra/ogpds

Closes IS-388
Closes IS-389

## Solution
1. add new component to structure the layout - this is `Editable.Section`
2. migrate old files over to `ts` 
3. shift handrolled components to use ogpds/chakra instead.
4. minor copy changes (`validators`/form labels etc)

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-08-16 at 2 39 23 PM

**AFTER**:
Screenshot 2023-08-16 at 2 38 59 PM

## Tests

- [ ] edit a field for infopic
- [ ] the preview should change
- [ ] click save
- [ ] the change should persist
- [ ] repeat ^ for infobar/resources 

## Notes
This PR does not tackle shifting out of logic for the form. This is still done using the various handlers at present, which isn't ideal.


src/layouts/EditHomepage.jsx
```diff
@@ -1025,20 +1025,19 @@ const EditHomepage = ({ match }) => {
                                 draggableId={`resources-${sectionIndex}-draggable`}
                                 index={sectionIndex}
                                 tag={<Tag variant="subtle">Resources</Tag>}
-                                title={section.resources.title}
+                                title="New resource widget"
```
alexanderleegs, line 1028: Is this title meant to be fixed?

src/layouts/components/Homepage/InfobarBody.tsx
```diff
@@ -0,0 +1,100 @@
...
+    <Button
+      id={`section-${index}`}
+      onClick={onClick}
+      alignSelf="center"
+      variant="clear"
+      colorScheme="critical"
+      mt="1rem"
+    >
+      Delete infobar
+    </Button>
```
alexanderleegs, line 98: I think `onClick` being used for the action taken (which is a delete) is not super intuitive to me? Not sure how the other engineers feel about this

src/layouts/components/Homepage/InfopicBody.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <Button
+        id={`section-${index}`}
+        onClick={onClick}
+        alignSelf="center"
+        variant="clear"
+        colorScheme="critical"
+        mt="1rem"
+      >
+        Delete infopic
+      </Button>
```
alexanderleegs, line 129: Same as above! Can resolve if not changing

src/layouts/components/Homepage/ResourcesBody.tsx
```diff
@@ -0,0 +1,90 @@
...
+    <Button
+      id={`section-${index}`}
+      onClick={onClick}
+      alignSelf="center"
+      variant="clear"
+      colorScheme="critical"
+      mt="1rem"
+    >
+      Remove resource widget
+    </Button>
```
alexanderleegs, line 88: Same as above! Can resolve if not changing

src/layouts/components/Homepage/InfobarBody.tsx
```diff
@@ -0,0 +1,100 @@
...
+  errors,
+}: InfobarBodyProps) => (
+  <Editable.Section>
+    <FormControl isRequired isInvalid={!!errors.subtitle}>
+      <FormLabel>Subtitle</FormLabel>
+      <Input
+        placeholder="This subtitle appears above the title"
+        id={`section-${index}-infobar-subtitle`}
+        value={subtitle}
+        onChange={onChange}
```
alexanderleegs, line 45: (Info, no change required)
New sections are created with default text in most fields, so we don't actually see the placeholder value for most of these components unless the user completely removes the text - just checking that design is aware!

alexanderleegs (approved):
lgtm, minor nits/clarifications

src/layouts/components/Homepage/InfobarBody.tsx
seaerchin: yeap! i initially thought it wasn't gonna be fixed too but i clarified with design - the wording is intended!

src/layouts/EditHomepage.jsx
seaerchin: yeap! i initially thought it wasn't gonna be fixed too but i clarified with design - the wording is intended!

src/layouts/components/Homepage/InfobarBody.tsx
seaerchin: was done cos there's only 1 button and this is the convention - open to changing if other ppl find this confusing also

PR #1422, opened 2023-08-16
ref(editable): style fixes + error hover

seaerchin:
## Problem
There are a couple of styling issues which do not line up with design. These will be described below. In addition, an error border is added to the accordion when collapsed.

**Design issues**
1. Hover states are applied across the **whole accordion** - this means that on expansion, the hover state is applied to the body also
2. padding issues wrt the `DragHandle` 

## Solution
1. For hover states, apply the `hover` conditionally only on the `Flex` wrapper for the accordion header (idk if this term makes sense). This necessitates the render prop pattern to access internal state on the `AccordionItem` - see [here](https://chakra-ui.com/docs/components/accordion#accessing-the-internal-state) 
    - current (and correct) behaviour is to have `hover` **only on collapsed states**
2. For the `DragHandle`, it has to display over the accordion such that when it is clicked/hovered, the interaction states on the accordion header does not trigger. Hence, this was given `position: absolute` + styling to fix it in place.
3. For the error border, we cannot use an actual border due to it cutting into the card, causing the items to realign themselves and throw alignment between acc

## Screenshots
**Jiggling items with border**
Recording 2023-08-16 at 13 27 50

**current behaviour**
Recording 2023-08-22 at 00 57 58

## Tests

- [ ] Go to homepage
- [ ] Add/expand infobar
- [ ] for each field, type >30 chars (or if textarea, >160) to trigger error state
- [ ] toggle close
- [ ] border should appear

### DIFF
--- a/src/layouts/EditHomepage.jsx
+++ b/src/layouts/EditHomepage.jsx
@@ -1055,7 +1055,7 @@ const EditHomepage = ({ match }) => {
                               <Editable.DraggableAccordionItem
                                 index={sectionIndex}
                                 tag={<Tag variant="subtle">Infobar</Tag>}
-                                title={section.infobar.title}
+                                title={section.infobar.title ?? "New infobar"}
                                 isInvalid={_.some(
                                   errors.sections[sectionIndex].infobar
                                 )}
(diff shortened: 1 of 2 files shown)


alexanderleegs (approved):
lgtm

PR #1423, opened 2023-08-16
fix: conditional for checkCSP 

alexanderleegs:
This PR fixes an issue where `checkCSP` could be run before the csp was retrieved. This PR adds an additional check that both csp and editorvalue have loaded before the editPage content is populated.

Also fixes an issue where we were searching for the wrong string for *

This issue doesn't seem to occur consistently - it seems that the cause might be a delay in retrieving csp, so we can trigger it by setting an extra delay on the getCsp hook. 

Before: content loads immediately, images show up as blocked

https://github.com/isomerpages/isomercms-frontend/assets/22111124/7388554d-823e-4dc9-9045-8843b9f6e4f9

Before (even longer delay for csp to load): 

https://github.com/isomerpages/isomercms-frontend/assets/22111124/4cce1d2b-430f-4e7b-a9a7-d7f3f94231b4

After: content takes time to load in (waiting for csp to be retrieved), images show up properly

https://github.com/isomerpages/isomercms-frontend/assets/22111124/51bb539f-9aa4-48e1-878e-3bc77473b4c1

### DIFF
--- a/src/layouts/EditPage.jsx
+++ b/src/layouts/EditPage.jsx
@@ -9,6 +9,7 @@ import {
 import { Button } from "@opengovsg/design-system-react"
 import axios from "axios"
 import DOMPurify from "dompurify"
+import _ from "lodash"
 import { marked } from "marked"
 import PropTypes from "prop-types"
 import { useEffect, useState } from "react"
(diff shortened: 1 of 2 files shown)


harishv7:
@alexanderleegs should we test E2E staging on this to check if it fixes the issue?

harishv7 (approved):

PR #1427, opened 2023-08-17
fix(e2e): dont show error overlay

kishore03109:
## Problem

Linked to #1418, where there are annoying overlays. This PR does not need to be merged, we can discuss the approach and decide if which solution we want to go towards.
 

## Solution

change webpack config to ignore overlay. This will only affect e2e tests, but not local dev, while #1418 impacts both local dev + e2e tests. 

Screenshot 2023-08-17 at 9 08 54 AM

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -43,6 +43,11 @@ export default defineConfig({
                 },
               ],
             },
+            devServer: {
+              client: {
+                overlay: true,
+              },
+            },
           },
         })
       )


cypress.config.ts
```diff
@@ -43,6 +43,11 @@ export default defineConfig({
                 },
               ],
             },
+            devServer: {
+              client: {
+                overlay: true,
```
harishv7, line 48: should this be true or false? we want to disable overlays for local E2E right?

kishore03109:
oof false alarm - turns out that I didint see it for the `move` test spec by pure luck, this solution does not work. Since this is a timeboxed effort, closing pr

PR #1428, opened 2023-08-17
fix(e2e): reduce mem usage

kishore03109:
## Problem

E2e keeps crashing -> not productive 

## Solution

redce mem usage + enable experimental feature. Not too worries about the experimental feature since the testing of e2e is internal and as such we can be more risk adverse on this. 

Before this, I wasnt able to even complete 4 tests for `move`, after this change the entire test suite was able to run. 
Screenshot 2023-08-17 at 9 08 54 AM

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -8,7 +8,8 @@ export default defineConfig({
   viewportHeight: 768,
   viewportWidth: 1366,
   projectId: "nxbty1",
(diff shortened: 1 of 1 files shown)


harishv7 (approved):
Tried this and cypress does seem faster than before! Thanks for looking into this @kishore03109 !

PR #1429, opened 2023-08-17
ref(homepage): adds new dropdown component for hero

seaerchin:
## Problem
Currently, our hero component has an outdated design. Design has a new version of the hero dropdown, which is implemented in this PR.

Closes IS-386

## Solution
1. Define new component, `HeroBody`, that wraps both `Dropdown` and `Highlights` 
    - this could potentially take `children` and only do styling + the 3 forms which, in hindsight, seems like a better solution tbh
2. Define new component, `HeroDropdownSection`, that does the actual dnd stuff + handles the form fields for the dropdown as a whole. 
3. Adds a new `DragDropContext` - we require this because `Editable.Draggable` requires a `onDragEnd` prop, which is defined _very_ high up the component tree but is the sole "source of truth" for dnd logic. This is similar in function to `react-hook-form`'s `FormProvider`, which just does a re-export of `useForm`'s return value. (also where i got the idea for the API from lmao)

**BEFORE**:
Screenshot 2023-08-17 at 12 54 06 PM

**AFTER**:
Recording 2023-08-17 at 12 52 57

**Empty state**
Screenshot 2023-08-17 at 12 54 34 PM

## Tests
- [ ] Go to homepage
- [ ] type in the title/subtitle fields, this should work + preview should chnage
- [ ] type in something extremely long in ^, should trigger validation with error message
- [ ] select a background image, this should work
- [ ] select dropdown
- [ ] add dropdown option
- [ ] perform steps 2-3, this should work + preview should chnage
- [ ] add another dropdown option - previwe should chnage
- [ ] drag and drop the options, preview should change + UI should reflect change

### DIFF
--- a/src/layouts/components/Homepage/HeroBody.tsx
+++ b/src/layouts/components/Homepage/HeroBody.tsx
@@ -1,0 +1,186 @@
+import {
+  FormControl,
+  Flex,
+  Icon,
+  Text,
+  Box,
+  Divider,
+  HStack,
+  RadioGroup,
+} from "@chakra-ui/react"
+import {
+  FormErrorMessage,
+  FormLabel,
+  Input,
+  Radio,
+} from "@opengovsg/design-system-react"
+import { useState } from "react"
+import { BiInfoCircle } from "react-icons/bi"
+
+import { FormContext, FormError, FormTitle } from "components/Form"
+import FormFieldMedia from "components/FormFieldMedia"
+
+import { Editable } from "layouts/components/Editable"
+
+import { EditorHeroDropdownSection, EditorHomepageState } from "types/homepage"
+
+import {
+  HeroDropdownFormFields,
+  HeroDropdownSection,
+} from "./HeroDropdownSection"
+
+export interface HeroBodyFormFields {
+  title: string
+  subtitle: string
+  background: string
+  dropdown: HeroDropdownFormFields
+}
+
+interface HeroBodyProps extends HeroBodyFormFields {
+  index: number
+  onChange: () => void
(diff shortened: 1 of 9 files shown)


src/layouts/components/Homepage/HeroBody.tsx
```diff
@@ -0,0 +1,191 @@
...
+                .hero as EditorHeroDropdownSection
+            }
+            errors={errors}
+            onCreate={() => createHandler({ target: { id: "dropdownelem" } })}
+            onChange={onChange}
+            onClick={(event) => deleteHandler(event, "Dropdown Element")}
+          />
+        ) : (
+          // <HeroHighlightSection />
+          <div />
```
alexanderleegs, line 186: Highlights section is missing!

src/types/homepage.ts
```diff
@@ -112,7 +113,7 @@ export interface EditorHomepageState {
 
 export interface EditorHeroDropdownSection {
   dropdown: {
-    options: []
+    options: { title: string; url: string }[]
```
alexanderleegs, line 116: I think title and url here might be optional!

src/layouts/components/Homepage/HeroBody.tsx
seaerchin: resolving this - added in next PR

src/types/homepage.ts
seaerchin: updated typing

alexanderleegs (approved):

harishv7 (approved):

PR #1431, opened 2023-08-17
chore(homepage): style fixes and cleanup

seaerchin:
## Problem
Some style fixes + refactors to avoid prop drilling + behavioural changes. These will be detailed below.

## Solution
1. add additional styling onto `EditableSidebar` - this is taken from our current css styling to avoid having an additional wrapper + `classNames` so that we can be fully using ogpds
2. add error border onto `EditableAccordionItem` + pass `isInvalid` - done so that a collapsed accordion item would have the error border styling on error
3. add `noOfLines = {1}` on `AccordionItems` so that on title overflow,  it truncates properly and doesn't interfere with UI
4. use render prop for `HeroBody` to avoid having to drill props down
5. add more power into the `EditableContext` + have various components get behaviour from said context. Done so that UI components can remain ignorant of behaviour + avoid prop drilling
6. add `scrollTo` wrapper for `scrollIntoView` so that we can have `scrollBehaviour: smooth` as defining the behaviour doesn't work on **sync** function calls 
7. minor code organization for neatness
 
## Tests
- [ ] Go to homepage
- [ ] add both `highlights` and `dropdown` for the hero section
- [ ] type a very long title
- [ ] it should truncate the title 
- [ ] the error message for the character count should come below the field
- [ ] collapsing the draggable item (in hero section) should have error border styling
- [ ] collapsing hero section should have error border styling 
- [ ] expanding the accordion should scroll smoothly to that section
- [ ] adding an item should scroll smoothly to that section
- [ ] repeat the above steps but for `Infobar/pic/resources`


src/layouts/components/Homepage/InfobarBody.tsx
seaerchin, line 1: line changes due to using context

src/layouts/components/Homepage/ResourcesBody.tsx
seaerchin, line 1: line changes due to using context

src/contexts/EditableContext.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { createContext, PropsWithChildren, useContext } from "react"
+
+interface EditEvent {
+  target: {
+    id: string
+    value?: string
+  }
+}
+
+interface EditableContextReturn {
```
harishv7, line 11: sounds like a verb, can make a noun? EditableContextProperties?

src/contexts/EditableContext.tsx
```diff
@@ -0,0 +1,43 @@
...
+  // NOTE: Assumed to be 0, which is the
+  // hero element if not provided
+  onDisplay: (elemType: string, index?: number) => void
+}
+
+export const EditableContext = createContext<null | EditableContextReturn>(null)
+
+export const useEditableContext = (): EditableContextReturn => {
+  const context = useContext(EditableContext)
+  if (context === null) {
```
harishv7, line 28: nit: can use !context?

src/layouts/EditHomepage.jsx
```diff
@@ -43,22 +43,59 @@ import {
...
 
-// Constants
-// ==========
 const RADIX_PARSE_INT = 10
 
+// NOTE: `scrollIntoView` does not work when called synchronously
+// to avoid this problem, we do a `setTimeout` to wrap it.
+// This calls it after 1ms, which allows it to work.
+const scrollTo = (ref) => {
+  setTimeout(() =>
```
harishv7, line 66: where is the setTimeout's time delay here? wouldn't this execute immediately?

src/layouts/EditHomepage.jsx
```diff
@@ -43,22 +43,59 @@ import {
...
+    })
+  )
+}
+
+const getHasError = (errorArray) =>
+  _.some(errorArray, (err) =>
+    _.some(err, (errorMessage) => errorMessage.length > 0)
+  )
+
+const getHasErrors = (errors) => {
```
harishv7, line 78: nit: is the get necessary, can just be hasErrors ?

src/layouts/EditHomepage.jsx
```diff
@@ -959,214 +966,190 @@ const EditHomepage = ({ match }) => {
...
-                      bg="base.canvas.alt"
-                      p="1.5rem"
-                      spacing="1.5rem"
-                      alignItems="flex-start"
+                    {heroSection.map((section, sectionIndex) => {
+                      return (
+                        <>
+                          <Editable.EditableAccordionItem
+                            title="Hero section"
+                            // TODO: Add `isInvalid` prop to `EditableAccordionItem`
```
harishv7, line 992: this prop is already added right?

seaerchin: Oops, will delete!

src/layouts/EditHomepage.jsx
seaerchin: It is because we're extracting a derived value (`hasError`) from a base value. The `get` is there to hint that this is an action

src/layouts/EditHomepage.jsx
seaerchin: Yeap! When omitting the time delay or when it's less than 1, it defaults to 1 (millisecond). 

This is the behaviour that we want - when a user expands an accordion item, they scroll to that section immediately!

src/contexts/EditableContext.tsx
seaerchin: Sounds good

src/contexts/EditableContext.tsx
seaerchin: Actually this is convention because the type is for the return value of `useContext`! 

See [here](https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/editable/src/use-editable.ts#L408C6-L408C6)

harishv7: gotcha

src/layouts/EditHomepage.jsx
harishv7: makes sense!

src/layouts/components/Homepage/HeroDropdownSection.tsx
```diff
@@ -64,7 +60,8 @@ export const HeroDropdownSection = ({
         <Text mt="0.5rem" textStyle="body-2" textColor="base.content.medium">
           Drag and drop dropdown options to rearrange them
         </Text>
-        <Editable.Accordion>
+        {/* TODO: Add `displayHandler` */}
+        <Editable.Accordion onChange={() => onDisplay("dropdownelem")}>
```
alexanderleegs, line 64: Is `index` not required for `onDisplay` (since we're passing `displayHandler`)?

src/layouts/EditHomepage.jsx
```diff
@@ -295,43 +333,10 @@ const EditHomepage = ({ match }) => {
 
   useEffect(() => {
     if (scrollRefs.length > 0) {
-      scrollRefs[frontMatter.sections.length - 1].current.scrollIntoView()
+      scrollTo(scrollRefs[frontMatter.sections.length - 1])
     }
   }, [scrollRefs, frontMatter.sections.length])
```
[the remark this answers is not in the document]
seaerchin: going to [this](https://cms.isomer.gov.sg/sites/a-test-v4/homepage) site has the same behaviour on production.

this was never noticeable because we instantly snapped there but now cos of smooth scrolling, we do know @__@

cc @jandonyada and @sehyunidaaa if this should be fixed (imo, yes if easy but can drop if hard cos is improvement)

src/layouts/components/Homepage/HeroDropdownSection.tsx
seaerchin: we never actually pass `displayHandler` HAHAHAHHA

seaerchin: it's only required for `section`! for `dropdown/highlight`, it automatically scrolls to index 0 (ie, the hero section)!

src/layouts/components/Editable/Editable.tsx
```diff
@@ -89,13 +90,24 @@ export const EmptySection = ({
...
-    <Flex flexDir="column" bg="base.canvas.alt">
+    <Flex
+      flexDir="column"
+      bg="base.canvas.alt"
+      width="450px"
+      overflowY="scroll"
+      pb="100px"
+      // NOTE: We reserve 80px **each** for
+      // both the header and the footer
+      h="calc(100vh - 160px - 1rem)"
```
harishv7, line 108: whats the 1rem for?

seaerchin: good question - i'm not entirely sure too. i've taken the styling from the existing editor styles [here](https://github.com/isomerpages/isomercms-frontend/blob/develop/src/styles/isomer-cms/pages/Editor.module.scss#L98). 

i didn't investigate this because it was tangential to the feature

src/layouts/EditHomepage.jsx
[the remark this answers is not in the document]
seaerchin: removing - will release as separate PR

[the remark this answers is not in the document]
harishv7: @seaerchin isn't this just doing: 

```
const [isFirstLoadComplete, setIsFirstLoadComplete] = useState(false)

useEffect(() => setIsFirstLoadComplete(true), [])

 useEffect(() => {
    if (scrollRefs.length > 0 && isFirstLoadComplete) {
      scrollTo(scrollRefs[frontMatter.sections.length - 1])
    }
  }, [scrollRefs, frontMatter.sections.length])
```

src/layouts/components/Editable/Editable.tsx
harishv7: was just weird to see all 3 units in same expression hahaha! but since we alr had this before i think is okay to keep it

src/layouts/EditHomepage.jsx
```diff
@@ -59,17 +59,6 @@ import { ResourcesBody } from "./components/Homepage/ResourcesBody"
 
 const RADIX_PARSE_INT = 10
 
-// NOTE: `scrollIntoView` does not work when called synchronously
```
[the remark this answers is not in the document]
seaerchin: yes - see slack thread [here](https://opengovproducts.slack.com/archives/CK68JNFHR/p1692770079054749)

src/layouts/EditHomepage.jsx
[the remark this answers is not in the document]
alexanderleegs: Most likey this doesn't work because the sections are generated from the frontmatter some time after the initial page load - this should work instead

```
const [isFirstLoadComplete, setIsFirstLoadComplete] = useState(false)

  useEffect(() => {
    if (scrollRefs.length > 0) {
      if (!isFirstLoadComplete) {
        setIsFirstLoadComplete(true)
      } else {
        scrollTo(scrollRefs[frontMatter.sections.length - 1])
      }
    }
  }, [scrollRefs, frontMatter.sections.length])
```

harishv7 (approved):
lgtm, tested with test plan as well

PR #1432, opened 2023-08-18
chore(homepage): flag edit homepage

seaerchin:
**Huge LoC is due to copy pasting current `EditHomepage` into `LegacyEditHomepage`**

## Problem
No feature flagging atm for `EditHomepage` - this PR adds the feature flag in

## Solution
1. define a list of testing sites and check if a site is iether an ogp site or a testing site, then conditionally render.
2. we have to define a `LegacyEditHomepage` + `FeatureFlaggedHomepage` because PRs on `EditHomepage` will alter the actual `EditHompage.jsx` file, which means that the "new" file name **must not change**.  

ticket for deletion is [here](https://isomeropengov.atlassian.net/browse/IS-477?atlOrigin=eyJpIjoiNzExODU1OGM3MzI2NGI2NzhlZDQwNTE0NTA3YzEzZmEiLCJwIjoiaiJ9)


src/constants/sites.ts
```diff
@@ -0,0 +1,22 @@
+export const TESTING_SITES = [
```
kishore03109, line 1: Hmm why are we not going with the env var approach like SL and GGs is going with? Is there a long wait time before this feature gets rolled out to public?

src/layouts/FeatureFlaggedHomepage.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import { TESTING_SITES } from "constants/sites"
+
+import EditHomepage from "./EditHomepage"
+import LegacyEditHomepage from "./LegacyEditHomepage"
+
+export const FeatureFlaggedHomepage = (): JSX.Element => {
+  const params = useParams<{ siteName: string }>()
+  const isOgpSite =
+    params.siteName.startsWith("ogp-") ||
```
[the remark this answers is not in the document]
seaerchin: hmm - there are 82 `ogp-` sites, which is q alot of hard coding. is `ogp-nyjc` a moe site?

src/constants/sites.ts
seaerchin: do those exist on frontend or backend? do note that this is supposed to roll out in phases, so for the most part, i don't think using an env var > this has any significant benefits

src/layouts/FeatureFlaggedHomepage.tsx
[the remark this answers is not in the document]
seaerchin: if they are an moe site, more tempted to do a rename of their site to `moe-nyjc`. the `ogp-nyjc` breaks convention and is mental overhead.

src/constants/sites.ts
kishore03109: sl in fe, gg in be 
nit would be to keep the approach same for all ff-ed features 
but ye, if you dont need dynamic reload then no high value

src/layouts/FeatureFlaggedHomepage.tsx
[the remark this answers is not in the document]
kishore03109: rename = site down temp so got ops overhead ya.
plus we also dont know how many of the 82 sites also have edge cases like this. highly suggest whitelisting here.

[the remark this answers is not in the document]
kishore03109: note: ogp site doesnt start with `ogp-` because its just `ogp`, need to add this edge case in

src/constants/sites.ts
seaerchin: resolving - not gonna do this because this is probably gonna be deleted by next sprint (2 weeks)

QiluXie (commented):
lgtm, will approve after the whitelisting of ogp sites

src/layouts/FeatureFlaggedHomepage.tsx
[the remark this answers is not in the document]
seaerchin: updated to whitelisting

src/constants/sites.ts
```diff
@@ -0,0 +1,32 @@
...
+  "rcd-test-repo",
+  "rcd-test-repo2",
+  "rcd-test-repo3",
+  "test-jacksontest",
+  "zhongjun-test",
+  "zhongjun-test-amplify",
+  "zhongjun-test-emailam",
+]
+
+export const OGP_SITES = [
```
harishv7, line 24: what if we need to rollout to more sites maybe a few agencies? we need a hotfix rollout like for ggs right? Not sure if this could be a possibility. do clarify if it is a possibility, and if so let's move to env var. Else, fine with this temp approach

kishore03109 (approved):
Maybe we need to add this to LD and empower XFN to enable this feature + measure impact, do we have a plan for this rollout for the edit homepage stuff? 

That being said, its more impt that the code goes in by tmr, approving this in light of that first

src/constants/sites.ts
seaerchin: to be clear here - i **want** this to be a constant in code, rather than an env var. i'll outline my reasoning below

env vars are primarily useful for _ad-hoc_ updates - when we want to adjust things on the fly without triggering a full deploy. this is useful for minor features (subjective but ok) where we don't really care about having, for instance, tracking values between rollouts. this, however, does introduce overhead to eg: the release process where the oncall needs to update said values.

having a constant in code means that our rollout is tied to our weekly releases. this trade off is acceptable to me for a **phased** rollout, where the phase are going to be in weeks anyway (ie, i release to 10% this week, 20% next week for example). additionally, it's very explicit and clear about what's gonna get the new features and we get things liek type checking, git history etc so it's even more clear. 

me relying on constants also rests on the assumption that growthbook/ld is going to be in either this week or the next - this is just a short lived measure for clarity that won't be here in a few weeks.

QiluXie (approved):
lgtm

PR #1437, opened 2023-08-23
fix(chromatic): trigger build for push to develop for Chromatic baseline

dcshzj:
## Problem

Chromatic builds do not run on pushes to the `develop` branch, which causes Chromatic to be unable to update its baseline.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Chromatic now builds for pushes to the `develop` branch, so that the baseline can be updated on Chromatic
    - It is important that this gets done, as we no longer build Chromatic automatically on every pull request. This means that there may be frontend changes that land into the `develop` branch, but Chromatic does not know. We cannot guarantee that everyone will force a Chromatic build for their PRs, which may cause the baseline to be incorrect over time.

## Tests

I hope it works, can only know after merging.

## Deploy Notes

*None*

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -6,13 +6,15 @@ on:
   issue_comment:
     types: [created, edited]
 
+  push:
+    branches:
+      - develop
+
 env:
   # The full comment text to match to trigger this workflow
   ISOMER_TRIGGER_COMMENT: "!run-chromatic"
   # The slug for the Isomer core team
   ISOMER_CORE_TEAM_SLUG: core
-  # The file name of this workflow, should match this file name
-  ISOMER_COMMENT_WORKFLOW_NAME: chromatic.yml
   # Use GitHub Token
   GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
   # Required for the chromatic action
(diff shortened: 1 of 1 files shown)


.github/workflows/chromatic.yml
```diff
@@ -6,13 +6,15 @@ on:
   issue_comment:
     types: [created, edited]
 
+  push:
```
kishore03109, line 9: uhm wait why would this build the baseline ah? 
this just says run this when something is pushed to dev, we are not doing smt like 'make dev as the baseline' since that is a state that is held within chromatic no?

kishore03109 (commented):
abit confused why this would work, can huddle if my comment is unclear

kishore03109:
+ lets time box this uh, if cannot just go back to previous push based, and just inform devs that if they have giant prs with lots of commits can pause building chromatic so as to not take up the credits

.github/workflows/chromatic.yml
dcshzj: Hmm Chromatic has knowledge that the baseline is the `develop` branch, so every build on this branch would be updating this baseline. We would need to run this on every push to the `develop` branch so that we can get Chromatic to update the baseline they have stored, and PRs that trigger their respective Chromatic builds will be compared against this baseline.

This is known from the build here: https://www.chromatic.com/review?appId=61fcc8e14c3af9003a102ac9&number=1435&type=linked&view=changes. You can notice that there are no changes detected, even though that PR has introduced a new story. This is because previously we have been building branches as if they are `develop`, causing the baselines to keep being updated and thus we see no changes detected.

kishore03109 (approved):

PR #1439, opened 2023-08-23
Feat/is 494 smooth scroll

seaerchin:
(edited by @alexanderleegs)
## Problem

Building off the work @seaerchin for smooth scroll - the issue we were facing was that we wanted to scroll to newly created sections by checking for times when `scrollRefs`/`frontMatter.sections.length` change, but only after the first paint. However, since we populate `scrollRefs`/`frontMatter.sections.length` via its own useEffect hook (from legacy code), this means that we also can't rely on a pure `hasLoaded` variable which is toggled on mount.

The solution is to have the check for first load inside the useEffect which controls the scrolling. Note that this variable cannot be included in the dependency array, otherwise it would immediately trigger twice and cause the scroll to the end.

Testing

- [ ] Navigate to homepage
  - [ ]  Preview should remain at hero and not scroll down

### DIFF
--- a/src/hooks/useAfterFirstLoad.ts
+++ b/src/hooks/useAfterFirstLoad.ts
@@ -1,0 +1,22 @@
+import { useState } from "react"
+
+/**
+ * This custom hook is used as a helper for effects which should only take place on the second time onwards,
+ * e.g. if data is not fully populated at the time of mounting.
+ * It wraps a method in a conditional to check that it has been executed once before.
+ */
+export const useAfterFirstLoad = (
+  wrappedMethod: (...args: unknown[]) => void
+) => {
+  const [isFirstLoadComplete, setIsFirstLoadComplete] = useState(false)
+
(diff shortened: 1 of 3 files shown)


src/layouts/EditHomepage.jsx
```diff
@@ -320,10 +331,17 @@ const EditHomepage = ({ match }) => {
     loadPageDetails()
   }, [homepageData])
 
+  const [isFirstLoadComplete, setIsFirstLoadComplete] = useState(false)
```
seaerchin, line 334: could we shift this into a custom hook? this doesn't seem too sustainable to me because we have a large amount of state in this file - would be more comfortable having this in another hook!

alexanderleegs: Shifted in https://github.com/isomerpages/isomercms-frontend/pull/1439/commits/0ae03dc7afc029741b9e2484d9060ea5b6c8c79b!

PR #1440, opened 2023-08-23
Feat/fixSLStyling

kishore03109:
## Problem

Currently storybooks cannot toggle the steps, design found it harder to review + write a guide for SL 
This also fixes trivial copy issues. 

## Solution

Change API so the `increaseStepNumber` and the `decreaseStepNumber` both come from the context directly, similar to the `increasePageNumber` + `decreaseStepNumber`. 
For Verbosity, I have also added 4 screens for design to review + increase possibilities of catching regression when other devs develop SL in the future.  

**Breaking Changes**

  - Details ...

## Before & After Screenshots

**AFTER**:
Storybook check boxes should work

New Domain:
https://github.com/isomerpages/isomercms-frontend/assets/42832651/981f7417-4c6f-48d8-a34c-61d2609f38c9

Old Domain:
https://github.com/isomerpages/isomercms-frontend/assets/42832651/6cee2d80-3afe-4609-b2b3-c01844e59661

4 added storybooks: 
A: `new-domain-non-www-dns-records`
Screenshot 2023-08-23 at 3 34 43 PM

B: `new-domain-www-dns-records`
Screenshot 2023-08-23 at 3 34 59 PM

C: `old-domain-non-www-dns-records`
Screenshot 2023-08-23 at 3 35 09 PM

D: `old-domain-www-dns-records`
Screenshot 2023-08-23 at 3 35 20 PM

Now copied got tooltip:
Screenshot 2023-08-23 at 3 32 35 PM

## Tests

None, mostly to sped up reviewal + trivial copy changes

### DIFF
--- a/src/contexts/SiteLaunchContext.tsx
+++ b/src/contexts/SiteLaunchContext.tsx
@@ -83,7 +83,6 @@ export const SiteLaunchProvider = ({
   })
 
   const { data: siteLaunchDto } = useGetSiteLaunchStatus(siteName)
-  console.log({ siteLaunchDto })
   const queryClient = useQueryClient()
   const refetchSiteLaunchStatus = () => {
     /**
--- a/src/contexts/SiteLaunchContext.tsx
+++ b/src/contexts/SiteLaunchContext.tsx
@@ -9,6 +9,7 @@ import { useGetSiteLaunchStatus } from "hooks/siteDashboardHooks"
 import { launchSite } from "services/SiteLaunchService"
 
 import {
+  NEW_DOMAIN_SITE_LAUNCH_TASKS_LENGTH,
   SiteLaunchBEStatus,
   SiteLaunchFEStatus,
   SiteLaunchFEStatusType,
@@ -17,6 +18,7 @@ import {
   SiteLaunchTaskTypeIndex,
   SITE_LAUNCH_PAGES,
   SITE_LAUNCH_TASKS,
+  SITE_LAUNCH_TASKS_LENGTH,
 } from "types/siteLaunch"
 
 interface SiteLaunchContextProps {
@@ -29,6 +31,8 @@ interface SiteLaunchContextProps {
   increasePageNumber: () => void
   decreasePageNumber: () => void
   pageNumber: SiteLaunchPageIndex
+  handleIncrementStepNumber: () => void
+  handleDecrementStepNumber: () => void
 }
 
 interface SiteLaunchProviderProps {
@@ -79,7 +83,7 @@ export const SiteLaunchProvider = ({
   })
 
   const { data: siteLaunchDto } = useGetSiteLaunchStatus(siteName)
-
+  console.log({ siteLaunchDto })
   const queryClient = useQueryClient()
   const refetchSiteLaunchStatus = () => {
     /**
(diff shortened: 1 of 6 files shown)


src/contexts/SiteLaunchContext.tsx
```diff
@@ -79,7 +83,7 @@ export const SiteLaunchProvider = ({
   })
 
   const { data: siteLaunchDto } = useGetSiteLaunchStatus(siteName)
-
+  console.log({ siteLaunchDto })
```
seaerchin, line 86: Sir.

src/contexts/SiteLaunchContext.tsx
```diff
@@ -29,6 +31,8 @@ interface SiteLaunchContextProps {
   increasePageNumber: () => void
   decreasePageNumber: () => void
   pageNumber: SiteLaunchPageIndex
+  handleIncrementStepNumber: () => void
+  handleDecrementStepNumber: () => void
```
seaerchin, line 35: we need to take care to prevent props explosion here. this is not an issue for now because idt we will have active work going on for site launch, but we want to avoid the case where we just put everything here and ignore proper component structure, which will impact maintainability in the future.

tl;dr: is ok but we should avoid adding stuff here already.

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
```diff
@@ -64,25 +70,54 @@ const getTextProps = (index: number, tasksDone: number): TextProps => {
...
+        placement="right-start"
+      >
+        <PopoverTrigger>
+          <Box display="flex" alignItems="center" maxW="16rem">
+            <Text
+              textStyle="body-2"
+              color="black"
+              whiteSpace="nowrap"
+              overflow="hidden"
+              textOverflow="ellipsis"
```
seaerchin, line 89: nit: prefer `noOfLines`

seaerchin (approved):
no functional blockers but we should avoid temptation to stuff everything into context

src/layouts/SiteLaunchPad/components/SiteLaunchCheckList.tsx
seaerchin: have to check the styling - the `Text` hasn't overflowed its bounding box yet i think

seaerchin: cos from the docs [here](https://chakra-ui.com/docs/components/text/usage#truncate-text), it _should_ display ellipsis (and in prior experience, it has)

src/contexts/SiteLaunchContext.tsx
kishore03109: can, but for future ref may I trouble you to explain how I could make this more maintainable + allow storybook to be versatile enough to be able to click thr the checkboxes? I am abit unsure of how I could have made this PR better in the future

seaerchin: without pair programming/actually working on this portion of the codebase, it's bit difficult to actually find something that works well. 1 sign that you could use to signal for help could be that you find yourself needing the context for everything, which shouldn't be the case.

1 thing that you can try is to see which parts of the state your child components truly need and refactoring so that you pass that state to your children. i've done htis [here](https://github.com/isomerpages/isomercms-frontend/pull/1431/commits/b82157bc034c48799679131bf1c76e94831ca7bc) because the two nested components were p disjoint in their props but only really needed the `selectedOption`

PR #1444, opened 2023-08-23
fix(legacyedithomepage): add back missing imports

seaerchin:
## Problem
Previously, when we merged the feature flagged PR, i only copy pasted the `EditHomepage.jsx` file. however, within `EditHomepage.jsx`, it was also referencing other files that had been previously deleted. This wasn't an issue as it branched off a base that had said legacy files, but on merge back to `develop`, the legacy files got deleted, which led to an error.

## Solution
1. add legacy files in
2. fix import

## Tests
- [ ] go to [this](http://localhost:3000/sites/a-test-v4/homepage) site. it should display the new components
- [ ] go to [this](http://localhost:3000/sites/ogp-designers/homepage) site. toggle the dropdown/highlight in `HeroSection` - it should be ok.

### DIFF
--- a/src/layouts/FeatureFlaggedHomepage.tsx
+++ b/src/layouts/FeatureFlaggedHomepage.tsx
@@ -3,7 +3,7 @@ import { useParams } from "react-router-dom"
 import { OGP_SITES, TESTING_SITES } from "constants/sites"
 
 import EditHomepage from "./EditHomepage"
-import LegacyEditHomepage from "./LegacyEditHomepage"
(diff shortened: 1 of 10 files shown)


harishv7 (commented):
@seaerchin to confirm, these were previously existing files which were added in without any functionality/logic changes right?

QiluXie (approved):
lgtm

seaerchin:
> @seaerchin to confirm, these were previously existing files which were added in without any functionality/logic changes right?

only change is in the imports. i got these files off `master` .-.

PR #1450, opened 2023-08-24
fix(edithomepage): remove optional tags

seaerchin:
## Problem
`notifications` and `infopic` had `optional` tags, which should be removed per design

## Solution
set `isRequired` to `true` even though it isn't really :(

# Screenshots
## Before
Screenshot 2023-08-30 at 2 09 03 PM

## After
Screenshot 2023-08-24 at 1 01 37 PM

## Tests

- [ ] open hero section
- [ ] notification should not have `optional` tag after the form's label
- [ ] add `infobar` 
- [ ] infobar should not have `optional` tag after the form's label

### DIFF
--- a/src/layouts/components/Homepage/HeroBody.tsx
+++ b/src/layouts/components/Homepage/HeroBody.tsx
@@ -71,7 +71,7 @@ export const HeroBody = ({
   return (
     <>
       <Editable.Section spacing="1.25rem">
-        <FormControl>
+        <FormControl isRequired>
           <FormLabel>Notification Banner</FormLabel>
           <Input
             // TODO: Remove the `id/onChange`
--- a/src/layouts/components/Homepage/InfopicBody.tsx
+++ b/src/layouts/components/Homepage/InfopicBody.tsx
@@ -46,7 +46,7 @@ export const InfopicBody = ({
 
   return (
     <Editable.Section>
-      <FormControl isInvalid={!!errors.subtitle}>
+      <FormControl isRequired isInvalid={!!errors.subtitle}>
         <FormLabel>Subtitle</FormLabel>
         <Input
           placeholder="This subtitle appears above the title"
@@ -56,7 +56,7 @@ export const InfopicBody = ({
         />
         <FormErrorMessage>{errors.subtitle}</FormErrorMessage>
       </FormControl>
-      <FormControl isInvalid={!!errors.title}>
+      <FormControl isRequired isInvalid={!!errors.title}>
         <FormLabel>Title</FormLabel>
         <Input
           placeholder="Your infopic title goes here"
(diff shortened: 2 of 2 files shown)


alexanderleegs (commented):
Actually the notification banner shouldn't be compulsory though? Not all sites want to have the banner, could we check in with design again on this

Or does this mean that we don't have the optional text but it's actually still optional haha

src/layouts/components/Homepage/HeroBody.tsx
```diff
@@ -71,7 +71,7 @@ export const HeroBody = ({
   return (
     <>
       <Editable.Section spacing="1.25rem">
-        <FormControl>
+        <FormControl isRequired>
           <FormLabel>Notification Banner</FormLabel>
```
harishv7, line 75: I am not seeing the optional on InfoBar on staging, I think you mean Infopic in your description
Screenshot 2023-08-29 at 5 38 46 PM

src/layouts/components/Homepage/HeroBody.tsx
```diff
@@ -71,7 +71,7 @@ export const HeroBody = ({
   return (
```
harishv7, line 71: will be good to have a before ss also to compare while reviewing

src/layouts/components/Homepage/InfopicBody.tsx
```diff
@@ -46,7 +46,7 @@ export const InfopicBody = ({
 
   return (
     <Editable.Section>
-      <FormControl isInvalid={!!errors.subtitle}>
+      <FormControl isRequired isInvalid={!!errors.subtitle}>
```
harishv7, line 49: > notifications and infobar had optional tags, which should be removed per design

@seaerchin so this will mandate people to key in some value? What happens to existing sites where ppl didnt fill them in?

seaerchin: nope, this is entirely UI only HAHAHA

src/layouts/components/Homepage/HeroBody.tsx
seaerchin: added

src/layouts/components/Homepage/HeroBody.tsx
seaerchin: updated PR desc

seaerchin:
> Actually the notification banner shouldn't be compulsory though? Not all sites want to have the banner, could we check in with design again on this
> 
> Or does this mean that we don't have the optional text but it's actually still optional haha

the latter - no text, still optional. enforcement comes later

QiluXie:
> > Actually the notification banner shouldn't be compulsory though? Not all sites want to have the banner, could we check in with design again on this
> > Or does this mean that we don't have the optional text but it's actually still optional haha
> 
> the latter - no text, still optional. enforcement comes later

Enforcement comes later means it will be changed to compulsory later?

seaerchin:
> > > Actually the notification banner shouldn't be compulsory though? Not all sites want to have the banner, could we check in with design again on this
> > > Or does this mean that we don't have the optional text but it's actually still optional haha
> > 
> > 
> > the latter - no text, still optional. enforcement comes later
> 
> Enforcement comes later means it will be changed to compulsory later? A bit confused here, is it our goal to make notification compulsory eventually?

we have some fields that are supposed to be compulsory, but the existing validation has a bug. this is fixed in the next PR

alexanderleegs (approved):

PR #1451, opened 2023-08-24
Fix/placeholder inputs

seaerchin:
## Problem
Previously, we always _prefilled_ sections on initial add - this allows people to slip up and accidentally click save even when it was the default values. 

## Solution
- Separate _prefill_ values and _placeholder_ values. This then allows us to add an empty block (on left panel) while still displaying a sane preview value on the preview panel. This was done through first filtering by which properties have an empty value then merging with default values on render. 
- fix existing validation - we always check if `length < 0`, which will never be true. 
- extract stuff out from current validation to reuse - this was because the validation also **sets** some additional properties on an error object.
- write util function to compute `hasError` from the `formState` - this is required because **errors are only set on change**. this means that, for example, users can add an empty block and save instantly (as no change was made), which runs counter to what we want. 
- add `onBlur = onChange`. as the value is controlled by us and there's no value change on blur, this only serves to retrigger validation on blur events so that we inform the user about what failed validation (empty input on initial create for eg) 

## Tests
- [ ] trigger validation on existing items
    - [ ] this can be done through exceeding max limit or deleting stuff on required fields
- [ ] create a new block
- [ ] save should be disabled 
- [ ] expand accordion
- [ ] click on input
- [ ] click away
- [ ] error message should pop up 
- [ ] save **should still** be disabled 

## Notes 
Feels like we're inventing the wheel here already. i don't want to put more effort into this but our validation strategy feels abit off as closing the accordion, for example, **does not** trigger validation. 

maybe can consider shifting to a form lib that does this for us but is a significant effort so i think we should just live with this as an intermediate measure **until we decide that we want to nudge website builders in the right dir**

@sehyunidaaa - maybe we can prompt the `isOptional` already otherwise feels like users might grapple with this


src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -522,7 +502,8 @@ const EditHomepage = ({ match }) => {
 
       switch (elemType) {
         case "section": {
-          const val = enumSection(value, false)
+          // NOTE: setting `isError` to true to get an empty object
+          const val = enumSection(value, true)
           const err = enumSection(value, true)
```
alexanderleegs, line 507: Hmm if we're always retrieving an empty object now, do we even need `enumSection` to take in the boolean `isError`

I'm fine with leaving it if it takes a lot of effort to remove though

src/templates/homepage/HeroSection.jsx
```diff
@@ -37,7 +42,7 @@ const HeroButton = ({ button }) => (
 
 const HeroDropdownElem = ({ title }) => (
   <div className={editorStyles["bp-dropdown-item"]}>
-    <h5>{title}</h5>
+    <h5>{title || DROPDOWN_ELEMENT_SECTION.title}</h5>
   </div>
```
alexanderleegs, line 46: Is there a reason why we pass in hero section this way instead of like the other preview sections? (destructuring existing object so that only non-blank params will overwrite)

src/utils/validators.js
```diff
@@ -169,329 +169,304 @@ export const MEDIA_SETTINGS_TITLE_MIN_LENGTH = 6
...
       // Title is too short
-      if (value.length < HIGHLIGHTS_TITLE_MIN_LENGTH) {
-        errorMessage = `The title should be longer than ${HIGHLIGHTS_TITLE_MIN_LENGTH} characters.`
+      if (value.length <= HIGHLIGHTS_TITLE_MIN_LENGTH) {
+        return `The title should be longer than ${HIGHLIGHTS_TITLE_MIN_LENGTH} characters.`
       }
       // Title is too long
       if (value.length > HIGHLIGHTS_TITLE_MAX_LENGTH) {
-        errorMessage = `The title should be shorter than ${HIGHLIGHTS_TITLE_MAX_LENGTH} characters.`
+        return `The title should be shorter than ${HIGHLIGHTS_TITLE_MAX_LENGTH} characters.`
```
alexanderleegs, line 195: The max length comparisons on this page should be `>=` to match the error message!

src/layouts/EditHomepage/EditHomepage.jsx
seaerchin: idts but i lazy to remove HAHA

src/templates/homepage/HeroSection.jsx
seaerchin: don't the other preview adopt this pattern also?

seaerchin: oh cos this way was less work - i'm open to keeping it consistent tbh

src/utils/validators.js
seaerchin: lmao good catch - 4 years in the making HAHAHAHA

src/layouts/EditHomepage/EditHomepage.jsx
seaerchin: removed

src/templates/homepage/HeroSection.jsx
seaerchin: this is abit annoying to do here cos we have to do a type cast + overwrite a child array. lmk if this is a blocker for u

src/templates/homepage/InfopicLeftSection.jsx
```diff
@@ -12,25 +12,16 @@ import { fetchImageURL } from "utils"
...
-    title,
-    subtitle,
-    description,
-    button,
-    sectionIndex,
-    imageUrl,
-    imageAlt,
-    siteName,
-  },
+  { title, subtitle, description, button, sectionIndex, image, alt, siteName },
```
QiluXie, line 15: nit: 
- Why did we rename `imageUrl` to `image`? Is it because `imageUrl` is actually the image file name? 
- if `alt` is the alt image url that can be loaded, is it better to name it as `altUrl`?

src/layouts/EditHomepage/utils.ts
```diff
@@ -1,7 +1,72 @@
...
+  const heroSection = (initialSection as HeroFrontmatterSection).hero
+
+  let hasError = false
+  if (isHighlightSection(heroSection)) {
+    hasError = heroSection.key_highlights.reduce((acc, cur) => {
+      const curErr = _(cur)
+        .entries()
+        .map(([key, value]) => validateHighlight(key, value))
+        .some()
+        .valueOf()
```
QiluXie, line 35: Do we need `.valueOf()` since `.some()` will return a boolean? 
boolean? 
```suggestion
        .some()
```

src/layouts/EditHomepage/utils.ts
```diff
@@ -1,7 +1,72 @@
...
+
+  // NOTE: This is mutually exclusive with the above check
+  // as our hero section only has these 2 options.
+  if (isDropdownSection(heroSection)) {
+    hasError = heroSection.dropdown.options.reduce((acc, cur) => {
+      const curError = _(cur)
+        .entries()
+        .map(([key, value]) => validateDropdownElem(key, value))
+        .some()
+        .valueOf()
```
QiluXie, line 48: Do we need `.valueOf()` since `.some()` will return a boolean? 
```suggestion
        .some()
```

src/layouts/EditHomepage/utils.ts
```diff
@@ -1,7 +1,72 @@
...
+      const curErr = _(cur)
+        .entries()
+        .map(([key, value]) => validateHighlight(key, value))
+        .some()
+        .valueOf()
+      return acc || curErr
+    }, hasError)
+  }
+
+  // NOTE: This is mutually exclusive with the above check
```
QiluXie, line 40: Can `heroSection` be both `DropdownSection` and `HighlightSection`?
How is this mutual exclusiveness enforced?

src/layouts/EditHomepage/utils.ts
```diff
@@ -1,7 +1,72 @@
...
+    hasError = heroSection.dropdown.options.reduce((acc, cur) => {
+      const curError = _(cur)
+        .entries()
+        .map(([key, value]) => validateDropdownElem(key, value))
+        .some()
+        .valueOf()
+      return acc || curError
+    }, hasError)
+  }
+
```
QiluXie, line 52: Looking at the logic, if `hasError` is `true` at this point, we can directly return `true`

src/layouts/EditHomepage/utils.ts
```diff
@@ -1,7 +1,72 @@
...
+    hasError = heroSection.key_highlights.reduce((acc, cur) => {
+      const curErr = _(cur)
+        .entries()
+        .map(([key, value]) => validateHighlight(key, value))
+        .some()
+        .valueOf()
+      return acc || curErr
+    }, hasError)
+  }
+
```
QiluXie, line 39: Looking at the logic, if `hasError` is `true` at this point, we can directly return `true`

src/layouts/EditHomepage/utils.ts
seaerchin: it's enforced in `HeroBody` via a radio - there are only 2 options, and the `setState` in hero body also only accepts a discriminated union (dropdown/highlight)

src/layouts/EditHomepage/utils.ts
seaerchin: not inclined to do this - seems abit confusing to always do `if (hasError) return true` and the size is small enough that runtime isn't a concern.

lmk if you feel strongly about this.

src/templates/homepage/HeroSection.jsx
alexanderleegs: I'm okay with leaving it as is if it's annoying to implement!

src/layouts/EditHomepage/utils.ts
QiluXie: Sounds reasonable, not feeling that strongly about this, can leave it as it is

PR #1455, opened 2023-08-29
IS-508: Introduce Interface for GrowthBook

harishv7:
## Problem

Previously, we used a constants file to keep track of the allowed feature flags. 

Closes IS-508

## Solution

This PR uses the recommended approach of using an interface instead. This still allows for type safety during development.

**Breaking Changes**

  - Details ...


src/types/featureFlags.ts
```diff
@@ -1,15 +1,7 @@
...
-
-export type FeatureFlagTypes = typeof FeatureFlagSupportedTypes[keyof typeof FeatureFlagSupportedTypes]
-
-export type FeatureFlag = {
-  key: string
-  type: FeatureFlagTypes
+// Example usage: const gb = useGrowthBook<FeatureFlags>();
+export interface FeatureFlags {
+  sampleFlag: string
+  sampleFlag2: string
```
seaerchin, line 4: are these supposed to be actual feature flags?

seaerchin (commented):
not in favour of having example code in prod - would be ok with this PR if the sample stuff was commented otu

src/types/featureFlags.ts
harishv7: @seaerchin these are just samples. we will need to modify this when we actually create flags on growthbook

seaerchin (approved):
lgtm. please do comment out the samples

PR #1457, opened 2023-08-29
Feat/is 443 nav bar styling

alexanderleegs:
## Problem
This PR updates our outdated Edit Navbar page to make use of the newly introduced drag and drop components, as well as update the styling of the page.

Resolves IS-443

Feature flagging will be done in a future PR

## Solution
This PR swaps out most of the existing legacy components in favour of the new `Editable` components with updated styles - most of the logic has been retained from the existing component.

Of particular note is the change to `DraggableAccordionItem`'s icon button - this is required for nested drag and drop to work consistently. This is because nested drag and drop is not natively supported by the library we're using, but components identified as interactive by the library will prevent the underlying draggable to be selected - we're working around this by falsely declaring the second level items as a `contentEditable` section

Before:
Screenshot 2023-08-29 at 3 53 15 PM
Screenshot 2023-08-29 at 3 54 03 PM

After:
Screenshot 2023-08-29 at 3 54 33 PM
Screenshot 2023-08-29 at 3 55 00 PM

## Tests

- [ ] Go to edit nav bar
- [ ] Add sections, all of them should work
- [ ] Type in the section fields, preview should update
  - [ ] Changing the folder in a folder section should change all the related sublinks
- [ ] Drag and drop should work (first level)
- [ ] Deleting a section should work
- [ ] Type in a very long string for title - should produce error, minimizing component should show error bar on the side
- [ ] Menu group tests
  - [ ] Add section should work
  - [ ] drag and drop (2nd level should work), try this multiple times to make sure the right component can be consistently selected to drag
  - [ ] type in a very long string for title of subitem - should produce error, minimizing submenu should show error bar, minimizing entire menu group component should show error
  - [ ] Delete section should work

### DIFF
--- a/src/layouts/EditNavBar.jsx
+++ b/src/layouts/EditNavBar.jsx
@@ -31,7 +31,7 @@ import { validateLink } from "utils/validators"
 import { getEditNavBarData } from "api"
 import { DEFAULT_RETRY_MSG, deslugifyDirectory, isEmpty } from "utils"
 
-import { CustomiseSectionsHeader, Editable } from "./components/Editable"
+import { Editable } from "./components/Editable"
 import { AddSectionButton } from "./components/Editable/AddSectionButton"
 import { FolderMenuBody } from "./components/NavBar/FolderMenuBody"
 import { GroupMenuBody } from "./components/NavBar/GroupMenuBody"
@@ -649,7 +649,11 @@ const EditNavBar = ({ match }) => {
                       alignItems="flex-start"
                     >
                       <VStack spacing="0.5rem" alignItems="flex-start">
-                        <CustomiseSectionsHeader />
+                        <Text textStyle="h5">Menu Items</Text>
+                        <Text textStyle="body-2">
+                          You can specify a folder or resource room to
+                          automatically populate its links.
+                        </Text>
                       </VStack>
                       <DragDropContext onDragEnd={onDragEnd}>
                         <Editable.Droppable
@@ -760,7 +764,7 @@ const EditNavBar = ({ match }) => {
                   </Editable.Accordion>
 
                   <Box p="1.5rem">
-                    <AddSectionButton>
+                    <AddSectionButton sectionType="menu item">
                       <AddSectionButton.List>
                         {/* NOTE: Check if the site contains any collections in `options` 
                             if it does not, prevent creation of a `folder` section
--- a/src/layouts/EditNavBar.jsx
+++ b/src/layouts/EditNavBar.jsx
@@ -764,7 +764,7 @@ const EditNavBar = ({ match }) => {
                   </Editable.Accordion>
 
                   <Box p="1.5rem">
-                    <AddSectionButton sectionType="menu item">
+                    <AddSectionButton buttonText="Add menu item">
                       <AddSectionButton.List>
                         {/* NOTE: Check if the site contains any collections in `options` 
                             if it does not, prevent creation of a `folder` section
--- a/src/layouts/EditNavBar.jsx
+++ b/src/layouts/EditNavBar.jsx
@@ -1,19 +1,20 @@
 import { useDisclosure, Box, Text, HStack, VStack } from "@chakra-ui/react"
-import { Button } from "@opengovsg/design-system-react"
+import { DragDropContext } from "@hello-pangea/dnd"
+import { Button, Tag } from "@opengovsg/design-system-react"
 import update from "immutability-helper"
 import _ from "lodash"
 import PropTypes from "prop-types"
 import { useEffect, useState } from "react"
-import { DragDropContext } from "react-beautiful-dnd"
 import { useQuery } from "react-query"
 
 import { Footer } from "components/Footer"
 import Header from "components/Header"
 import { LoadingButton } from "components/LoadingButton"
-import NavSection from "components/navbar/NavSection"
 import { WarningModal } from "components/WarningModal"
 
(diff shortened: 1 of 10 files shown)


dcshzj:
@alexanderleegs should we feature flag this?

alexanderleegs:
> @alexanderleegs should we feature flag this?

Yup, planning to add in a separate PR

src/layouts/components/Editable/AddSectionButton.tsx
```diff
@@ -2,7 +2,15 @@ import { HStack, Icon, MenuItemProps, Text } from "@chakra-ui/react"
...
 import { BiPlus } from "react-icons/bi"
 
-export const AddSectionButton = ({ children, ...rest }: MenuButtonProps) => {
+interface AddSectionButtonProps extends MenuButtonProps {
+  sectionType?: string
+}
+
+export const AddSectionButton = ({
+  children,
+  sectionType = "section",
```
seaerchin, line 11: think zj made a change here to put the prop as `buttonText` - more inclined to just use `buttonText` cos less restrictive

src/layouts/components/NavBar/GroupMenuBody.tsx
```diff
@@ -0,0 +1,164 @@
...
+                      </Editable.DraggableAccordionItem>
+                    </VStack>
+                  ))}
+                </VStack>
+              </Editable.Accordion>
+            </Editable.EmptySection>
+          </Editable.Droppable>
+        </DragDropContext>
+      </FormControl>
+      <Box w="100%" pt="0.5rem">
```
seaerchin, line 140: nit: non-functionality related, not important to fix but was wondering why not just pass the props on `Box` to the button directly

src/layouts/EditNavBar.jsx
```diff
@@ -672,43 +633,215 @@ const EditNavBar = ({ match }) => {
...
+                                          ...errors.links[linkIndex],
+                                          sublinks: errors.sublinks[linkIndex],
+                                        }}
+                                      />
+                                    </Editable.DraggableAccordionItem>
+                                  )}
+                                  {!link.resource_room &&
+                                    !link.collection &&
+                                    !link.sublinks && (
+                                      <Editable.DraggableAccordionItem
```
seaerchin, line 738: could you point me to the corresponding section the legacy code that has this condition? just wanted to ensure we aren't causing a regression

seaerchin (commented):
looks good - only want to double check condition

seaerchin (approved):
looks good - only want to double check condition

src/layouts/EditNavBar.jsx
alexanderleegs: It's not exactly as it was before, the closest section to this is `NavElem` in `NavSection`

In the past, the different types of navigation elements actually were all the same type of component, with fields being conditionally rendered based on the presence/absence of params like `resource_room` and `collection`

Specifically, in the past, to render a url field the equivalent code is
```
{!collection && !isResource && (
  <FormContext hasError={!!linkErrors.url}>
    <FormTitle>Permalink</FormTitle>
    <FormField
      placeholder="Permalink"
      id={`link-${linkIndex}-url`}
      value={url}
      onChange={onFieldChange}
    />
    <FormError>{linkErrors.url}</FormError>
  </FormContext>
)}
```

But this was also because the `submenu` type of menu item also contains the `url` param. Since these items are all separate components now, in order to definitively identify this component as a `single page` type of menu, we can only check this via the absence of the distinguishing features of the other 3 types

src/layouts/components/Editable/AddSectionButton.tsx
alexanderleegs: Oh yeah that works better, changed in https://github.com/isomerpages/isomercms-frontend/pull/1457/commits/6c91fb77d62f110ead8d875b0bce976159cb64cc

src/layouts/components/NavBar/GroupMenuBody.tsx
alexanderleegs: Good point, fixed in https://github.com/isomerpages/isomercms-frontend/pull/1457/commits/40369daaa9268ba3779bf4604619b21a25a55100

src/layouts/EditNavBar.jsx
seaerchin: my main fear is that this might lead to issues down the line because the conditionals don't match exactly. things we'd need to check here are:
1. is `isResource` actually a `boolean` or a falsy value? because `!{}` evaluates to `false`, if `isResource` is a `boolean`, we might have a regression (if `resource_room` is `{}`)
2. there's an additional check for `sublinks` which wasn't present before - not sure if this will cause any issues.

alexanderleegs: 1. `isResource` corresponds to `link.resource_room` previously, and we're checking for `!link.resource_room` now! The actual check in our template code is assuming `resource_room` is a boolean value (in fact, the only value it can take is `true`)
2. `sublinks` is to catch the menu group type items - previously the field was rendered in both menu group and single page, so we didn't have a check for that.

I think what we can verify here is that

- Three of the sections have a unique identifying detail (`Group` - `sublinks`, `Resource room` - `resource_room`, `Folder` - `collection`), but `Group` is a superset of `Single page`, so it has no unique identifying feature - it thus becomes the fallback component
- These details are also what we're searching for in our template, so it should be okay

seaerchin: sgtm, merge whenever

PR #1461, opened 2023-08-30
Fix/homepage images on private repos

alexanderleegs:
## Problem

This PR fixes an issue where images on the homepage for private/GGS repos were not showing up as intended. The reason for this issue was the preview components using the outdated `v1` endpoint to fetch images - they have been modified to use the updated hook calling the `v2` endpoint instead, since the media url is handled by our backend in v2 instead of being offloaded to the frontend.

Resolves IS-527

## Test

- [ ] Go to edithomepage for any private repo (e.g. a-test-v4)
- [ ] Verify that images show up on 
  - [ ] Hero banner
  - [ ] Left infopic
  - [ ] Right infopic
- [ ] Verify that nested images work as well (e.g. `/images/album/nested.jpg`)
- [ ] When creating a new infopic section, no error toast should appear

### DIFF
--- a/src/templates/homepage/HeroSection.jsx
+++ b/src/templates/homepage/HeroSection.jsx
@@ -4,15 +4,13 @@
 /* eslint-disable jsx-a11y/anchor-is-valid */
 
 import PropTypes from "prop-types"
-import { forwardRef, useState, useEffect } from "react"
+import { forwardRef } from "react"
 
-import { useGetMediaHook } from "hooks/mediaHooks"
+import { useFetchPreviewMedia } from "hooks/useFetchPreviewMedia"
 
 import editorStyles from "styles/isomer-cms/pages/Editor.module.scss"
 
 import { getClassNames } from "templates/utils/stylingUtils"
-
-import { getImageDetails } from "utils/images"
 
 /* eslint
   react/no-array-index-key: 0
@@ -176,22 +174,10 @@ const KeyHighlights = ({ highlights }) => (
 )
 
 const TemplateHeroSection = (
-  { hero, siteName, dropdownIsActive, toggleDropdown },
+  { hero, dropdownIsActive, toggleDropdown },
   ref
 ) => {
-  const [loadedImageURL, setLoadedImageURL] = useState("")
-  const { fileName, imageDirectory } = getImageDetails(hero.background)
-  const { data: mediaData } = useGetMediaHook({
-    siteName,
-    mediaDirectoryName: imageDirectory || "images",
-    fileName,
-  })
-
-  useEffect(() => {
-    if (mediaData) {
-      setLoadedImageURL(mediaData.mediaUrl)
-    }
-  }, [mediaData])
+  const loadedImageURL = useFetchPreviewMedia(hero.background)
 
   const heroStyle = {
     // See j08691's answer at https://stackoverflow.com/questions/21388712/background-size-doesnt-work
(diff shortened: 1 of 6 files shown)


src/templates/homepage/HeroSection.jsx
```diff
@@ -178,14 +179,19 @@ const TemplateHeroSection = (
...
+    mediaDirectoryName: "images",
+    fileName,
+  })
+
+  useEffect(() => {
+    if (mediaData) {
+      setLoadedImageURL(mediaData.mediaUrl)
     }
-  )
+  }, [mediaData])
```
seaerchin, line 194: why can't we just rename `data` to `loadedImageURL` here?

src/utils/images.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Util method to retrieve image path from /images folder from the relative file path,
+ * e.g. "/images/album/picture.jpg" -> "album/picture.jpg"
+ */
+export const getImagePath = (imageLink: string) => {
+  const cleanImagePath = decodeURI(imageLink).replace(/^\//, "")
+  const filePathArr = cleanImagePath.split("/")
+  const filePath = filePathArr[filePathArr.length - 1]
+  return filePath
```
seaerchin, line 9: won't this return the last item in the array post split? ie, `"/images/album/picture.jpg" -> "picture.jpg"`
Screenshot 2023-08-30 at 12 20 44 PM

seaerchin (commented):
same pattern is repeatedly seen - inclined to move to a custom hook for this or, if `useGetMediaHook` is only used here, to update the hook.

if renaming `data` to `loadedImageURL` works, then that's ok too

src/templates/homepage/HeroSection.jsx
alexanderleegs: `data` consists of a lot of image related data, not just the url!

src/utils/images.ts
alexanderleegs: Good catch, i only checked this on first level images - have made a fix in https://github.com/isomerpages/isomercms-frontend/pull/1461/commits/62d436c34d3bc02653de03ff1ae39fa2a27bf7b6

src/utils/images.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Util method to retrieve image details from /images folder from the relative file path,
+ * e.g. "/images/album%/picture$.jpg" -> { imageDirectory: "images%2Falbum%24", fileName: "picture%24.jpg" }
+ */
+export const getImageDetails = (imageLink: string) => {
```
QiluXie, line 5: nit: we can add return types since this is a `ts` file
```suggestion
export const getImageDetails = (imageLink: string): { fileName: string; imageDirectory: string } => {
```

src/utils/images.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Util method to retrieve image details from /images folder from the relative file path,
+ * e.g. "/images/album%/picture$.jpg" -> { imageDirectory: "images%2Falbum%24", fileName: "picture%24.jpg" }
```
QiluXie, line 3: Minor mistake in the comment. `imageDirectory` should be `images%2Falbum%25`. Note `%` is encoded to `%25` instead of `%24`
```suggestion
 * e.g. "/images/album%/picture$.jpg" -> { imageDirectory: "images%2Falbum%25", fileName: "picture%24.jpg" }
```

alexanderleegs:
> same pattern is repeatedly seen - inclined to move to a custom hook for this or, if `useGetMediaHook` is only used here, to update the hook.
> 
> if renaming `data` to `loadedImageURL` works, then that's ok too

Shifted out in https://github.com/isomerpages/isomercms-frontend/pull/1461/commits/5cb437acfd02601fd8e030b1cb3871a413ba6f73

seaerchin (approved):

harishv7 (approved):

PR #1466, opened 2023-08-31
Chore/fix edit nav bar styles

alexanderleegs:
## Problem

This PR fixes some minor styling issues for edit nav bar.

Preview component - added additional spacing to make preview more accurate. Note that this is a slight deviation from the existing template - the reason for these changes is that our preview component doesn't actually comprise the entire nav bar template code, and the differences result in styling which is slightly different in our preview component. The actual changes are to give more padding in mobile view, and to center the nav component in desktop view.
Before:
(desktop preview)
Screenshot 2023-08-31 at 3 40 20 PM

(mobile preview)
Screenshot 2023-08-31 at 3 39 26 PM

After:
(desktop preview)
Screenshot 2023-08-31 at 3 33 45 PM
(mobile preview)
Screenshot 2023-08-31 at 3 34 04 PM

Actual component:
(desktop)
Screenshot 2023-08-31 at 3 34 53 PM
(mobile)
Screenshot 2023-08-31 at 3 35 17 PM

Legacy Edit Nav block component - added spacing to the add section buttons:
Before:
Screenshot 2023-08-31 at 3 41 13 PM

After:
Screenshot 2023-08-31 at 3 36 11 PM

### DIFF
--- a/src/components/navbar/NavSection.jsx
+++ b/src/components/navbar/NavSection.jsx
@@ -1,3 +1,4 @@
+import { Box } from "@chakra-ui/react"
 import { Button, IconButton } from "@opengovsg/design-system-react"
 import PropTypes from "prop-types"
 import { useState, useRef } from "react"
@@ -269,7 +270,7 @@ const NavSection = ({
           </div>
         )}
       </Droppable>
-      <div className="d-flex justify-content-between mt-4">
+      <Box d-flex justify-content-between mt="2rem">
         <Select
           ref={selectInputRef}
           className="w-50"
@@ -278,6 +279,7 @@ const NavSection = ({
           options={sectionCreationOptions}
         />
         <Button
+          mt="1rem"
           onClick={sectionCreationHandler}
           isDisabled={!newSectionType}
           variant="solid"
@@ -285,7 +287,7 @@ const NavSection = ({
         >
           Create New Menu
         </Button>
-      </div>
+      </Box>
       <span className={elementStyles.info}>
         {`Note: you can specify a folder ${
           hasResourceRoom ? `or resource room ` : ``
--- a/src/components/navbar/NavSection.jsx
+++ b/src/components/navbar/NavSection.jsx
@@ -270,7 +270,7 @@ const NavSection = ({
           </div>
         )}
       </Droppable>
-      <Box d-flex justify-content-between mt="2rem">
+      <Box mt="2rem" mb="1rem">
         <Select
           ref={selectInputRef}
           className="w-50"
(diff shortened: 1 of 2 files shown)


src/components/navbar/NavSection.jsx
```diff
@@ -269,7 +270,7 @@ const NavSection = ({
           </div>
         )}
       </Droppable>
-      <div className="d-flex justify-content-between mt-4">
+      <Box d-flex justify-content-between mt="2rem">
```
seaerchin, line 273: Don't you still need the `className` here?

src/templates/NavBar.jsx
```diff
@@ -15,9 +15,10 @@ const NavDropdownSection = ({ children, link }) => (
 
 const TemplateNavBar = ({ links, collectionInfo, resources }) => (
   <nav className="navbar is-transparent flex-fill">
-    <div className="bp-container">
+    <div className="bp-container" style={{ display: "flex" }}>
```
seaerchin, line 18: Hmm, does the flex itself suffice? Just noting also that we deviate from using `classNames` here, but I'm unsure of how important consistency is here.

alexanderleegs: Yeah I opted to deviate slightly away from the template here - the reason is that `flex` is provided by a parent component in the template which doesn't exist in the partial version we have in the preview, so I think this is the closest approximation to that

src/components/navbar/NavSection.jsx
alexanderleegs: This is for the sidebar which doesn't need to follow the existing template styles - so i think we can drop the `className` as long as the styling is fine

seaerchin: ok then pls drop the `d-flex...` as those are unused and leaving them in is confusing

src/templates/NavBar.jsx
seaerchin: what i meant was that we could instead define a css class that provides flex and give the classname as `bp-container flex`

src/components/navbar/NavSection.jsx
alexanderleegs: Removed in https://github.com/isomerpages/isomercms-frontend/pull/1466/commits/eb6e949f13de868115b0ad726338e1d6209601c9

src/templates/NavBar.jsx
alexanderleegs: Ah right, changed in https://github.com/isomerpages/isomercms-frontend/pull/1466/commits/5e95895ede0aaba087a61b171a48876b3b71b6f7

seaerchin (approved):

PR #1472, opened 2023-09-05
Chore/fix title text

alexanderleegs:
Changes the title to `IsomerCMS` instead of `Isomer CMS`.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -29,7 +29,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>Isomer CMS</title>
+    <title>IsomerCMS</title>
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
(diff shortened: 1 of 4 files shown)


public/index.html
```diff
@@ -29,7 +29,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>Isomer CMS</title>
```
harishv7, line 32: why do we not have the space? Isomer and CMS are 2 different words right?

alexanderleegs: Yup, but to be consistent with our comms - see [this thread](https://opengovproducts.slack.com/archives/C04S8QA6ES3/p1693875975226159)

dcshzj:
@alexanderleegs nit: the `AllSitesHeader.tsx` and `Header.jsx` files contains the `alt` attribute that should also have the space removed, so we can be consistent. Same as the `VerifyUserDetailsModal.jsx` file.

alexanderleegs:
> @alexanderleegs nit: the `AllSitesHeader.tsx` and `Header.jsx` files contains the `alt` attribute that should also have the space removed, so we can be consistent. Same as the `VerifyUserDetailsModal.jsx` file.


Good catch, fixed in https://github.com/isomerpages/isomercms-frontend/pull/1472/commits/0e14bca2ee63d5dd9289c963ccd9830ac9aa3566

dcshzj (approved):

PR #1473, opened 2023-09-05
feat(herobody): add first pass of layout

seaerchin:
## Problem
we need to add new layout options for hero banner

## Solution
- Add panels on the hero section first
- changed hero body so that it contains the new options now

## Screenshots
Recording 2023-09-08 at 14 39 45

## Tests
- [ ] go to homepage
- [ ] click on the hero section
- [ ] select the various new layouts
- [ ] the panel should reflect the new options
- [ ] the **preview should not work**

### DIFF
--- a/src/layouts/components/Homepage/HeroBody.tsx
+++ b/src/layouts/components/Homepage/HeroBody.tsx
@@ -152,7 +152,7 @@ type SectionAlignment = "left" | "right"
 
 type SectionBackgroundColor = "black" | "white" | "translucent gray"
 
-interface HeroSideSectionProps {
+interface HeroSideSectionProps extends HeroCenteredLayoutProps {
   background: string
   index: number
   errors: {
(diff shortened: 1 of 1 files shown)


seaerchin:
ps, missing 1 title/subtitle on side layuot - adding now

src/layouts/components/Homepage/HeroBody.tsx
```diff
@@ -93,50 +364,41 @@ export const HeroBody = ({
...
+              return <HeroImageOnlyLayout {...rest} />
+            }
+
+            if (currentSelectedOption === HERO_LAYOUTS.SIDE_SECTION.value) {
+              return (
+                <HeroSideSectionLayout
+                  {...rest}
+                  size="half"
+                  alignment="left"
+                  backgroundColor="black"
```
alexanderleegs, line 384: This resets the values every time the user selects a different option then comes back - are we okay with this behaviour?

src/layouts/components/Homepage/HeroBody.tsx
```diff
@@ -37,6 +42,276 @@ export interface HeroBodyFormFields {
...
+  background,
+  index,
+  errors,
+  title,
+  subtitle,
+  size = "half",
+  alignment = "left",
+  backgroundColor = "black",
+}: HeroSideSectionProps) => {
+  const [, setSectionSize] = useState(size)
```
harishv7, line 178: why do we need the set state without declaring the state?

src/layouts/components/Homepage/HeroBody.tsx
```diff
@@ -37,6 +42,276 @@ export interface HeroBodyFormFields {
...
+            <Radio value="right" size="xs" w="50%" allowDeselect={false}>
+              Right
+            </Radio>
+          </HStack>
+        </Radio.RadioGroup>
+      </Box>
+      <Box w="100%">
+        <Text textStyle="subhead-1">Section background colour</Text>
+        <HStack spacing="0.75rem" alignItems="flex-start">
+          <IconButton
```
harishv7, line 232: maybe can store as [{}] and .map

src/layouts/components/Homepage/HeroBody.tsx
seaerchin: this is gonna be fixed in an upcoming PR - we're gonna roll out the new stuff totally (behind a feature flag) so this intermediate state won't be seen

src/layouts/components/Homepage/HeroBody.tsx
seaerchin: ps, i only added the panel - this means that the preview stuff here isn't working (it's a stop gap).

the actual preview component for these are gonna be added in a follow up and htis will be removed

src/layouts/components/Homepage/HeroBody.tsx
seaerchin: ps, could you go into a bit more detail her?

seaerchin: i will put as a function cos it relies on the icon colour (black/gray/white)

alexanderleegs (approved):
lgtm

PR #1476, opened 2023-09-06
Fix/edit nav nits

alexanderleegs:
## Problem

Resolves Nav specific issues in https://www.notion.so/opengov/Restyling-feedback-3a4dc84523ff44529e0644ae85b4e0ca. 

Resolves IS-533


dcshzj (approved):
No issues with the styling changes, but @alexanderleegs can check for the delete buttons, I realised yours isn't 100% width, should it be the full width or just the current width (which is based on the text)?

PR #1477, opened 2023-09-06
fix(homepage): various styling fixes 

seaerchin:
## Problem
See the restyling notion page [here](https://www.notion.so/opengov/Restyling-feedback-3a4dc84523ff44529e0644ae85b4e0ca?pvs=4), specifically those that were tagged to me

## Solution
Mostly style fixes + copy changes, should be self explanatory


alexanderleegs (approved):
lgtm

PR #1479, opened 2023-09-06
fix(edithomepage): move error msg extraction into component

seaerchin:
## Problem
previously we try to extract error msg prior to data load

## Solution
shift error msg extraction into component

## Tests
- [ ] visit hoempage
- [ ] should load without issue


seaerchin:
not doing this anymore - see [here](https://opengovproducts.slack.com/archives/CK68JNFHR/p1694056094285909)

PR #1482, opened 2023-09-07
fix: styling

seaerchin:
## Problem
nested cards (yeet) merged ytd but only for contact-us 
buttons are full width on hover for contact-us 

## Solution
use `isNested` prop + fix `boxShadow` for nested cards
remove `w = 100%` for full width buttons
make infopic take full width for form field media

### DIFF
--- a/src/layouts/components/ContactUs/ContactCard.tsx
+++ b/src/layouts/components/ContactUs/ContactCard.tsx
@@ -232,7 +232,6 @@ export const ContactCard = ({
 
         <Button
           variant="clear"
-          w="100%"
           id={`contacts-${index}`}
           onClick={() => onDelete(`contacts-${index}`, "contact information")}
           alignSelf="center"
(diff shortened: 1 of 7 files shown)


src/layouts/components/ContactUs/ContactCard.tsx
```diff
@@ -232,7 +232,6 @@ export const ContactCard = ({
 
         <Button
           variant="clear"
-          w="100%"
```
harishv7, line 235: don't we need this for full-width button?

or is it that the button is *not* supposed to be full-width?

seaerchin: the button is **not** supposed to be full width, clarified with designer

alexanderleegs (approved):

harishv7 (approved):

dcshzj:
The `_hover` broke :( let me fix that separately

PR #1483, opened 2023-09-07
Fix/style nits

seaerchin:
## Problem
see [here](https://opengovproducts.slack.com/archives/CK68JNFHR/p1694064374781469)

Closes [insert issue #]

## Solution
1. add consistent spacing of 0.75
2. add padding on accordion when `isNested=true`


src/layouts/components/NavBar/GroupMenuBody.tsx
seaerchin, line 1: change made was spacing + remove extra `VStack`

alexanderleegs (approved):

PR #1485, opened 2023-09-08
feat(preview): add preview for new homepage variant

seaerchin:
## Problem
This adds a new homepage variant without the image.

## Solution
1. break down **Template**HeroSection -> use diff components, so that it's easier to use/maintain
2. convert template from `.js` to `.ts` -> `PropTypes` already done, so not much effort
3. **for template**, i copied over the styling from the template repo **as-is** and used `.css` to do this to conform to the style of handling in the template repository.

## Screenshots
Recording 2023-09-08 at 14 38 01


src/styles/isomer-template.scss
```diff
@@ -1,4 +1,6 @@
 @charset "UTF-8";
+@use 'templates/components/hero.scss';
```
dcshzj, line 2: I realised that changing this file meant that the CSS classes that are only intended for the editor preview gets leaked, which may cause unintended side effects on other aspects of the CMS. Can we put this inside `preview-panel.scss` instead?

src/templates/homepage/HeroSection/HeroImageOnlyLayout.tsx
```diff
@@ -0,0 +1,27 @@
...
+      >
+        <div
+          className={getClassNames(editorStyles, [
+            "row",
+            "is-vcentered",
+            "is-centered",
+            "ma",
+          ])}
+        >
+          <div className={getClassNames(editorStyles, ["min-height-mobile"])} />
```
dcshzj, line 22: Should this tag be closed as `<div></div>` instead of being a self-closing tag? `div` is not a [void element](https://html.spec.whatwg.org/multipage/syntax.html#void-elements), which may cause rendering issues.

dcshzj (changes requested):

src/styles/isomer-template.scss
seaerchin: actually not sure if that's the fix - maybe we should remove the extra import insetad? wdyt

src/templates/homepage/HeroSection/HeroImageOnlyLayout.tsx
seaerchin: on auto save, `<div> </div` auto formats to `<div />` ._. it seems like the two should be equivalent?

src/styles/isomer-template.scss
dcshzj: @alexanderleegs Do you have more context on why we import `isomer-template.scss` globally in `src/index.jsx`?

src/templates/homepage/HeroSection/HeroImageOnlyLayout.tsx
dcshzj: Ah okay, I think not super important.

src/styles/isomer-template.scss
alexanderleegs: As discussed offline - we're still using this for preview for regular pages, because that preview doesn't have an isolated set of styles. The compromise we made last time was that all styles imported from `isomer-template` would at leat be prefixed with a unique class, so we don't have leakage

dcshzj: Discussed offline, we should not add/modify this file and instead, should just add directly to `preview-panel.scss` if it is a homepage component.

seaerchin: :sadge:

seaerchin: fixed

src/templates/homepage/HeroSection/HeroSection.tsx
```diff
@@ -0,0 +1,158 @@
...
+        >
+          {variant === "center" ? (
+            <HeroCenteredLayout
+              hero={hero}
+              dropdownIsActive={dropdownIsActive}
+              toggleDropdown={toggleDropdown}
+            />
+          ) : (
+            <HeroImageOnlyLayout />
+          )}
```
alexanderleegs, line 149: Do we have a fallback for when variant doesn't exist?

seaerchin: i'll raise a PR at the end so that i can account for all the variants!

alexanderleegs (approved):
lgtm

dcshzj (approved):

PR #1491, opened 2023-09-13
feat(hero-section): adds dropdown

seaerchin:
## Problem
The existing hero section does not have a dropdown. This adds the preview component for the dropdown in.

## Solution
Copy paste over the dropdown from the template code (with modification so it can be consumed by react)

## Tests 
- [ ] Go to the hero section
- [ ] select the side variant
- [ ] select dropdown 
- [ ] the options there should be preserved, and the preview should be updated
- [ ] change the dropdown options/title
- [ ] the preview should also update accordingly
- [ ] try this in different view ports (technically we don't support)


alexanderleegs (approved):
lgtm

PR #1496, opened 2023-09-14
feat(homepage): add preview for announcements component

dcshzj:
## Problem

Closes IS-517.

## Solution

**Breaking Changes**

**Features**:

- The homepage editor now has the ability to preview a new announcements component.

## Before & After Screenshots

Screenshot 2023-09-14 at 3 41 04 PM

## Tests

- [ ] Smoke tests

No tests are currently available, as the announcements component can only be previewed once the left sidebar is updated to support editing of the new announcements component.

## Deploy Notes

*None*

### DIFF
--- a/src/styles/isomer-template/homepage/announcements.scss
+++ b/src/styles/isomer-template/homepage/announcements.scss
@@ -1,0 +1,37 @@
+.announcements-divider {
+  height: 1px;
+  color: #f9f9f9;
+}
+
+.announcements-announcement-title {
+  font-size: 1.625rem;
+  font-weight: 700;
+  line-height: 2rem;
+}
+
+.announcements-announcement-subtitle {
+  font-size: 1rem;
+  font-weight: 400;
+  line-height: 1.375rem;
+}
+
+.announcements-announcement-link {
+  text-transform: unset !important;
+  letter-spacing: 0.27px !important;
+}
+
+@media screen and (min-width: 1024px) {
+  .announcements-announcement-row-px-desktop {
+    padding-left: 0.75rem;
+    padding-right: 0.75rem;
+  }
+
+  .announcements-announcement-col-px-desktop {
+    padding-left: 1.5rem;
+    padding-right: 1.5rem;
+  }
+
+  .announcements-announcement-title-mr-desktop {
+    margin-right: 1.5rem;
+  }
+}
--- a/src/styles/isomer-template/homepage/announcements.scss
+++ b/src/styles/isomer-template/homepage/announcements.scss
@@ -16,22 +16,20 @@
 }
 
 .announcements-announcement-link {
-  text-transform: unset !important;
-  letter-spacing: 0.27px !important;
+  color: var(--site-secondary-color);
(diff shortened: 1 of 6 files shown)


src/types/homepage.ts
```diff
@@ -49,7 +49,7 @@ export interface InfopicSection {
   description?: string
   button?: string
   url?: string
-  images?: string
```
kishore03109, line 52: Is this intended?

src/types/homepage.ts
```diff
@@ -91,27 +104,26 @@ export type HomepageEditorHeroSection =
   | EditorHeroHighlightsSection
 
 export type HeroFrontmatterSection = { hero: HomepageEditorHeroSection }
-// TODO: add properties here instead of typing as `Record<string, unknown>`
-// we can find them in `HomepagePreview`
-export type ResourcesFrontmatterSection = { resources: Record<string, unknown> }
+export type ResourcesFrontmatterSection = { resources: ResourcesSection }
```
kishore03109, line 107: sanity check here, we are certain about the shape for this ya?

kishore03109: This + all the rest of the `infopic` + `infobar`, etc

src/styles/isomer-template/homepage/announcements.scss
```diff
@@ -0,0 +1,37 @@
...
+}
+
+.announcements-announcement-subtitle {
+  font-size: 1rem;
+  font-weight: 400;
+  line-height: 1.375rem;
+}
+
+.announcements-announcement-link {
+  text-transform: unset !important;
```
kishore03109, line 19: May I get more context to this? why do we need to `!important` here ah? I was under the impression this is what we would be actively moving away from, is this assumption wrong?

dcshzj: This will probably change! I originally did it this way to get access to the site-specific secondary colour + hover (could only access it using `.bp-button`, which had the `text-transform` set). Will adjust the template + this preview once the other changes are completed.

src/types/homepage.ts
dcshzj: Yup, the frontmatter for infopic uses `image` instead of `images`. The InfopicSection.jsx uses `image` too.

src/types/homepage.ts
dcshzj: Yup, it should follow the frontmatter format already used in the homepage, but for sanity check let me confirm with @seaerchin.

src/styles/isomer-template/homepage/announcements.scss
dcshzj: Adjusted in bb34fae!

src/types/homepage.ts
seaerchin: i _think_ so? but this was done quite awhile back tbh... if there are no build issues, it shuold be ok cos i remember i used `Record<string, unknown>` to avoid having to trace the properties (for velocity)

seaerchin (approved):
lgtm

PR #1497, opened 2023-09-14
Feat/announcement block

kishore03109:
## Problem

Announcement block editing in homepage according to [figma](https://www.figma.com/file/7OtNYkc4s9BkFk05ConEbd/%E2%AD%90-V1.5-Master?type=design&node-id=14719-162567&mode=design&t=vyPgS9YRM4DGGoo0-4). 

## Solution
In the interest of consistency, similar approach to key_highlights in the hero banner. 

**Breaking Changes**

  - Details ...

https://github.com/isomerpages/isomercms-frontend/assets/42832651/ecfc0d90-ea10-446f-9652-2c2b6090fb57

## Tests
- [ ] Go into edit homepage 
- [ ] Click on add new announcement
- [ ] Ensure that when new announcements are created, they are created from the top
- [ ] Ensure that save button is not available when errors are present
- [ ] Ensure that when there are errors in the announcement item there is a red outline in both
       1. The annoucemnet item
       2. The entire announcement block
- [ ] You can drag and drop announcements 
- [ ] When one announcement block is clicked, you cannot add more announcement blocks (only allowed <= 1 announcement block)
- [ ] You can only add up to 5 announcement items (only allowed <= 5 announcement items) 
- [ ] When save is clicked + page is refreshed, the announcement persists -> suggests that the changes were sucessfully published to gh. 

## Known limitations 
- Announcement does not open up on create. (I attempted to use `displayAnnouncements` for this but somehow it is not working as intended. Filing a ticket here.)
- Implementing unique keys in `AnnouncementBody.tsx`. Since we are rendering components over a map, each element would require a unique key. However, when I did do it the re-renders were visible + triggered an off-focus 😱. Have anyone solved this problem before? 
Screenshot 2023-09-15 at 2 53 03 PM. 
This key was removed in 1556850fde16a2ed911a90becfecd28caa624486

## Blockers
I think it is difficult for design to iterate give a green check here since we don't have storybook. I tried (a timeboxed) attempt it b5b41cccb6a4ae563d0098804953a4b0186f0af6 but the paddings and all were off so this attempt does not yield the benefits that I was hoping for. Since this to be shipped soon, for now Ill document the flows in video form to get Design's [@sehyunidaaa] approval first. 

## Note to reviewers
This PR is too bulky for my taste, I am seeking feedback on how I could have split this PR up (or if this PR size is actually ok).


src/utils/validators.js
```diff
@@ -803,13 +915,6 @@ const validateDayOfMonth = (month, day) => {
   }
 }
 
-const validateNonFutureDate = (dateStr) => {
```
kishore03109, line 806: Note: this is not removed, but shifted up + renamed to group similar functions tgt

src/layouts/components/Homepage/AnnouncementBody.tsx
```diff
@@ -0,0 +1,224 @@
...
+    onChange,
+    onCreate,
+    onDelete,
+    onDisplay,
+  } = useEditableContext()
+  return (
+    <Box w="full">
+      <DragDropContext onDragEnd={onDragEnd}>
+        <Editable.Droppable editableId="announcement">
+          <VStack spacing="1.25rem" align="flex-start" p={0} />
```
dcshzj, line 44: Hmm was this self-closing `<VStack>` intended?

src/layouts/components/Homepage/AnnouncementBody.tsx
```diff
@@ -0,0 +1,224 @@
...
+          <Text mt="1.5rem" textStyle="h6">
+            Announcements
+          </Text>
+          <Text mt="0.5rem" textStyle="body-2" textColor="base.content.medium">
+            You can display up to 5 announcements at a time. Newly added
+            announcements are shown on the top of the list
```
dcshzj, line 50: Can we use `MAX_ANNOUNCEMENTS` instead of "5" here?

src/layouts/components/Homepage/AnnouncementBody.tsx
```diff
@@ -0,0 +1,224 @@
...
+      </DragDropContext>
+      <Button
+        id={`announcement-${announcementItems.length}-create`}
+        onClick={() => onCreate({ target: { id: "announcement" } })}
+        variant="outline"
+        w="full"
+        leftIcon={<BiPlus fontSize="1.5rem" />}
+        isDisabled={announcementItems.length >= MAX_ANNOUNCEMENTS}
+      >
+        Add Announcement
```
dcshzj, line 220: Might need to check with design on this text copy ("Add announcement" or "Add new announcement").

src/layouts/components/Homepage/AnnouncementBody.tsx
```diff
@@ -0,0 +1,224 @@
...
+              <Editable.Section px={0} spacing="0.75rem" py="1.5rem">
+                {announcementItems.map(
+                  (
+                    {
+                      title: announcementTitle,
+                      date: announcementDate,
+                      announcement: announcementContent,
+                      link_text: announcementLinkText,
+                      link_url: announcementLinkUrl,
+                    },
```
dcshzj, line 68: Not a functional comment but just wondering if it would be better to just have `announcement` object here, then call things like `announcement.title` below.

src/layouts/components/Homepage/AnnouncementBody.tsx
```diff
@@ -0,0 +1,224 @@
...
+                          </FormControl>
+                          <FormControl
+                            isInvalid={
+                              !!errors.announcementItems[announcementIndex].date
+                            }
+                            isRequired
+                          >
+                            <FormLabel>Date</FormLabel>
+                            <DatePicker
+                              placeholder="announcement description"
```
dcshzj, line 111: Might not be needed? The placeholder texts might need to adjust.

dcshzj (commented):
Not a complete review yet, left EditHomepage.jsx but have some concerns regarding frontmatter

src/types/homepage.ts
```diff
@@ -146,6 +154,10 @@ export const EditorHomepageFrontmatterSection = {
     section: EditorHomepageFrontmatterSection
   ): section is AnnouncementsFrontmatterSection =>
     !!(section as AnnouncementsFrontmatterSection).announcements,
+  isAnnouncement: (
+    section: PossibleEditorSections
+  ): section is AnnouncementOption =>
+    !!(section as AnnouncementOption).announcement,
```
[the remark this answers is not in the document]
kishore03109: ill remove

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -139,6 +157,7 @@ const EditHomepage = ({ match }) => {
     image: "",
     notification: "",
     sections: [],
+    announcements: {},
```
dcshzj, line 160: Should `announcements` be inside `sections` instead? It is on the same level as `infopic` and `infobar`.

Actually I quite confused, the `onFieldChange` part puts the announcements inside `sections`, which is the correct behaviour. Perhaps this part is not needed?

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -267,6 +292,32 @@ const EditHomepage = ({ match }) => {
...
+          if (section.announcements) {
+            sectionsErrors.push({
+              announcements: getErrorValues(ANNOUNCEMENT_BLOCK),
+            })
+            announcementItemErrors = _.map(
+              section.announcements.announcement_items,
+              () => getErrorValues(ANNOUNCEMENT_SECTION)
+            )
+            if (!section.announcements.announcement_items) {
+              // define an empty array to announcement_items to prevent error
```
dcshzj, line 304: Just checking, will we allow for having the announcements block without any announcement item?

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -1109,6 +1315,19 @@ const EditHomepage = ({ match }) => {
                         subtitle={INFOBAR_SECTION.subtitle}
                         onClick={() => onClick(INFOBAR_SECTION.id)}
                       />
+                      {/* NOTE: Check if the sections contain any `announcement` 
+                                and if it does, prevent creation of another `resources` section
```
dcshzj, line 1319: nit: `resources` -> `announcements`

src/utils/validators.js
```diff
@@ -171,6 +178,20 @@ export const MEDIA_SETTINGS_TITLE_MIN_LENGTH = 6
...
+  const chosenDate = new Date(dateStr)
+  const daysDiff = today - chosenDate
+  return daysDiff >= 0
+}
+
+const validateNonFutureAnnouncementDate = (dateStr) => {
+  const today = new Date(moment().tz("Asia/Singapore").format("YYYY-MM-DD"))
+  const IsoDateStr = dateStr.split("/").reverse().join("-")
+  const chosenDate = new Date(IsoDateStr)
+  return today > chosenDate
```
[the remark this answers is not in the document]
kishore03109: Oh semantically they are the same! But for consistency sake, I have modified this to be the same! 
To be clear, if the today's date is the default date, it should be valid (part of the testing plan + video should have this functionality!)

src/utils/validators.js
```diff
@@ -218,6 +239,57 @@ const validateHighlights = (highlightError, field, value) => {
   return newHighlightError
 }
 
+const validateAnnouncementItems = (announcementError, field, value) => {
+  const newAnnouncementError = announcementError
+  let errorMessage = ""
+  switch (field) {
+    case "title": {
+      // Title is too short
+      if (value.length < ANNOUNCEMENT_TITLE_MIN_LENGTH) {
```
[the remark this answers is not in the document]
kishore03109: good catch df7b377d50ae034d5c7ac0de8d557ae9959ec107

src/utils/validators.js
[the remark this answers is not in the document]
kishore03109: done here df7b377d50ae034d5c7ac0de8d557ae9959ec107

src/utils/validators.js
```diff
@@ -112,10 +114,16 @@ const HERO_TITLE_MIN_LENGTH = 0
...
-// const HERO_BUTTON_TEXT_MAX_LENGTH = 30;
 const HERO_DROPDOWN_MIN_LENGTH = 0
 const HERO_DROPDOWN_MAX_LENGTH = 30
+// Announcement
+const ANNOUNCEMENT_BLOCK_SUBTITLE_MAX_LENGTH = 30
+const ANNOUNCEMENT_BLOCK_TITLE_MAX_LENGTH = 50
+const ANNOUNCEMENT_TITLE_MIN_LENGTH = 0
+const ANNOUNCEMENT_TITLE_MAX_LENGTH = 100
+const ANNOUNCEMENT_DESCRIPTION_MIN_LENGTH = 0
+const ANNOUNCEMENT_DESCRIPTION_MAX_LENGTH = 100
```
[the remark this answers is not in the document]
kishore03109: xiexie df7b377d50ae034d5c7ac0de8d557ae9959ec107

src/layouts/components/Homepage/AnnouncementBody.tsx
kishore03109: nope, thanks for the catch 2c733f2cb227aed0b72daf0043228d64a01454cd

src/layouts/components/Homepage/AnnouncementBody.tsx
kishore03109: good idea 2c733f2cb227aed0b72daf0043228d64a01454cd

src/layouts/components/Homepage/AnnouncementBody.tsx
kishore03109: hey i rather destructure at the top bah. 
i didnt want to do announcement.announcement for clarity

src/types/homepage.ts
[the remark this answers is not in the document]
kishore03109: done! 82b389584a01a2eb37412da41eb416e7e00f11a0

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -344,6 +396,8 @@ const EditHomepage = ({ match }) => {
               },
             },
           })
+          console.log({ newSections })
+          console.log({ sections })
```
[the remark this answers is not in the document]
kishore03109: f5ad91ade834432e20a36698f1a5c1fd30133e6f

src/layouts/components/Homepage/AnnouncementBody.tsx
kishore03109: Checked with @sehyunidaaa, it is indeed `Add announcement`

src/layouts/EditHomepage/EditHomepage.jsx
kishore03109: yes this is the case

src/layouts/EditHomepage/EditHomepage.jsx
kishore03109: d035a847adeb7e59df19dd11c70882ceba8627a9

src/layouts/EditHomepage/EditHomepage.jsx
kishore03109: hey good catch, you are right this is not needed, removed it here 342581a2cfa4620a056b0334bf5187c31f9b198b

src/hooks/useDrag.tsx
```diff
@@ -241,6 +279,61 @@ const updateHomepageState = (
         newHighlightErrors
       )
     }
+    case "announcement": {
+      const announcementKeyExist = !_.isEmpty(
```
seaerchin, line 283: nit: `doesAnnouncementKeyExist`

src/hooks/useDrag.tsx
```diff
@@ -406,6 +545,45 @@ export const onDelete = (
...
+      )
+      const newAnnouncementErrors = deleteElement(
+        errors.announcementItems,
+        indexToDelete
+      )
+
+      const newDisplayAnnouncements = deleteElement(
+        displayAnnouncementItems,
+        indexToDelete
+      )
```
seaerchin, line 577: nit: some amount of repetition here as it's very similar to the above. not a big issue for now as intent is clear but we might benefit it this model extends also to other section types

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -267,6 +291,32 @@ const EditHomepage = ({ match }) => {
...
+                    announcements: {
+                      announcement_items: {
+                        $set: [],
+                      },
+                    },
+                  },
+                },
+              })
+            }
+          }
```
seaerchin, line 318: seems like we're following the style of the existing code here rgiht?

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -441,6 +492,68 @@ const EditHomepage = ({ match }) => {
           scrollTo(scrollRefs[0])
           break
         }
+        case "announcement": {
+          // The field that changed belongs to a announcements
```
seaerchin, line 496:
```suggestion
          // The field that changed belongs to announcements
```

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -843,6 +984,20 @@ const EditHomepage = ({ match }) => {
         return newSection
       })
 
+      // Modify announcements date time object
+      if (filteredFrontMatter.announcements) {
+        filteredFrontMatter.announcements.announcement_items = filteredFrontMatter.announcements.announcement_items.map(
+          (announcement) => {
+            return announcement.$set(
```
seaerchin, line 991: is this actually intended? `$set` doesn't exist as a primitive. do also note that this will only ever appear on our backend + templates -> will there be a visual drift? (displayed dt format on template is different from preview)

if so, maybe this modification should take place inside `onFieldChange` to preserve parity

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -1082,6 +1237,54 @@ const EditHomepage = ({ match }) => {
                                       />
                                     </Editable.DraggableAccordionItem>
                                   )}
+
+                                  {section.announcements &&
+                                    errors.sections[sectionIndex]
+                                      .announcements && (
```
seaerchin, line 1243: is this intended? this seems to only render if there's an error at that specific section

src/layouts/EditHomepage/constants.ts
```diff
@@ -3,6 +3,27 @@ export const RESOURCES_SECTION = {
   subtitle: "Add a preview and link to your Resource Room",
   id: "resources",
 } as const
+export const LocalDateTimeNow = new Date()
```
seaerchin, line 6: nit: either lowercase this or screaming snake (more inclined to  prev) 

**potential bug**: this seems to only construct the date **once** - ie, if i stay on this page and call this `const` again, it might not update. perhaps this should be a function instead?

src/layouts/components/Homepage/AnnouncementBody.tsx
```diff
@@ -0,0 +1,222 @@
...
+                              !!errors.announcementItems[announcementIndex].date
+                            }
+                            isRequired
+                          >
+                            <FormLabel>Date</FormLabel>
+                            <DatePicker
+                              id={`announcements-${announcementIndex}-date`}
+                              inputValue={announcementDate}
+                              onInputValueChange={(value) => {
+                                onChange({
```
seaerchin, line 113: feels like i shuold've wrapped the `onChange` ._.

src/layouts/components/Homepage/AnnouncementBody.tsx
seaerchin, line 1: skimmed - seems to be ok

src/templates/homepage/AnnouncementsSection.tsx
```diff
@@ -112,7 +113,9 @@ export const TemplateAnnouncementsSection = forwardRef<
                                   ]
                                 }
                               >
-                                {announcement.date}
+                                {moment(announcement.date, "DD/MM/YYYY").format(
+                                  "DD MMMM YYYY"
+                                )}
```
seaerchin, line 118: noted that this will reconcile the diff in view but it still feels weird to have it here (as a view level thing) rather than just doing it at the data level. this offloads responsibility to the dev next time to know this @__@

src/utils/validators.js
```diff
@@ -171,6 +178,21 @@ export const MEDIA_SETTINGS_TITLE_MIN_LENGTH = 6
...
+const validateNonFutureResourceDate = (dateStr) => {
+  const today = new Date(moment().tz("Asia/Singapore").format("YYYY-MM-DD"))
+  const chosenDate = new Date(dateStr)
+  const daysDiff = today - chosenDate
+  return daysDiff >= 0
+}
+
+const validateNonFutureAnnouncementDate = (dateStr) => {
+  const today = new Date(moment().tz("Asia/Singapore").format("YYYY-MM-DD"))
+  const IsoDateStr = dateStr.split("/").reverse().join("-")
```
seaerchin, line 190:
```suggestion
  const isoDateStr = dateStr.split("/").reverse().join("-")
```

src/utils/validators.js
```diff
@@ -218,6 +240,57 @@ const validateHighlights = (highlightError, field, value) => {
   return newHighlightError
 }
 
+const validateAnnouncementItems = (announcementError, field, value) => {
```
seaerchin, line 243: could we separate out the validation from the set? this makes it easier to reuse in the future, which was 1 blocker i faced when trying to use the existing validators

src/hooks/useDrag.tsx
kishore03109: ae0d2a82bac8433be1a4795c4dfd989ec505fa11

src/layouts/EditHomepage/EditHomepage.jsx
kishore03109: yup, this is correct

src/layouts/EditHomepage/EditHomepage.jsx
kishore03109: differing meaning, modified to this instead for clarity ce2460f28f5f85fe4c380b5d1c15b16b184313b0

src/layouts/EditHomepage/constants.ts
kishore03109: you are right, c1bb6644bb42dc6e734bf522ec550a434e4f1750

src/utils/validators.js
kishore03109: 8b5b38b2436bfee4b9b02441dc2a9e75cbccd66a

src/utils/validators.js
kishore03109: validation from the set? May I get your help to expand more on this/direct me to the part of the codebase that uses the pattern you are trying to explain?

src/layouts/EditHomepage/EditHomepage.jsx
kishore03109: Good catch, this was due to an error that I was facing where the `errors.sections[sectionIndex].announcements` was set to `undefined` for 2 renders. I wasnt able to pinpoint the exact reason here, but had to timebox the effort here + fix it temp but wrapping it around a check.

src/templates/homepage/AnnouncementsSection.tsx
kishore03109: 352dda83e9ad59e741de145591ada8630fdc7fa1

src/layouts/EditHomepage/EditHomepage.jsx
kishore03109: Hey, you make sense, mutating it only at the component level as a neater solution here, thanks for this! 352dda83e9ad59e741de145591ada8630fdc7fa1

src/hooks/useDrag.tsx
kishore03109: deprio for now ya

src/utils/validators.js
seaerchin: abit unclear wording on my end, but `validateAnnouncementItems` sets the corresponding index. in this case, it does **not** simply validate (ie, return bool + msg) but also does a side effect (setting the corresponding index). this isn't particularly important to fix _now_ because it follows the existing style of code but when we want to do simple validation that is unaware of the _structure_ of the error, will be an issue

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -205,17 +230,18 @@ const EditHomepage = ({ match }) => {
       }
 
       try {
-        const {
+        let {
           content: { frontMatter },
-          sha,
         } = homepageData
+        const { sha } = homepageData
```
seaerchin, line 236: nit: this change appears to have different semantics; `content` now appears to be writable now whereas previously it was only readable

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -1082,6 +1226,55 @@ const EditHomepage = ({ match }) => {
...
+                                          ) ||
+                                          getHasError(errors.announcementItems)
+                                        }
+                                      >
+                                        <AnnouncementSection
+                                          {...section.announcements}
+                                          title={section.announcements.title}
+                                          subtitle={
+                                            section.announcements.subtitle
+                                          }
```
seaerchin, line 1258: nit: seems like a redundant spread

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -1082,6 +1226,55 @@ const EditHomepage = ({ match }) => {
                                       />
                                     </Editable.DraggableAccordionItem>
                                   )}
+
+                                  {showNewLayouts &&
+                                    section.announcements &&
+                                    errors.sections[sectionIndex]
+                                      .announcements && (
```
seaerchin, line 1233: (can't see prev comment) - this was the hotfix to an issue you found previously right? could you just add a comment here explaining the issue + fix so that we have clarity going forward

src/layouts/EditHomepage/constants.ts
```diff
@@ -1,9 +1,44 @@
+import moment from "moment"
+
 export const RESOURCES_SECTION = {
   title: "Resources",
   subtitle: "Add a preview and link to your Resource Room",
   id: "resources",
 } as const
 
+export type AnnouncementSectionType = {
```
seaerchin, line 9: clarifying your intent here - the type here is nearly a constant, with the **only** possible change being `date`. this means that, for example, `title` cannot be a `string` - it has to be the _constant_ string "New Announcement". if this is intended then mark this as resolved, otherwise we might benefit from giving a more relaxed type (eg: `string`)

src/layouts/components/Homepage/AnnouncementBody.tsx
```diff
@@ -0,0 +1,240 @@
...
+  return formattedDate
+}
+
+const toCmsPanelDateFormat = (date?: string) => {
+  const formattedDate = moment(date, "DD MMMM YYYY", true).format("DD/MM/YYYY")
+  if (formattedDate === "Invalid date") {
+    return date
+  }
+  return formattedDate
+}
```
seaerchin, line 44: nit: quite hard to tell apart at first glance, can just add a comment on either one stating the difference so ppl don't edit/delete accidentally

src/types/homepage.ts
```diff
@@ -20,6 +20,16 @@ export type HighlightOption = {
...
 
+export type AnnouncementOption = {
+  title: string
+  date: string
+  announcement: string
+  link_text: string
+  link_url: string
+}
+
+export type AnnouncementError = AnnouncementOption
```
seaerchin, line 31: not important _now_ but hte lack of a discriminant property (eg: `kind: 'error' | 'option'`) means that checking if this is actualy error or not is upheld by devs rather than the type system.

src/layouts/components/Homepage/InfobarBody.tsx
```diff
@@ -89,7 +89,7 @@ export const InfobarBody = ({
       </FormControl>
       <Button
         id={`section-${index}`}
-        onClick={() => onDelete(`section-${index}`, "Infobar Section")}
```
seaerchin, line 92: can i get context on why did we delete this?

seaerchin (commented):
mostly sanity checks

src/layouts/components/Homepage/AnnouncementBody.tsx
```diff
@@ -0,0 +1,240 @@
...
+                            isRequired
+                          >
+                            <FormLabel>Date</FormLabel>
+                            <DatePicker
+                              id={`announcements-${announcementIndex}-date`}
+                              inputValue={toCmsPanelDateFormat(
+                                announcementDate
+                              )}
+                              onInputValueChange={(value) => {
+                                console.log({ value })
```
seaerchin, line 130: could we also remove this?

src/layouts/components/Homepage/InfobarBody.tsx
kishore03109: The modal already renders section. 
Rn in prod we repeat section twice 
Screenshot 2023-09-20 at 7 12 28 PM

src/layouts/components/Homepage/AnnouncementBody.tsx
kishore03109: c9816a6d18975d38286812cac138cd43c341f91b

src/layouts/EditHomepage/constants.ts
kishore03109: hey yup you are right, relaxed [here](70c135bee8c2cf96af40ded17b92e980c3eab16d)

src/layouts/components/Homepage/AnnouncementBody.tsx
kishore03109: 4883ebf13a13fc7a253afb1dd7634a9649f2c32c

src/layouts/components/Homepage/InfobarBody.tsx
seaerchin: good catch!

src/layouts/EditHomepage/EditHomepage.jsx
kishore03109: Hey yes this is the case. There was a bug when the key was undefined, but its still valid frontmatter state. As such, I just refine an empty array so that our FE does not throw any error.

src/layouts/EditHomepage/EditHomepage.jsx
kishore03109: bc5d4392058ab27614786f80b8713db69293800a

src/layouts/EditHomepage/EditHomepage.jsx
kishore03109: ef614d16eb6a7c4eed4d74329c310e5f4a00cb69

src/types/homepage.ts
kishore03109: ok, to prior if we touch this part in the future + find it painful ya

src/utils/validators.js
kishore03109: hmm ok, am inclined to keep it as is for consistency with the rest of the local code, will modify if future pain points warrant it?

seaerchin (approved):
lgtm (i think); given the scope of the PR, maybe a better structure would've been to:

1. PR for changes in logical model (`useDrag` + related files) 
2. add in components for views 
3. add in additional functionality like feature flag etc

kishore03109:
@seaerchin ps need to your look at small additional commit (the others shd be trivial)  
[8c46a2c](https://github.com/isomerpages/isomercms-frontend/pull/1497/commits/8c46a2cbdefc5e6db696da515595321f7e95e6f4)

seaerchin (approved):

PR #1503, opened 2023-09-19
fix(styles): introduce site colors as CSS variables

dcshzj:
## Problem

The site colours are not available in SCSS as they are only available at runtime. This meant that we cannot make use of CSS classes that needs to change based on the site colours defined by the user.

## Solution

**Breaking Changes**

**Improvements**:

- This change adds the primary and secondary colours that can be set by the user for the site as CSS variables.
- The secondary colour hover has also been implemented the same way as the template, which is to lighten by 10%.

## Tests

- [x] Smoke tests
    - [ ] Visit any site's homepage
    - [ ] See that the `--site-primary-color`, `--site-secondary-color` and `--site-secondary-color-hover` variables are defined in the root element

## Deploy Notes

**New dependencies**:

- `tinycolor2` : Library for performing SCSS's lighten on a provided HEX code

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -71,6 +71,7 @@
         "sass": "^1.51.0",
         "sgds-govtech": "^1.3.16",
         "slugify": "^1.6.5",
+        "tinycolor2": "^1.6.0",
         "toml": "^3.0.0",
         "turndown": "^7.1.1",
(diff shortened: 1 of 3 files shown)


seaerchin (approved):
i'm approving this to unblock but we should really relook at how we maintain and handle injection of run time variables :(

PR #1505, opened 2023-09-20
Feat/text cards panel

alexanderleegs:
This PR introduces the text cards panel to our homepage editor. Resolves IS-454.

https://github.com/isomerpages/isomercms-frontend/assets/22111124/2a1859ba-9350-4b5f-904a-663e8a504bd3

### Tests

- [ ] Go into edit homepage
- [ ] Click on add new section -> text cards
- [ ] New text card section should be editable
- [ ] Errors should cause section to have error bar when minimized
- [ ] Should be able to add a new card to this section
- [ ] Should have a maximum of 4 cards in this section, after which add card button becomes disabled
- [ ] Cards should be draggable
- [ ] Errors in card section show up when card is minimized, and also when section is minimized
- [ ] Errors should block saving
- [ ] On save, changes should persist after refresh

### Known limitations

- On initial creation of a new text card section, no cards are created by default - desired behaviour is for 3 to be created.

### DIFF
--- a/src/components/Editable/Editable.tsx
+++ b/src/components/Editable/Editable.tsx
@@ -167,6 +167,7 @@ type HomepageDroppableZone =
   | "leftPane"
   | "highlight"
   | "announcement"
+  | `textcardcard-${number}`
 type ContactUsDroppableZone =
   | "locations"
   | "contacts"
--- a/src/components/Editable/Editable.tsx
+++ b/src/components/Editable/Editable.tsx
@@ -167,7 +167,7 @@ type HomepageDroppableZone =
   | "leftPane"
   | "highlight"
   | "announcement"
-  | `textcardcard-${number}`
+  | `textCardItem-${number}`
 type ContactUsDroppableZone =
   | "locations"
   | "contacts"
--- a/src/hooks/useDrag.tsx
+++ b/src/hooks/useDrag.tsx
@@ -9,12 +9,17 @@ import {
   EditorHeroHighlightsSection,
   EditorHomepageElement,
   EditorHomepageState,
+  EditorTextcardCardsSection,
+  EditorTextcardSection,
   HeroFrontmatterSection,
   PossibleEditorSections,
   EditorHomepageFrontmatterSection,
   AnnouncementsFrontmatterSection,
   AnnouncementOption,
+  TextcardFrontmatterSection,
 } from "types/homepage"
+
+const RADIX_PARSE_INT = 10
 
 const updatePositions = <T,>(
   section: T[],
@@ -44,11 +49,24 @@ const deleteElement = <T,>(section: T[], indexToDelete: number): T[] => {
   })
 }
 
+const updateElement = <T,>(
+  section: T[],
+  elem: T,
+  indexToUpdate: number
+): T[] => {
+  return update(section, {
+    [indexToUpdate]: {
+      $set: elem,
+    },
+  })
+}
+
 const updateEditorSection = (
   homepageState: EditorHomepageState,
   newDisplaySections: unknown[],
   newFrontMatterSection: EditorHomepageState["frontMatter"]["sections"],
-  newSectionErrors: unknown[]
+  newSectionErrors: unknown[],
+  newTextcardErrors: unknown[][]
 ): EditorHomepageState => ({
   ...homepageState,
   displaySections: newDisplaySections,
@@ -56,7 +74,11 @@ const updateEditorSection = (
     ...homepageState.frontMatter,
     sections: newFrontMatterSection,
   },
-  errors: { ...homepageState.errors, sections: newSectionErrors },
+  errors: {
+    ...homepageState.errors,
+    sections: newSectionErrors,
+    textcards: newTextcardErrors,
+  },
 })
 
 const updateAnnouncementSection = (
@@ -125,6 +147,37 @@ const updateHighlightsSection = (
   errors: { ...homepageState.errors, highlights: newHighlightErrors },
 })
 
+const updateTextCardsCardSection = (
+  homepageState: EditorHomepageState,
+  sectionIndex: number,
+  newTextCards: unknown[],
+  newTextCardErrors: unknown[]
+): EditorHomepageState => ({
+  ...homepageState,
+  frontMatter: {
+    ...homepageState.frontMatter,
+    sections: _.set(
+      // NOTE: Deep clone here to avoid mutation
+      _.cloneDeep(homepageState.frontMatter.sections),
+      [sectionIndex, "textcards", "cards"],
+      newTextCards
+    ),
+  },
+  errors: {
+    ...homepageState.errors,
+    textcards: _.set(
+      // NOTE: Deep clone here to avoid mutation
+      _.cloneDeep(homepageState.errors.textcards),
+      [sectionIndex],
+      newTextCardErrors
+    ),
+    // {
+    //   ...homepageState.errors.textcards,
+    //   [sectionIndex]: newTextCardErrors,
+    // },
+  },
+})
+
 type OnDragEndResponseWrapper = (
   state: EditorHomepageState
 ) => (result: DropResult) => EditorHomepageState
@@ -164,7 +217,8 @@ const updateHomepageState = (
   )
     return homepageState
 
-  switch (type) {
+  const typeArray = type.split("-")
+  switch (typeArray[0]) {
     case "editor": {
       const draggedElem = frontMatter.sections[source.index]
       const newSections = updatePositions(
@@ -181,6 +235,14 @@ const updateHomepageState = (
         draggedError
       )
 
+      const draggedTextcardError = errors.textcards[source.index]
+      const newCardErrors = updatePositions(
+        errors.textcards,
+        source.index,
+        destination.index,
+        draggedTextcardError
+      )
+
       const displayBool = displaySections[source.index]
       const newDisplaySections = updatePositions(
         displaySections,
@@ -193,7 +255,8 @@ const updateHomepageState = (
         homepageState,
         newDisplaySections,
         newSections,
-        newSectionErrors
+        newSectionErrors,
+        newCardErrors
       )
     }
     // inner dnd for hero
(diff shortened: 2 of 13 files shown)


src/hooks/useDrag.tsx
```diff
@@ -170,7 +223,8 @@ const updateHomepageState = (
   )
     return homepageState
 
-  switch (type) {
+  const typeArray = type.split("-")
+  switch (typeArray[0]) {
```
seaerchin, line 227: sadly this seems really finicky and prone to error. it appears that we only need the initial element so maybe we should just omit the `index` from `textcard-{index}`?

src/hooks/useDrag.tsx
```diff
@@ -478,7 +595,8 @@ export const onCreate = <E,>(
 export const onDelete = (
   homepageState: EditorHomepageState,
   elemType: EditorHomepageElement,
-  indexToDelete: number
+  indexToDelete: number,
+  subindexToDelete = 0
```
seaerchin, line 599: i might be missing something here - could you point me to why we are adding a new argument here? i know it's used down below but wouldn't it be possible to pass in `subelemType` + `subIndex`?

src/hooks/useDrag.tsx
```diff
@@ -333,7 +396,35 @@ const updateHomepageState = (
         announcementsIndex
       )
     }
+    case "textcardcard": {
+      const parentId = parseInt(typeArray[1], RADIX_PARSE_INT)
+      const draggedElem = ((frontMatter.sections[
+        parentId
+      ] as TextcardFrontmatterSection).textcards as EditorTextcardSection)
+        .cards[source.index]
```
seaerchin, line 404: would it be possible to just cast once at the top so it's easier to follow?

seaerchin (commented):
we are adding quite abit of tech debt here but it's ok for now because it's in the same style as the existing legacy code so any future refactors would probably just require a one pass to set groundwork.

what concerns me here is that the logical model for dealing with homepage (`useDrag`) has been altered - we now require 2 indexes for doing dnd which means that we cannot simply "drag and drop" an array item; would this be possible to refactor so that the underlying `useDrag` still accepts **only a singular index**?

src/components/Editable/Editable.tsx
```diff
@@ -162,7 +162,11 @@ const EditableSidebar = ({
   )
 }
 
-type HomepageDroppableZone = "dropdownelem" | "leftPane" | "highlight"
+type HomepageDroppableZone =
+  | "dropdownelem"
+  | "leftPane"
+  | "highlight"
+  | `textcardcard-${number}`
```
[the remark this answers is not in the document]
alexanderleegs: Hmm was trying to reduce the confusion between `textcard` (the section) vs `textcardcard` (the card, which is an item in the textcards section) - happy to take suggestions which make this distinction clear!

src/hooks/useDrag.tsx
```diff
@@ -94,6 +116,37 @@ const updateHighlightsSection = (
...
+    textcards: _.set(
+      // NOTE: Deep clone here to avoid mutation
+      _.cloneDeep(homepageState.errors.textcards),
+      [sectionIndex],
+      newTextCardErrors
+    ),
+    // {
+    //   ...homepageState.errors.textcards,
+    //   [sectionIndex]: newTextCardErrors,
+    // },
```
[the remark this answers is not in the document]
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/1505/commits/48ec71600e9f74eecaf06fe8a37e0f4fe742dc7a

src/hooks/useDrag.tsx
```diff
@@ -39,19 +44,36 @@ const deleteElement = <T,>(section: T[], indexToDelete: number): T[] => {
...
+  elem: T,
+  indexToUpdate: number
+): T[] => {
+  return update(section, {
+    [indexToUpdate]: {
+      $set: elem,
+    },
+  })
+}
+
```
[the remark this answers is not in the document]
alexanderleegs: Whoops this is actually unnecessary, removed in https://github.com/isomerpages/isomercms-frontend/pull/1505/commits/ff1415fffe61efe39f90ee445fb11e9e2ca83712

src/hooks/useDrag.tsx
```diff
@@ -50,19 +55,36 @@ const deleteElement = <T,>(section: T[], indexToDelete: number): T[] => {
   })
 }
 
+const updateElement = <T,>(
```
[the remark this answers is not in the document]
alexanderleegs: Good catch, https://github.com/isomerpages/isomercms-frontend/pull/1505/commits/ff1415fffe61efe39f90ee445fb11e9e2ca83712

src/hooks/useDrag.tsx
alexanderleegs: As discussed offline - this issue is due to textcards not being unique, so they're the first section in homepage that requires both `index` and `subindex` to uniquely identify! Will be fixed in a followup PR as discussed by standardizing the type to `textcardcard-{index}` and swapping to if/else instead of switch/case

src/hooks/useDrag.tsx
alexanderleegs: See above!

src/hooks/useDrag.tsx
alexanderleegs: Good catch,  https://github.com/isomerpages/isomercms-frontend/pull/1505/commits/215a1ff08d571782ed4873b7fa16efd23343088a

src/hooks/useDrag.tsx
```diff
@@ -50,19 +55,36 @@ const deleteElement = <T,>(section: T[], indexToDelete: number): T[] => {
...
   displaySections: newDisplaySections,
   frontMatter: {
     ...homepageState.frontMatter,
     sections: newFrontMatterSection,
   },
-  errors: { ...homepageState.errors, sections: newSectionErrors },
+  errors: {
+    ...homepageState.errors,
+    sections: newSectionErrors,
+    textcards: newTextcardErrors,
```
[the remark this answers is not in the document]
alexanderleegs: I think Chin has the same concern! will be making a follow up PR which will hopefully reduce the complexity

src/components/Editable/Editable.tsx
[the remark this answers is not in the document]
seaerchin: why not `textCardSection` for the section and `textCardItem` for the card itself

src/hooks/useDrag.tsx
```diff
@@ -54,15 +59,20 @@ const updateEditorSection = (
...
+  },
 })
```
seaerchin, line 77: why do we need the extra argument? if `homepageState.errors.textcards` is already updated then the spread of `...homepageState.errors` should be sufficient, no?

src/hooks/useDrag.tsx
```diff
@@ -494,12 +600,14 @@ export const onDelete = (
...
+      const newTextcardErrors = deleteElement(errors.textcards, indexToDelete)
 
       return updateEditorSection(
         homepageState,
         newDisplaySections,
         sections,
-        newErrorSections
+        newErrorSections,
+        newTextcardErrors
       )
```
seaerchin, line 611: if my understanding here is correct, `homepageState.errors.textcards[i]` would already have been deleted, no? so we don't need to pass in this argument

src/hooks/useDrag.tsx
```diff
@@ -494,12 +600,14 @@ export const onDelete = (
       const sections = deleteElement(frontMatter.sections, indexToDelete)
       const newErrorSections = deleteElement(errors.sections, indexToDelete)
       const newDisplaySections = deleteElement(displaySections, indexToDelete)
+      const newTextcardErrors = deleteElement(errors.textcards, indexToDelete)
```
seaerchin, line 603: i'm assuming this should be `subindexToDelete` - could you check downstream if the deletion is done properly also?

src/layouts/components/Homepage/TextCardsBody.tsx
seaerchin, line 1: skimmed, seems to follow same structure as existing code

src/types/homepage.ts
```diff
@@ -78,6 +78,19 @@ export interface AnnouncementsBlockSection {
   subtitle?: string
   announcement_items: AnnouncementOption[]
 }
+export interface TextcardsCard {
```
seaerchin, line 81: suggest renaming to TextCardItem

seaerchin (commented):
we should aim to preserve the API of the existing code in `useDrag` as much as possible because it makes a guarantee that we can compute the current state from the previous state. 

if possible, we should pass in an updated copy and have the method update it rather than changing the API of the method itself

src/hooks/useDrag.tsx
alexanderleegs: This call is made when a new section is created - `homepageState.errors.textcards` needs to be modified here because the length of the textcards errors should also match the length of sections

src/hooks/useDrag.tsx
alexanderleegs: Ah nope, `errors.textcards` is an nested array - each textcard error is identified by parent index and subitem index, and the only time the first level of `errors.textcards` is modified is when a section is created/deleted/moved

src/hooks/useDrag.tsx
alexanderleegs: See above! This is for deleting a section, so the first layer of errors in `errors.textcards` should be deleted

src/types/homepage.ts
alexanderleegs: Changed in https://github.com/isomerpages/isomercms-frontend/pull/1505/commits/21b7ee1e6bc696c409c65b4d95b9638dcc458779!

src/components/Editable/Editable.tsx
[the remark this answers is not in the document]
alexanderleegs: changed in https://github.com/isomerpages/isomercms-frontend/pull/1505/commits/21b7ee1e6bc696c409c65b4d95b9638dcc458779!

seaerchin (approved):

PR #1507, opened 2023-09-20
chore(hero): add feature flag for hero banner

seaerchin:
## Problem
we want to gate our hero banner variant behind a flag for now so that users **cannot** toggle and anyhow change the variant.

## Solution
1. add feature flag in
2. add default args to template (previously caller passed in default); for **preview**, there's already a default added last time (unseen cos no diff)

## Tests 
- [ ] go to **an actual production site from our agency** (<- we want to guarantee that there's no template related changes made)
- [ ] expand hero section
- [ ] **should not** have single select for toggling variant

### DIFF
--- a/src/constants/featureFlags.ts
+++ b/src/constants/featureFlags.ts
@@ -1,3 +1,4 @@
 export const FEATURE_FLAGS = {
   STYLING_REVAMP: "styles",
+  HOMEPAGE_TEMPLATES: "homepage_new_templates",
 } as const
--- a/src/layouts/components/Homepage/HeroBody.tsx
+++ b/src/layouts/components/Homepage/HeroBody.tsx
@@ -9,6 +9,7 @@ import {
   HStack,
   Spacer,
 } from "@chakra-ui/react"
+import { useFeatureIsOn } from "@growthbook/growthbook-react"
 import {
   FormErrorMessage,
   FormLabel,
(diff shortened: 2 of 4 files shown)


src/layouts/components/Homepage/HeroBody.tsx
```diff
@@ -291,29 +293,34 @@ const HeroLayoutForm = ({
...
-                id: "section-0-hero-variant",
-                value: val as HeroBannerLayouts,
-              },
-            })
-          }}
-        />
-      </FormControl>
+      <Text textStyle="h5">{`Customise ${
+        showNewLayouts ? "Layout" : "Hero"
+      }`}</Text>
```
seaerchin, line 302: checked with design - avoid confusion

src/layouts/components/Homepage/HeroBody.tsx
```diff
@@ -291,29 +293,34 @@ const HeroLayoutForm = ({
...
+      {showNewLayouts && (
```
seaerchin, line 303: only change made is conditional, rest is formatting

alexanderleegs (approved):

src/constants/featureFlags.ts
```diff
@@ -1,3 +1,4 @@
 export const FEATURE_FLAGS = {
   STYLING_REVAMP: "styles",
+  HOMEPAGE_TEMPLATES: "homepage_new_templates",
```
kishore03109, line 3: Just to confirm ya moving forward we use this FF? Ie any agency will either see all of the homepage changes or none of the homepage changes?

kishore03109 (approved):

src/constants/featureFlags.ts
seaerchin: yeap, cleared with @jandonyada for first pass

PR #1513, opened 2023-09-20
fix(chromatic): only do path checking on push

dcshzj:
## Problem

Path checking is done for both push and issue_comment currently, which is not ideal since if I do a `!run chromatic`, I really want Chromatic to run, regardless of whether there is a need to. We as humans need to take back control from the system.

## Solution

**Breaking Changes**

**Improvements**:

- Restrict the path changed checking to only push events.

## Tests

- [ ] Smoke tests

Minor fix, only testable after merge.

## Deploy Notes

*None*

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -62,7 +62,8 @@ jobs:
         with:
           fetch-depth: 0 # 👈 Required to retrieve git history
 
-      - name: Check for changes
+      - name: Check for changes (push)
+        if: ${{ github.event_name == 'push' }}
         uses: dorny/paths-filter@v2
         id: filter
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/chromatic.yml
```diff
@@ -62,7 +62,8 @@ jobs:
         with:
           fetch-depth: 0 # 👈 Required to retrieve git history
 
-      - name: Check for changes
+      - name: Check for changes (push)
+        if: ${{ github.event_name == 'push' }}
         uses: dorny/paths-filter@v2
         id: filter
         with:
```
kishore03109, line 69: wait for the filters do we need package json? asking cuz of the failure [here](https://github.com/isomerpages/isomercms-frontend/actions/runs/6247004407/job/16958757263#step:7:23), but not 100% sure if needed

dcshzj: Yup I think that's a good idea, added in 48e10d7!

PR #1527, opened 2023-09-27
chore: refactor homepage to if/else to preserve existing behaviour

alexanderleegs:
Replaces behaviour in `useDrag` for homepage to allow for handling section subitems without changing the existing behaviour. This PR modifies the `switch/cases` to `if/else` instead to allow for types of the form `<subitem name>-<number>`.

### DIFF
--- a/src/hooks/useDrag.tsx
+++ b/src/hooks/useDrag.tsx
@@ -191,7 +191,9 @@ type UpdateHomepageType =
   | "announcement"
   | `textCardItem-${number}`
 
-const isUpdateHomepageType = (value: any): value is UpdateHomepageType => {
+const isUpdateHomepageType = (
+  value: any
+): value is UpdateHomepageType | EditorHomepageElement => {
   if (typeof value === "string" && value.startsWith("textCardItem-")) {
     const valArr = value.split("-")
     const possibleCardIndex = valArr[1]
@@ -206,7 +208,8 @@ const isUpdateHomepageType = (value: any): value is UpdateHomepageType => {
       value === "editor" ||
       value === "dropdownelem" ||
       value === "highlight" ||
-      value === "announcement"
+      value === "announcement" ||
+      value === "section"
     )
   }
 
@@ -414,6 +417,7 @@ const updateHomepageState = (
     )
   }
   if (type.startsWith("textCardItem")) {
+    // We've validated that type can only be such that the second item is a number
     const parentId = parseInt(type.split("-")[1], RADIX_PARSE_INT)
     const textCardsItem = (frontMatter.sections[
       parentId
@@ -460,6 +464,8 @@ export const onCreate = <E,>(
     displayHighlights,
     displayAnnouncementItems,
   } = homepageState
+
+  if (!isUpdateHomepageType(elemType)) return homepageState
 
   if (elemType === "section") {
     const sections = createElement(frontMatter.sections, val)
@@ -578,6 +584,7 @@ export const onCreate = <E,>(
     )
   }
   if (elemType.startsWith("textCardItem")) {
+    // We've validated that type can only be such that the second item is a number
     const parentId = parseInt(elemType.split("-")[1], RADIX_PARSE_INT)
     const sectionInfo = (frontMatter.sections[
       parentId
@@ -617,6 +624,7 @@ export const onDelete = (
     displayAnnouncementItems,
   } = homepageState
 
+  if (!isUpdateHomepageType(elemType)) return homepageState
   if (elemType === "section") {
     const sections = deleteElement(frontMatter.sections, indexToDelete)
     const newErrorSections = deleteElement(errors.sections, indexToDelete)
@@ -708,6 +716,7 @@ export const onDelete = (
     )
   }
   if (elemType.startsWith("textCardItem")) {
+    // We've validated that type can only be such that the second item is a number
     const parentId = parseInt(elemType.split("-")[1], RADIX_PARSE_INT)
     const newTextCards = deleteElement(
       ((frontMatter.sections[parentId] as TextcardFrontmatterSection)
--- a/src/hooks/useDrag.tsx
+++ b/src/hooks/useDrag.tsx
@@ -194,7 +194,7 @@ type UpdateHomepageType =
 const isUpdateHomepageType = (value: any): value is UpdateHomepageType => {
   if (typeof value === "string" && value.startsWith("textCardItem-")) {
     const valArr = value.split("-")
-    const number = valArr[1]
+    const possibleCardIndex = valArr[1]
     return (
       valArr.length === 2 &&
       !!possibleCardIndex &&
--- a/src/hooks/useDrag.tsx
+++ b/src/hooks/useDrag.tsx
@@ -184,6 +184,31 @@ export const useDrag: OnDragEndResponseWrapper = (
   return updateHomepageState(result, homepageState)
 }
 
+type UpdateHomepageType =
+  | "editor"
+  | "dropdownelem"
+  | "highlight"
+  | "announcement"
+  | `textcardcard-${number}`
+
+const isUpdateHomepageType = (value: any): value is UpdateHomepageType => {
+  if (typeof value === "string" && value.startsWith("textcardcard-")) {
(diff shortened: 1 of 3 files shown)


src/hooks/useDrag.tsx
```diff
@@ -184,6 +184,31 @@ export const useDrag: OnDragEndResponseWrapper = (
...
+  | "dropdownelem"
+  | "highlight"
+  | "announcement"
+  | `textcardcard-${number}`
+
+const isUpdateHomepageType = (value: any): value is UpdateHomepageType => {
+  if (typeof value === "string" && value.startsWith("textcardcard-")) {
+    const valArr = value.split("-")
+    const number = valArr[1]
+    return valArr.length === 2 && !Number.isNaN(Number(number))
```
harishv7, line 198: Nit: I think can just use `isNaN` directly without `Number.isNaN`?

harishv7: Is `!Number.isNaN(Number(number))` to assert if its a number?

In the case where value is just `textcardcard-`, split gives `['textcardcard', '']`

`!Number.isNaN(Number(number))` evaluates to true here

src/hooks/useDrag.tsx
```diff
@@ -184,6 +184,31 @@ export const useDrag: OnDragEndResponseWrapper = (
...
+  }
+  if (typeof value === "string") {
```
harishv7, line 200: nit: can be an else if

src/hooks/useDrag.tsx
```diff
@@ -184,6 +184,31 @@ export const useDrag: OnDragEndResponseWrapper = (
   return updateHomepageState(result, homepageState)
```
harishv7, line 184: this PR mainly changes the condiitonal logic from switch-case to if-else right? Why does switch-case not support
`<subitem name>-<number>`?

seaerchin: ^switch case not as powerful as `match` sadly enough :/

src/hooks/useDrag.tsx
```diff
@@ -184,6 +184,31 @@ export const useDrag: OnDragEndResponseWrapper = (
...
 }
 
+type UpdateHomepageType =
```
seaerchin, line 187: instead of defining a separate type and needing to maintain this + `HomepageDroppableZone`, could we instead just use the latter since **they should be the same**

src/hooks/useDrag.tsx
```diff
@@ -184,6 +184,31 @@ export const useDrag: OnDragEndResponseWrapper = (
...
+  | "editor"
+  | "dropdownelem"
+  | "highlight"
+  | "announcement"
+  | `textcardcard-${number}`
+
+const isUpdateHomepageType = (value: any): value is UpdateHomepageType => {
+  if (typeof value === "string" && value.startsWith("textcardcard-")) {
+    const valArr = value.split("-")
+    const number = valArr[1]
```
seaerchin, line 197: nit: it might be confusing to see `number = valArr[1]` because the variable name `number` isn't describing its usage but rather its value. 

could we use `possibleCardIndex` instead?

src/hooks/useDrag.tsx
```diff
@@ -211,211 +236,208 @@ const updateHomepageState = (
...
+    return updateAnnouncementSection(
+      homepageState,
+      newDisplayAnnouncementItems,
+      newAnnouncementsOptions,
+      newAnnouncementErrors,
+      announcementsIndex
+    )
   }
+  if (type.startsWith("textcardcard")) {
+    const parentId = parseInt(type.split("-")[1], RADIX_PARSE_INT)
```
seaerchin, line 413: nit: add a comment here on why we can do this (cos it's been validated above to have the correct type so the second item is always a number)

src/hooks/useDrag.tsx
```diff
@@ -436,155 +457,152 @@ export const onCreate = <E,>(
...
+
+    return updateAnnouncementSection(
+      homepageState,
+      newDisplayAnnouncementItems,
+      announcements,
+      newAnnouncementErrors,
+      announcementsIndex
+    )
+  }
+  if (elemType.startsWith("textcardcard")) {
```
seaerchin, line 576: we also need the same check here - `if (!isUpdateHomepageType)`

src/hooks/useDrag.tsx
```diff
@@ -595,121 +613,113 @@ export const onDelete = (
...
+
+    return updateAnnouncementSection(
+      homepageState,
+      newDisplayAnnouncements,
+      newAnnouncementOptions,
+      newAnnouncementErrors,
+      announcementsIndex
+    )
+  }
+  if (elemType.startsWith("textcardcard")) {
```
seaerchin, line 706: also need the check here

src/hooks/useDrag.tsx
alexanderleegs: They're not :(

HomepageDroppableZone is 
```
type HomepageDroppableZone =
  | "dropdownelem"
  | "leftPane"
  | "highlight"
  | "announcement"
  | `textCardItem-${number}`
```
UpdateHomepageType is
```
type UpdateHomepageType =
  | "editor"
  | "dropdownelem"
  | "highlight"
  | "announcement"
  | `textCardItem-${number}`
```

src/hooks/useDrag.tsx
alexanderleegs: My linter is removing the else ifs here actually, since the if statement returns haha

src/hooks/useDrag.tsx
alexanderleegs: Eslint prefers`Number.isNaN` because of the implicit coercion inside `isNaN`! I opted to use Number.isNan due to this

Good catch for the empty string, fixed in https://github.com/isomerpages/isomercms-frontend/pull/1527/commits/4cf4b018a4197a4a9fce80138f35577911cd81d0

src/hooks/useDrag.tsx
alexanderleegs: Makes sense, https://github.com/isomerpages/isomercms-frontend/pull/1527/commits/e1e69692b2e410b3ef8953faf7a8b2631a30abda

src/hooks/useDrag.tsx
alexanderleegs: Added comments in https://github.com/isomerpages/isomercms-frontend/pull/1527/commits/96d0dddc2f7d06b599739083fad3887ffef41b9c

src/hooks/useDrag.tsx
alexanderleegs: Added type guard in https://github.com/isomerpages/isomercms-frontend/pull/1527/commits/96d0dddc2f7d06b599739083fad3887ffef41b9c

src/hooks/useDrag.tsx
alexanderleegs: Added type guard in https://github.com/isomerpages/isomercms-frontend/pull/1527/commits/96d0dddc2f7d06b599739083fad3887ffef41b9c

src/hooks/useDrag.tsx
seaerchin: sgtm

src/hooks/useDrag.tsx
```diff
@@ -184,6 +184,38 @@ export const useDrag: OnDragEndResponseWrapper = (
...
 
+type UpdateHomepageType =
+  | "editor"
+  | "dropdownelem"
+  | "highlight"
+  | "announcement"
+  | `textCardItem-${number}`
+
+const isUpdateHomepageType = (
+  value: any
```
seaerchin, line 195: nit: chaneg to `unknown`

seaerchin (approved):
lgtm - i think we've accumulated quite abit of tech debt HAHA i think is a good time to relook at how we wanna handle dnd in these pages. 

just adding on doesn't look to be sustainable :/

src/hooks/useDrag.tsx
alexanderleegs: https://github.com/isomerpages/isomercms-frontend/pull/1527/commits/8e0b01728961af0602a429a857a7f8648f961103

PR #1529, opened 2023-09-27
IS-577: Move site privatisation to growth book

harishv7:
## Problem

Site privatisation is feature-flagged via env var now with just a boolean. We want to move this to growthbook and to allow site-level whitelist.

Closes IS-577

## Solution

Moved to growthbook with new feature flag and json response of all the whitelisted sites for repo privatisation feature.

**Breaking Changes**

  - Details ...

## Tests

- [ ] Smoke tests
    - [ ] On growthbook, whitelist your test-repo for the `dev` stage only or if on staging, for `staging` only
    - [ ] On CMS, visit site settings and ensure you see the Privacy Settings option

## Deploy Notes

**Removed environment variables**:

- `REACT_APP_IS_SITE_PRIVATISATION_ACTIVE` : removed in favour of new feature flag on GB

### DIFF
--- a/src/hooks/settingsHooks/useGetSettings.ts
+++ b/src/hooks/settingsHooks/useGetSettings.ts
@@ -6,7 +6,6 @@ import { SETTINGS_CONTENT_KEY } from "constants/queryKeys"
 
 import * as SettingsService from "services/SettingsService"
 
-import { PrivatisationWhitelist } from "types/featureFlags"
 import {
   BackendPasswordSettings,
   BackendSiteSettings,
@@ -129,7 +128,7 @@ const extractPassword = (
 export const useGetSettings = (
   siteName: string,
   isEmailLogin?: boolean,
-  repoPrivatisationWhitelist?: PrivatisationWhitelist
+  isPrivatisationAllowed?: boolean
 ): UseQueryResult<SiteSettings> => {
   const shouldGetPrivacyDetails =
     isEmailLogin === undefined ? false : isEmailLogin
@@ -138,7 +137,7 @@ export const useGetSettings = (
     async () => {
       const siteSettings = await SettingsService.get({ siteName })
       const isSitePrivatisationActive =
-        repoPrivatisationWhitelist?.repos.includes(siteName) || false
+        isPrivatisationAllowed === undefined ? false : isPrivatisationAllowed
       let passwordSettings
       if (shouldGetPrivacyDetails && isSitePrivatisationActive) {
         passwordSettings = await SettingsService.getPassword({ siteName })
--- a/src/hooks/settingsHooks/useGetSettings.ts
+++ b/src/hooks/settingsHooks/useGetSettings.ts
@@ -136,8 +136,7 @@ export const useGetSettings = (
     [SETTINGS_CONTENT_KEY, siteName, shouldGetPrivacyDetails],
     async () => {
       const siteSettings = await SettingsService.get({ siteName })
-      const isSitePrivatisationActive =
-        isPrivatisationAllowed ?? false
+      const isSitePrivatisationActive = isPrivatisationAllowed ?? false
       let passwordSettings
       if (shouldGetPrivacyDetails && isSitePrivatisationActive) {
         passwordSettings = await SettingsService.getPassword({ siteName })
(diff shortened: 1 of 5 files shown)


src/types/featureFlags.ts
```diff
@@ -1,8 +1,11 @@
 import { FEATURE_FLAGS } from "constants/featureFlags"
 
+export type PrivatisationWhitelist = { repos: string[] }
+
 // Example usage: const gb = useGrowthBook<FeatureFlags>();
 export interface FeatureFlags {
   [FEATURE_FLAGS.STYLING_REVAMP]: boolean
+  [FEATURE_FLAGS.REPO_PRIVATISATION]: PrivatisationWhitelist
```
kishore03109, line 8: Hmm why are we not using a boolean value like styling revamp? This way we are exposing info to the client on which sites use this feature, which seems something that we can avoid by switching to boolean type instead, would this be feasible?

src/layouts/Settings/Settings.tsx
```diff
@@ -37,6 +41,14 @@ import { PrivacySettings } from "./PrivacySettings"
 import { SocialMediaSettings } from "./SocialMediaSettings"
 
 export const Settings = (): JSX.Element => {
+  const growthbook = useGrowthBook<FeatureFlags>()
+  const fallbackWhitelist: PrivatisationWhitelist = { repos: [] }
+  const repoPrivatisationWhitelist =
+    growthbook?.getFeatureValue(
+      FEATURE_FLAGS.REPO_PRIVATISATION,
+      fallbackWhitelist
+    ) || fallbackWhitelist
```
kishore03109, line 50: Hmm abit a bit confused here, why would growthbook be undefined ah? 
+ how do we reason about `fallbackWhiteList`? if this list does not contain anything, should this still live in production line ya?

src/types/featureFlags.ts
harishv7: We want to gate the rollout slowly. See Slack convo for more info

src/layouts/Settings/Settings.tsx
harishv7: Return of useGrowthbook is this:

```
 useGrowthBook<FeatureFlags>(): GrowthBook<FeatureFlags> | undefined
```

It can be undefined because under the hood it uses a react context and growthbook if not yet initialised will return undefined

The fallbackWhitelist helps in 2 scenarios:
1. If the useGrowthbook call returns undefined
2. If growthbook is initialised but the fetching from growthbook fails

Hence, it should be there for defensiveness

src/types/featureFlags.ts
kishore03109: Hmm correct me if I am wrong ya but we can still rollout per site basis without returing the entire list to the client right? eg here:
Screenshot 2023-09-27 at 5 27 08 PM

src/layouts/Settings/Settings.tsx
kishore03109: So moving forward we are going to maintain 2 sot arh? more in favour of just || [], and we expect growthbook to work most of the time, but if it doesnt we have a sane value to prevent app from crashing rather than maintaining 2 sot!

src/types/featureFlags.ts
harishv7: Oh that's what you meant! I wanted to be explicit, but actually you have a very good point. We can off-load the conditional part to growthbook! Let me try this approach 👍🏼

src/layouts/Settings/Settings.tsx
harishv7: Wait, how did you get 2 sot? There is still 1 sot on growthbook. The fallback value here is already the mechanism for preventing the app from crashing

src/types/featureFlags.ts
harishv7: fixed

src/hooks/settingsHooks/useGetSettings.ts
```diff
@@ -131,17 +127,19 @@ const extractPassword = (
...
+  isPrivatisationAllowed?: boolean
 ): UseQueryResult<SiteSettings> => {
   const shouldGetPrivacyDetails =
     isEmailLogin === undefined ? false : isEmailLogin
   return useQuery<SiteSettings>(
     [SETTINGS_CONTENT_KEY, siteName, shouldGetPrivacyDetails],
     async () => {
       const siteSettings = await SettingsService.get({ siteName })
+      const isSitePrivatisationActive =
+        isPrivatisationAllowed === undefined ? false : isPrivatisationAllowed
```
kishore03109, line 140:
```suggestion
        isPrivatisationAllowed = isPrivatisationAllowed ?? false
```

kishore03109 (approved):
lgtm w nit

PR #1536, opened 2023-09-28
fix(e2e): utlise new backend endpoint to reset e2e repos

dcshzj:
## Problem

Part of IS-446.

## Solution

**Breaking Changes**

**Improvements**:

- Resetting e2e test repos now uses a new backend endpoint instead of calling GitHub API directly. This was done in https://github.com/isomerpages/isomercms-backend/pull/950.

## Tests

- [x] Smoke tests
    - [ ] Ensure your backend has the new reset repo endpoint
    - [ ] Run `npm run reset-e2e`
    - [ ] Notice that there are now 4 repos being reset

## Deploy Notes

**REMOVED dev dependencies**:

- `btoa`

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -11,7 +11,7 @@ const {
 } = process.env
 if (!CYPRESS_COOKIE_NAME || !CYPRESS_COOKIE_VALUE) {
   throw new Error(
-    `E2E tests require env vars CYPRESS_COOKIE_NAME (${CYPRESS_COOKIE_NAME}) and CYPRESS_COOKIE_VALUE (${CYPRESS_COOKIE_VALUE}) to be set.`
+    `E2E tests require env vars CYPRESS_COOKIE_NAME and CYPRESS_COOKIE_VALUE to be set.`
   )
 }
 
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -2,23 +2,28 @@
 const { spawn } = require("child_process")
 
 const axios = require("axios")
-const btoa = require("btoa")
 
-// login credentials
-const { PERSONAL_ACCESS_TOKEN, USERNAME } = process.env
-if (!PERSONAL_ACCESS_TOKEN || !USERNAME) {
+// Login credentials
+const {
+  CYPRESS_COOKIE_NAME,
+  CYPRESS_COOKIE_VALUE,
+  CYPRESS_BACKEND_URL,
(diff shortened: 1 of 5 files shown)


scripts/build.js
```diff
@@ -2,23 +2,28 @@
...
+// Login credentials
+const {
+  CYPRESS_COOKIE_NAME,
+  CYPRESS_COOKIE_VALUE,
+  CYPRESS_BACKEND_URL,
+} = process.env
+if (!CYPRESS_COOKIE_NAME || !CYPRESS_COOKIE_VALUE) {
   throw new Error(
-    `E2E tests require env vars PERSONAL_ACCESS_TOKEN (${PERSONAL_ACCESS_TOKEN}) and USERNAME (${USERNAME}) to be set.`
+    `E2E tests require env vars CYPRESS_COOKIE_NAME (${CYPRESS_COOKIE_NAME}) and CYPRESS_COOKIE_VALUE (${CYPRESS_COOKIE_VALUE}) to be set.`
```
kishore03109, line 14: Nit: lets not throw/log the actual content of an env var, better practice here would just be to mention that `E2E tests require env vars CYPRESS_COOKIE_NAME and CYPRESS_COOKIE_VALUE to be set`

scripts/build.js
```diff
@@ -83,7 +79,8 @@ const runE2eTests = async () =>
         // NOTE: Cypress uses the exit code to show the number of failed tests.
         // If we do not exit here, the child process (that's running the test) will not be able to feedback
         // to Github Actions if the tests pass.
-        process.exit(exitCode)
+        // If exitCode is null, means that the child process was killed by a signal.
+        process.exit(exitCode || 1)
```
kishore03109, line 83: Hmm are we unintentionally changing the semantics of the code here? 
the comment mentions that the exit code is used to show the number of failed tests, if the child process was killed by a signal, should we report that 0 tests passed rather than 1?

dcshzj: Hmm we can't do an `exit 0` as it would mean the build passed, so I opted to default to `1` so that the build can fail correctly when it is killed externally.

scripts/build.js
dcshzj: Fixed in c9c68b837c0485d6f7d4859b9eaac88a3d2c6c14!

harishv7 (approved):

PR #1559, opened 2023-10-03
Fix/is 470 edit page upload image modal

alexanderleegs:
## Problem

This PR fixes IS-470, as well as several other media modal related issues. Previously, the issue was that uploaded images were not showing up properly in the modal which also asked users to specify alt text. This was actually due to a combination of factors:

- After an image was successfully uploaded, users were brought back to the select page, with the image already selected - but this was not indicated to users in any way, so users would normally click into the image they wanted again
- Clicking on the image in this screen actually deselects the image, but the proceed button was bugged to not get disabled in this case, and the item appears selected due to faulty display css
- The image selected now has missing data in some places, which prevents the preview from working

To fix these issues, this PR makes the following changes:

- The image upload modal immediately transitions to the modal asking for alt text to insert - this is to prevent users from having to go back through their list of images to find and select the image
- The deselect behaviour has been fixed
- The outline css has been tweaked to be active for the selected item, and not on focus.

Fixed behaviour:
https://github.com/isomerpages/isomercms-frontend/assets/22111124/db49e816-fe50-46c8-9980-69ff5ea4e778

### DIFF
--- a/src/components/media/MediaModal.jsx
+++ b/src/components/media/MediaModal.jsx
@@ -42,7 +42,7 @@ const MediaModal = ({ onClose, onProceed, type, showAltTextModal = false }) => {
 
   const onMediaSelect = (media) => {
     if (methods.watch("selectedMedia")?.name === media.name) {
-      methods.setValue("selectedMedia", "")
+      methods.setValue("selectedMedia", undefined)
       methods.setValue("selectedMediaPath", "")
     } else {
       methods.setValue(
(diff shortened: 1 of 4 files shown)


src/components/media/MediaModal.jsx
```diff
@@ -42,7 +42,7 @@ const MediaModal = ({ onClose, onProceed, type, showAltTextModal = false }) => {
 
   const onMediaSelect = (media) => {
     if (methods.watch("selectedMedia")?.name === media.name) {
-      methods.setValue("selectedMedia", {})
+      methods.setValue("selectedMedia", "")
```
dcshzj, line 45: A little confused here, shouldn't it be an empty object and not an empty string?

alexanderleegs: The check for no item selected here is actually just an existence check, so `{}` mistakenly passes! I've changed this value to `undefined` to more clearly convey the meaning though

dcshzj (approved):

dcshzj:
🥇 for a clean solution

PR #1570, opened 2023-10-06
refactor(collections): upgrade third nav edit page to use v2 endpoint

dcshzj:
## Problem

The third nav edit page still uses the v1 collection pages endpoint.

## Solution

**Breaking Changes**

**Improvements**:

- The underlying hook to get the list of collection pages has been upgraded to use the v2 endpoint.
- Unused functions related to the collections endpoint has been removed.

## Tests

- [x] Smoke tests
    - [ ] Navigate to any subfolder of any site
    - [ ] Load the edit page screen
    - [ ] Verify that the third level navigation still appears on the left side of the preview, and should match the folder structure of that particular subfolder.

## Deploy Notes

*None*


dcshzj:
!run e2e

src/hooks/directoryHooks/useGetAllDirectoryPages.ts
```diff
@@ -0,0 +1,60 @@
...
+        return item.children.map((child) => `${item.name}/${child}`)
+      })
+      .flatMap((item) => item)
+  })
+}
+
+export const useGetAllDirectoryPages = (
+  params: DirectoryParams,
+  queryOptions?: Omit<UseQueryOptions<string[]>, "queryFn" | "queryKey">
+): UseQueryResult<string[]> => {
```
alexanderleegs, line 36: This hook is called by all `EditPages`, including unlinked and resource pages, which lack a `collectionName` (required by `DirectoryParams`). I think this currently doesn't affect those pages though, since this hook's primary function seems to be for retrieving data to build a page's third nav in the preview, so we can probably avoid calling them? Currently users always see an error popup when loading unlinked/resource pages

dcshzj: Ah I missed that error toast when testing. Fixed in 71b89a5e!

src/templates/pageComponentsV2/LeftNav.jsx
```diff
@@ -2,12 +2,19 @@
...
 import { useEffect, useState } from "react"
 
+import { useGetAllDirectoryPages } from "hooks/directoryHooks"
+
 import { generateLeftNav } from "templates/utils/leftnavGeneration"
 
-export const LeftNav = ({ dirData, fileName }) => {
+export const LeftNav = ({ siteName, fileName, collectionName }) => {
+  const params = {
+    siteName,
```
alexanderleegs, line 11: Nit: we probably don't need `siteName` to be drilled here because we can retrieve it from 
`useParams()`

alexanderleegs (approved):
minor nit, but otherwise lgtm

src/templates/pageComponentsV2/LeftNav.jsx
dcshzj: Forgot that `useParams()` existed, fixed in bd527cb!

PR #1571, opened 2023-10-06
chore(api): remove unused functions in api that use v1 endpoints

dcshzj:
## Problem

There are many unused functions in `src/api.js`, which can be confusing especially if they are using the v1 API endpoints.

## Solution

**Breaking Changes**

**Improvements**:

- Removed all unused functions in `src/api.js`. This is done by doing a global search for each function name and if they are not used beyond `src/api.js`, they are removed.

## Tests

- [ ] Smoke tests

## Deploy Notes

*None*


src/api.js
```diff
@@ -34,266 +34,13 @@ const getLastUpdated = async (siteName) => {
...
-  const apiUrl = `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/folders/${folderName}/subfolder/${subfolderName}/rename/${newSubfolderName}`
-  return axios.post(apiUrl)
-}
-
 const getResourceRoomName = async (siteName) => {
   const apiUrl = `${process.env.REACT_APP_BACKEND_URL}/sites/${siteName}/resource-room`
   const resp = await axios.get(apiUrl)
   const { resourceRoom } = resp.data
   return resourceRoom
 }
```
alexanderleegs, line 42: Is this method still used? I think we might be able to remove it as well

src/api.js
```diff
@@ -590,37 +130,10 @@ export {
   getDirectoryFile,
   setDirectoryFile,
```
alexanderleegs, line 131: Saw that these were removed in your other PR

alexanderleegs (approved):
Minor clarification on something that wasn't removed, but otherwise lgtm! Thanks for looking through the file!

I think there might be others we can drop such as `getMediaDetails` because it's only used in an unused method in `legacy.js`, but we can clean up other files in a later PR

src/api.js
dcshzj: Oops yup, removed in c142579e!

src/api.js
dcshzj: Yup, at least `getDirectoryFile` is being used currently but I figured would be cleaner if I removed them in the other PR since they are related.

PR #1572, opened 2023-10-09
refactor(lastupdated): upgrade get last updated to v2 endpoint

dcshzj:
## Problem

The last updated hook still uses the v1 API endpoint.

## Solution

**Breaking Changes**

**Improvements**:

- The underlying hook to get the last updated information about the site has been upgraded to use the v2 endpoint.
- Unused functions related to the last updated information endpoint has been removed.

## Tests

- [x] Smoke tests
    - [ ] Navigate to the workspace of any site
    - [ ] Verify that the last updated information on the left sidebar still shows the relative time correctly.

## Deploy Notes

*None*


src/hooks/siteDashboardHooks/useGetLastUpdated.ts
```diff
@@ -0,0 +1,46 @@
...
+  const errorToast = useErrorToast()
+
+  return useQuery<string>(
+    [LAST_UPDATED_KEY, siteName],
+    () => getRelativeLastUpdated(siteName),
+    {
+      ...queryOptions,
+      retry: false,
+      onError: (err) => {
+        console.log(err)
```
harishv7, line 33: was this intentional to log to console?

src/services/SiteDashboardService.ts
```diff
@@ -1,11 +1,19 @@
...
   SiteDashboardInfo,
   SiteDashboardReviewRequest,
 } from "types/siteDashboard"
 
 import { apiService } from "./ApiService"
 
+export const getLastUpdated = async (
+  siteName: string
+): Promise<LastUpdatedInfo> => {
+  const endpoint = `/sites/${siteName}/lastUpdated`
```
harishv7, line 13: if possible endpoints should be made last-updated. but non-blocking, this shld be standardised for a couple of endpoints actually

src/hooks/siteDashboardHooks/useGetLastUpdated.ts
```diff
@@ -0,0 +1,46 @@
...
+        if (isAxiosError(err) && err.response && err.response.status === 404) {
+          setRedirectToNotFound()
+        } else {
+          errorToast({
+            id: "get-last-updated-error",
+            description: `There was a problem retrieving the site last updated time.`,
```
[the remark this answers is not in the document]
dcshzj: Fixed in a495c470!

src/hooks/siteDashboardHooks/useGetLastUpdated.ts
dcshzj: I don't have a clear answer for this because there are some other hooks that do this, but I removed it in a495c470.

src/services/SiteDashboardService.ts
dcshzj: True that yeah, I think we can perform this migration after all the v1 endpoints have been upgraded to v2.

harishv7 (approved):

PR #1577, opened 2023-10-12
refactor(editpage): abstract out overwrite changes warning modal

dcshzj:
## Problem

Part of ticket 629 (intentionally not linking since JIRA will think the task is done).

## Solution

**Breaking Changes**

**Improvements**:

- The overwrite changes warning modal that used to be present in EditPage has now been abstracted out into its own component. This is to allow other editing pages to be able to use the same override modal.

## Tests

- [x] Smoke tests
    - [ ] Go to any site
    - [ ] Open the page editor for a single page on two tabs (both tabs editing the same page)
    - [ ] Make changes and save on the first tab, ensure that the page has saved successfully
    - [ ] Make changes on the second tab and save, verify that the override changes warning modal appears as intended
    - [ ] Click on Override, verify that the changes are indeed overwritten on the repository

## Deploy Notes

*None*

### DIFF
--- a/src/components/OverwriteChangesModal/OverwriteChangesModal.tsx
+++ b/src/components/OverwriteChangesModal/OverwriteChangesModal.tsx
@@ -1,0 +1,51 @@
+import { Box, Text, VStack } from "@chakra-ui/react"
+import { Button } from "@opengovsg/design-system-react"
+
+import { WarningModal } from "components/WarningModal"
+
+interface OverwriteChangesModalProps {
+  isOpen: boolean
+  onClose: () => void
+  onProceed: () => void
+}
+
+export const OverwriteChangesModal = (
+  props: OverwriteChangesModalProps
+): JSX.Element => {
+  const { isOpen, onClose, onProceed } = props
+
+  return (
+    <WarningModal
+      isOpen={isOpen}
+      onClose={onClose}
+      displayTitle="Override Changes"
+      displayText={
+        <Box>
+          <VStack gap="1.5rem">
+            <Text>
+              A different version of this page has recently been saved by
+              another user. You can choose to either override their changes, or
+              go back to editing.
+            </Text>
+            <Text>
+              We recommend you to make a copy of your changes elsewhere, and
+              come back later to reconcile your changes.
+            </Text>
+          </VStack>
+        </Box>
+      }
+    >
+      <Button
+        variant="clear"
+        colorScheme="secondary"
+        textColor="text.title.alt"
+        onClick={onClose}
+      >
+        Back to Editing
+      </Button>
+      <Button colorScheme="critical" onClick={onProceed}>
+        Override
+      </Button>
+    </WarningModal>
+  )
+}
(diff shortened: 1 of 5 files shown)


src/components/OverwriteChangesModal/OverwriteChangesModal.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import { WarningModal } from "components/WarningModal"
+
+interface OverwriteChangesModalProps {
+  isOpen: boolean
+  onClose: () => void
+  onProceed: () => void
+}
+
+export const OverwriteChangesModal = (
```
seaerchin, line 12: could i get some insight on why we're doing this instead of just changing `WarningModal`? is this done for convenience due to how much `WarningModal` is used?

src/components/OverwriteChangesModal/OverwriteChangesModal.tsx
```diff
@@ -0,0 +1,51 @@
...
+  props: OverwriteChangesModalProps
+): JSX.Element => {
+  const { isOpen, onClose, onProceed } = props
```
seaerchin, line 15: nit: destructure inline

seaerchin (approved):
lgtm with 1 question; do note that this does increase maintenance debt, and we might not be revisiting this area of the codebase sometime soon.

src/components/OverwriteChangesModal/OverwriteChangesModal.tsx
dcshzj: Fixed in 717246b1!

dcshzj:
!run chromatic

src/components/OverwriteChangesModal/OverwriteChangesModal.tsx
dcshzj: This is done mainly because the overwrite modal would not be used on solely the EditPage, but also on the other special pages (starting with the EditHomepage). WarningModal is also used in other places as well, which means changing it would not be possible.

seaerchin: Ah so to be clear, it's done as a new component for convenience due to the large scope of change for refactoring the `WarningModal`?

I think this is ok then, maybe we might wanna file an issue so we actually fix the root cause (`WarningModal` being trash)

dcshzj: Wait sorry I'm a bit confused. This new OverrideChangesModal still uses WarningModal under the hood, so it isn't really a refactoring per se but rather just moving it out of EditPage so that it can be used in other pages.

seaerchin: yeap, i just thought that the "ideal" way to do this would be to change `WarningModal` rather than wrapping `WarningModal` which is why i wanted clarity on why this was done

dcshzj:
!run chromatic

PR #1578, opened 2023-10-12
refactor(media): paginate images

seaerchin:
## Problem
Previously, the frontend fetched all the images at once without pagination. This resulted in a large blob size being transported over the network, which resulted in an extremely long loading time.

## Solution
1. Paginate the required content. This is done through adding 2 hooks, 1 to fetch subdirectories (of images), another to fetch the files contained inside that subdirectory (ie, the actual content, be it images or files). 
    - no pagination done on subdirectories (we can) as i think there's no case yet of having a huge # of directories 
    - this is done using page and limit, which dictates the # of pages and the # of items per page respectively
    - we use 2 hooks instead of 1 because we want to **fetch subdirs and content separately**. Lumping them together would result in both refetching whenever the page changes, which isn't good
2. Adding a pagination component in images that retriggers whenever we change the page.

## Tests 
- [ ] Go to any repo (preferably a-test-v4 or nlb)
- [ ] click on the images
- [ ] if there are >15 images, there should be a pagination component
- [ ] clicking on a diff page should refetch 
- [ ] **clicking on a page that was previously rendered should not trigger a refetch**
- [ ] uploading the image should proceed normally
- [ ] clicking on an image should proceed normally (+ show preview)
- [ ] go to a content page
- [ ] selecting an image to insert/uplodaing image shuold proceed normally
- [ ] repeat the above for files


dcshzj (commented):
I think generally is okay but wanted to know if design is okay with this pagination approach.

I'm asking because it doesn't seem to gel well with the bulk uploading/actions approach proposed by design, as they can't see all their images at a glance, and if they were to select multiple images they would have to toggle between pages. Why not load the next 15 images on scroll instead (similar to infinite scrolling but we stop once we got all the available images)?

src/layouts/Media/Media.tsx
```diff
@@ -59,15 +59,51 @@ const getMediaLabels = (mediaType: "files" | "images"): MediaLabels => {
   }
 }
 
+// NOTE: This is synced with backend
+const MEDIA_PAGINATION_SIZE = 15
```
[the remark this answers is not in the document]
seaerchin: my bad here - the comment is more on the default on the backend (also 15); i can remove the note if it's confusing!

seaerchin:
> I think generally is okay but wanted to know if design is okay with this pagination approach.
> 
> I'm asking because it doesn't seem to gel well with the bulk uploading/actions approach proposed by design, as they can't see all their images at a glance, and if they were to select multiple images they would have to toggle between pages. Why not load the next 15 images on scroll instead (similar to infinite scrolling but we stop once we got all the available images)?

i can add this in first then refine the approach later? shouldn't let the users deal with slow load times in anticipation of a future change. wdyt?

dcshzj:
> i can add this in first then refine the approach later? shouldn't let the users deal with slow load times in anticipation of a future change. wdyt?

Hmm yeah I think that is okay too. I meant like instead of using pagination to solve the slow load times problem, why not use a load-after-scroll-to-end kind of solution? I feel like it still achieves the objective of loading faster for users and also works well for future changes.

Definitely not a blocker here but just not sure if using a paginator is ideal.

seaerchin:
> > i can add this in first then refine the approach later? shouldn't let the users deal with slow load times in anticipation of a future change. wdyt?
> 
> Hmm yeah I think that is okay too. I meant like instead of using pagination to solve the slow load times problem, why not use a load-after-scroll-to-end kind of solution? I feel like it still achieves the objective of loading faster for users and also works well for future changes.
> 
> Definitely not a blocker here but just not sure if using a paginator is ideal.

checked with design on this - no blockers; i think changing from paginator -> infinite scroll (if required) is less effort than going from current state -> inf scroll.

wdyt?

src/layouts/Media/Media.tsx
```diff
@@ -143,17 +181,27 @@ export const Media = (): JSX.Element => {
...
                 if (mediaType === "images") {
                   return <ImagePreviewCard name={name} mediaUrl={mediaUrl} />
                 }
                 return <FilePreviewCard name={name} />
               })}
             </SimpleGrid>
+            {mediaFolderFiles && mediaFolderFiles?.total > 0 && (
+              <Center mt="1rem">
+                <Pagination
+                  totalCount={mediaFolderFiles?.total || 0}
```
dcshzj, line 198: nit: would the `0` ever happen? Seems like the conditional rendering would have eliminated the `0` here.

src/layouts/Media/Media.tsx
```diff
@@ -1,12 +1,15 @@
-import { SimpleGrid, Box, Text, Skeleton } from "@chakra-ui/react"
-import { Button } from "@opengovsg/design-system-react"
+import { SimpleGrid, Box, Text, Skeleton, Center } from "@chakra-ui/react"
+import { Button, Pagination } from "@opengovsg/design-system-react"
 import _ from "lodash"
+import { useEffect, useState } from "react"
```
kishore03109, line 4: nit: useState not being used

src/hooks/usePaginate.ts
```diff
@@ -0,0 +1,14 @@
+import { useState } from "react"
+
+export const usePaginate = (
+  initialPage = 0
```
kishore03109, line 4: Hmm why will there be a case when initial page is not 0 ah?

src/layouts/Media/Media.tsx
```diff
@@ -63,15 +64,50 @@ const getMediaLabels = (mediaType: "files" | "images"): MediaLabels => {
...
+
+  useEffect(() => {
+    // NOTE: Because this component is shared between different media types + subfolders,
+    // we need to reset the page number to 1 when the url changes.
+    // If this is not done, we might end up with erroneous pagination
+    // such as being on page 4 of a 1 page list.
+    // This results in the back button being clickable,
+    // but the number always being 1.
+    setCurPage(1)
+    // NOTE: **NOT** adding `setCurPage` as a dependency here
```
kishore03109, line 87: Could we add an eslint ignore if we are going against the recommended usage?

src/hooks/usePaginate.ts
```diff
@@ -0,0 +1,14 @@
...
+): [number, (page: number) => void] => {
+  const [curPage, setCurPage] = useState(initialPage)
+
+  return [
+    curPage + 1,
```
kishore03109, line 9: Hmm is there a reason for this offset? 
worried that it might be confusing to expect a currPage to be = 1 after the caller calls setCurrPage(2)

src/services/DirectoryService/DirectoryService.ts
```diff
@@ -65,9 +68,45 @@ export const getWorkspacePages = (siteName: string): Promise<PageData[]> => {
 export const getMediaData = ({
   siteName,
   mediaDirectoryName,
-}: MediaDirectoryParams): Promise<DirectoryData[]> => {
+  // NOTE: These defaults are set to adhere to our current behaviour.
+  // Github has a fixed upper limit of 1000 items and we return
+  // every item up to the upper limit.
+  // Hence, this behaviour is essentially a single page of up to 1000 items.
```
kishore03109, line 74: wait sanity check here, out limit is not the github limit per se right? do we not do pagination in our backend code when calling github?

src/types/directory.ts
```diff
@@ -45,3 +45,13 @@ export type ResourcePageData = Required<Omit<PageData, "type">> & {
 export interface ResourceRoomNameUpdateProps {
   newDirectoryName: string
 }
+
+export interface GetMediaFoldersDto {
+  directories: DirectoryData[]
+  files: MediaData[]
+  total: number
```
kishore03109, line 52: nit could we use `totalFiles` or `totalItems`? just to distinguish this from total directories?

src/services/DirectoryService/DirectoryService.ts
kishore03109: Also, from a perf pov, should we use a lower amount (ideally should be = to the number of files typically displayed to a user before any scrolling occurs) to fully utilise the benefits of pagination?

kishore03109 (commented):
>we could also fetch the media file in parallel on the frontend rather than doing it on the backend. Doing this on the backend blocks until all files are fetched, making the loading time scale linearly with the largest file size. we shuold just map over the images in the frontend and fetch individually from the backend so that we can display any images that load.

heads up this might not be possible if our repos are privatised (unless we use signed urls or a variant to achieve this)

src/layouts/Media/Media.tsx
seaerchin: done here to satisfy ts cos `totalCount` is given as a number. it was between this or using `!` to hint that it will always be present (but more prone to errors down the line)

src/layouts/Media/Media.tsx
seaerchin: this is removed downstream #1582

src/hooks/usePaginate.ts
seaerchin: lets say that we want to store current page view between loads etc. 

it's a case where it's actually more effort to remove (and less functionality) than to just have it there

src/layouts/Media/Media.tsx
seaerchin: sure

src/services/DirectoryService/DirectoryService.ts
seaerchin: > do we not do pagination in our backend code when calling github

github doesn't have pagination for repository contents :/

> should we use a lower amount (ideally should be = to the number of files typically displayed to a user before any scrolling occurs) to fully utilise the benefits of pagination 

the caller handles the limit/page. this default setting is done **to preserve compat with existing hooks**

src/types/directory.ts
seaerchin: this is changed downstream so that `total` only exists on 

```ts
export interface GetMediaFilesDto {
  files: MediaFilePathData[]
  total: number
}
```

src/hooks/usePaginate.ts
seaerchin: `Pagination` component is offset by 1.

src/hooks/directoryHooks/useListMediaFolderFiles.ts
```diff
@@ -0,0 +1,43 @@
...
+      ...queryOptions,
+      retry: false,
+      onError: (err) => {
+        console.log(err)
+        if (isAxiosError(err) && err.response && err.response.status === 404) {
+          setRedirectToNotFound()
+        } else {
+          errorToast({
+            id: "get-media-directory-error",
+            description: `There was a problem retrieving directory contents. ${DEFAULT_RETRY_MSG}`,
```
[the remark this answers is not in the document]
seaerchin: i'll change for this and the directory one as the `useGetMediaFolders` is removed. 

will change to indicate if it's pages/directory and the name

seaerchin:
> heads up this might not be possible if our repos are privatised (unless we use signed urls or a variant to achieve this)

hmm i actually did this in the subsequent PR without issue.

src/hooks/directoryHooks/useListMediaFolderSubdirectories.ts
```diff
@@ -0,0 +1,46 @@
...
+      ...queryOptions,
+      retry: false,
+      onError: (err) => {
+        console.log(err)
+        if (isAxiosError(err) && err.response && err.response.status === 404) {
+          setRedirectToNotFound()
+        } else {
+          errorToast({
+            id: "get-media-directory-error",
+            description: `There was a problem retrieving directory contents. ${DEFAULT_RETRY_MSG}`,
```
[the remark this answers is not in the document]
seaerchin: changed

dcshzj (approved):
LGTM!

PR #1579, opened 2023-10-12
chore(styles): remove styles feature flagging for special pages

dcshzj:
## Problem

The new styles have been rolled out to all users for a while now, and keeping them adds maintenance debt and makes it difficult to adjust the hooks that these special pages use.

## Solution

**Breaking Changes**

**Improvements**:

- The styles feature flagging has been removed.

## Tests

- [x] Smoke tests
    - [ ] Navigate to any site
    - [ ] Verify that the EditHomepage page shows the new styles
    - [ ] Verify that the EditNavbar page shows the new styles
    - [ ] Verify that the EditContactUs page shows the new styles

⚠️ Remove the styles feature flag on Growthbook after deployment ⚠️ 

## Deploy Notes

**REMOVED feature flag**
- styles


seaerchin (approved):
lgtm - thanks for doing this; was lazy to do and wanted to check in on priority for this (which, as it turns out, wasn't too high)

could we also take note to delete the flag on growthbook post deployment

dcshzj:
@seaerchin Added a warning in the PR description. Agreed that this wasn't really a high priority but when working on the override changes modal on the homepage, I realised I might need to change the useUpdateHomepageHook, which would be more difficult to do with this feature flag.

PR #1580, opened 2023-10-13
ref(media): upgrade getMediaDetails from v1 to v2 API endpoint

dcshzj:
## Problem

This PR is tackling a couple of issues at once:
1. The `getMediaDetails` function is still using the v1 API endpoint, which is deprecated.
2. The `prependImageSrc` function used by EditPage is extremely slow for a private repo, which causes the preview to render very slowly.

## Solution

**Breaking Changes**

**Improvements**:

- The `prependImageSrc` logic in EditPage has been changed entirely to take advantage of React Queries, so that we can cache responses from the backend and improve preview rendering speed.
    - Because [hooks need to be called within a functional component](https://react.dev/warnings/invalid-hook-call-warning), the logic behind how the preview is rendered needed to change in order to take advantage of React Queries. This is done by breaking it down into a three-step process.
    - When the user makes a change on the editor, the `editorValue` gets updated. This triggers a state change, which in turn triggers the `useEffect` hook to be called. Previously, this would trigger a single function call which parses the Markdown, prepends the image src and does a sanitisation of the HTML rendered. Now, this only triggers the parsing of Markdown and sets the state of `mediaSrcs` with an array of strings, which is the `src` attribute of all internal `src` links.
    - Since `mediaSrcs` is a state variable that is hooked into `useGetMultipleMediaHook`, this triggers the React Query hook to run and retrieve the correct image src from the backend. Once that is completed, the result is stored into `mediaData`.
    - Since `mediaData` has changed, the `useEffect` hook that is watching this variable gets triggered, which parses the Markdown from the editor, update internal `src` links with the results obtained from the backend and renders the final HTML on the preview.
- To support the above, a new `useGetMultipleMediaHook` was introduced to retrieve the media data from the backend and to make use of React Query's caching abilities. This allows the images to be loaded once for the user while the preview can update multiple times depending on the user's input into the Markdown editor.
- Now that `prependImageSrc` is no longer used, the supporting legacy functions can be removed, and the v1 API endpoints are no longer used.

## Tests

- [x] Smoke tests
    - [ ] Navigate to any site that is privatised (such as a-test-v4).
    - [ ] Edit a page and add any valid image within the repo. The image should load as expected in the preview.
    - [ ] Continue editing, the preview should refresh very quickly on every keystroke.
    - [ ] Add another image in a specific album within the repo. The image should load as expected in the preview.
    - [ ] Continue editing, the preview should refresh very quickly on every keystroke.
    - [ ] Add another image that does not exist on the repo. The placeholder image should load as expected.
    - [ ] Continue editing, the preview should refresh very quickly on every keystroke, and the preview does not keep jumping, and the placeholder image does not keep refreshing.
    - [ ] Add another image to load from an external URL. The image should load as expected in the preview.
    - [ ] Continue editing, the preview should refresh very quickly on every keystroke.
    - [ ] Add another image from an invalid external URL. The placeholder image should load as expected.
    - [ ] Continue editing, the preview should refresh very quickly on every keystroke.
    - [ ] Save the page, the page should be able to save as expected.

## Deploy Notes

*None*


dcshzj:
!run e2e

src/types/media.ts
```diff
@@ -0,0 +1,4 @@
+export interface MultipleMediaParams {
+  siteName: string
+  mediaSrcs: string[]
```
kishore03109, line 3: should we use a set instead?

src/utils/images.ts
```diff
@@ -18,3 +24,52 @@ export const getImageDetails = (
...
+ * Util method to extract all image srcs that are internal from the provided HTML.
+ */
+export const getMediaSrcsFromHtml = (html: string): string[] => {
+  const $ = cheerio.load(html)
+  const mediaSrcs: string[] = []
+
+  $("img").each((_, element) => {
+    const src = $(element).attr("src")
+
+    if (src && isLinkInternal(src)) {
```
kishore03109, line 69: why do we want to introuduce this logic in the first place ah? 
could we not treat this as long as it is an image, we cache it unless it needs to be invalidted?

src/utils/images.ts
```diff
@@ -18,3 +24,52 @@ export const getImageDetails = (
     imageDirectory,
   }
 }
+
+/**
+ * Util method to adjust image src in provided HTML, so that they can load from
```
kishore03109, line 29: nit: this is not just HTML right? includes Mkdwn as well?

kishore03109 (commented):
sanity check here, does this work for both markdown + html?

src/utils/images.ts
dcshzj: Hmm this adjustment of the image src is done on the generated HTML provided by `marked.parse()`, so we will be working on the HTML only and not the underlying Markdown.

src/utils/images.ts
dcshzj: Hmm caching still happens, just on the browser:
1. If it is an external link, the browser will fetch it once and cache it on subsequent re-renders on the preview.
2. If it is a non-private site, the frontend will query the backend, which will return the raw.githubusercontent.com image path, so the browser will fetch it once and cache it on subsequent re-renders on the preview.
3. If it is a private site, the frontend will query the backend, which will return the base64 encoding of the image loaded from GitHub/GGS. The caching is done by react-query using the image src URL specified by the user.

Overall caching will still happen, just that we are tackling the third situation as the browser is spending time converting the raw.githubusercontent.com result into base64.

src/types/media.ts
dcshzj: Adjusted in e8649b95!

src/hooks/mediaHooks/useGetMultipleMediaHook.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { MediaData } from "types/directory"
+import { MultipleMediaParams } from "types/media"
+
+const getMultipleMedia = (params: MultipleMediaParams) => {
+  const mediaService = new MediaService({ apiClient: apiService })
+
+  return Promise.all(
+    params.mediaSrcs.map((mediaSrc) => {
+      const mediaUrlTokens = mediaSrc.split("/")
+      // Note: The mediaUrl will always start with a "/"
```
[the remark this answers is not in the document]
dcshzj: Hmm it is more of a note although we are somewhat guaranteed about this behaviour. I'm not really inclined to assert here since it is safely ignored when the preview is rendered.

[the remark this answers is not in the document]
kishore03109: hmm if we are not sure, should we still put this as comments ah?

src/utils/images.ts
kishore03109: Hmm why depend on the browser here ah? we do have control over the cache header for our base 64, if we are relying on the browser should we then instead just make our cache header longer? 
i think using one solution might be better here than using 2 different solutions for the same problem!

src/utils/misc.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Util method for determining if a particular URL link is a relative link that
+ * should point to a resource on the same site.
+ */
+export const isLinkInternal = (url: string) => {
+  const tempLink = document.createElement("a")
+  tempLink.href = url
+  return tempLink.hostname === window.location.hostname
```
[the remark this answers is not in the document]
kishore03109: wait but these links are in reference to the user site no? 
eg. `/images/blah.png`, which doesnt belong in the cms? 

am actually ok with this cuz the edge case here is rare, but asking this for sake of completeness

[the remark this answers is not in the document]
dcshzj: No it shouldn't be in reference to the user's site. We are not trying to determine if the image is from the user's site or not here, but rather we are only determining if a particular image link is a relative or absolute URL. When the HTML is generated from `marked.parse`, because the user may have loaded images using either a relative or absolute URL. This function is then required to determine if the image is a relative URL.

There is no loading of images right at this step, which is why the translation of the mediaSrc is necessary, and that is done at a later stage once we get the correct URLs/base64 from the backend.

src/utils/images.ts
dcshzj: The two solutions is because one is a URL and the other is a base64 image. If the image src is a URL, the browser automatically does this caching for us, and we should leave it to the browser to determine their caching strategy, etc. depending on the user's preferences too.

In the other case, we are loading a base64 image, which by definition is not cached (since it is not loading the image from any external source, so there is nothing to cache here). The problem originally is because we always regenerate the base64 on every render, which because there is no caching involved here, the preview rendering is slow. Instead, we rely on react-query to perform the caching on the relative image URL as the key and offload the base64 generation to the backend, so we don't need to keep regenerating the base64 on every render unless the user made a change to the image URL, which is the key used for the cache.

Essentially they are two separate problems, so two separate solutions are needed.

kishore03109 (approved):

PR #1581, opened 2023-10-13
ref(media): update media hooks + components

seaerchin:
## Problem
Right now, we return the image fully per call. This isn't optimal because we have to load the image fully even when we don't need it. As an example, consider the image move - we don't require the images (only titles), but we do a full load inclusive of the image data.

## Solution
- Split up the image fetching into another hook (`useListMediaFolderFiles`) instead of stuffing it all into `useGetMediaFolders`
- similarly, split up fetching sub directories into another hook also (`useListMediaFolderSubdirectories`). 
    - both of these combined allows us to load separate content faster than when they were combined (no need to transport such a big amount of data over network) 
- for components that actually require the image to be displayed, another hook has been created (`useGetAllMediaFiles`) that abstracts over each media query. this allows us to load each image when they are fetched, **instead of all at once**. This allows progressive loading into the UI, giving the illusion of a faster loading time (slowest still same time, just that we are no longer blocked on render until it loads)
- Components have also been refactored to use the most specific hook they can 
- `MediaSelectModal` has been updated to use chakra ui so it's less jank.

## Videos 
https://github.com/isomerpages/isomercms-frontend/assets/44049504/f41de0ea-8c50-4782-8a83-1d23222cb9d5

## Screenshots
Screenshot 2023-10-13 at 6 15 43 PM

  

**Breaking Changes**

  - Requires [this](https://github.com/isomerpages/isomercms-backend/pull/967)
 PR 

## Tests
- [ ] go into `a-test-v4` repo
- [ ] upload an image
- [ ] UI should re-render to display new image
- [ ] rename image
- [ ] UI should re-render to display new name (only 1 item should re-render) 
- [ ] delete an image 
- [ ] UI should re-render to remove item 
- [ ] move the image into a subdirectory
- [ ] UI should re-render to remove item
- [ ] subdirectory should have item
- [ ] repeat teh above steps but on a sub-directory (ie, try to perform CRUD operations on a sub-directory)


seaerchin:
!run-e2e

seaerchin:
!run e2e

src/components/media/MediasSelectModal.jsx
```diff
@@ -150,41 +195,63 @@ const MediasSelectModal = ({
...
+            </Skeleton>
+          </div>
+        </ModalBody>
+
+        <ModalFooter>
+          <VStack w="100%">
+            {files && mediaFolderFiles && total > 0 && (
+              <Center w="100%">
+                <Pagination
+                  totalCount={total || 0}
```
dcshzj, line 235: Is this `|| 0` needed? Seems like it is unreachable.

src/components/media/MediasSelectModal.jsx
```diff
@@ -150,41 +195,63 @@ const MediasSelectModal = ({
...
+                      w="100%"
+                      h={isLoading ? "4.5rem" : "fit-content"}
+                      isLoaded={!isLoading}
+                    >
+                      <MediaCard
+                        type={mediaRoom}
+                        media={data}
+                        mediaItemIndex={mediaItemIndex}
+                        onClick={() => onMediaSelect(data)}
+                        key={data.name}
```
dcshzj, line 219: The key should be set on the `<Skeleton>` tag instead.

src/hooks/directoryHooks/useCreateDirectory.ts
```diff
@@ -13,7 +13,7 @@ import { extractCreateDirectoryInfo } from "./utils"
 // This is because the return type isn't used by empty resource room
 // and will trigger a refetch due to query key invalidation
 // NOTE: This is also used in directory creation screen with different params
-export const useCreateDirectory = (
+export const useCreateResourceDirectory = (
```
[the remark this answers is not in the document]
seaerchin: good catch! will fix

src/components/media/MediasSelectModal.jsx
seaerchin: good catch, removed!

src/components/media/MediasSelectModal.jsx
seaerchin: not entirely sure why here, could you explain this to me?

dcshzj: Because each array item [needs to have a unique key](https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key) that can uniquely identify it among other items, and in this case the `<Skeleton>` tag is now the one used to compare siblings, instead of `<MediaCard>` which was originally the case.

dcshzj (approved):
LGTM!

PR #1585, opened 2023-10-16
feat(infocols-panel): IS-512: Implement infocols left panel

harishv7:
Note: To review in conjunction with https://github.com/isomerpages/isomercms-frontend/pull/1586 which will need to be merged into this PR first

## Problem

Infocols does not have CMS implementation yet

Closes IS-512

## Solution

Implement Infocols CMS left panel and help overlay.

**Breaking Changes**

  - Details ...

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots
Screenshot 2023-10-16 at 11 54 00 AM
Screenshot 2023-10-16 at 11 53 55 AM
Screenshot 2023-10-16 at 11 53 48 AM
Screenshot 2023-10-16 at 11 53 44 AM

## Tests

- [ ] Smoke tests


harishv7:
Note: Lint fixes are done in the child PR - IS-513 https://github.com/isomerpages/isomercms-frontend/pull/1586

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -793,6 +882,27 @@ const EditHomepage = ({ match }) => {
               )
             }
             setHomepageState(intermediateHomepageState)
+          } else if (val.infocols) {
+            // Create 3 infoboxes by default
+            const parentId =
```
seaerchin, line 887: could i get some clarity on why the last element is the parent?

src/types/homepage.ts
```diff
@@ -92,6 +92,18 @@ export interface TextcardsSection {
   cards: TextCardItem[]
 }
 
+export interface Infobox {
```
seaerchin, line 95: nit: abit weird to not prefix with `InfoCol` when it is done so elsewhere

seaerchin (commented):
minor comments

src/layouts/EditHomepage/EditHomepage.jsx
harishv7: this happens within the create handler. so at the point of creation of a new block, this is being added to the last index of the sections array
hence, for the added element, if it is infocols, the parentId of this will be the last index of the sections array

harishv7: same is being done for textcards as well

seaerchin: gotcha - no issue with this, just that the variable `parentId` is abit confusing. could we just add this comment there

harishv7: added as comment

src/types/homepage.ts
harishv7: refactored

src/hooks/useDrag.tsx
```diff
@@ -197,6 +235,17 @@ const isUpdateHomepageType = (
       !Number.isNaN(Number(possibleCardIndex))
     )
   }
+
+  if (typeof value === "string" && value.startsWith("infocolInfobox-")) {
+    const valArr = value.split("-")
+    const possibleInfoboxIndex = valArr[1]
+    return (
+      valArr.length === 2 &&
```
kishore03109, line 243: Hmm why suddenly go 2 here ah? could we constant it vs using magic numbers?

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -383,8 +401,23 @@ const EditHomepage = ({ match }) => {
             textCardItemErrors.push(
               _.map(cards, () => getErrorValues(TEXTCARDS_ITEM_SECTION))
             )
+          }
+
+          // TODO: seems to load an existing state incorrectly into the errors object
```
kishore03109, line 406: I faced this error, i think it might have something to do with the useEffect updating errors. My concern is not mainly is the root problem found and solved, but more of is this a concern of is this to impt for production? ie are we sure that the error states are correctly shown to users?

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -820,6 +896,25 @@ const EditHomepage = ({ match }) => {
               )
             }
             setHomepageState(intermediateHomepageState)
+          } else if (val.infocols) {
+            // Create 3 infoboxes by default
+
+            // his happens within the create handler. so at the point of creation
```
kishore03109, line 902: this*

src/layouts/EditHomepage/EditHomepage.jsx
kishore03109: wait if it is handed over to the last element of the array, what happens during a reorder of the elements? if infocols and text cards are reordered, does this mean that n && n +1 swaps places with m && m + 1 respectively, where n in the index of textcard and m is the index of infocol?

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -820,6 +896,25 @@ const EditHomepage = ({ match }) => {
...
+            // of a new block, this is being added to the last index of the sections array
+            const parentId =
+              updatedHomepageState.frontMatter.sections.length - 1
+            let intermediateHomepageState = updatedHomepageState
+            for (let i = 0; i < 3; i += 1) {
```
kishore03109, line 907: I am quite confused about this... if we only have 3, should we use this as a constant instead?

src/layouts/components/Homepage/InfocolsBody.tsx
```diff
@@ -0,0 +1,209 @@
...
+                {infoboxes.map((infobox, infoboxIndex) => (
+                  <Editable.DraggableAccordionItem
+                    draggableId={`infocolInfobox-${index}-${infoboxIndex}-draggable`}
+                    index={infoboxIndex}
+                    title={infobox.title}
+                    isInvalid={_.some(infoboxErrors[infoboxIndex])}
+                    isNested
+                  >
+                    <Editable.Section>
+                      {/* infobox Title */}
```
kishore03109, line 125: infocol*

src/layouts/components/Homepage/InfocolsBody.tsx
```diff
@@ -0,0 +1,209 @@
...
+                          id={`infocolInfobox-${index}-${infoboxIndex}-title`}
+                          value={infobox.title}
+                          onChange={onChange}
+                        />
+                        <FormErrorMessage>
+                          {infoboxErrors[infoboxIndex]?.title}
+                        </FormErrorMessage>
+                      </FormControl>
+
+                      {/* infobox Description */}
```
kishore03109, line 142: infocol*

src/utils/validators.js
```diff
@@ -692,6 +701,82 @@ const validateTextcard = (cardError, field, value) => {
...
+    }
+    case "linktext": {
+      // Link text is too long
+      if (value.length >= INFOCOLS_BLOCK_LINKTEXT_MAX_LENGTH) {
+        errorMessage = `Text should be shorter than ${INFOCOLS_BLOCK_LINKTEXT_MAX_LENGTH} characters.`
+      }
+      break
+    }
+    case "url": {
+      // if (!permalinkRegexTest.test(value)) {
```
kishore03109, line 734: hmm could we just do a lose validation in the meantime? hunch here is that a basic incomplete one > no validation

src/hooks/useDrag.tsx
harishv7: Hmm i don't feel this is a magic number. If read within the context, we are doing a split on line 240. Hence, the expectation is to receive an array of length 2

src/layouts/EditHomepage/EditHomepage.jsx
harishv7: Oops, I fixed this already. Was updated in the child PR. Will remove this comment

src/layouts/EditHomepage/EditHomepage.jsx
harishv7: fixed

src/layouts/components/Homepage/InfocolsBody.tsx
harishv7: is supposed to be infobox here. infocol is the outer section whereas the infobox is the inner section

src/layouts/components/Homepage/InfocolsBody.tsx
harishv7: same as above

src/layouts/EditHomepage/EditHomepage.jsx
harishv7: not always 3. if u see the function where this is being run, is the create operation. When a user creates a infocol section, we by default create 3 infoboxes within it

src/layouts/EditHomepage/EditHomepage.jsx
harishv7: If u see src/hooks/useDrag.tsx, we will need to re-compute the state of the specific section (updatePositions) - so we do a shifting of the indices from source index to the new destination index

src/utils/validators.js
harishv7: Preferring to do this in a separate PR - none of the components in this file have url validation. should update all at once

src/layouts/EditHomepage/EditHomepage.jsx
kishore03109: Would prefer this to be a func abstracted out and not in this chunky file similar to what `getDefaultAnnouncementSection` does. 
could we at least define a `DEFAULT_NUMBER_OF_INFO_COLS` for this ya?

src/layouts/EditHomepage/EditHomepage.jsx
kishore03109: lets huddle over this to speed up review

kishore03109: Is this needed because there could be more than one infocol?

kishore03109 (approved):
Approving this on the assumption that ther reason for the additional completxiy in parent id is due to because there exists more than one infocol. Please dismiss my review if this assumption is incorrect

harishv7:
> Approving this on the assumption that ther reason for the additional completxiy in parent id is due to because there exists more than one infocol. Please dismiss my review if this assumption is incorrect

@kishore03109 yes, we need a reference as to which infocol index (if more than 1) the infobox belongs to

src/layouts/EditHomepage/EditHomepage.jsx
harishv7: done

harishv7:
Merging as design approved as well @sehyunidaaa

src/layouts/EditHomepage/EditHomepage.jsx
harishv7: yea, hence we need to track to which section the infobox belongs to

PR #1586, opened 2023-10-16
feat(infocols-previews): IS-513: Implement infocols previews

harishv7:
## Problem

Infocols has no preview yet

Closes IS-513

## Solution

Implements template previews for infocols on CMS

Screenshot 2023-10-16 at 11 40 36 PM

**Breaking Changes**

  - Details ...


src/templates/homepage/InfocolsSection.tsx
```diff
@@ -0,0 +1,163 @@
...
+      title,
+      subtitle,
+      linktext,
+      url,
+      infoboxes,
+      sectionIndex,
+    }: TemplateTextCardsSectionProps,
+    ref: ForwardedRef<HTMLDivElement>
+  ) => {
+    const getInfoBox = (infoboxTitle: string, infoboxDescription?: string) => {
```
harishv7, line 29: follows same structure as in templates. infobox is an includes file in templates which has been mimicked here with a extracted function

seaerchin: not a real issue, just curious why you chose to use a function over a component

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -373,23 +374,23 @@ const EditHomepage = ({ match }) => {
             textCardItemErrors.push(
               _.map(cards, () => getErrorValues(TEXTCARDS_ITEM_SECTION))
             )
-            if (textCardItemErrors.length === 0) {
-              textCardItemErrors.push([])
-            }
```
seaerchin, line 378: what happens now when there are no items?

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -373,23 +374,23 @@ const EditHomepage = ({ match }) => {
...
             infocolInfoboxErrors.push(
               _.map(infoboxes, () => getErrorValues(INFOCOLS_INFOBOX_SECTION))
             )
-            if (infocolInfoboxErrors.length === 0) {
-              infocolInfoboxErrors.push([])
-            }
+          } else {
+            // populate for nested components
+            textCardItemErrors.push([])
+            infocolInfoboxErrors.push([])
```
seaerchin, line 393: do we not need to do the checks here? (length === 0)

src/layouts/EditHomepage/HomepagePreview.tsx
```diff
@@ -200,12 +201,16 @@ export const HomepagePreview = ({
...
           {EditorHomepageFrontmatterSection.isInfocols(section) && (
             <>
-              <div
+              <TemplateInfocolsSection
                 key={`section-${sectionIndex}`}
+                title={section.infocols.title}
+                subtitle={section.infocols.subtitle}
+                linktext={section.infocols.linktext}
+                url={section.infocols.url}
+                infoboxes={section.infocols.infoboxes}
```
seaerchin, line 210: not a blocker, wouldn't using a spread over `section.infocols` be better here?

src/styles/isomer-template/components/homepage/_infocols.scss
seaerchin, line 1: skimmed css review, leaving this to design to check for parity

seaerchin (approved):
lgtm

src/layouts/EditHomepage/EditHomepage.jsx
harishv7: the length checks were never there. I added it in the parent PR (IS-512) which was wrong. This logic exists to keep track of the errors with the indices representing the section

src/layouts/EditHomepage/EditHomepage.jsx
harishv7: Replied below - this logic is to basically maintain sectionIdx - errors mapping

src/layouts/EditHomepage/HomepagePreview.tsx
harishv7: true, was following the same format as the above components. updated now

src/templates/homepage/InfocolsSection.tsx
seaerchin: Oh ps, I was ok with the function, just curious on the reason!!! Don't feel pressured to do this

harishv7: done

src/layouts/EditHomepage/HomepagePreview.tsx
harishv7: done

PR #1588, opened 2023-10-16
feat(homepage): allow user to override changes when there is a conflict

dcshzj:
## Problem

Closes IS-629.

## Solution

**Breaking Changes**

**Features**:

- The homepage now has an override modal to override changes if a 409 Conflict is detected.

## Before & After Screenshots

Screenshot 2023-10-16 at 13 43 35

## Tests

- [x] Smoke tests
    - [ ] Navigate to any site and open 2 tabs of the Homepage editor
    - [ ] Make a valid change on the first tab and save. Verify that the changes can be saved successfully.
    - [ ] Switch to the second tab and make another change that is not the same as the first one.
    - [ ] Save the changes. Verify that the override modal (screenshot above) appears, and there is no error toast shown behind the warning modal.
    - [ ] Click on "Save my edits anyway", verify that the second set of changes have been saved successfully.

## Deploy Notes

*None*

### DIFF
--- a/src/components/OverwriteChangesModal/OverwriteChangesModal.tsx
+++ b/src/components/OverwriteChangesModal/OverwriteChangesModal.tsx
@@ -27,8 +27,8 @@ export const OverwriteChangesModal = ({
               edits now, their changes will be lost.
             </Text>
             <Text>
-              We recommend you to make a copy of your changes elsewhere and come
-              back later.
+              We recommend you to make a copy of your changes elsewhere, and
+              come back later to reconcile your changes.
             </Text>
           </VStack>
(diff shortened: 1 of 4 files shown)


kishore03109 (approved):
Mostly ok, but the copy feels off. 
rather than saying come back later, should it instead be 
save changes locally -> refresh page -> re-input their changes again?

kishore03109:
+ it may not necessarily be another user right? imagine one user with two tabs.. is this a concern?

dcshzj:
@kishore03109 restored the original copy instead after discussing with design

PR #1592, opened 2023-10-17
feat(navbar): upgrade navbar to use v2 endpoints

dcshzj:
## Problem

Closes IS-626.

## Solution

**Breaking Changes**

**Features**:

- The navigation bar has been upgraded to use the v2 API endpoints.
    - This required the removal of the `getEditNavBarData` function, which essentially retrieves the navigation bar data from 4 different sources (/navigation, /collections, /resources and individual subcollections).
    - This data is now retrieved through 4 different hooks, and the existing `useEffect` hook is enhanced to take advantage of these new hooks.

**Bug Fixes**:

- The logic to block back navigation on the homepage has also been adjusted, so that the warning modal only appears if the user actually made changes.

## Tests

- [x] Smoke tests
    - [ ] Navigate to any site's Edit Navbar page
    - [ ] Verify that the page loads as expected, with the structure of the sidebar and preview being the same as before.
    - [ ] Create a folder item and verify that (a) all existing collections are available in the dropdown, and (b) each collection shows the correct pages in the preview dropdown.
    - [ ] Create a resource room item and verify that the preview dropdown has all the resource room categories.
    - [ ] Create a custom menu group with some submenu items and verify that the preview dropdown shows the added items correctly.
    - [ ] Save the page and refresh, verify that the changes made are persistent, the success toast appears and the page does not refresh itself.

## Deploy Notes

*None*


src/layouts/EditNavBar.jsx
```diff
@@ -634,239 +625,236 @@ const EditNavBar = ({ match }) => {
...
-                            You can specify a folder or resource room to
-                            automatically populate its links.
-                          </Text>
-                        </VStack>
-                        <DragDropContext onDragEnd={onDragEnd}>
-                          <Editable.Droppable
-                            width="100%"
-                            editableId="link"
-                            onDragEnd={onDragEnd}
+        <Skeleton isLoaded={hasLoaded}>
```
dcshzj, line 628: This is only indentation changes after adding the `<Skeleton>` tag.

src/hooks/directoryHooks/useGetSecondLevelFoldersAndPages.ts
```diff
@@ -0,0 +1,45 @@
...
+  SecondLevelFoldersAndPagesParams,
+} from "types/directory"
+import { isDirectoryData } from "utils"
+
+interface CollectionOrder {
+  collectionName: string
+  order: string[]
+}
+
+const getCollectionOrder = (
```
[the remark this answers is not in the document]
seaerchin: This API should take the data directly and return the collection order instead of wrapping over directory service.

This is because the method is not concerned with the fetching of data, but the view of said data

src/hooks/directoryHooks/useGetSecondLevelFoldersAndPages.ts
```diff
@@ -0,0 +1,45 @@
...
+export const useGetSecondLevelFoldersAndPages = ({
+  siteName,
+  collectionsData = [],
+}: SecondLevelFoldersAndPagesParams): UseQueryResult<CollectionOrder[]>[] => {
+  const directoryQueries = collectionsData
+    .filter<DirectoryData>((item): item is DirectoryData =>
+      isDirectoryData(item)
+    )
+    .map(({ name }) => ({
+      queryKey: [DIR_SECOND_LEVEL_CONTENT_KEY, siteName, name],
```
seaerchin, line 38: Should we also add `collectionName` here

src/hooks/navHooks/useUpdateNavHook.ts
```diff
@@ -39,13 +43,13 @@ export const useUpdateNavHook = (
     },
     {
       onSuccess: () => {
-        queryClient.invalidateQueries([GET_NAV_KEY, siteName])
+        queryClient.invalidateQueries([NAVIGATION_CONTENT_KEY, siteName])
```
seaerchin, line 46: We're not required to update second nav contents cos only the title may change right?

src/layouts/EditNavBar.jsx
seaerchin, line 1: Skimmed, assuming only API changes from old hook to new

seaerchin (commented):
Minor comments, query key is potentially an issue as users might want to invalidate a specific collection only

src/constants/queryKeys.ts
```diff
@@ -1,5 +1,6 @@
 export const PAGE_CONTENT_KEY = "page-contents"
 export const DIR_CONTENT_KEY = "dir-contents"
+export const DIR_SECOND_LEVEL_CONTENT_KEY = "dir-second-level-contents"
```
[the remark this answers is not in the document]
dcshzj: Yup it's just the second level directories, adjusted in 758738cc00a190f8c83d69cc8a76a4e1966c880d!

src/hooks/directoryHooks/useGetSecondLevelFoldersAndPages.ts
```diff
@@ -0,0 +1,45 @@
...
+import * as DirectoryService from "services/DirectoryService/index"
+
+import {
+  DirectoryData,
+  PageData,
+  SecondLevelFoldersAndPagesParams,
+} from "types/directory"
+import { isDirectoryData } from "utils"
+
+interface CollectionOrder {
```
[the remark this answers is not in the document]
dcshzj: Adjusted in 1e8b8b84f3c2aa810746aea11ae3077bb6bce490!

src/hooks/directoryHooks/useGetSecondLevelFoldersAndPages.ts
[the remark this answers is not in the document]
dcshzj: Merged this function into the parent hook in 1e8b8b84f3c2aa810746aea11ae3077bb6bce490.

src/hooks/directoryHooks/useGetSecondLevelFoldersAndPages.ts
dcshzj: The `name` variable is the `collectionName`.

src/hooks/navHooks/useUpdateNavHook.ts
dcshzj: Hmm we still need to invalidate this key because the user may have made changes to the structure of the navbar data object. Invalidating here ensures the user gets the latest SHA from the backend after the change was made.

seaerchin: not this key, i meant `DIR_SECOND_LEVEL_CONTENT_KEY`

seaerchin (approved):
lgtm if we don't need to invalidate `DIR_SECOND_LEVEL_CONTENT_KEY`

harishv7 (approved):

dcshzj:
Yup, discussed offline. React query would probably not refetch the data (i.e. DirectoryService.getCollection() will not be called again if the query key is a cache hit), so invalidating it would be necessary.

PR #1602, opened 2023-10-18
fix(media): adjust query key to load image in preview on upload

dcshzj:
## Problem

Image preview does not load newly uploaded images correctly.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Adjust query key to load media that were just uploaded.

## Tests

- [x] Smoke tests
    - [ ] Navigate to any site and edit a page
    - [ ] Upload a new image using the editor and add the newly uploaded image to the page
    - [ ] Verify that the newly uploaded image will load after a while without any additional keystrokes/actions

## Deploy Notes

*None*

### DIFF
--- a/src/hooks/mediaHooks/useGetMultipleMediaHook.tsx
+++ b/src/hooks/mediaHooks/useGetMultipleMediaHook.tsx
@@ -40,7 +40,7 @@ export const useGetMultipleMediaHook = (
   queryOptions?: Omit<UseQueryOptions<MediaData[]>, "queryFn" | "queryKey">
 ): UseQueryResult<MediaData[]> => {
   return useQuery<MediaData[]>(
-    [MEDIA_MULTIPLE_CONTENT_KEY, params],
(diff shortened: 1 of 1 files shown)


alexanderleegs (approved):

harishv7:
@dcshzj is momentary refresh of all images instead of just the latest one expected?


https://github.com/isomerpages/isomercms-frontend/assets/5507822/01703057-594e-4523-bd0e-0c254f5e1208

alexanderleegs:
> @dcshzj is momentary refresh of all images instead of just the latest one expected?
> 
>  Screenshare.-.2023-10-18.10_07_25.PM.mp4

This should be because the query only gets cached if all images are the same - uploading a new image will cause a fresh refetch

dcshzj:
Yup, exactly what @alexanderleegs said. We might be able to optimise it further by giving each image their own cache key but I felt that it might add quite a bit of complexity. If it is important I can take this up in a separate PR!

PR #1603, opened 2023-10-18
feat(blocks): rollback handler

seaerchin:
## Problem
We have no way of rolling back blocks at present, which means that we are unable to undo changes safely. 

## Solution
Add a hook that sanitises the frontmatter if the given blocks are not inside the whitelist. This is done through: 
1. checking if the current blocks provided by frontmatter are permitted by our whitelist (and storing it inside `localStorage`) 
2. checking if existing blocks in `localStorage` are now permitted and merging with ^ 
3. presenting this back to the user. 

## Notes
added **all** blocks to feature flag, not just new blocks. lets us rollback as required. also, `Hero` doesn't rollback the variants but the block as a whole :monkas: maybe not ideal but the alternative is quite abit of work and doesn't add that much value too

## Tests
- [ ] go to `a-test-v4` on staging
- [ ] go to homepage
- [ ] make sure all the blocks are present/can be added
- [ ] go to growthbook and disable some of the flags
- [ ] go back to the cms and reload
- [ ] the unflagged blocks should now be gone
- [ ] check `localStorage` (dev console -> application) 
- [ ] blocks should be there
- [ ] undo the changes to growthbook
- [ ] reload
- [ ] blocks should be present

### DIFF
--- a/src/constants/featureFlags.ts
+++ b/src/constants/featureFlags.ts
@@ -3,4 +3,5 @@ export const FEATURE_FLAGS = {
   REPO_PRIVATISATION: "repo_privatisation",
   BANNER: "banner",
   NPS_FORM: "nps_form",
+  ENABLED_BLOCKS: "homepage_enabled_blocks",
 } as const
--- a/src/constants/featureFlags.ts
+++ b/src/constants/featureFlags.ts
@@ -2,7 +2,7 @@ export const FEATURE_FLAGS = {
   REPO_PRIVATISATION: "repo_privatisation",
   BANNER: "banner",
   NPS_FORM: "nps_form",
-  ENABLED_BLOCKS: "homepage_enabled_blocks",
+  HOMEPAGE_ENABLED_BLOCKS: "homepage_enabled_blocks",
 } as const
 
 // NOTE: Only have 4 default blocks:
--- a/src/constants/featureFlags.ts
+++ b/src/constants/featureFlags.ts
@@ -1,7 +1,10 @@
 export const FEATURE_FLAGS = {
-  HOMEPAGE_TEMPLATES: "homepage_new_templates",
   REPO_PRIVATISATION: "repo_privatisation",
   BANNER: "banner",
   NPS_FORM: "nps_form",
   ENABLED_BLOCKS: "homepage_enabled_blocks",
 } as const
+
+// NOTE: Only have 4 default blocks:
+// hero/infobar/infopic/resources
+export const NUM_DEFAULT_HOMEPAGE_BLOCKS = 4
--- a/src/constants/localStorage.ts
+++ b/src/constants/localStorage.ts
@@ -1,4 +1,5 @@
 export enum LOCAL_STORAGE_KEYS {
+  FrontmatterSections = "frontmatter-sections",
   SitesIsPrivate = "sites-is-private",
   Email = "email",
   Announcements = "announcements",
--- a/src/constants/localStorage.ts
+++ b/src/constants/localStorage.ts
@@ -1,5 +1,5 @@
 export enum LOCAL_STORAGE_KEYS {
-  FrontmatterSections = "frontmatter-sections",
+  HomepageFrontmatterSections = "homepage-frontmatter-sections",
   SitesIsPrivate = "sites-is-private",
   Email = "email",
   Announcements = "announcements",
--- a/src/hooks/homepageHooks/useFrontmatterSections.ts
+++ b/src/hooks/homepageHooks/useFrontmatterSections.ts
@@ -1,5 +1,6 @@
 import { useFeatureValue } from "@growthbook/growthbook-react"
 import _ from "lodash"
+import { useCallback } from "react"
 
 import { FEATURE_FLAGS } from "constants/featureFlags"
 import { LOCAL_STORAGE_KEYS } from "constants/localStorage"
@@ -18,9 +19,11 @@ const getIsFeatureEnabled = (
   return _.intersection(blockKeys, enabledBlocks).length > 0
 }
 
-export const useFrontmatterSections = (
-  baseFrontmatterSections: EditorHomepageFrontmatterSection[]
-): EditorHomepageFrontmatterSection[] => {
+type UseFrontmatterSectionsReturn = (
+  sections: EditorHomepageFrontmatterSection[]
+) => EditorHomepageFrontmatterSection[]
+
+export const useFrontmatterSections = (): UseFrontmatterSectionsReturn => {
   const [storedFrontmatter, setStoredFrontmatter] = useLocalStorage<
     EditorHomepageFrontmatterSection[]
   >(LOCAL_STORAGE_KEYS.FrontmatterSections, [])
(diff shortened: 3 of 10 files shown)


src/constants/featureFlags.ts
```diff
@@ -1,6 +1,10 @@
 export const FEATURE_FLAGS = {
-  HOMEPAGE_TEMPLATES: "homepage_new_templates",
   REPO_PRIVATISATION: "repo_privatisation",
   BANNER: "banner",
   NPS_FORM: "nps_form",
+  ENABLED_BLOCKS: "homepage_enabled_blocks",
```
kishore03109, line 5: Hey may I understand why splitting this up into 
`homepage_enabled_blocks` and non `homepage_enabled_blocks` would be non-trivial work? Feels like a quick win here when things are buggy!

src/constants/localStorage.ts
```diff
@@ -1,4 +1,5 @@
 export enum LOCAL_STORAGE_KEYS {
+  FrontmatterSections = "frontmatter-sections",
```
kishore03109, line 2: nit: rename to homepage-frontmatter-sections for clarity

src/constants/featureFlags.ts
```diff
@@ -1,6 +1,10 @@
...
 } as const
+
+// NOTE: Only have 4 default blocks:
+// hero/infobar/infopic/resources
+export const NUM_DEFAULT_HOMEPAGE_BLOCKS = 4
```
kishore03109, line 10: Hmm I am abit confused how this works, should we not list out the default blocks here and have a util function that tells us if this particular block is indeed a default block or not?

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -316,7 +324,10 @@ const EditHomepage = ({ match }) => {
         const scrollRefs = []
         const announcementScrollRefs = []
 
-        frontMatter.sections.forEach((section) => {
+        const sections = computeFrontmatter(frontMatter.sections)
+
+        sections.forEach((section) => {
+          console.log(section)
```
kishore03109, line 330: nit: acciedntal commit of section

src/hooks/homepageHooks/useFrontmatterSections.ts
```diff
@@ -0,0 +1,76 @@
...
+      )
+
+      setStoredFrontmatter(
+        _.uniqBy(
+          [...rolledBackBlocks, ...remainingStoredFrontmatterBlocks],
+          (obj) => JSON.stringify(obj)
+        )
+      )
+
+      console.log("===", enabledStoredFrontmatterBlocks, actualBlocks)
```
kishore03109, line 70: unintended console log

src/hooks/homepageHooks/useFrontmatterSections.ts
```diff
@@ -0,0 +1,76 @@
...
+
+      return [...actualBlocks, ...enabledStoredFrontmatterBlocks]
+    },
+    [blocks]
```
kishore03109, line 74: are we missing deps here?

src/layouts/EditHomepage/EditHomepage.jsx
```diff
@@ -1551,8 +1566,7 @@ const EditHomepage = ({ match }) => {
                                       </Editable.DraggableAccordionItem>
                                     )}
 
-                                    {showNewLayouts &&
-                                      section.announcements &&
+                                    {section.announcements &&
```
kishore03109, line 1569: is the removal of feature flag here intended?

src/layouts/Sites.tsx
```diff
@@ -163,7 +166,10 @@ export const Sites = (): JSX.Element => {
     )
   }, [siteRequestData])
 
-  const showNewLayouts = useFeatureIsOn(FEATURE_FLAGS.HOMEPAGE_TEMPLATES)
+  const { blocks } = useFeatureValue(FEATURE_FLAGS.ENABLED_BLOCKS, {
```
kishore03109, line 169: Hmm why are we deviating away from a simple boolean showNewLayouts at the growthbook level and adding this layer of logic at the code level ya? is there a benefit that we are trying to achieve here?

src/layouts/components/Homepage/HeroBody.tsx
```diff
@@ -390,7 +396,10 @@ export const HeroBody = ({
     initialSectionType
   )
   const { onChange } = useEditableContext()
-  const showNewLayouts = useFeatureIsOn(FEATURE_FLAGS.HOMEPAGE_TEMPLATES)
+  const { blocks } = useFeatureValue(FEATURE_FLAGS.ENABLED_BLOCKS, {
+    blocks: [],
+  })
+  const showNewLayouts = blocks.length > NUM_DEFAULT_HOMEPAGE_BLOCKS
```
kishore03109, line 402: This code block seems repeated, make this into util func?

kishore03109 (commented):
Hey Chin, am abit confused about the change in structure about this feature flag. Perviously, we did a per site basis of enabling all the new homepage blocks, but now it seems to have deviated into for all sites, a list of allowed homepage blocks. This approach seems more inflexible than the status quo, eg if a repo has issues, now everyone will have to be rolled back because our code did not take into account this one repo's edge case, forcing everyone to go through this rollback handler logic which introduces an additional moving part. Am I misunderstanding something here?

src/types/featureFlags.ts
```diff
@@ -8,6 +8,7 @@ export interface FeatureFlags {
     message: string
   }
   [FEATURE_FLAGS.NPS_FORM]: boolean
+  [FEATURE_FLAGS.ENABLED_BLOCKS]: string[]
```
harishv7, line 11: if this is only for homepage, can we name as HOMEPAGE_ENABLED_BLOCKS?

src/layouts/Sites.tsx
harishv7: we need to know which blocks have been rolled out and which are not in the event of a rollback

src/layouts/components/Homepage/HeroBody.tsx
harishv7, line 402: can explain what is this check for?

src/layouts/Sites.tsx
kishore03109: could we use this + site level? I think say rolling back on one site helps reduce complexity of moving parts here, thoughts?

src/constants/featureFlags.ts
seaerchin: i only ever want 1 check to be in the codebase - my thought process here is if we have to rollback, we'll rollback by block. this is because it's q likely that, for example, an issue on the block for 1 site would also be similarly applicable to other sites (ie, that blocking by site doesn't add much value).

because of this, i would rather simplify the checks required rather than adding complexity here. wdyt on this!

src/constants/localStorage.ts
seaerchin: sgtm!

src/constants/featureFlags.ts
seaerchin: actually something that i thought about was just to remove the 4 default blocks from the growthbook, but i thought that we might wanna rollback those also. (this is marginal, i don't mind just removing from gb) 

we need a util here that determines if the feature is on, and what happened previously was determining if the flag is on/off. given that the flag no longer exists, i just decided to check if all the default blocks are in (a good enough check()

src/layouts/EditHomepage/EditHomepage.jsx
seaerchin: good catch~!

src/hooks/homepageHooks/useFrontmatterSections.ts
seaerchin: gud ketch

src/hooks/homepageHooks/useFrontmatterSections.ts
seaerchin: actually, no.

src/layouts/EditHomepage/EditHomepage.jsx
seaerchin: yes!

src/layouts/Sites.tsx
seaerchin: see above

src/layouts/components/Homepage/HeroBody.tsx
seaerchin: shure

seaerchin: quoted from above
> we need a util here that determines if the feature is on, and what happened previously was determining if the flag is on/off. given that the flag no longer exists, i just decided to check if all the default blocks are in (a good enough check)

src/constants/featureFlags.ts
kishore03109: non blocking: hmm i think complexity introduced in storing + merging from local storage > maintaining a model of feature per site, but I dont feel too strong about this we can go ahead with this implementation

src/constants/featureFlags.ts
kishore03109: nit: if the code is not intended to rollback default blocks, could we just list out the blocks so that we are explict in the blocks that we support? 


non blocking since code is functional: 

Hmm could we just do away with this? 
the feature flag just contains a list of allowed homepage blocks -> then FE displays blocks to users based on that. 

I cannot reason how our fe behaves if say GB states 

`hero`, `infobar`, `infopic` + say `announcements`. in this case feature flag does not get enabled, and the default 4 blocks are still shown, right?

src/types/featureFlags.ts
kishore03109: +1

kishore03109 (approved):
approving with nits, with a preference for removing this completely from gb and explictly defining at the code level the blocks that we support since we dont intend to support rollback of production ready default blocks

src/constants/featureFlags.ts
seaerchin: hey apologies if it's unclear - i wanted to also have the ability to rollback default blocks because the drag/drop refactor also went from js -> ts. 

so, for example, if we have the 4 mentioned above, the modal **will not show** but the only blocks that are shown are `hero/infobar/infopic/announcements`. i'm not super sure of the incremental value of adding the ff back cos of additional complexity + idt we will rollback all the blocks until we have less than 4

src/hooks/homepageHooks/useFrontmatterSections.ts
seaerchin: added comment

src/types/featureFlags.ts
seaerchin: done

src/constants/featureFlags.ts
kishore03109: Wait this is not true no? cuz later in the code we check the number of allowed blocks (not the unique identifiers per block) and if === 4, that means that the feature flag is not turned on? did I misunderstand this?

seaerchin: actually you're right - i skimmed over this part and did a simple replacement. i'll update so that the check is not a blanket check but `blocks.includes(<feature_name>)`

PR #1609, opened 2023-10-19
fix(settings): use env var for spec command

seaerchin:
## Problem
Previously, we only had 1 e2e test repo. now, we have 3, which results in the command not hitting the correct repo. this results in an await call that never happens, leading to cascading failures.

## Solution
use env var


dcshzj (approved):
LGTM!

harishv7 (approved):

PR #1613, opened 2023-10-20
feat(urls): upgrade endpoint to get site URL to v2

dcshzj:
## Problem

The method to get the site URL is still using a v1 endpoint.

## Solution

**Breaking Changes**

**Features**:

- The method to get the site URL has been upgraded to use a v2 endpoint. For instances that use the site URL, a higher preference is given to the site URL specified by the user in site settings.
- All existing methods to get the staging and site URLs have been consolidated to avoid code duplication.
- Deprecated methods have been removed.

## Tests

- [x] Smoke tests
    - [ ] Navigate to a site where the site URL is defined in the settings page
    - [ ] Go to that site's workspace and change any page settings
    - [ ] Verify that the domain name under the "Page URL" shows the site URL defined in settings.
    - [ ] Navigate to a site where the site URL is NOT defined in the settings page
    - [ ] Go to that site's workspace and change any page settings
    - [ ] Verify that the domain name under the "Page URL" shows the Netlify/Amplify prod/master URL.

## Deploy Notes

*None*


src/hooks/siteDashboardHooks/useGetSiteUrl.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import * as SiteDashboardService from "services/SiteDashboardService"
+
+const getSiteUrl = async (siteName: string): Promise<string> => {
+  return SiteDashboardService.getUserProvidedSiteUrl(siteName)
+    .then((url) => {
+      if (url) return `https://${url}`
+
+      return SiteDashboardService.getSiteUrl(siteName)
+    })
```
alexanderleegs, line 13: Just curious: why do we prefer the user set url provided by settings here?

alexanderleegs (approved):
lgtm, but minor clarification on site url!

src/hooks/siteDashboardHooks/useGetSiteUrl.ts
dcshzj: Hmm it's because we should treat the user's provided URL as the truth (www.agency.gov.sg) and only fallback to the site URL stored inside the DB (master.gibberish.amplifyapp.com) if we don't have that information available.

PR #1616, opened 2023-10-23
fix(header): adjust siteName function params

dcshzj:
## Problem

The siteName param was given as an object to the `useGetStagingUrl` hook, which causing the "Open Staging" button to not work when editing pages.

## Solution

**Breaking Changes**

**Bug Fixes**:

- The params input has been fixed and the "Open Staging" button works now.

## Tests

- [x] Smoke tests
    - [ ] Navigate to any site and edit a page
    - [ ] Verify that the "Open Staging" button works and brings you to the correct staging site

## Deploy Notes

*None*


kishore03109 (approved):
haiz this is why we need typed code :sadge:

PR #1618, opened 2023-10-25
feat(editor): use tiptap for new editor

seaerchin:
**Note: about 1.2k line changes are from `package.json` + style changes**
next PR in chain is #1625 which handles the toggling

## Problem
The old editing experience wasn't ideal. This PR adds a new rich text editor together with full functionality in order to improve user experience. The structure of the PR will be laid out below.

## Solution
- This PR relies heavily on `tiptap` (see [here](https://tiptap.dev/))) 
    - as tiptap is _headless_ (without a UI, just provides state), we have to style the editor ourselves, together with the styling _within_ the editor (ie, how the various elements are displayed inside the editor). This is done through the usage of `scss` in `Editor/styles.scss`. The scss there will apply fully to the html tags in the editor (scoped to `.tip-tap`) 
-  `EditPage` has too much responsibilities - it is responsible not just for the editor's internal state but also various formatting/redirection responsibilities. In order to make the page less bulky, the layout stuff has been delegated to a new `EditPageLayout`.
    - This handles page level stuff **that is not directly related to editing.** Examples of these are: handling the save properly, the format of the page (header, footer etc), together with disabling writes if required + redirecting out. 
    - `EditPage` handles the `Editor` together with the `PagePreview`; no significant change here. However, 2 new contexts have been added to `EditPage`. The new contexts don't have additional logic but handles making modals/editors available to deeply nested components. This was done as the `MenuBar` requires the modals/editor instance in order to dispatch certain actions (eg: showing media modal on button click, updating internal state of editor)
- 1 new extension added to the editor. This handles copy pasting of `iframes` from their clipboard into the editor. On a high level, this just uses a regex to check if the text they are pasting starts/ends with `<iframe>` and converts it appropriately.  
    - This also has an associated view in the editor, done to ease over a pain point at present (iframes take up alot of space without presenting useful visual information). The view just shows the `src` property together wtih a label (`iframe`)
- 2 new bubble menus have been added. 
    - The link bubble menu shows when the `link` property is active (ie, the cursor is inside a link); This allows users to change the link via clicking on the button inside the menu.
    - The table bubble menu allows users to do create/delete operations on rows/cols and shows up when the `table` property is active (anywhere inside table).
    - do take note that the `link` bubble menu will display preferentially over the table; should **not** be a problem cos they can jsut click elsewhere to insert 
   
**Breaking Changes**
This PR does **not** do the toggle - this is done in a later PR. For now, this PR is technically backward compat. This is because content saved via this editor will also be loadable by our old editor as the format is html.

**Tests**
- [ ] load the editor 
- [ ] click on the file/image buttons and make sure they all work


src/layouts/components/Editor/LegacyBlockView.tsx
seaerchin, line 1: renamed to `IframeView` in later PR

src/contexts/EditorModalContext.tsx
```diff
@@ -0,0 +1,22 @@
...
+interface EditorModalContextProps {
+  showModal: (modalVariant: "images" | "files") => void
+}
+
+const EditorModalContext = createContext<null | EditorModalContextProps>(null)
+
+export const useEditorModal = (): EditorModalContextProps => {
+  const editorContext = useContext(EditorModalContext)
+  if (!editorContext)
+    throw new Error("useLoginContext must be used within an LoginProvider")
```
seaerchin, line 12: fixed downstream

src/layouts/components/Editor/Iframe.ts
```diff
@@ -0,0 +1,118 @@
...
+declare module "@tiptap/core" {
+  interface Commands<ReturnType> {
+    iframe: {
+      /**
+       * Add an iframe
+       */
+      setIframe: (options: { src: string }) => ReturnType
+    }
+  }
+}
```
seaerchin, line 23: required so that other files can access the command like so: `editor.commands.iframe.setIframe`

src/layouts/components/Editor/LegacyBlockView.tsx
```diff
@@ -0,0 +1,66 @@
+import { Box, Text } from "@chakra-ui/react"
+import {
+  NodeViewContent,
+  NodeViewProps,
+  NodeViewRendererProps,
+  NodeViewWrapper,
+  ReactNodeViewRendererOptions,
+} from "@tiptap/react"
+
+// .react-component {
```
seaerchin, line 10: removed in later PRs

src/layouts/components/Editor/LegacyBlockView.tsx
```diff
@@ -0,0 +1,66 @@
...
+        bgColor="#0d0d0d"
+        textStyle="caption-1"
+        fontWeight="bold"
+        color="#fff"
+        position="absolute"
+        top="0"
+        padding="0.25rem 0.75rem"
+        borderRadius="0 0 0.5rem 0.5rem"
+      >
+        Legacy
```
seaerchin, line 59: updated to `src` in later PRs

src/layouts/components/Editor/MenuItem.tsx
```diff
@@ -0,0 +1,33 @@
...
+    bgColor={isActive && isActive() ? "gray.200" : "transparent"}
+    border="none"
+    h="1.75rem"
+    w="1.75rem"
+    p="0.25rem"
+    aria-label={title || "divider"}
+  >
+    <svg className="remix" height="1.25rem" width="1.25rem">
+      <use xlinkHref={`${remixiconUrl}#ri-${icon}`} />
+    </svg>
```
seaerchin, line 31: updated in later PRs to use `icon` prop of parent, for now this means the icons are huge

src/layouts/components/Editor/styles.scss
seaerchin, line 1: feel free to skip - this will also change down the line + change alr approved by design

src/layouts/EditPage/EditPage.tsx
```diff
@@ -0,0 +1,260 @@
...
+  getPageBody,
+  children,
+}: PropsWithChildren<EditPageLayoutProps>) => {
+  const params = useParams<{ siteName: string }>()
+  const decodedParams = getDecodedParams(params)
+  const {
+    mutateAsync: updatePageHandler,
+    isLoading: isSavingPage,
+  } = useUpdatePageHook(params, {
+    // NOTE: Not deleting this as this is important enough
```
kishore03109, line 77: Assuming this is fixed upsteam?

src/layouts/EditPage/EditPage.tsx
```diff
@@ -0,0 +1,260 @@
...
+    // to leave here so that we avoid regression.
+    // onError: (err) => {
+    //   if (err.response.status === 409) onOverwriteOpen()
+    // },
+  })
+
+  const { siteName } = decodedParams
+
+  const { setRedirectToNotFound } = useRedirectHook()
+  const { data: pageData, isLoading: isLoadingPage } = useGetPageHook(params, {
```
kishore03109, line 87: nit: not used

src/layouts/EditPage/EditPage.tsx
```diff
@@ -0,0 +1,260 @@
...
+        <Footer>
+          <Button
+            onClick={() => {
+              updatePageHandler(({
+                pageData: {
+                  frontMatter: pageData?.content?.frontMatter,
+                  pageBody: getPageBody(),
+                  sha: pageData.sha,
+                },
+              } as unknown) as void)
```
kishore03109, line 127: why do we need to type this to void ah?

src/layouts/EditPage/EditPage.tsx
```diff
@@ -0,0 +1,260 @@
...
+            }}
+            isLoading={isSavingPage}
```
kishore03109, line 129: are we missing a `isLoading={isSavingPage || isLoading}` here?

src/layouts/components/Editor/Editor.tsx
```diff
@@ -0,0 +1,53 @@
...
+          alignItems="center"
+          flex="0 0 auto"
+          flexWrap="wrap"
+          p="0.25rem 0.75rem"
+        >
+          <Text>
+            {editor.storage.characterCount.characters()} characters
+            <br />
+            {editor.storage.characterCount.words()} words
+          </Text>
```
kishore03109, line 48: [Design] Do we feel that the user cares about this? If no, shall we just omit this?

src/layouts/components/Editor/Iframe.ts
```diff
@@ -0,0 +1,118 @@
+import { Node, nodeInputRule, nodePasteRule } from "@tiptap/core"
+import { ReactNodeViewRenderer } from "@tiptap/react"
+import _ from "lodash"
+
+import { LegacyBlockView } from "./LegacyBlockView"
+
+export interface IframeOptions {
+  allowFullscreen: boolean
+  HTMLAttributes: {
+    [key: string]: any
```
kishore03109, line 10: [non-blocking]: type to unknown?

src/layouts/components/Editor/Iframe.ts
kishore03109: rather than redeclaring this here, could we instead add a thin wrapper around this and call this `IsomerTipTapCore`? Now devs have to remember to export from here rather than the default tiptap core that is exposed by the library.

src/layouts/components/Editor/Iframe.ts
```diff
@@ -0,0 +1,118 @@
...
+
+        return true
+      },
+    }
+  },
+
+  addPasteRules() {
+    return [
+      nodePasteRule({
+        find: /<iframe (.*)><\/iframe>/g,
```
kishore03109, line 98: what happens if the user has an iframe? 
do we input the iframe? do we care cases like `<iframe></iframe>` (this will not be parsed cuz there is no spacing bet the `e` and hte closing anchor tag)

src/layouts/components/Editor/Iframe.ts
```diff
@@ -0,0 +1,118 @@
...
+        type: this.type,
+        getAttributes: (match) => {
+          const srcIndex = match.input?.search("src=")
```
kishore03109, line 101: hmm am uncomfortable with this regex parsing :( Are we controlling the inputs here ah? eg if user puts `src = blah` this will fail + now we introduce stricter syntax rules when developing tiptap.

src/layouts/components/Editor/Iframe.ts
```diff
@@ -0,0 +1,118 @@
...
+          if (srcIndex && srcIndex > -1) {
+            const substring = match.input?.substring(srcIndex + 5)
+            const src = _.takeWhile(substring, (char) => char !== '"').join("")
```
kishore03109, line 104: uhm whats happening here ya, could we have a comment to explain this?

src/layouts/components/Editor/LinkBubbleMenu.tsx
```diff
@@ -0,0 +1,93 @@
...
+          <ModalFooter>
+            <Button
+              colorScheme="blue"
+              mr={3}
+              onClick={() => {
+                editor
+                  .chain()
+                  .focus()
+                  // NOTE: Force `https` by default
+                  .setLink({ href: `https://${href}` })
```
kishore03109, line 67: what about relative paths ah?

src/layouts/components/Editor/MenuBar.tsx
```diff
@@ -0,0 +1,190 @@
...
+      bgColor="gray.50"
+      flex="0 0 auto"
+      flexWrap="wrap"
+      p="0.25rem"
+      borderBottom="1px solid"
+      borderColor="base.divider.strong"
+      borderTopRadius="0.25rem"
+    >
+      {items.map((item, index) => (
+        // eslint-disable-next-line react/no-array-index-key
```
kishore03109, line 174: quick win to just use uuid here? we alr have the lib in our fe iirc

src/layouts/components/Editor/MenuItem.tsx
```diff
@@ -0,0 +1,33 @@
...
+interface MenuItemProps {
+  icon?: string
+  title?: string
+  action?: MouseEventHandler<HTMLButtonElement>
+  isActive?: null | (() => boolean)
+}
+
+export default ({ icon, title, action, isActive = null }: MenuItemProps) => (
+  <IconButton
+    // variant="unstyled"
```
kishore03109, line 15: nit: acciendtal comment

src/layouts/components/Editor/MenuItem.tsx
```diff
@@ -0,0 +1,33 @@
...
+    type="button"
+    className="menu-item"
+    _hover={{ bg: "gray.100" }}
+    _active={{ bg: "gray.200" }}
+    onClick={action}
+    title={title}
+    bgColor={isActive && isActive() ? "gray.200" : "transparent"}
```
kishore03109, line 22: could we rename the isActives to represent what each type of `isActive` here refers to?

src/layouts/components/Editor/styles.scss
```diff
@@ -0,0 +1,165 @@
...
+    height: auto;
+    max-width: 100%;
+  }
+
+  hr {
+    margin: 1rem 0;
+  }
+
+  blockquote {
+    border-left: 2px solid rgba(#0d0d0d, 0.1);
```
kishore03109, line 136: nits: use rem + colour pallets

src/styles/index.scss
```diff
@@ -8,6 +8,10 @@ body {
   height: 100%;
 }
 
+div[id="root"] {
```
kishore03109, line 11: May I get the context behind this change?

src/templates/pageComponentsV2/PageBody.jsx
```diff
@@ -1,11 +1,7 @@
...
-
-export const PageBody = ({ chunk }) => (
-  <div className="col is-8 is-offset-1-desktop is-12-touch print-content page-content-body">
-    <div className="content" dangerouslySetInnerHTML={{ __html: chunk }} />
-  </div>
-)
-
-PageBody.propTypes = {
-  chunk: PropTypes.string,
+export const PageBody = ({ chunk }) => {
```
kishore03109, line 1: nit: quick win change to tsx?

kishore03109 (commented):
Lets discuss today to help speed up review ya
Nit: understand that this was a rushed pr, i do think a quick win could be to take screenshots of what this pr achieves as a quick sanity check esp with prs this large, I find myself skimming more than reviewing :sadge:

src/layouts/components/Editor/Iframe.ts
seaerchin: not sure i'm understanding you here. the point of this declaration here is for [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) so that the default editor used by the library has the typing on it.

src/layouts/EditPage/EditPage.tsx
seaerchin: yes

src/layouts/EditPage/EditPage.tsx
seaerchin: the underlying hook is in jsx, which has auto inferred types.

src/layouts/EditPage/EditPage.tsx
seaerchin: not entirely sure here - took it as-is from the existing code. i can add it in but have to be aware that the extra check didn't exist initially

src/layouts/components/Editor/Editor.tsx
seaerchin: i added for feature parity with prod

src/layouts/components/Editor/MenuItem.tsx
seaerchin: fixed!

src/layouts/components/Editor/MenuItem.tsx
seaerchin: no, because we require a function call as otherwise this will be static.

src/layouts/components/Editor/styles.scss
seaerchin: we don't use `rem` for fixed size things (border, border-radius etc). 

these should not be tied to font-size, so a fixed unit is preferable.

seaerchin: see [here](https://www.shecodes.io/athena/28078-when-to-use-ems-rems-and-px-in-css#:~:text=Use%20px%20when%20you%20need,font%20size%20of%20the%20browser.)

src/styles/index.scss
seaerchin: our `root` div (where react injects the app) is not giving 100% height oddly enough, which _might_ have led to overflow on workspace

i think i fixed the overflow issue downstream so i'm ok with removing this

src/templates/pageComponentsV2/PageBody.jsx
seaerchin: leaving as is. don't want to deal with import updates + not that much incremental value (things already work). 

ideally, we make chunk `TrustedHTML` and this allows us to guarantee that all our HTML is sanitised/safe. otherwise, shifting from `jsx -> tsx` here is abit of a hassle for me

src/layouts/components/Editor/Iframe.ts
seaerchin: we never input anything. we just parse the source text (if going from old -> new) and check if there exists `iframe` tags.

if there are iframe tags per your example, they never did anything in the first place ._.

src/layouts/components/Editor/Iframe.ts
seaerchin: it'll just be outputted as the most general matching node, which is `Text` here, so the user doesn't lose info.

i'm not inclined to make this change _now_ - most users copy paste their iframe without editing; this is just to allow them a view of their iframe. this seems like an edge case (user actually edits their iframe, specifically the `src` property) which we can deal with when it happen.

do feel free to add this in/a stricter regex if you want it to happen that way. fwiw, on my end, i think it's good to change but maybe not the best use of time now

src/layouts/components/Editor/Iframe.ts
seaerchin: sure

src/layouts/components/Editor/LinkBubbleMenu.tsx
seaerchin: hmm good catch! for files, the modal itself uses relative links so it's ok. this _would_ be a problem for updating file extensions. perhaps i'll change so that this happens for absolute links, and not relative.

src/layouts/components/Editor/MenuBar.tsx
seaerchin: sure

src/layouts/EditPage/EditPage.tsx
seaerchin: removed

src/layouts/components/Editor/LinkBubbleMenu.tsx
```diff
@@ -0,0 +1,110 @@
...
+          if (
+            (editor.getAttributes("link").href as
+              | string
+              | undefined)?.startsWith("http")
+          ) {
+            onOpen()
+          } else {
+            // Otherwise, show the file modal
+            // and let the user select a file to link to.
+            showModal("files")
```
dcshzj, line 50: What about relative links to other pages of the site?

src/layouts/components/Editor/LinkBubbleMenu.tsx
```diff
@@ -0,0 +1,110 @@
...
+        Change link
+      </button>
+      <Modal isOpen={isOpen} onClose={onClose} size="sm">
+        <ModalOverlay />
+        <ModalContent>
+          <ModalHeader>Update link</ModalHeader>
+          <ModalCloseButton />
+          <ModalBody>
+            <InputGroup size="sm">
+              <InputLeftAddon>https://</InputLeftAddon>
```
dcshzj, line 63: I think similar comment as below, this assumption might be too strong to make that:
1. Users only add external links - they can add relative ones too
2. Users always add links to https - some sites might not have https, or the user might be linking to an email (i.e `mailto:test@example.com`)

src/layouts/components/Editor/MenuBar.tsx
```diff
@@ -0,0 +1,191 @@
...
+      flex="0 0 auto"
+      flexWrap="wrap"
+      p="0.25rem"
+      borderBottom="1px solid"
+      borderColor="base.divider.strong"
+      borderTopRadius="0.25rem"
+    >
+      {items.map((item) => (
+        // eslint-disable-next-line react/no-array-index-key
+        <Fragment key={uuid()}>
```
dcshzj, line 176: Can remove the eslint disable?

src/layouts/components/Editor/TableBubbleMenu.tsx
```diff
@@ -0,0 +1,63 @@
...
+import remixiconUrl from "remixicon/fonts/remixicon.symbol.svg"
+
+import { useEditorContext } from "contexts/EditorContext"
+
+export const TableBubbleMenu = () => {
+  const { editor } = useEditorContext()
+
+  return (
+    <BubbleMenu
+      shouldShow={() => editor.isActive("table") && !editor.isActive("link")}
```
dcshzj, line 12: Should have one central utility function to determine this. Otherwise if we add new bubble menus in the future we got n places to update.

src/layouts/components/Editor/styles.scss
dcshzj: Can I check why we don't use a styles provider?

src/layouts/components/Editor/LinkBubbleMenu.tsx
kishore03109: uhm so this is changed upstream? this one impt ya cuz agencies use relative path quite frequently

kishore03109 (approved):
Please resolve the links comments + ensure that we are not introducing bugs by forcing `https://` in front of _**all**_ links

src/layouts/components/Editor/LinkBubbleMenu.tsx
seaerchin: sure, i can remove this! actually this opens up another can of worms (users linking to insecure sites) but i think we're better off keeping the existing behaviour for now.

src/layouts/components/Editor/MenuBar.tsx
seaerchin: sure

src/layouts/components/Editor/TableBubbleMenu.tsx
seaerchin: this is intended - different bubble menus shuold show based on their surrounding context. unless you mean something like `showBubbleMenu("table")` and `showBubbleMenu("link")`?

dcshzj: Yup, so we only have a central location (i.e. `showBubbleMenu`) to edit when we add more bubble menus in the future.

src/layouts/components/Editor/styles.scss
dcshzj: Not blocking btw so can just ignore

src/layouts/components/Editor/LinkBubbleMenu.tsx
seaerchin: good catch. i'll add this in on another PR (the one where behaviour got changed) to avoid rebase errors

dcshzj (approved):

PR #1623, opened 2023-10-26
fix(editpage): change query key from object to array

dcshzj:
## Problem

The edit page does not invalidate the correct key for some reason.

## Solution

**Breaking Changes**

**Bug Fixes**:

- The edit page query key has been adjusted to put the site name and file name in the query key directly instead of being nested inside the object. I'm not sure why objects do not work, [when the docs claim that they do](https://tanstack.com/query/v4/docs/react/guides/query-keys#query-keys-are-hashed-deterministically), might be an upstream bug.

## Tests

- [x] Smoke tests
    - [ ] Navigate to any site and edit the page.
    - [ ] Save the page, verify that the latest changes will stay on the editor.

## Deploy Notes

*None*

### DIFF
--- a/src/hooks/pageHooks/useGetPageHook.jsx
+++ b/src/hooks/pageHooks/useGetPageHook.jsx
@@ -12,8 +12,10 @@ import { DEFAULT_RETRY_MSG } from "utils"
 export function useGetPageHook(params, queryParams) {
   const { pageService } = useContext(ServicesContext)
   const errorToast = useErrorToast()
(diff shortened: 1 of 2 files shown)


seaerchin (approved):
lgtm with 1 caveat - we need to change all call sites of `PAGE_CONTENT_KEY` to use the array structure. could you just double check that this is done!

dcshzj:
@seaerchin yup, only these two hooks use this query key.

PR #1627, opened 2023-10-27
chore: update select media modal

alexanderleegs:
## Problem

This PR modifies the styling of the Select media modal seen in edit pages, as well as the settings modal of various pages and the settings page (via the insert image modal).

[figma](https://www.figma.com/file/7OtNYkc4s9BkFk05ConEbd/%E2%AD%90-V1.5-Master?type=design&node-id=16627%3A9465&mode=design&t=aSyMe69nzapWNYvm-1)

Resolves IS-687

Before:
Screenshot 2023-10-27 at 9 43 55 AM

After:
Screenshot 2023-10-27 at 9 42 56 AM

Tests:

- [ ] Ensure the insert image/file modals are working properly for
  - [ ] Edit page (via insert image/file)
  - [ ] Page and resource page settings (via insert image/file)
  - [ ] Settings (via image fields)

### DIFF
--- a/src/components/media/MediasSelectModal.jsx
+++ b/src/components/media/MediasSelectModal.jsx
@@ -13,14 +13,21 @@ import {
   ModalFooter,
   ModalBody,
   ModalCloseButton,
+  HStack,
+  BreadcrumbItem,
+  BreadcrumbLink,
 } from "@chakra-ui/react"
-import { Button, Searchbar, Pagination } from "@opengovsg/design-system-react"
+import {
+  Button,
+  Searchbar,
+  Pagination,
+  Breadcrumb,
+} from "@opengovsg/design-system-react"
 import { useState, useEffect } from "react"
 import { useFormContext } from "react-hook-form"
 import { useRouteMatch } from "react-router-dom"
 
 import { FolderCard } from "components/FolderCard"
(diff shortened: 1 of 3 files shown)


src/components/media/MediasSelectModal.jsx
```diff
@@ -52,35 +57,45 @@ const MediasSelectModal = ({
...
+    if (listMediaDirectoriesData)
+      setMediaFolderSubdirectories(listMediaDirectoriesData.directories)
+  }, [listMediaDirectoriesData])
+
+  useEffect(() => {
+    if (listMediaFilesData) {
+      setMediaFolderFiles(listMediaFilesData.files)
+      setTotal(listMediaFilesData.total)
+    }
+  }, [listMediaFilesData])
```
seaerchin, line 98: not very sure why this is required but in alex's absence, will approve for now cos it doesn't impact functionality

seaerchin (approved):

src/components/media/MediasSelectModal.jsx
alexanderleegs: This was blocking the image display until after all images were loaded, since we were defaulting the return initially to `[]`  - moving it into it's own separate useEffect allows us to track the load progress independently

PR #1634, opened 2023-11-01
feat(media): add delete multiple media modal

dcshzj:
## Problem

Closes IS-720.

## Solution

**Breaking Changes**

**Features**:

- Added functionality to delete multiple media at once.

## Before & After Screenshots
Screenshot 2023-11-01 at 17 58 25

## Tests

- [x] Smoke tests
    - [ ] Go to any site and enter the media workspace
    - [ ] Select 1 image, go to the "Edit selected" menu and select "Delete all"
    - [ ] Verify that the delete modal appears and the image can be deleted
    - [ ] Select multiple images, go to the "Edit selected" menu and select "Delete all"
    - [ ] Verify that the delete modal appears and the images can be deleted
    - [ ] Go to an individual image and click on the kebob menu, then click "Delete image"
    - [ ] Verify that the delete modal appears and the image can be deleted
    - [ ] Enter the file workspace
    - [ ] Go to an individual image and click on the kebob menu, then click "Delete file"
    - [ ] Verify that the delete modal appears and the file can be deleted

## Deploy Notes

*None*


dcshzj:
!run chromatic

kishore03109:
>  File sizes now only display the raw number, need to convert to human-friendly format
uhm can we either do this or not put any number ah? number without the units abit confusing! 
also why cant we do this now ah, just math.floor() then keep all to mb? is there a complexity that I am missing here?

src/components/DeleteMediaModal/DeleteMediaModal.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+Default.args = {
+  selectedMedia: [
+    { filePath: "/images/hero-banner.png", size: 1234, sha: "sha1234" },
+  ],
+}
+
+export const MultipleMedia = deleteMediaModalTemplate.bind({})
+MultipleMedia.args = {
+  selectedMedia: [
+    { filePath: "/images/hero-banner.png", size: 1234, sha: "sha1234" },
```
kishore03109, line 55: lets add units please + round to nearest number 2sf?

src/components/DeleteMediaModal/DeleteMediaModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+                  <Text textStyle="caption-2" mt="0.25rem">
+                    {selectedMedia.length} {pluralMediaLabel} will be deleted
+                  </Text>
+                  <TableContainer
+                    mt="0.75rem"
+                    overflowY="scroll"
+                    maxH="14.5rem"
+                  >
+                    <Table variant="simple">
+                      <Tbody borderTop="1px" borderColor="gray.100">
```
kishore03109, line 104: lets use rem?

dcshzj:
> uhm can we either do this or not put any number ah? number without the units abit confusing! also why cant we do this now ah, just math.floor() then keep all to mb? is there a complexity that I am missing here?

Err cos chin did it alr I just waiting for his PR to be merged. That's done in #1633.

src/hooks/mediaHooks/useDeleteMultipleMediaHook.tsx
```diff
@@ -0,0 +1,84 @@
...
+            const { sha, ...rest } = curr
+            mediaService.delete(rest, { sha }).catch((error) => {
+              // We want to continue even if some of the media files fail to delete
+              // because there is no turning back now if we fail (the earlier files
+              // would have been deleted)
+              return error
+            }) as Promise<void>
+          })
+          // This wait is necessary to avoid the repo lock
+          .then(() => new Promise((resolve) => setTimeout(resolve, 500))),
```
kishore03109, line 50: quick qn why cant we just do sequentially? why have this 500 ms wait?

kishore03109: ^ non blocking if this is just done after the return. Else blocking since we might have an edge case here right if the request takes longer than 500?

src/layouts/Media/Media.tsx
```diff
@@ -288,27 +367,32 @@ export const Media = (): JSX.Element => {
                             color="interaction.sub.default"
                             ml="0.5rem"
                           >
-                            {selectedImages.length}
+                            {selectedMedia.length}
                           </Badge>
                         </Menu.Button>
                         <Menu.List>
+                          {/* FIXME: To add back when flow is available
```
kishore03109, line 374: sanity check here, is this blocking push to production?

src/types/media.ts
```diff
@@ -1,4 +1,20 @@
+import { MediaService } from "services"
+
 export interface MultipleMediaParams {
   siteName: string
   mediaSrcs: Set<string>
 }
+
+export interface MediaLabels {
+  articleLabel: "a" | "an"
```
kishore03109, line 9: LMAO

kishore03109 (approved):
While content of the pr look ok, am curious if 
1. this can go in independently to prod
2. can cherry pick chin's fix for the numbers for a quite win here

src/components/DeleteMediaModal/DeleteMediaModal.stories.tsx
dcshzj: Hmm it's supposed to be random data, not supposed to mean anything.

src/components/DeleteMediaModal/DeleteMediaModal.tsx
dcshzj: Wait this needs to be exactly 1 px to match how the other intermediate dividers are defined within the Tbody.

src/hooks/mediaHooks/useDeleteMultipleMediaHook.tsx
dcshzj: This is inside a `.then()` call, so it would happen after the delete request is sent and has complete. This is for the backend to catch up between requests.

src/layouts/Media/Media.tsx
dcshzj: Nope, they just won't see those options but the existing flow still exists where they can move their files individually.

src/components/DeleteMediaModal/DeleteMediaModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+            <ModalBody mt="1rem">
+              <VStack spacing="1.5rem">
+                <Text as="p" textStyle="body-1">
+                  Are you sure you want to delete {selectedMedia.length}{" "}
+                  {pluralMediaLabel}?
+                  <br />
+                  <br />
+                  If you used these {pluralMediaLabel} on any page, site
+                  visitors may see a broken {singularMediaLabel}. This cannot be
```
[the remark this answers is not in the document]
dcshzj: Ah good catch yes, thanks! Updated in d4b771c907e5ee88861f37ef3e5f9cb6cbd70507.

src/hooks/mediaHooks/useDeleteMultipleMediaHook.tsx
kishore03109: we could have used await here tho right?

PR #1647, opened 2023-11-03
fix(protected-route): add missing deps

seaerchin:
## Problem
Previously we didn't have a dep in protected route, which might have led to `siteName` not being in growthbook

## Solution
Add to `useEffect`

### DIFF
--- a/src/routing/ProtectedRoute.tsx
+++ b/src/routing/ProtectedRoute.tsx
@@ -53,7 +53,16 @@ export const ProtectedRoute = ({
       }
       growthbook.setAttributes(gbAttributes)
     }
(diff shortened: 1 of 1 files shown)


src/routing/ProtectedRoute.tsx
```diff
@@ -53,7 +53,16 @@ export const ProtectedRoute = ({
...
-  }, [userId, userType, email, displayedName, contactNumber, currPath])
+  }, [
+    userId,
+    userType,
+    email,
+    displayedName,
+    contactNumber,
+    currPath,
+    growthbook,
+    siteNameFromPath,
```
seaerchin, line 64: onlly these 2

harishv7 (approved):

PR #1652, opened 2023-11-05
feat(media): introduce new move media modal

dcshzj:
**This PR is quite large, please review by commits instead if you prefer**

## Problem

Closes IS-684.

## Solution

**Breaking Changes**

**Features**:

- A new move media modal has been introduced to support moving of multiple images in a single click.
- A new hook has been added to support the moving of multiple images.

**Improvements**:

- The MediaBreadcrumbs component has been generalised, so that the same breadcrumbs logic can be used in more places while providing the flexibility to customise the output.

**Bug Fixes**:

- The delete modal previously did not reset the checkbox properly. This has now been fixed.

## Before & After Screenshots

**BEFORE**:

Screenshot 2023-11-06 at 00 08 35

**AFTER**:

Screenshot 2023-11-06 at 00 08 57

## Tests

- [x] Smoke tests

**Smoke test steps:**
- [ ] Navigate to any site and access the Images workspace
- [ ] Select a single image in the root folder, click on the "Edit selected" menu on the top right and click "Move images to album"
- [ ] Verify that the modal shown loads the images directory and its contents correctly
- [ ] Move the image to another album
- [ ] Verify that the image is correctly moved to the above album
- [ ] Select multiple images in any of the albums (have a mixture of images from different albums), click on the "Edit selected" menu on the top right and click "Move images to album"
- [ ] Move the images to another album
- [ ] Verify that the images are correctly moved to the above album
- [ ] Navigate to the Files workspace
- [ ] Click on the kebob menu and select "Move to"
- [ ] Move the file to another directory
- [ ] Verify that the file is correctly moved to the above directory

## Deploy Notes

*None*


dcshzj:
!run chromatic

src/components/move/MoveMenuItem.tsx
```diff
@@ -43,40 +44,42 @@ export const DirMenuItem = ({
...
-          <Icon
-            id={`moveModal-forwardButton-${name}`}
-            w="24px"
-            h="24px"
-            as={BxChevronRight}
-          />
-        </Flex>
-      </Button>
-    </>
+    <Button
```
dcshzj, line 47: Only indentation changes, removed the unnecessary `<></>`

src/layouts/Media/Media.tsx
```diff
@@ -325,62 +371,56 @@ export const Media = (): JSX.Element => {
...
-                        onClick={() => setSelectedMedia([])}
-                      >
-                        Deselect all
-                      </Button>
-                    </Box>
-                    <Box mt="auto">
-                      <Menu
-                        isStretch={false}
-                        onOpen={() => setIndividualMedia(null)}
+                <Box mt="auto">
```
dcshzj, line 378: Indentation changes, mainly just adjusted the conditions in which the action buttons will appear.

src/components/MoveMediaModal/MoveMediaModal.tsx
```diff
@@ -0,0 +1,326 @@
...
+              <Text textStyle="body-2">
+                {/* Display the file name when only one file is moved */}
+                {selectedMedia.length === 1 && (
+                  <Breadcrumbs
+                    items={[
+                      ...moveTo.split("%2F").map((item) => ({
+                        title: _.upperFirst(decodeURIComponent(item)),
+                      })),
+                      {
+                        title: selectedMedia[0].filePath.split("/").pop() || "",
```
kishore03109, line 265: hey am worried about the maintance of this approach of encoding -> splitting by `%2F` for some operations and not for others :( 
Could we leverage on `path` to help us do the [splitting](https://nodejs.org/api/path.html) + [encoding](https://nodejs.org/api/path.html#pathnormalizepath) instead?

src/components/move/MoveMenuItem.tsx
```diff
@@ -18,22 +16,25 @@ export const FileMenuItem = ({
...
+        <Text textStyle="body-1">{pageFileNameToTitle(name, isResource)}</Text>
+      </HStack>
+    </Box>
   )
 }
 
 export interface DirMenuItemProps {
   name: string
-  id: string
+  id: string | number
```
kishore03109, line 37: hmmmm  wait why got 2 data types ah

src/hooks/moveHooks/useMoveMultipleMediaHook.tsx
```diff
@@ -0,0 +1,111 @@
...
+              items: curr.items,
+            }
+            return moveService.move(params, body).catch((error) => {
+              // We want to continue even if some of the media files fail to
+              // move because there is no turning back now if we fail (the
+              // earlier files would have been moved)
+              return error
+            }) as Promise<void>
+          })
+          .then(() => new Promise((resolve) => setTimeout(resolve, 500))),
```
kishore03109, line 71: hey why do we have a timeout here ah?

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <Icon
+          as={BiChevronRight}
+          color="interaction.links.neutral-default"
+          lineHeight="0.5rem"
+          textStyle={textStyle}
+        />
+      }
+      lineHeight="0.5rem"
+    >
+      {items.map(({ title, url }, idx) => {
```
kishore03109, line 52: chakra might have this inbuilt that we could use! 
https://dev.to/estheragbaje/how-to-truncate-text-using-chakra-ui-4mpj

src/components/MoveMediaModal/MoveMediaModal.tsx
```diff
@@ -0,0 +1,326 @@
...
+    mediaDirectoryName: moveTo,
+  })
+
+  const onModalClose = () => {
+    setMoveTo(mediaType)
+    onClose()
+  }
+
+  useEffect(() => {
+    // Note: This useEffect is needed to reset the state of the modal when it
```
kishore03109, line 96: Hmm given that this entire modal gets re-rendered when isOpen changes, could we just put 
`setMoveTo(mediaType)` without the useEffect instead?

src/components/MoveMediaModal/MoveMediaModal.tsx
dcshzj: Hmm the one you linked is for NodeJS so we can't use it here on the frontend, but I took note about this confusion and made some adjustments in 87efbfff:
1. The `moveTo` state variable will always be dealing with the decoded version (i.e. uses `/` instead of `%2F`) to match how the `filePath` is received from the backend. Only at the final step before sending to the backend does it get encoded.
2. The splitting/joining stuff has been abstracted out and a utility function is used as much as possible.

src/components/move/MoveMenuItem.tsx
dcshzj: Hmm cos the ID can come from the index when doing a `.map()`, which is a `number`.

src/hooks/moveHooks/useMoveMultipleMediaHook.tsx
dcshzj: Added a comment in 87efbfff but it is the same reasoning as the delete multiple hook, which is to allow the backend to catch up. Some optimisations have already been made to move multiple images from the same source folder in a single API call.

src/components/Breadcrumbs/Breadcrumbs.tsx
dcshzj: Hmm yes but unfortunately we need to perform this truncation ourselves as the page you linked is for texts, and it is when there is not enough space to display all the text. However, in this case we are showing the first item and the last 2 items, then compressing anything in-between into a single ellipsis.

src/components/MoveMediaModal/MoveMediaModal.tsx
dcshzj: Hmm no unfortunately not possible, because the modal is rendered immediately on mount. The `isOpen` variable only sets the `display: none` CSS style, but in the background the modal is already loaded.

src/components/move/MoveMenuItem.tsx
```diff
@@ -43,40 +44,42 @@ export const DirMenuItem = ({
...
+      w="100%"
+      variant="clear"
+      colorScheme="slate"
+      color="base.content.strong"
+      borderWidth={0}
+      pl="1rem"
+      pr="1.25rem"
+      py="1.5rem"
+      display="flex"
+      data-cy={id}
```
seaerchin, line 57: stonks

src/hooks/moveHooks/useMoveMultipleMediaHook.tsx
```diff
@@ -0,0 +1,112 @@
...
+            }
+            return moveService.move(params, body).catch((error) => {
+              // We want to continue even if some of the media files fail to
+              // move because there is no turning back now if we fail (the
+              // earlier files would have been moved)
+              return error
+            }) as Promise<void>
+          })
+          // This wait is necessary to avoid the repo lock
+          .then(() => new Promise((resolve) => setTimeout(resolve, 500))),
```
seaerchin, line 72: should it be `finally` just in case we fail early on?

src/layouts/Media/Media.tsx
```diff
@@ -305,14 +351,16 @@ export const Media = (): JSX.Element => {
...
           {/* Header section */}
           <HStack w="100%">
             {/* Page title segment */}
             <VStack spacing="0.5rem" align="left" w="40%">
               <MediaBreadcrumbs />
               <Text as="h4" textStyle="h4" mt="0.75rem">
-                All {mediaType}
+                {mediaType === mediaDirectoryName
+                  ? `All ${mediaType}`
+                  : _.upperFirst(mediaDirectoryName.split("%2F").pop())}
```
seaerchin, line 363: could we do a `.get` as this alters `mediaDirectoryName`

dcshzj: Hmm but we do a `split()` first so that would create a new copy as an array.

src/components/Breadcrumbs/Breadcrumbs.tsx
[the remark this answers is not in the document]
dcshzj: Hmm this component covers the case with the most requirements (i.e. truncate intermediate and underline last child). It is assigned default values to cater to usages where any of these 2 requirements are not needed, so I think breaking it up might be quite difficult (and I think, hopefully, there are no additional requirements needed in the future).

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,108 @@
...
+        // replaced with an ellipsis
+        if (
+          items.length > maxBreadcrumbsLength &&
+          idx > 1 &&
+          idx < items.length - (maxBreadcrumbsLength - 1)
+        ) {
+          return <></>
+        }
+
+        const isEllipsis = items.length > maxBreadcrumbsLength && idx === 1
```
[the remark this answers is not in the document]
dcshzj: Consider this case: We have 6 levels to show but we want to truncate such that only 3 levels are shown. With truncation, a total of 4 items will need to be shown: Item 1 > ... > Item 5 > Item 6.

Creating these 4 items is done in the earlier if-else (number of items exceeded limit of 3, so find item with index 2 and 3 and "drop" them), preserving items with index 0, 1, (length - 2), (length - 1). At this stage, we will have Item 1 > Item 2 > Item 5 > Item 6.

Now, we want to preserve only the first and last 2 items. This `isEllipsis` identifies the "Item 2" (index 1) that we want to replace, so we can change "Item 2" to show "...". This is how we get the truncation we want above.

If the number of items is less than the maximum defined, we do not want to do any truncation, so step 1 (dropping of items) and step 2 (replacement to ellipsis) will not happen.

src/hooks/moveHooks/useMoveMultipleMediaHook.tsx
dcshzj: Hmm we caught any failures earlier, so we will always execute this to completion, hence we are safe to use `.then` to continue here.

src/components/Breadcrumbs/Breadcrumbs.tsx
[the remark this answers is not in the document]
seaerchin: i'm fine with this for now - but the component makes the judgement here (whether to underline/truncate), which leads to bloat. ideally we can make smaller subcomponents that take this logic out (UI only), and use a hook or expose a wrapper for people that require this formatting logic. 

don't sweat it tho, is o kfor now

src/hooks/moveHooks/useMoveMultipleMediaHook.tsx
seaerchin: sgtm

src/components/Breadcrumbs/Breadcrumbs.tsx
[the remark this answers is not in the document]
seaerchin: can, not an issue. this isn't immediately obvious and q confusing (which led to my question) but it can be addressed later ba

seaerchin (approved):
lgtm - seems like no functional blockers. we can address unclear code later on, but it's a win for now

PR #1653, opened 2023-11-06
fix(scss): importing of styles

kishore03109:
## Problem

Text colors are the same for preview + staging sites
Screenshot 2023-11-06 at 4 52 29 PM

Screenshot 2023-11-06 at 4 52 15 PM

is-Secondary change is abit scary, after doing a string search on `.is-secondary` noted that this is used for

`.bp-button`
`.bp-sec`
`input`
`textarea`
`select`
`file`
`help`
`bp-notification`
`navbar`
`bp-hero`

 
nav bar + incols was ok
 the search bar in the staging site looked alright + the files not used in blueprint.scss

For reference, in `blueprint.scss` has the following rules: 

Screenshot 2023-11-10 at 4 28 03 PM

### DIFF
--- a/src/utils/siteColorUtils.js
+++ b/src/utils/siteColorUtils.js
@@ -63,7 +63,7 @@ const createPageStyleSheet = (repoName, primaryColor, secondaryColor) => {
 
   // EditHomepage: hero button - secondary color
   customStyleSheet.insertRule(
-    `.is-secondary { background-color: ${secondaryColor} !important;}`,
+    `.is-secondary { background-color: ${secondaryColor} !important; color: #fff !important;}`,
     0
   )
 


src/utils/siteColorUtils.js
```diff
@@ -63,7 +63,7 @@ const createPageStyleSheet = (repoName, primaryColor, secondaryColor) => {
 
   // EditHomepage: hero button - secondary color
   customStyleSheet.insertRule(
-    `.is-secondary { background-color: ${secondaryColor} !important;}`,
+    `.is-secondary { background-color: ${secondaryColor} !important; color: #fff !important;}`,
```
seaerchin, line 66: can i check where you got this from?

seaerchin (commented):
change is ok, just wanted to make sure of unintended regression

src/utils/siteColorUtils.js
kishore03109: Screenshot 2023-11-09 at 1 06 21 PM
hmm am worried about regression too, got it from this line here. is this sound?

seaerchin: We should check the scss file itself to see the condition. 

This also requires an additional class `bp-button` so the conditions aren't exactly the same

kishore03109: to be clear thats from the live site, the one in our preview is here!
Screenshot 2023-11-10 at 4 20 51 PM

seaerchin (approved):

PR #1654, opened 2023-11-06
feat(images): bulk upload

seaerchin:
## Problem
Right now, we don't have a bulk upload flow for images - this leads to bad UX because users want to be able to upload images in bulk and not 1 by 1.

## Solution
Make bulk upload flow - this is done using a few things, which will be outlined below: 
- we add a new editor that handles state transitions + portion data out to each individual component 
- various sub components that handle the title/content/footer. didn't want to do the content alone because the title/footer might change between states (success vs not) and eventually it seemed like the alternative was to make a function which makes it slightly harder to trace
- add a new hook that works via chaining off the old hook - this is **sequential** in order, and would be **slower** than just uploading everything at 1 shot. the reason this is done is because github doesn't handle mass uploads well, which might lead to unintuitive error cases :/
- add new method to serve to create single media file - done because existing one is indecipherable. 

## Before & After Screenshots
**AFTER**:
**Initial upload**
Screenshot 2023-11-07 at 12 39 32 PM

users will be unable to select invalid file types if they click; if they drag + drop, can have invalid. will still be rejected anyways and displayed as scrollable table. 

if >5 valid, will reject (design ok for now)

**Uploading** 
Screenshot 2023-11-07 at 12 41 38 PM

this is **sequential** - github might fail if multiple concurrent; maybe btr solution is we expose an endpoint for bulk upload **onto ggs** and we manually invoke a push later.

**Success**
Screenshot 2023-11-07 at 12 43 44 PM

**Error**
same as success -> just that msg and icon change

cancelling here will not do anything -> files will still be uploaded.

## Tests

- [ ] Smoke tests


src/hooks/mediaHooks/useCreateMultipleMedia.ts
```diff
@@ -0,0 +1,89 @@
...
+              ...prev,
+              getAxiosErrorMessage(axiosError),
+            ])
+          })
+          .then(() =>
+            queryClient.invalidateQueries([
+              // invalidates media directory
+              LIST_MEDIA_DIRECTORY_FILES_KEY,
+              { siteName, mediaDirectoryName },
+            ])
```
dcshzj, line 80: non-blocking comment: why do we need to invalidate on every upload? Can we invalidate onSuccess instead?

src/components/Attachment/AttachmentFileInfo.tsx
```diff
@@ -60,11 +60,15 @@ export const AttachmentFileInfo = forwardRef<AttachmentFileInfoProps, "div">(
         )}
         <Flex sx={styles.fileInfo}>
           <Icon as={BxImage} />
-          <Text ml="2rem">{file.name}</Text>
+          <Text ml="2rem" textStyle="caption-2" noOfLines={1}>
```
dcshzj, line 63: nit: should add `word-break: break-all` else we get:

Screenshot 2023-11-08 at 15 19 18

src/components/MediaCreationModal/MediaCreationModal.tsx
```diff
@@ -0,0 +1,350 @@
...
+  setFileRejections,
+  isDisabled,
+  onUpload,
+  variant,
+}: MediaDropzoneProps) => {
+  const { onClose } = useModalContext()
+
+  return (
+    <>
+      <ModalHeader>Upload files</ModalHeader>
```
dcshzj, line 72: suggestion(ux, non-blocking): should use `getMediaLabels()` to get the appropriate labels here, since this is shared between images and files.

src/components/MediaCreationModal/MediaCreationModal.tsx
```diff
@@ -0,0 +1,350 @@
...
+              : IMAGE_UPLOAD_ACCEPTED_MIME_TYPES
+          }
+          onChange={(curUploadedFiles, rejections) => {
+            setUploadedFiles([...uploadedFiles, ...curUploadedFiles])
+            setFileRejections([...fileRejections, ...rejections])
+          }}
+          value={uploadedFiles}
+          name=""
+          // NOTE: 5MB - maxSize is in bytes
+          maxSize={5 * 10 ** 6}
```
dcshzj, line 93: nit: we should store the `5` value as a constant somewhere else (seen this being used multiple times)

src/components/MediaCreationModal/MediaCreationModal.tsx
```diff
@@ -0,0 +1,350 @@
...
+}
+
+interface MediaDropzoneProps {
+  fileRejections: FileRejection[]
+  uploadedFiles: File[]
+  setUploadedFiles: (files: File[]) => void
+  setFileRejections: (rejections: FileRejection[]) => void
+  isDisabled?: boolean
+  onUpload: (files: File[]) => Promise<void>
+  variant: UploadVariant
```
dcshzj, line 56: nit: would be good to standardise this to use `mediaType` instead of `variant`

src/hooks/mediaHooks/useCreateMultipleMedia.ts
seaerchin: if people exit halfway, it'll lead to a bug

src/components/Attachment/AttachmentFileInfo.tsx
seaerchin: can but my pathological case on `break-word` led to it overflowing to the side. i'll change it!

src/components/MediaCreationModal/MediaCreationModal.tsx
seaerchin: can, will update

src/components/MediaCreationModal/MediaCreationModal.tsx
seaerchin: oso can

src/components/MediaCreationModal/MediaCreationModal.tsx
seaerchin: will update

src/components/MediaCreationModal/MediaCreationModal.tsx
seaerchin: swapped to `mediaType` for naming, but couldn't find the type available?

dcshzj: Lol shucks that's in my PR #1652. Can dup that type inside `src/types/media.ts`.

src/hooks/mediaHooks/useCreateMultipleMedia.ts
dcshzj: Can, make sense, thanks!

src/components/Attachment/AttachmentFileInfo.tsx
seaerchin: swapped to enforcing `w=100%`, which prevented this case

src/components/MediaCreationModal/MediaCreationModal.tsx
seaerchin: actually i asked our designer for a [copy](https://opengovproducts.slack.com/archives/CK68JNFHR/p1699349295935939?thread_ts=1699331583.799259&cid=CK68JNFHR) that worked across files/images so not inclined to do this tbh @__@

dcshzj (approved):
LGTM

PR #1656, opened 2023-11-07
feat(tiptap): blocks editor

seaerchin:
**NOTE: We will not be doing block editing the way this PR is done, but rather it will be integrated as part of the editor itself. Simple text editing will be done normally but complex blocks will have a border around them with a different flow for changing their contents. --@dcshzj**

## Problem
This PR implements a first pass block editor with draggables

## Solution
- add a custom hook to add the preview editor 
- migrate `Preview` to a `PagePreview` component -> this is because we want to nest the editor inside the preview 
- Add a custom draggable extension -> this provides the drag handle and draggable functionality
- Add a custom context to handle teh various click interaction - this is to transition the states to different views 
- We add in a prop for `editor` -> we choose the prop preferentially over the context's; this is because we need to support multiple editors at present (:sadge:)


src/layouts/EditPage/EditPage.tsx
```diff
@@ -137,64 +147,66 @@ export const EditPage = () => {
...
-                    `<a target="_blank" rel="noopener noreferrer nofollow" href="${selectedMediaPath}">file</a>`
-                  )
-                  .run()
-              else {
-                editor
-                  .chain()
-                  .focus()
-                  .setLink({ href: selectedMediaPath })
-                  .run()
+        <BlocksContextProvider blocks={[DEFAULT_BLOCKS.text]}>
```
seaerchin, line 150: indentation from new context

PR #1657, opened 2023-11-08
fix(badge): flickering issue

kishore03109:
## Problem

Fixed. Design havent gvivne green check yet, will prioritise + make design changes if needed on monday. 

## Solution

shouldnt flicker on refresh (which was solved using portal, using only useMemo did not solve it) 
Screenshot 2023-11-08 at 10 05 52 AM

### DIFF
--- a/src/components/Header/StatusBadge.tsx
+++ b/src/components/Header/StatusBadge.tsx
@@ -3,9 +3,9 @@ import {
   HStack,
   Icon,
   Popover,
+  PopoverBody,
   PopoverContent,
   PopoverTrigger,
-  Portal,
   Text,
 } from "@chakra-ui/react"
 import { Badge } from "@opengovsg/design-system-react"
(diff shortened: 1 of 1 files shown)


src/components/Header/StatusBadge.tsx
```diff
@@ -82,60 +76,70 @@ const StagingPopoverContent = ({ status }: StagingPopoverContentProps) => {
...
+export const StatusBadge = (): JSX.Element => {
+  const { siteName } = useParams<{ siteName: string }>()
+  const { data: getStagingStatusData } = useGetStagingStatus(siteName)
+  const status = getStagingStatusData?.status
+
+  const badge = useMemo(() => {
+    return StatusBadgeComponent(status)
+  }, [status])
+
+  return badge
```
seaerchin, line 144: not important to fix, but this `useMemo` is redundant - the `StatusBadgeComponent` will re-render on props change, which consists only of `status`. similarly, the `useMemo` will recompute on deps change, which is also `status`. 

this is identical to 

```ts
return <StatusBadgeComponent status = {status} />
```

seaerchin (approved):
lgtm, problem might have been because of rerender, which causes `Popover` to rerender too due to it being inside `Box`.

PR #1659, opened 2023-11-09
fix(menubar): remove uuid

seaerchin:
## Problem
The menu bar was rendering twice, which required 2 clicks to active modals. 

Closes IS-742

## Solution
Removed the `uuid`. This is because on state change (clicking etc), the uuid function fires again, which leads to a new key. This, in turn, causes react to re-render the component, leading to the first click being dropped.

### DIFF
--- a/src/layouts/components/Editor/components/MenuBar.tsx
+++ b/src/layouts/components/Editor/components/MenuBar.tsx
@@ -176,7 +176,7 @@ export const MenuBar = ({ editor }: { editor: Editor }) => {
       borderTopRadius="0.25rem"
     >
       {items.map((item) => (
-        <Fragment key={uuid()}>
+        <>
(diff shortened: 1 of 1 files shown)


kishore03109 (approved):
Hmm am a bit concerned about this ya, why was it rendering twice? Strict mode does help us figure out potential bugs, are we missing a bug here?

seaerchin:
> Hmm am a bit concerned about this ya, why was it rendering twice? Strict mode does help us figure out potential bugs, are we missing a bug here?

the same thing happens on `staging`, which has strict mode off. the reason for it rendering twice is given in the write-up. you can read more [here](https://stackoverflow.com/questions/75737838/should-i-use-uuid-to-provide-id-on-the-fly-inside-a-react-component)

PR #1663, opened 2023-11-10
fix(preview): use site colours for headings in preview

dcshzj:
## Problem

The preview does not show the headings properly in the site colours.

Fixes IS-567.

## Solution

**Breaking Changes**

**Bug Fixes**:

- The site colours CSS rule is now properly injected in the site colour utilities. This rule is taken from the custom.scss files that every repo should be using: https://github.com/isomerpages/site-creation-base/blob/b7b29561760891bdfa5ba7f7d5fb78e9725d53a2/misc/custom.scss#L257-L272

## Before & After Screenshots

**BEFORE**:

Screenshot 2023-11-10 at 09 23 55

**AFTER**:

Screenshot 2023-11-10 at 09 24 24

## Tests

- [x] Smoke tests
    - [ ] Navigate to any site with a custom colour set for their secondary colour.
    - [ ] Verify that pages that use headings and quotes now use the secondary colour correctly.
    - [ ] Change the secondary colour of the site
    - [ ] Verify that the same pages using the headings and quotes now use the new secondary colour correctly.

## Deploy Notes

*None*

### DIFF
--- a/src/utils/siteColorUtils.js
+++ b/src/utils/siteColorUtils.js
@@ -126,6 +126,21 @@ const createPageStyleSheet = (repoName, primaryColor, secondaryColor) => {
     `.content h5 { color: ${secondaryColor} !important;}`,
     0
   )
+
+  // EditPage: Blockquotes
+  customStyleSheet.insertRule(
+    `.content blockquote { border-left-color: ${secondaryColor} !important;}`,
+    0
+  )
+  customStyleSheet.insertRule(
+    `.content blockquote > p { color: ${secondaryColor} !important;}`,
+    0
+  )
+  customStyleSheet.insertRule(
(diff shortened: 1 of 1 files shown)


seaerchin (commented):
Code changes are fine, just curious where did we get this stylesheet from?

Also, is it reactive? Ie, if user changes and goes back to the page, would the colour also change?

dcshzj:
> Code changes are fine, just curious where did we get this stylesheet from?

This is from site-creation-base, which we use as a template when creating new sites. Essentially, most of the sites that are CMS-compatible would have came from this repo, so this stylesheet can be used as a reference for the ones used on most of our sites.

> Also, is it reactive? Ie, if user changes and goes back to the page, would the colour also change?

Yup, the styles are injected when the user loads the page.

alexanderleegs (approved):
lgtm!

PR #1669, opened 2023-11-10
fix(editor): refactor to use `tooltip` 

seaerchin:
## Problem
Previously, we were relying on the `title` prop to show the tooltip (this is HTML spec). However, this led to a significant delay before the tooltip showed.

Closes IS-743

## Solution
Change to use chakra tooltip. Design system tooltip **does not work** (see [here](https://github.com/opengovsg/design-system/issues/546))

## Test
- [ ] go to new editor
- [ ] hover over tooltip
- [ ] should appear after 0.5s


alexanderleegs (approved):
lgtm

PR #1673, opened 2023-11-11
Fix/buildStatusBadgeAPI

kishore03109:
## Problem

Forgot to remove one last place where api calls were made. Functionally, api to be should only be made iif whitelisted.

Verified that the only other place that `useGetStagingStatus` gets called is in the component `StatusBadge` which only gets rendered conditionally based on feature flag
Screenshot 2023-11-11 at 4 42 18 PM

Storybooks no longer worked post adding feature flag to fe, modified the code to allow for this.
Screenshot 2023-11-11 at 4 36 56 PM

### DIFF
--- a/src/components/Header.jsx
+++ b/src/components/Header.jsx
@@ -133,7 +133,13 @@ const Header = ({
           </Flex>
         ) : null}
         <HStack flex={1} justifyContent="flex-end">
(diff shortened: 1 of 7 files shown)


kishore03109:
!run chromatic

harishv7 (approved):

PR #1674, opened 2023-11-14
feat/addReportingBUtton

kishore03109:
Screenshot 2023-11-14 at 2 08 50 PM

Screenshot 2023-11-14 at 2 08 28 PM

### DIFF
--- a/src/components/Header/AllSitesHeader.tsx
+++ b/src/components/Header/AllSitesHeader.tsx
@@ -10,7 +10,7 @@ import {
 
 import { AvatarMenu } from "components/Header/AvatarMenu"
 
-import { ISOMER_GUIDE_LINK } from "constants/config"
+import { ISOMER_GUIDE_LINK, ISOMER_REPORT_ISSUE_LINK } from "constants/config"
 
 import { useLoginContext } from "contexts/LoginContext"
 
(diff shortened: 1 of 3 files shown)


alexanderleegs (commented):
Just checking - is the button only for all sites and dashboard pages? It doesn't exist on any of the workspace pages and beyond currently

harishv7 (approved):

kishore03109:
> Just checking - is the button only for all sites and dashboard pages? It doesn't exist on any of the workspace pages and beyond currently

@sehyunidaaa please! Inclined to still merge current state in first, then add if needed

sehyunidaaa:
@kishore03109 @harishv7 

Just talked to @shazlithebestie about this, previously wasn't sure how prominent he wanted the option to be! After talking, I think the option can be more prominent.

I think we can have a floating button (see a working design [here](https://www.figma.com/file/owgl8SFMKQd2jrCJMVrukR/CMS-%7C-Q42023?type=design&node-id=298%3A8465&mode=design&t=ZgQvubQD0aCVZcrt-1)) so it can be always visible across 1) Dashboard 2) Workspace and 3) Page editing. 

One risk is that people spam the forms because it's a little more accessible now; Shaz is okay with this risk for now because the reporting workflow will be more robust soon! Second risk is discoverability, which we can minimise by exploring the other design options.

Let me know whether we want to tackle this soon or if we want to consider a simpler approach in the short run.

PR #1677, opened 2023-11-15
Fix/is724-apply-styles

kishore03109:
Headings + table borders now match 
Screenshot 2023-11-16 at 8 57 00 AM

opting to show the vertical lines cuz I feel its easier for our user? lmk if we dont agree on this 
Screenshot 2023-11-16 at 8 57 41 AM

### DIFF
--- a/src/layouts/EditPage/EditPage.tsx
+++ b/src/layouts/EditPage/EditPage.tsx
@@ -70,7 +70,7 @@ export const EditPage = () => {
         pluginKey: "tableBubble",
       }),
       Table.configure({
-        resizable: true,
+        resizable: false,
       }),
       TableRow,
       TableHeader,
--- a/src/styles/isomer-cms/pages/Editor.module.scss
+++ b/src/styles/isomer-cms/pages/Editor.module.scss
@@ -47,42 +47,42 @@
   h1 {
     font-size: 3.375rem;
     line-height: 4.5rem;
-    font-weight: 400;
+    font-weight: 400 !important;
     letter-spacing: -0.022em;
   }
 
   h2 {
     font-size: 2.75rem;
     line-height: 4rem;
-    font-weight: 400;
+    font-weight: 400 !important;
     letter-spacing: -0.022em;
   }
 
   h3 {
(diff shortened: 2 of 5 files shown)


src/styles/isomer-cms/pages/Editor.module.scss
```diff
@@ -47,42 +47,42 @@
   h1 {
     font-size: 3.375rem;
     line-height: 4.5rem;
-    font-weight: 400;
+    font-weight: 400 !important;
```
alexanderleegs, line 50: Hmm these !importants are because of the values in `base.scss` right? I think we can actually remove the h1-h6 styles there - I looked through the template code, and couldn't find reference to that particular weight, and it doesn't seem to be used outside the editor currently. 

Non-blocking comment - if this turns out to break stuff i'm fine with reverting to `!important`, I just think we should try to avoid this as far as possible

alexanderleegs (approved):
lgtm, non-blocking comments! Can merge once cleared by design

src/styles/isomer-cms/pages/Editor.module.scss
harishv7: Yes is good practice to avoid this. Can use specificity of CSS to help here?

src/layouts/EditPage/EditPage.tsx
```diff
@@ -70,7 +70,7 @@ export const EditPage = () => {
         pluginKey: "tableBubble",
       }),
       Table.configure({
-        resizable: true,
+        resizable: false,
```
dcshzj, line 73: for clarification: why do we disable resizing of the table? Seems like we are using HTML now when saving the page, so users should be able to adjust the size of their tables to fit their requirements.

kishore03109: hey good qn here, the users were confused about the resizability (doest translate to the actual preview)

dcshzj: Makes sense yeah ok, thanks!

PR #1685, opened 2023-11-21
IS-782-Editor-File-text-always-comes-out-as-file-regardless-of-text

kishore03109:
## Problem

IS-713 + is-782 + is-783

https://github.com/isomerpages/isomercms-frontend/assets/42832651/9b42566c-ce36-4478-b9b3-5f98917fd980

### DIFF
--- a/src/layouts/components/Editor/extensions/Image/Image.tsx
+++ b/src/layouts/components/Editor/extensions/Image/Image.tsx
@@ -4,14 +4,6 @@ import { ReactNodeViewRenderer } from "@tiptap/react"
 import { ImageView } from "./ImageView"
 
 export const IsomerImage = Image.extend({
-  parseHTML() {
-    return [
-      {
-        tag: "img",
-      },
-    ]
-  },
-
   renderHTML({ HTMLAttributes }) {
     return [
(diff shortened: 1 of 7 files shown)


dcshzj:
Logic-wise I think is okay, only issue is that text placed after the image does not render properly. Probably need to wrap the image inside `<div>` tags so that it does not interfere with the other HTML.

src/layouts/components/Editor/extensions/Image/Image.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { ImageView } from "./ImageView"
+
+export const IsomerImage = Image.extend({
+  parseHTML() {
+    return [
+      {
+        tag: "img",
+      },
+    ]
+  },
```
dcshzj, line 13: question: I think this would not be needed now, since we are not parsing for our own wrapper.

kishore03109: 8524e38b0ae96562f801ffbbecd9f318bb7468e4
forgot we overloading

dcshzj (approved):

PR #1688, opened 2023-11-21
feat(tiptap): add Instagram embed proper node handler

dcshzj:
## Problem

The Instagram embed is done in a hacky manner whereby we replace it with our own custom iframe, which we cannot guarantee to support.

## Solution

**Breaking Changes**

**Features**:

- The Instagram embed is now a full Tiptap Node.

## Tests

- [x] Smoke tests
    - [ ] Navigate to a site that has the Tiptap editor enabled and edit a page with the new Tiptap editor.
    - [ ] Insert an Instagram embed (both captioned and non-captioned) using the embed modal.
    - [ ] Verify that the Instagram post can be inserted correctly and loads in the final staging site.

## Deploy Notes

*None*

### DIFF
--- a/src/components/EditorEmbedModal/EditorEmbedModal.tsx
+++ b/src/components/EditorEmbedModal/EditorEmbedModal.tsx
@@ -18,7 +18,6 @@ import {
   Link,
   ModalCloseButton,
 } from "@opengovsg/design-system-react"
-import * as Cheerio from "cheerio"
 import { useEffect } from "react"
 import { FormProvider, useForm } from "react-hook-form"
 import * as Yup from "yup"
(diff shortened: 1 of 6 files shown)


kishore03109 (approved):
Hmm this is not a blocking comment since this is functional + serves the problem that we currently have in production line. But, I think having these sort of defined in line make it 
1) fragile, since we depend on the immutability of the tree structure of the insta embed for this to work
2) counter-intuitive, what we are doing here is dropping all the nodes and then recreating them on the fly to the preview and on save. I fear that this would trip other devs up


Proposed solution: 
Create a node views for all the elements here (briefly looking seems like divs + svgs)+ code out the attributes that we want to preserve. 
This would also help with devs/ops debugging, in case we need to change stuff directly in gh/efs

dcshzj:
> fragile, since we depend on the immutability of the tree structure of the insta embed for this to work

This part isn't entirely true. We only depend on the main `<blockquote>` with the presence of the `data-instgrm-permalink` attribute to determine if it is an Instagram embed. Hence, the structure inside how we render the Instagram embed is free to be updated, as long as the two attributes above are preserved.

If the Instagram-embed code changes in the future, then yes, we will need to update the way we parse, but I am going to gamble that our parse rule is general enough to handle future cases.

> counter-intuitive, what we are doing here is dropping all the nodes and then recreating them on the fly to the preview and on save. I fear that this would trip other devs up

I think yes this is true, but only if there were significant differences between the original embed code and the one that we are generating, which for most cases shouldn't be the case (unless users are actually modifying the embed code itself, which also gives us possibly an added protection in that we do not support these customisations).

PR #1693, opened 2023-11-22
fix: submitting state for react-hook-form

alexanderleegs:
## Problem

Resolves IS-777.

This PR fixes an issue introduced by a change to the behaviour of `react-hook-form` - it no longer returns `isSubmitting` to false when an error is thrown (see [link](https://github.com/orgs/react-hook-form/discussions/9879)). As we make use of react-query, which handles promises rejection (and requires promises to throw on failure), we need to add an additional catch statement for each of the places where we make use of `isSubmitting` to determine whether the submit button is disabled. Note that error handling is already tackled in each of these places, hence the catch without doing anything else - this is purely to ensure that react-hook-form returns to the correct state.

Affected components:

- [ ] Email Login (both send otp and verify otp)
- [ ] Verification of contact number
- [ ] Comment submission

### DIFF
--- a/src/layouts/ReviewRequest/components/Comments/SendCommentForm.tsx
+++ b/src/layouts/ReviewRequest/components/Comments/SendCommentForm.tsx
@@ -39,17 +39,13 @@ export const SendCommentForm = ({
   const {
     mutateAsync: updateNotifications,
     error: updateNotificationsError,
+    isLoading: isUpddatingComments,
   } = useUpdateComments()
 
   const queryClient = useQueryClient()
 
   const handleUpdateNotifications = async ({ comment }: CommentFormProps) => {
-    try {
-      await updateNotifications({ siteName, requestId, message: comment })
-    } catch (e) {
-      // Needed for react-hook-form to reset isSubmitting
-      return
-    }
+    await updateNotifications({ siteName, requestId, message: comment })
     resetField("comment")
     queryClient.invalidateQueries([COMMENTS_KEY, siteName])
   }
@@ -67,7 +63,7 @@ export const SendCommentForm = ({
     <form onSubmit={handleSubmit(handleUpdateNotifications)}>
       <FormControl
         isInvalid={!!formState.errors.comment}
-        isReadOnly={formState.isSubmitting}
+        isReadOnly={isUpddatingComments}
         pb="1.5rem"
       >
         <HStack>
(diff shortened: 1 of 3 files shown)


src/layouts/ReviewRequest/components/Comments/SendCommentForm.tsx
```diff
@@ -44,7 +44,12 @@ export const SendCommentForm = ({
   const queryClient = useQueryClient()
 
   const handleUpdateNotifications = async ({ comment }: CommentFormProps) => {
-    await updateNotifications({ siteName, requestId, message: comment })
+    try {
+      await updateNotifications({ siteName, requestId, message: comment })
```
dcshzj, line 48: The `updateComments` hook is a mutation hook, would it be better in this case if we leveraged on react-query and update this section to use their `onSuccess` and `onError` callback functions?

dcshzj (approved):
Non-blocking comment, but I think we should move the otp functions over to react-query in the future. I believe they are the remaining legacy functions that still use axios directly.

src/layouts/ReviewRequest/components/Comments/SendCommentForm.tsx
alexanderleegs: Doesn't work unfortunately! `onError` still results in a thrown error, which messes with the new implementation of react-hook-form - the field straight up won't work anymore until it's in a no longer submitting state. Even if we use the error states to determine whether the button is useable, the user can't go back to change the field because the "unhandled error" is blocking further changes to the fields. So the only solution is to have these extra try catch blocks which are just there to consume the error

alexanderleegs:
> Non-blocking comment, but I think we should move the otp functions over to react-query in the future. I believe they are the remaining legacy functions that still use axios directly.

Wait sorry which methods? I think all of them are already on react-query! (which is why this bug exists)

src/layouts/ReviewRequest/components/Comments/SendCommentForm.tsx
alexanderleegs: Wait actually you're right - i thought it was being blocked by rhf - it's actually blocked by `isReadOnly`. Removed `isSubmitting` from the comments form! The others i've opted to leave in because they require prop drilling if we want to make use of `isLoading` from react-query instead

PR #1694, opened 2023-11-22
fix(images): ability to change img

kishore03109:
## Tests
- [ ] Go to a repo and add an image to a page that has rte enabled
- [ ] Hover over the image + notice the change image button 
- [ ] Click on the button + select another image
- [ ] Save the page, notice that the image is now saved with the new source


alexanderleegs (approved):
lgtm

PR #1695, opened 2023-11-22
Fix/is 778 creating media in subfolder in select modal

alexanderleegs:
## Problem

This PR fixes an issue with uploading images within subfolders in the select modal, which would upload the result to the root.  This issue was caused by the create modal taking url params instead of the params which were passed to the modal - it has been reconfigured to make use of the passed params.

It also refactors `getMediaDirectoryName` into typescript, and fixes an issue with modal closing after uploading an image on the media tab.

Also resolves IS-765 (makes the select modal taller).

## Tests

Check that uploading functionality works for
- [ ] Media tab, root
- [ ] Media tab, nested directory
- [ ] Select media modal (edit page/page settings), root
- [ ] Select media modal (edit page/page settings), nested directory

### DIFF
--- a/src/components/DirectorySettingsModal/DirectorySettingsModal.jsx
+++ b/src/components/DirectorySettingsModal/DirectorySettingsModal.jsx
@@ -14,13 +14,10 @@ import { LoadingButton } from "components/LoadingButton"
 import elementStyles from "styles/isomer-cms/Elements.module.scss"
 
 import { getDirectorySettingsType } from "utils/directoryUtils"
+import { getMediaDirectoryName } from "utils/media"
 import { isWriteActionsDisabled } from "utils/reviewRequests"
 
-import {
-  deslugifyDirectory,
-  getLastItemType,
-  getMediaDirectoryName,
-} from "utils"
+import { deslugifyDirectory, getLastItemType } from "utils"
 
 import { DirectorySettingsSchema } from "./DirectorySettingsSchema"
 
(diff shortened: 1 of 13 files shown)


dcshzj (approved):
Approving for now, but I still think this part of media is quite jank. Couple of reasons:

1. We should eventually move away from using routes to serve modals and just show them directly on the main page. Using routes messes with the browser history a little bit, which is not ideal when people click the back button.
2. We pass information around using the route params instead of directly using props.
3. For MediaSelectModal, the 3 modals are treated as separate modals and there is no information sharing between them.
4. All these stuff is in JSX and does not have type-safety.

I think it is not super important to fix now but this is probably a tech debt that we will need to pay down in the future.

alexanderleegs:
> Approving for now, but I still think this part of media is quite jank. Couple of reasons:
> 
> 1. We should eventually move away from using routes to serve modals and just show them directly on the main page. Using routes messes with the browser history a little bit, which is not ideal when people click the back button.
> 2. We pass information around using the route params instead of directly using props.
> 3. For MediaSelectModal, the 3 modals are treated as separate modals and there is no information sharing between them.
> 4. All these stuff is in JSX and does not have type-safety.
> 
> I think it is not super important to fix now but this is probably a tech debt that we will need to pay down in the future.

Hmm i think the only route param we're using with this modal currently is the siteName, which we can be reasonably sure exists! `MediaSelectModal` is supposed to pass around params, but it just wasn't being used prior to this change. Agreed on the JSX -> TSX issue though!

PR #1698, opened 2023-11-23
Fix/rm-.NpmRc

kishore03109:
## Problem

our tiptap editor uses a private registory, and involves committing a token to .npmrc

add .npmrc into gitignore + rm cache


alexanderleegs (approved):

kishore03109:
Closing, didnt expect the removal of .npmrc to break build process. Will find alternate solution

PR #1701, opened 2023-11-24
feat(tiptap): add card grid block

dcshzj:
## Problem

Card grid block

## Solution

**Breaking Changes**

**Features**:

- The card grid block can now be recognised by the Tiptap editor.

**Note: This PR is part of a stack of changes, refer to #1707 for the complete write-up:**
1. #1701 **<--- YOU ARE HERE**
2. #1704 
3. #1705 
4. #1706
5. #1707

## Before & After Screenshots

Screenshot 2023-11-24 at 10 33 11

## Tests

- [x] Smoke tests
    - [ ] To be updated

## Deploy Notes

*None*


src/layouts/EditPage/EditPage.tsx
```diff
@@ -73,12 +81,28 @@ export const EditPage = () => {
       TaskItem,
       CharacterCount,
       IsomerImage,
-      Link.configure({ openOnClick: false, protocols: ["mailto"] }),
+      Link.extend({
+        priority: 100,
```
kishore03109, line 85: is the change in priority req even after the parseHTML only does anything other than `.isomer-card`?

src/layouts/components/Editor/extensions/Cards/Cards.ts
```diff
@@ -0,0 +1,420 @@
...
+  const { nodes } = schema
+
+  const cards = Array(numberOfCards).fill(
+    createCard({
+      schema,
+      title: "This is a title for your card",
+      description: "This is body text for your card. Describe your card.",
+      footer: "This is a link for your card",
+      link: "https://www.isomer.gov.sg",
+      image: "https://placehold.co/600x400",
```
kishore03109, line 98: hmm shall we import from local? prefer not to have an unknown dep from a third party app

kishore03109 (approved):

src/layouts/components/Editor/extensions/Cards/CardsView.tsx
```diff
@@ -0,0 +1,32 @@
...
+            backgroundColor="#055AFF"
+            textColor="white"
+            p="0.25rem"
+          >
+            <Text textStyle="caption-1">Cards grid</Text>
+          </Box>
+        )}
+        <Box
+          outline={selected ? "2px solid #055AFF" : undefined}
+          py="0.75rem"
```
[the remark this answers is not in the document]
dcshzj: Hmm in this case should still use rem I think? The padding can scale based on the font size of the browser.

src/layouts/components/Editor/extensions/Cards/Cards.ts
dcshzj: For now I keep it first? Agree with you that we should use our own placeholder image but I need to check how it would affect the rendering since technically when they save the page and it gets built on staging, the image is not on the site for the card to reference to.

PR #1706, opened 2023-11-27
feat(cards): support undo/redo of cards editing

dcshzj:
## Problem

Card grid block

## Solution

**Breaking Changes**

**Features**:

- Undo/redo is supported with all editing of the card grid block treated as a single transaction. Closing the editor is without saving will reset the editor state correctly.

## Tests

- [ ] Smoke tests

## Deploy Notes

*None*

### DIFF
--- a/src/components/EditorCardsDrawer/EditorCardsDrawer.tsx
+++ b/src/components/EditorCardsDrawer/EditorCardsDrawer.tsx
@@ -5,6 +5,7 @@ import {
   HStack,
   Icon,
   Text,
+  useDisclosure,
   VStack,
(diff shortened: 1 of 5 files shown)


seaerchin (approved):
lgtm - can i trouble you to add some comments (JSDoc) to the command, specifically for teh part where we don't add to the history? 

this is because this command seems have a few implicit requirements to be used

dcshzj:
@seaerchin going to do that in the final PR of the stack, here: https://github.com/isomerpages/isomercms-frontend/pull/1707/commits/87d9f4ed692001fe383be025340f939958e5378c

PR #1705, opened 2023-11-27
feat(cards): allow editing of cards within drawer

dcshzj:
## Problem

Card grid block

## Solution

**Breaking Changes**

- [ ] Yes - this PR contains breaking changes
- [x] No - this PR is backwards compatible with ALL of the following feature flags in this [doc](https://www.notion.so/opengov/Existing-feature-flags-518ad2cdc325420893a105e88c432be5)

**Features**:

- Extracting the contents of the editor selection, making changes to the editor content and updating the preview when editing is now supported.
- Cards editing within the drawer is validated using react-hook-form
- Added support for nested card look and draggable cards.

**Note: This PR is part of a stack of changes, refer to #1707 for the complete write-up:**
1. #1701 
2. #1704 
3. #1705 **<--- YOU ARE HERE**
4. #1706
5. #1707

## Tests

- [ ] Unit tests (using `npm run tests`)
- [ ] e2e tests (comment on this PR with the text `!run e2e`)
- [ ] Smoke tests

## Deploy Notes

*None*

PR #1707, opened 2023-11-27
feat(cards): render preview the same way as template

dcshzj:
## Problem

Card grid block.

## Solution

**Breaking Changes**

**Features**:

- Render the card grid block the same way as how it is rendered on the template.

**Note: This PR is part of a stack of changes:**
1. #1701 
2. #1704 
3. #1705 
4. #1706
5. #1707 **<--- YOU ARE HERE**

In summary, these were the changes that were made to support card grid block editing:
1. The Isomer template has been updated to support styling of the card grid block. This was done in https://github.com/isomerpages/isomerpages-template/pull/360.
2. The card grid block is defined inside Tiptap so that we know how to parse and render the HTML for this block. This was done in #1701.
    1. This also included rendering the block within the editor itself and showing a blue border when the block is selected/active.
3. To support editing the card grid block itself, a new mini drawer was introduced (similar to Chakra's Drawer component but without the overlay). This was done in #1704.
    1. To support future complex blocks that may have their own drawers, a context provider was introduced.
    2. When any drawer is opened, the main Save button will be disabled.
4. The cards can now be edited within the mini drawer. This was done in #1705.
    1. Part 1 of this PR included extracting out the editor selection and parsing that content such that we know what cards are present inside the card grid block, and to also break down each card into its individual components. Support was also added to update the contents of the card grid block so that the preview is updated on each user's change (plus the changes can be saved).
    2. Part 2 of this PR included adding validation to each of the card's components using react-hook-forms.
    3. Part 3 of this PR included supporting drag-and-drop and accordion block, with most of the functionality reused from EditHomepage.
6. Editing of the card grid block is seen as a single transaction. This is done in #1706.
    1. To be clear, when the user edits inside the mini drawer, multiple transactions are still created. This is the default behaviour of Tiptap.
    2. However, when the user is inside the mini drawer, we disable saving to the history until the user saves. This is to prevent the user from entering an invalid state, closing the mini drawer and using the undo function to enter into the invalid state.
7. The preview renders the card grid block the same way as the template. This is done here in #1707.

## Tests

- [ ] Smoke tests

## Deploy Notes

*None*


src/layouts/components/Editor/extensions/Cards/Cards.ts
```diff
@@ -154,9 +154,13 @@ export const IsomerCards = Node.create<CardOptions>({
         if (dispatch) {
           const offset = tr.selection.anchor + 1
 
-          tr.replaceSelectionWith(node)
+          tr.insert(tr.selection.anchor, node)
```
kishore03109, line 157: May I get context behind this change?

src/styles/isomer-template.scss
```diff
@@ -1,5 +1,10 @@
 @charset "UTF-8";
 
+// This import is needed as we want the raw, un-scoped CSS to render the preview
```
kishore03109, line 3: Hmmm why are we not scoping it to just be in the editormodule.scss ah?

src/layouts/components/Editor/extensions/Cards/Cards.ts
dcshzj: The thinking is that when users click on add card grid, they don't expect their current contents to be removed. The `replaceSelectionWith` function does a removal of the existing selection, which would be potentially unexpected for the user.

Simply put, adding content is not the same as replacing content, so this change is to respect that mental model.

src/components/EditorCardsDrawer/EditorCardItem.tsx
```diff
@@ -29,145 +43,196 @@ export const EditorCardItem = ({
...
+        <VStack w="100%" spacing="1rem" p="0.5rem">
+          {/* Card image */}
+          {methods.watch("isDisplayImage") && (
+            <>
+              <Box w="100%">
+                <FormContext
+                  hasError={!!errors.cards?.[index]?.image}
+                  onFieldChange={(e) => {
+                    methods.setValue(`cards.${index}.image`, e.target.value)
+                    methods.trigger()
```
seaerchin, line 114: could i get some context on why we are manually trigger validation here? my understanding is that if our mode is `onTouch`, this shouldn't be required?

src/styles/isomer-template.scss
dcshzj: Good point! Didn't realise that we could scope it down from there, done in 5f8e7d6c53bee3a8881aae522f4505a89738fca2.

src/layouts/components/Editor/extensions/Cards/Cards.ts
```diff
@@ -154,9 +154,13 @@ export const IsomerCards = Node.create<CardOptions>({
...
             .scrollIntoView()
-            .setSelection(Selection.near(tr.doc.resolve(offset)))
+            .setSelection(
+              Selection.near(
+                tr.doc.resolve(tr.doc.resolve(offset).parentOffset)
+              )
+            )
```
seaerchin, line 163: could i get some context on why we need the parent offset? my understanding is that this is because our cards are nested components right?

src/styles/isomer-template/components/_cards.scss
```diff
@@ -0,0 +1,98 @@
+@use "sass:map";
```
seaerchin, line 1: casual reminder to duplicate over to template as the styles need to be replicated

seaerchin: skimming, didn't have issues based off what i saw. will leave design to review

src/templates/pageComponentsV2/PageBody.jsx
```diff
@@ -1,6 +1,6 @@
 export const PageBody = ({ chunk }) => {
   return (
-    <div className="col is-8 is-offset-1-desktop is-12-touch print-content page-content-body">
+    <div className="col is-offset-1-desktop is-12-touch print-content page-content-body">
```
seaerchin, line 3: assuming the deletion of `is-8` here is also replicated onto templates? also, could you check if this leads to teh side nav cutting into content?

seaerchin (approved):

seaerchin:
noting down that there's a bug for dragging child cards in teh drawer

src/templates/pageComponentsV2/PageBody.jsx
dcshzj: Hmm no, this is not a replication of the template. The size of the preview is too small with this `is-8` class, because the browser uses the entire screen width in its calculation instead of only the width of the preview. I opted to remove this so that the preview does not look distorted.

src/styles/isomer-template/components/_cards.scss
dcshzj: Yup, design reviewed the template earlier.

src/layouts/components/Editor/extensions/Cards/Cards.ts
dcshzj: For some weird reason the selection will use one of the child elements (the image for one of the cards iirc) and not the card grid block, so opted to do this instead to ensure that we are selecting the card grid.

src/components/EditorCardsDrawer/EditorCardItem.tsx
dcshzj: Hmm cos we are setting the values using `setValue`, which for some reason doesn't trigger a validation.

PR #1714, opened 2023-11-29
Feat/select media modal for external media

alexanderleegs:
## Problem

Resolves IS-787.

This PR adds back functionality for inserting images from external sources. Previously, users were able to manually modify the links for images to specify an external image, but with the introduction of the tiptap editor, this option was no longer available. This PR revamps the existing media insertion modal, and adds an additional tab to allow for external images within an editpage.

Screenshot 2023-11-29 at 1 10 35 PM

The other existing media select modals have also been revamped to use the new selection view, but without the ability to add external media.

Screenshot 2023-11-29 at 1 11 24 PM
Screenshot 2023-11-29 at 1 11 51 PM

## Tests

- [ ] Functionality for existing inserts should continue to work normally
  - [x]  Adding images to page previews
  - [x]  Adding files to resource page
  - [x]  Adding image to settings fields
  - [x]  Adding internal images in edit page
  - [x]  Adding internal files in edit page
- [x] Functionality for external images
  - [x] Images should be able to be inserted into the editor, and modified in the tiptap rte
  - [x] Images which are too large (e.g. [link](https://images.unsplash.com/photo-1682687221323-6ce2dbc803ab?q=80&w=3174&auto=format&fit=crop&ixlib=rb-4.0.3&ixid=M3wxMjA3fDF8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8fA%3D%3D)) should throw an error message and prevent modal submission
  - [x] Gibberish urls which do not link to an image should prevent modal submission

### DIFF
--- a/src/components/media/MediasSelectModal.jsx
+++ b/src/components/media/MediasSelectModal.jsx
@@ -277,7 +277,7 @@ const MediasSelectModal = ({
           <div
             className={`${mediaStyles.mediaCards} justify-content-center pt-3 pl-2`}
           >
-            <Flex w="100%" justifyContent="space-between" pb="1.25rem">
+            <Flex w="100%" justifyContent="space-between" pb="1.25rem" pt="1px">
               {/* Search medias */}
               <Box>
                 <Searchbar


src/components/pages/EditorModals.jsx
```diff
@@ -59,6 +59,27 @@ const EditorModals = ({
...
+      )
+    if (mediaType === "images") {
+      cm.replaceSelection(
+        `![${altText}](${selectedMediaPath.replaceAll(" ", "%20")})`
+      )
+    }
+    // set state so that rerender is triggered and image is shown
+    onSave(simpleMde.codemirror.getValue())
+    onClose()
+  }
```
seaerchin, line 81: nit: this doesn't appear to be any different from `onMediaSave` defined above?

src/layouts/EditPage/EditPage.tsx
```diff
@@ -161,6 +159,26 @@ export const EditPage = () => {
...
+        .chain()
+        .focus()
+        .setImage({
+          src: selectedMediaPath,
+          alt: altText,
+        })
+        .run()
+    }
+    onMediaModalClose()
+  }
```
seaerchin, line 180: this actually appears to be the same thing that we do by default?

src/components/media/MediaModal.jsx
```diff
@@ -1,30 +1,27 @@
...
 
 import { getMediaDirectoryName } from "utils/media"
 
-const MediaModal = ({ onClose, onProceed, type, showAltTextModal = false }) => {
+const MediaModal = ({
+  onClose,
+  onProceed,
+  type,
+  showAltTextModal = false,
+  onExternalProceed,
```
seaerchin, line 17: from what i can tell, in both the markdown and tiptap editor, `onExternalProceed` appears to be doing the exact same thing as `onProceed` - this seems to suggest that this parameter is superfluous and should be removed

src/components/media/MediasSelectModal.jsx
```diff
@@ -127,38 +159,103 @@ const MediasSelectModal = ({
...
-            <Flex w="100%" justifyContent="space-between">
+  const [activeTab, setActiveTab] = useState(0)
+
+  const handleTabChange = (index) => {
+    onMediaSelect("")
+    clearErrors()
+    setActiveTab(index)
+  }
+
+  const debouncedHandleSearch = _.debounce((value) => {
```
seaerchin, line 170: very nice.

src/components/media/MediasSelectModal.jsx
```diff
@@ -127,38 +159,103 @@ const MediasSelectModal = ({
...
+    setSearchedValue(value)
+    // Perform any other actions you want with the input value
+  }, 1000)
```
seaerchin, line 173: suggest shifting this out to a constant (could be inlined in this file or elsewhere) - this is because we want to add some meaning to this otherwise arbitrary number of 1000.

src/components/media/MediasSelectModal.jsx
```diff
@@ -127,38 +159,103 @@ const MediasSelectModal = ({
...
+                  setQueryParams((prevState) => {
+                    return {
+                      ...prevState,
+                      mediaDirectoryName: getMediaDirectoryName(
+                        queryParams.mediaDirectoryName,
+                        {
+                          end:
+                            queryParams.mediaDirectoryName.split("%2F").length -
+                            1,
+                        }
```
seaerchin, line 199: could i get some clarity on what this is doing? specifically, what the `end` parameter is doing here? 

i'm abit confused as to why we wouldn't just use the default `mediaDirectoryName` and have to add an additional `end` parameter

src/components/media/MediasSelectModal.jsx
```diff
@@ -291,16 +338,290 @@ const MediasSelectModal = ({
...
-            </LoadingButton>
-          </VStack>
-        </ModalFooter>
+          </div>
+        </Box>
+      </GridItem>
+    </Grid>
+  )
+
+  const FolderBreadcrumb = () => (
```
seaerchin, line 347: could we shift this outside? this appears to be an independent component

src/components/media/MediasSelectModal.jsx
```diff
@@ -291,16 +338,290 @@ const MediasSelectModal = ({
...
+    <VStack>
+      <Breadcrumb alignSelf="start">
+        {queryParams.mediaDirectoryName
+          ? queryParams.mediaDirectoryName.split("%2F").map((dir, idx) => (
```
seaerchin, line 351: nit: prefer `&&`

src/components/media/MediasSelectModal.jsx
```diff
@@ -291,16 +338,290 @@ const MediasSelectModal = ({
...
+                }
+              >
+                <BreadcrumbLink
+                  onClick={(e) => {
+                    e.preventDefault()
+                    setQueryParams((prevState) => ({
+                      ...prevState,
+                      mediaDirectoryName: getMediaDirectoryName(
+                        queryParams.mediaDirectoryName,
+                        { end: idx + 1 }
```
seaerchin, line 364: could i check why the end parameter is different here? previously it was `length - 1`

src/components/media/MediasSelectModal.jsx
```diff
@@ -291,16 +338,290 @@ const MediasSelectModal = ({
...
+                <TabPanels minHeight="16.5rem">
+                  <TabPanel>{internalMediaSelect}</TabPanel>
+                  <TabPanel>
+                    <Grid h="100%" templateColumns="1fr 1fr">
+                      <GridItem
+                        borderRight="1px solid"
+                        borderColor="base.divider.medium"
+                        pr="1.5rem"
+                      >
+                        {register && (
```
seaerchin, line 431: abit confused as to why this `register &&` is required

src/components/media/MediasSelectModal.jsx
```diff
@@ -291,16 +338,290 @@ const MediasSelectModal = ({
...
+                    !!errors.altText ||
+                    !!errors.externalImageValidation
+                  }
+                  onClick={handleSubmit(async (data) => {
+                    if (activeTab === 0) {
+                      onProceed(data)
+                      return
+                    }
+                    onExternalProceed(data)
+                  })}
```
seaerchin, line 579: in both use cases, `onProceed` and `onExternalProceed` appears to have no difference so we might not need this

src/components/pages/EditorModals.jsx
alexanderleegs: Yeah you're right, removed in 516be476!

src/components/media/MediaModal.jsx
alexanderleegs: Yep, removed in 516be476

src/components/media/MediasSelectModal.jsx
alexanderleegs: Good point, shifted in d71dd484

src/components/media/MediasSelectModal.jsx
alexanderleegs: Makes sense, d71dd484!

src/components/media/MediasSelectModal.jsx
alexanderleegs: 86e0cfe9!

src/components/media/MediasSelectModal.jsx
alexanderleegs: `idx` is the existing position of the element already (0-indexed)! So the stop point is one after this one

src/components/media/MediasSelectModal.jsx
alexanderleegs: It isn't! removed in 86e0cfe9

alexanderleegs:
> This PR appears to be functionally complete but `MediasSelectModal` was made _significantly_ more complex.
> 
> This is made worse because the file is also in JSX, which has no annotations and doesn't future proof well.
> 
> would it be possible to shift some functionality out or make 2 different modals? (note that the modal itself is 600 lines long atm :/)

Yeah agreed, I timeboxed the effort to shift to ts since i was going on leave soon, and the component was more complicated to migrate than expected, so I abandoned it. Still think it's worth doing though, maybe in a future PR?

src/components/media/MediaModal.jsx
```diff
@@ -18,13 +16,6 @@ const MediaModal = ({ onClose, onProceed, type, showAltTextModal = false }) => {
...
   const methods = useForm({
     mode: "onTouched",
-    resolver: yupResolver(
-      Yup.object({
-        altText: Yup.string()
-          // .required("Alt text is required")
-          .max(100, "Alt text should be less than 100 characters"),
-      })
-    ),
   })
```
seaerchin, line 20: sanity check - we no longer require validation for alt text?

src/layouts/EditPage/EditPage.tsx
```diff
@@ -161,17 +159,41 @@ export const EditPage = () => {
 
   if (!editor) return null
 
+  const onExternalMediaSave = async ({
```
seaerchin, line 162: i think we can remove this?

src/layouts/EditPage/EditPage.tsx
```diff
@@ -161,17 +159,41 @@ export const EditPage = () => {
...
+      return { mediaPath: src }
+    }
     const { fileName, imageDirectory } = getImageDetails(src)
-    const { mediaPath, mediaUrl } = await mediaService.get({
+    const { mediaPath } = await mediaService.get({
       siteName,
       mediaDirectoryName: imageDirectory || "images",
       fileName,
     })
     const nomalisedMediaPath = mediaPath.startsWith("images/")
```
seaerchin, line 193: nit: `normalisedMediaPath`

src/components/media/MediasSelectModal.jsx
```diff
@@ -54,8 +73,38 @@ const titleCase = (str) => {
...
+            </BreadcrumbLink>
+          </BreadcrumbItem>
+        ))}
+    </Breadcrumb>
+  </VStack>
+)
+
 const MediasSelectModal = ({
   onProceed,
+  onExternalProceed,
```
seaerchin, line 106: i think this parameter might be safe to remove with the new changes

src/components/media/MediasSelectModal.jsx
```diff
@@ -127,38 +189,103 @@ const MediasSelectModal = ({
...
+    clearErrors()
+    setActiveTab(index)
+  }
+
+  const debouncedHandleSearch = _.debounce((value) => {
+    setSearchedValue(value)
+    // Perform any other actions you want with the input value
+  }, DEBOUNCE_TIME)
+
+  const internalMediaSelect = (
```
seaerchin, line 205: could we just inline this? it's abit more confusing trying to figure out where it's used @__@

src/components/media/MediasSelectModal.jsx
```diff
@@ -291,16 +368,259 @@ const MediasSelectModal = ({
...
+      overflowY="hidden"
+    >
+      <ModalOverlay />
+      <ModalContent
+        padding="0.5rem"
+        paddingTop="1rem"
+        maxHeight="90%"
+        overflowY="hidden"
+      >
+        {mediaType === "images" && allowExternal ? (
```
seaerchin, line 394: maybe we should just make 2 modals? might help to simplify this. just offering suggestions, am ok with this atm as functionality is ok and it's not _super_ complicated.

seaerchin (approved):

seaerchin (commented):
Please remove the extra props + methods before merging. they don't seem to be required and will lead to confusion further on when refactoring (which, by virtue of the bloat we've added, is guaranteed to be required)

src/components/media/MediaModal.jsx
alexanderleegs: This has been moved into the alt text modal itself! react-hook-form doesn't allow a resolver + individual field validation to exist at the same time

src/components/media/MediasSelectModal.jsx
alexanderleegs: Err it's used in 2 places, which is why i had to move it out. I think ideally we put this in a separate component altogether, but i got tripped up when figuring out all the props required - probably should start from here when refactoring this component

src/components/media/MediasSelectModal.jsx
alexanderleegs: agreed!

src/components/media/MediasSelectModal.jsx
alexanderleegs: This is needed! The select modal has differing behaviour for ext and internal images - ext puts it in the file immediately, internal brings up the alt text modal

src/layouts/EditPage/EditPage.tsx
alexanderleegs: Whoops, removed in 4640532e!

src/layouts/EditPage/EditPage.tsx
alexanderleegs: Oh good catch haha, 4640532e

src/components/media/MediasSelectModal.jsx
seaerchin: sgtm

seaerchin (approved):

PR #1718, opened 2023-12-01
feat(tiptap): add outline blocks 

seaerchin:
## Problem
Right now, we got no clear way of knowing when a complex block is selected. This leads to confusion on the user end when they select it as they are not sure if it's selected. 

## Solution
Wrap the blocks in `BlockWrapper`. Do note that `Table` is **not** done here due to hidden complexities; will do in a separate PR


src/layouts/components/Editor/extensions/Iframe/IframeView.tsx
```diff
@@ -1,33 +1,37 @@
...
-        ml="1rem"
-        bgColor="#0d0d0d"
-        textStyle="caption-1"
-        fontWeight="bold"
-        color="#fff"
-        position="absolute"
-        top="0"
-        padding="0.25rem 0.75rem"
-        borderRadius="0 0 0.5rem 0.5rem"
+    <BlockWrapper name="Iframe" isSelected={selected}>
```
dcshzj, line 8: nit: Might be good to standardise as "Embed" and also add this BlockWrapper to InstagramView and FormSGView

seaerchin: oh didn't realise there were different views! will do, thanks for pointing it out

seaerchin:
conflicts becauase the `useEditor` got shifted into diff file, won't affect much

src/layouts/components/Editor/extensions/FormSG/FormSGView.tsx
```diff
@@ -1,36 +1,40 @@
...
-        ml="1rem"
-        bgColor="#0d0d0d"
-        textStyle="caption-1"
-        fontWeight="bold"
-        color="#fff"
-        position="absolute"
-        top="0"
-        padding="0.25rem 0.75rem"
-        borderRadius="0 0 0.5rem 0.5rem"
+    <BlockWrapper name="FormSG embed" isSelected={selected}>
```
dcshzj, line 10: nit: I think might be better to not differentiate them and just show as "Embed" to the end user? But this is not a strong opinion.

dcshzj (approved):

src/layouts/components/Editor/extensions/FormSG/FormSGView.tsx
seaerchin: oki i change to embed bah

PR #1722, opened 2023-12-04
fix(tables): update table behaviour

seaerchin:
## Problem
Tables are very ambiguous and it's not clear how to drag and drop our tables. This PR adds a drag handle and adds it to `BlockWrapper` and also to `Table`.

## Solution
- Add a draghandle to `BlockWrapper`. This has been cleared by design 
- Add the same drag handle to `Table`
- Reimplement `BlockWrapper` for `Table` in JS
    - this is required because using the `NodeView` in react adds a `div` between the `table` and the `tbody`, which cause our tables to be squished; the [doc](https://tiptap.dev/guide/node-views/react#changing-the-wrapping-dom-element) here is for v2.2.0, which is unreleased yet (we are on 2.1.12), which means that this wrapping doesn't work. manually upgrading to `2.2rc-7` causes interop issues with other extensions
    -  unable to implement `isSelected` state; using the `getPos` prop together with teh logic in `ReactNodeViewRenderer` does **not** work because the element isn't reactive. this leads to the view not being updated on click. 

## Screenshots
Screenshot 2023-12-04 at 6 22 42 PM


src/layouts/components/Editor/extensions/Table/Table.tsx
```diff
@@ -0,0 +1,46 @@
...
+import "../../components/BlockWrapper/styles.scss"
+
+export const Table = TiptapTable.extend({
+  draggable: true,
+  addNodeView() {
+    return () => {
+      const dom = document.createElement("div")
+      const contentDOM = document.createElement("table")
+
+      // NOTE: Append to `dom` as this should not be editable.
```
kishore03109, line 13: hmm i dont quite understand this comment :(
what does appending to the dom got to do with the mutability of the component arh?

kishore03109 (commented):
> unable to implement isSelected state; using the getPos prop together with teh logic in ReactNodeViewRenderer does not work because the element isn't reactive. this leads to the view not being updated on click.


hey i remember facing a similar quirk of tiptap for this. what do we feel about calculating this ourselves by getting the current selection location and seeing if a 
`const currNode = findParentNode((t) => t.type === this.type)(selection)`

exists?

src/layouts/components/Editor/extensions/Table/Table.tsx
seaerchin: we have to do this because tiptap will set a `contentEditable` with the corresponding attribute on the node view and subsequently, use that to control if editing is allowed or not.

see [here](https://tiptap.dev/guide/node-views/js#adding-a-content-editable) for more details

seaerchin:
@kishore03109 - i actually tried reimplementing `selected` using `getPos` (fwiw, their implementation does the same, i copied over0 but it didn't work as the `editor` instance provided is **not** reactive - hence, won't recompute.

i can try this but yea just take note of this quirk

kishore03109 (approved):

PR #1724, opened 2023-12-05
refactor(tiptap): add placeholder instead of prefill

seaerchin:
## Problem
Previously, we used to do a _prefill_ for user when their content was empty. This was not ideal as it allowed user to save said prefill.

## Solution
- add placeholder when empty 

## Screenshots
Screenshot 2023-12-05 at 3 19 41 PM


kishore03109 (approved):
I thought that placeholderwas intentional to show what they could do with tiptap lol

seaerchin:
> I thought that placeholderwas intentional to show what they could do with tiptap lol

yea but then i think the downsides outweighed the upside and i got asked to chagne it back AHHA

PR #1730, opened 2023-12-06
fix p1 accordion issues

kishore03109:
## Problem

This PR attempts to fix the functional issues wrt to Accordions. 
Will leave paddings as separate since those will be easier to fix + don't really need much context to be worked on. 

Fix bunch of stuff [here](https://www.notion.so/opengov/Accordion-1de9bd79bfa540daacfdc5b703a80b00?pvs=4)

To test template stuff, need to make changes to [kishore-test](https://staging-cms.isomer.gov.sg/sites/kishore-test/editPage/accordion%20page.md)

Corresponding template pr: https://github.com/isomerpages/isomerpages-template/pull/361 (doesnt block this pr since complex blocks are behind feature flag anyway)

deprior padding related fix as design is not 100% sure about it yet + padding are easier to fix so no biggie

Screenshot 2023-12-06 at 4 15 20 PM

## Reviewer notes

The approach for this PR would look weird considering the buttons are not bubble-menued. 
I did try it but it kept popping up on every individual cell (similar behaviour to our table stuff). I somehow couldnt get it to work since (note that there were 2 levels of abstraction here [bubbleMenu in react/src](https://github.com/ueberdosis/tiptap/blob/main/packages/react/src/BubbleMenu.tsx) -> [bubblemenu in the extensions](https://github.com/ueberdosis/tiptap/blob/main/packages/extension-bubble-menu/src/bubble-menu.ts)). As of now, the API is not extensible enough to allow for this to happen, and this wasnt just a trivial as a `.extend` that we do for our other extensions like `table` .There might be a way, please let me know if that is the case, but i time boxed this effort and just wrapped it around `NodeViewWrapper` to at least land something in main line first. 

This is still feature flagged and thus safe to merge to production for now, but hopefully p1 issues are all fixed and the pending stuff are just padding css stuff

### DIFF
--- a/src/layouts/components/Editor/components/BlockWrapper/BlockWrapper.tsx
+++ b/src/layouts/components/Editor/components/BlockWrapper/BlockWrapper.tsx
@@ -7,16 +7,19 @@ import "./styles.scss"
 interface BlockWrapperProps {
   name: string
   isSelected: boolean
+  otherButtons?: JSX.Element
 }
 
 export const BlockWrapper = ({
   name,
   isSelected,
+  otherButtons,
   children,
 }: PropsWithChildren<BlockWrapperProps>): JSX.Element => {
   return (
     <Box as={NodeViewWrapper}>
       <Box position="relative" maxW="36.5rem" data-group>
+        {isSelected && <>{otherButtons}</>}
         <Box
           _groupHover={{
             display: "block",
@@ -29,7 +32,9 @@ export const BlockWrapper = ({
           textColor="white"
           p="0.25rem"
         >
-          <Text textStyle="caption-1">{name}</Text>
+          <Text textStyle="caption-1" contentEditable="false">
+            {name}
+          </Text>
         </Box>
         <Box
           className="drag-handle"
--- a/src/layouts/components/Editor/components/BlockWrapper/BlockWrapper.tsx
+++ b/src/layouts/components/Editor/components/BlockWrapper/BlockWrapper.tsx
@@ -7,19 +7,19 @@ import "./styles.scss"
 interface BlockWrapperProps {
   name: string
   isSelected: boolean
-  otherButtons?: JSX.Element
+  childButtons?: JSX.Element
 }
 
 export const BlockWrapper = ({
   name,
   isSelected,
-  otherButtons,
+  childButtons,
   children,
 }: PropsWithChildren<BlockWrapperProps>): JSX.Element => {
   return (
     <Box as={NodeViewWrapper}>
       <Box position="relative" maxW="36.5rem" data-group>
-        {isSelected && <>{otherButtons}</>}
+        {isSelected && <>{childButtons}</>}
         <Box
           _groupHover={{
             display: "block",
--- a/src/layouts/components/Editor/extensions/Details/IsomerDetailGroup.ts
+++ b/src/layouts/components/Editor/extensions/Details/IsomerDetailGroup.ts
@@ -1,6 +1,5 @@
-import { ChainedCommands, findParentNode, Node } from "@tiptap/core"
+import { findParentNode, findParentNodeClosestToPos, Node } from "@tiptap/core"
 import { NodeRange, Slice } from "@tiptap/pm/model"
-import { EditorState } from "@tiptap/pm/state"
 import { mergeAttributes, ReactNodeViewRenderer } from "@tiptap/react"
 
 import { AccordionBackgroundType } from "types/editPage"
@@ -22,6 +21,8 @@ declare module "@tiptap/core" {
       changeDetailGroupBackground: (
         backgroundColor: AccordionBackgroundType
       ) => ReturnType
+      unsetDetailsGroup: () => ReturnType
+      removeDetail: () => ReturnType
     }
   }
 }
@@ -32,7 +33,7 @@ export const IsomerDetailsGroup = Node.create<IsomerDetailGroupOptions>({
   atom: true,
   draggable: true,
   content: "details+",
-  defining: true,
+  definingForContext: true,
   isolating: true,
   allowGapCursor: false,
 
@@ -91,7 +92,6 @@ export const IsomerDetailsGroup = Node.create<IsomerDetailGroupOptions>({
               ],
             }
           )
-          .setTextSelection(blockRange.start + 2)
           .run()
       },
       appendDetail: (startPos: number, endPos: number) => ({
@@ -152,6 +152,19 @@ export const IsomerDetailsGroup = Node.create<IsomerDetailGroupOptions>({
           )
           .run()
       },
+      removeDetail: () => ({ chain, state }) => {
+        const { selection } = state
+
+        const currNode = findParentNodeClosestToPos(selection.$anchor, (t) => {
+          return t.type.name === "details"
+        })
+
+        if (currNode === undefined) {
+          return false
+        }
+
+        return chain().deleteNode(currNode.node.type).run()
+      },
 
       changeDetailGroupBackground: (
         backgroundColor: AccordionBackgroundType
@@ -197,36 +210,12 @@ export const IsomerDetailsGroup = Node.create<IsomerDetailGroupOptions>({
           )
           .run()
       },
-      unsetDetailsGroup: () => ({
-        chain,
-        state,
-      }: {
-        state: EditorState
-        chain: ChainedCommands
-      }) => {
-        const { selection } = state
-        const currNode = findParentNode((t) => t.type === this.type)(selection)
-        if (!currNode) return false
-        const startPos = currNode.pos
-        const range = { from: startPos, to: startPos + currNode.node.nodeSize }
-
-        // TODO: check type
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore
-        return chain()
-          .insertContentAt(range, [])
-          .setTextSelection(startPos + 1)
-          .run()
+      unsetDetailsGroup: () => ({ tr, editor }) => {
+        editor.state.selection.replace(tr)
+        return true
       },
     }
   },
-
-  // TODO: add keyboard shortcuts for backspace
-  // addKeyboardShortcuts() {
-  //   return {
-  //     "Backspace": () => this.editor.commands.detailGroup.setDetailsGroup(),
-  //   }
-  // }
 
   addAttributes() {
     return {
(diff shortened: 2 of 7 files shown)


kishore03109:
Current dependencies on/for this PR:
* develop
  * **PR #1730** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-frontend/1730?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/isomerpages/isomercms-frontend/1730?utm_source=stack-comment).

src/layouts/components/Editor/components/BlockWrapper/BlockWrapper.tsx
```diff
@@ -7,16 +7,19 @@ import "./styles.scss"
...
 export const BlockWrapper = ({
   name,
   isSelected,
+  otherButtons,
   children,
 }: PropsWithChildren<BlockWrapperProps>): JSX.Element => {
   return (
     <Box as={NodeViewWrapper}>
       <Box position="relative" maxW="36.5rem" data-group>
+        {isSelected && <>{otherButtons}</>}
```
seaerchin, line 22: could i ask why we require this? i think an alternative that we might wish to consider here would be to create another component that uses `BlockViewWrapper` and does this functionality. the reason for doing so is that this `otherButtons` is kind of a unique use case and kinda pollutes the responsibility of `BlockWrapper` (thus tempting other devs to add mroe things here)

if that doesn't work well, i'm ok w/ this if we have a better fitting name

src/layouts/components/Editor/extensions/Details/IsomerDetailGroup.ts
```diff
@@ -32,7 +33,7 @@ export const IsomerDetailsGroup = Node.create<IsomerDetailGroupOptions>({
   atom: true,
   draggable: true,
   content: "details+",
-  defining: true,
+  definingForContext: true,
```
seaerchin, line 36: this is new to me, could you eli5 this for me? couldn't seem to find it on the tiptap docs oddly

src/layouts/components/Editor/extensions/Details/IsomerDetailGroup.ts
```diff
@@ -197,37 +210,13 @@ export const IsomerDetailsGroup = Node.create<IsomerDetailGroupOptions>({
...
-
-        // TODO: check type
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore
-        return chain()
-          .insertContentAt(range, [])
-          .setTextSelection(startPos + 1)
-          .run()
+      unsetDetailsGroup: () => ({ tr, editor }) => {
+        editor.state.selection.replace(tr)
```
seaerchin, line 214: sanity check - this just replaces the selection with an empty string right? what happens if we have a selection within the accordion and we call this? would only the selection (ie, the text) be replaced? 

iirc, tiptap uses `toggleNode` or smth for their exported nodes, maybe that might be worth checking out

src/layouts/components/Editor/extensions/Details/IsomerDetailGroupView.tsx
```diff
@@ -13,94 +15,112 @@ export const IsomerDetailGroupView = ({
...
+              </IconButton>
+            </Tooltip>
+          </Box>
+        </HStack>
+      </Box>
+      <Box
+        position="absolute"
+        bottom="0"
+        left="50%"
+        transform="translateX(-50%)"
```
seaerchin, line 84: double checking - are we trying to center the `box` here?

src/layouts/components/Editor/extensions/Details/IsomerDetailGroupView.tsx
```diff
@@ -13,94 +15,112 @@ export const IsomerDetailGroupView = ({
...
+        mb="-1.5rem"
```
seaerchin, line 85: nit: idt `mb` works when position is `absolute`

src/layouts/components/Editor/extensions/Details/IsomerDetails.ts
```diff
@@ -68,4 +68,27 @@ export const IsomerDetails = Details.extend({
...
+          return this.editor.commands.command(({ tr }) => {
+            const start = $anchor.pos - 1
+            const end = $anchor.pos
+            tr.delete(start, end)
+            return true
+          })
+        }
+        return this.editor.commands.removeDetail()
+      },
+    }
```
seaerchin, line 92: idt this works - adding a keyboard shortcut would invoke it for every associated keypress (`Backspace`) here (unless we copied over the backspace logic from tiptap)  

i think if the intended outcome is to delete the block as a whole, we might wanna just make the parent node `defining` so that backspace consumes it atomically.

src/layouts/components/Editor/components/BlockWrapper/BlockWrapper.tsx
kishore03109: Well, ideally we should be using bubble menu for this, but somehow I couldnt get the bubble menu to work to target the component as a whole, it targets the individual cells which does not fit our requirement. 

I am not sure if I understood your suggestion correctly, if I were to create another component wouldnt that copy over quite a bit of code ah? Maybe can quick pair if you have bandwidth please!

src/layouts/components/Editor/extensions/Details/IsomerDetailGroup.ts
kishore03109: sure, this is a prosemirror thing found [here](https://prosemirror.net/docs/ref/#model.NodeSpec.definingForContent). TLDR, had this weird bug when if you backspace on an accordion title the format screws up, so this flag j sets the parent as immutable

src/layouts/components/Editor/extensions/Details/IsomerDetailGroup.ts
kishore03109: You are right to call this out, there is an underlying assumption that this func is only called iff the entire accordion is selected. This is maintained (correct me if i am wrong here) since the delete is only called from pressing the button, which only appears on select of the entire accordion block. 


But out of curorsity, say this could actually be called when the selection is within the accordion, how will the `toggleNode` work ah? if > 1 accordion in the page, this toggle node will not work as expected right?

src/layouts/components/Editor/extensions/Details/IsomerDetailGroupView.tsx
kishore03109: yes

src/layouts/components/Editor/extensions/Details/IsomerDetails.ts
kishore03109: oh the purpose of this is not to delete the block as a whole, is to actually have a functional backspace key press for the titles, but I should not be able to backspace on an empty title to delete the entire block (deletion of the entire block should be in the button)!

src/layouts/components/Editor/extensions/Details/IsomerDetailGroupView.tsx
kishore03109: weirdly in this case it does leh 
Screenshot 2023-12-28 at 10 30 09 AM

Screenshot 2023-12-28 at 10 29 55 AM

seaerchin: oh cos we don't set `top`

src/layouts/components/Editor/components/BlockWrapper/BlockWrapper.tsx
seaerchin: if it's the bubble menu, wouldn't it be possible to check if the node that's active is the `detailSummary`?

src/layouts/components/Editor/extensions/Details/IsomerDetailGroup.ts
seaerchin: the base node is `detailSummary`, which we can call [toggle node](https://tiptap.dev/docs/editor/api/commands/toggle-node) on. alternatively, we could also consider just using tiptap's default [unsetDetails](https://tiptap.dev/docs/editor/api/nodes/details#unset-details), which is less destructive.

src/layouts/components/Editor/extensions/Details/IsomerDetails.ts
seaerchin: that's fine, but because this triggers for every keypress on backspace, it might destructively interfere with other nodes that don't adhere to the default backspace behaviour.

have we tried deleting:
1. nested nodes (cards) with this?
2. `defining` nodes?
3. plain text?

as this uses the `selection` and checks if the selection is empty, we might mess up on deletion on plain text w/o a selection (haven't tried)

src/layouts/components/Editor/components/BlockWrapper/BlockWrapper.tsx
kishore03109: Hmm the issue is not to check whether the node is active, the issue was revolving more around `detailSummary` being selected but the popper js targeting each detail summary rather than the entire detail group

src/layouts/components/Editor/extensions/Details/IsomerDetailGroup.ts
kishore03109: this was the original implementation! this had to be coded out so that we dont unset the details causing a backspace on a detail deleting the entire detail group into an inconsistent state that we do not expect

src/layouts/components/Editor/extensions/Details/IsomerDetails.ts
kishore03109: hey good point!
regarding the selection, the delete button selects the node as a whole, and therefore this should not accidently delete another selected text for example. 

regarding the 
- nested card, deletion works as expected, the cards is deleted as expected 
- defining nodes, eg text? abit confused about this point ps, can clarify here 
- plain text works too

seaerchin: `defining` nodes are those that should be deleted as a group! (i think iframe is, for one)

if this works, then am ok w/ it as chance of destructive interference seems low - tiptap might dispatch the event to the specific node itself, in which case idt my concerns stand.

seaerchin (approved):

PR #1731, opened 2023-12-06
fix(editpagelayout): shift context call elsewhere

seaerchin:
## Problem
we were crashign previously due to the context call in `EditPageLayout`. This was because `EditPageLayout` is a layout component that was shared between both legacy/current sites. 

## Solution
1. separate the call out as a prop
2. for `LegacyEditPage`, there's a default argument of `false` on `EditPageLayout` so no changes needed (callers should set **if required**) 
3. for the other two, just inlined the `isAnyDrawerOpen` functionality. (i'm not sure if `MarkdownEditPage` requires because technically it should not but who knows)
4. Take note that `useEditorDrawerContext` is only called for cards/details, which means that no existing functionality in `LegacyEditPage` requires this.

misc - style fix cos i noticed got some whitespace. set the editor to grow instead of giving fixed width. this means that when resizing, we take away the width of editor first if not enough. if there's extra space, the editor will grow to fit extra space.

## Test
- [ ] Go to a legacy site (or just off the flag for the env) 
- [ ] editor shud not crash
- [ ] turn on flag
- [ ] same site should work

### DIFF
--- a/src/layouts/EditPage/EditPageLayout.tsx
+++ b/src/layouts/EditPage/EditPageLayout.tsx
@@ -20,8 +20,6 @@ import Header from "components/Header"
 import { OverwriteChangesModal } from "components/OverwriteChangesModal"
 import { WarningModal } from "components/WarningModal"
 
-import { useEditorDrawerContext } from "contexts/EditorDrawerContext"
-
 import { useGetMultipleMediaHook } from "hooks/mediaHooks"
 import { useGetPageHook, useUpdatePageHook } from "hooks/pageHooks"
 import { useCspHook, useGetSiteColorsHook } from "hooks/settingsHooks"
@@ -37,6 +35,7 @@ import { sanitiseRawHtml, updateHtmlWithMediaData } from "./utils"
 interface EditPageLayoutProps {
   getEditorContent: () => string
   setEditorContent: (content: string) => void
+  shouldDisableSave?: boolean
   variant: PageVariant
 }
 
@@ -44,9 +43,9 @@ export const EditPageLayout = ({
   getEditorContent,
   setEditorContent,
   variant = "markdown",
+  shouldDisableSave = false,
   children,
 }: PropsWithChildren<EditPageLayoutProps>) => {
-  const { isAnyDrawerOpen } = useEditorDrawerContext()
   const params = useParams<{ siteName: string }>()
   const decodedParams = getDecodedParams(params)
   const [mediaSrcs, setMediaSrcs] = useState(new Set(""))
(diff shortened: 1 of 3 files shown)


src/layouts/EditPage/MarkdownEditPage.tsx
```diff
@@ -119,11 +120,14 @@ export const MarkdownEditPage = ({ togglePreview }: MarkdownPageProps) => {
...
   }, [mediaData, editorValue, updateHtmlWithMediaData])
 
+  const { isAnyDrawerOpen } = useEditorDrawerContext()
+
   return (
     <EditPageLayout
       variant="markdown"
       getEditorContent={() => editorValue}
       setEditorContent={(content: string) => setEditorValue(content)}
+      shouldDisableSave={isAnyDrawerOpen}
```
kishore03109, line 130: this is the old editor right? if yes can default to yes leh, since this drawer only applies to cards + details (new rte)

harishv7 (approved):

src/layouts/EditPage/EditPageLayout.tsx
```diff
@@ -37,16 +35,17 @@ import { sanitiseRawHtml, updateHtmlWithMediaData } from "./utils"
...
   setEditorContent: (content: string) => void
+  shouldDisableSave?: boolean
   variant: PageVariant
 }
 
 export const EditPageLayout = ({
   getEditorContent,
   setEditorContent,
   variant = "markdown",
+  shouldDisableSave = false,
```
kishore03109, line 46: nit: rename to `isAnyDrawerOpen`. 
there are other reasons why save should be disabled (eg js content in the editor), so save button being disabled does not necessarily mean the drawer is open.

kishore03109 (approved):
nit: reduce complexity here by defaulting `isAnyDrawerOpen` to false for older rte

src/layouts/EditPage/MarkdownEditPage.tsx
seaerchin: this is the markdown editor, prior to using tiptap. this is **feature flagged** but yea, as noted in the writeup, should be safe to enable but left as-is to maintain compat

kishore03109: can also

src/layouts/EditPage/EditPageLayout.tsx
seaerchin: not doing this change - this isn't meant to be a specific "is this drawer open" prop. this is more of "is my parent component telling me to disable save" of which a possible reason could be that the drawer is open. the context is given _in the component itself_, where we say that whether we disable save is based on whether the drawer is open.

hope this makes sense!

PR #1734, opened 2023-12-06
Hotfix/v0.63.2

seaerchin:
## Problem
width doesn't work well on flex. this allows the preview to be squished if the editor got alot of content. turns out we have to use `flex-basis` - this is done here using the `flex` shorthand.

## Solution
1. set `flex-basis` to be `0 0 62.5rem` (ie, minimum 62.5rem, don't take extra space when growing + shrinking). our editor has `flex-grow: 1` (grows when got extra space) and `flex-shrink: 1` (editor shrinks instead of preview)

## Screenshots
**before**
Screenshot 2023-12-06 at 3 27 27 PM

**after**
Screenshot 2023-12-06 at 3 26 14 PM


kishore03109:
Screenshots with the problmatic viewports + fixed change please

harishv7 (approved):

PR #1736, opened 2023-12-08
IS-712 Fixed image resizing issue in RTE

QiluXie:
## Problem

Closes [IS-712]
The existing (commented out) image resizing feature on the new RTE has 2 major issues:

1. When user enters a new percentage value and clicks save in the form, it creates an additional resized image node instead of resizing the original image node 
2. To make the feature work, the `width` property in `.content img` in the template scss file has to be removed, which affects the old non-RTE page and is not backward compatible 

## Solution

1. To avoid duplicating image node, `updateAttributes` is called on the image editor instead of `setImage`, which will update the attribute of the selected image node. 
2. To avoid removing the `width` property in `.content img` css, a locally scoped `style` tag is used to update the `width` of each image node which overrides the pre-defined css property. 
3. To make 2nd point possible, A modification is needed in `image` extension to accept `style` as incoming attribute. 

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- Fixed issues on image resizing feature in the new RTE 

## Before & After Screenshots

**BEFORE**:

https://github.com/isomerpages/isomercms-frontend/assets/14878879/5bbea17d-00bd-4bc7-b186-1307fccf788d

**AFTER**:

https://github.com/isomerpages/isomercms-frontend/assets/14878879/a3d5aee5-0a64-4467-a035-9d55fef87524

## Tests

Go to staging site and replicate the action in the screen recording under `AFTER` section. 

[IS-712]: https://isomeropengov.atlassian.net/browse/IS-712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
+++ b/src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
@@ -56,7 +56,11 @@ export const ResizeImagePopover = ({
       return
     }
 
-    editor.chain().focus().setImageStyle(`width: ${data.value}%;`).run()
+    const style = {
+      width: data.value,
+    }
+
+    editor.chain().focus().setImageStyle(style).run()
 
     onClose()
   }
--- a/src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
+++ b/src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
@@ -68,8 +68,6 @@ export const ResizeImagePopover = ({
       .updateAttributes("image", {
         // use locally scoped style to override width in template scss
         style: `width: ${data.value}%;`,
-        // To display the resized width value in the form
-        // width: `${data.value}%`,
         ...isomerImageAttrs,
       })
       .run()
@@ -82,6 +80,7 @@ export const ResizeImagePopover = ({
       const initialValue: string | number =
         editor.state.selection
           .content()
+          // regex to parse the width value from style tag
           .content.firstChild?.attrs.style?.match(/width: (\d+)%/)?.[1] ?? 100
 
       if (typeof initialValue === "string") {
--- a/src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
+++ b/src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
@@ -65,7 +65,10 @@ export const ResizeImagePopover = ({
     editor
       .chain()
       .focus()
-      .setImage({
+      .updateAttributes("image", {
+        // use locally scoped style to override width in template scss
+        style: `width: ${data.value}%;`,
+        // To display the resized width value in the form
         width: `${data.value}%`,
         ...isomerImageAttrs,
       })
@@ -88,7 +91,7 @@ export const ResizeImagePopover = ({
         methods.setValue("value", initialValue)
       }
     }
-  }, [isOpen])
+  }, [editor.state.selection, isOpen, methods])
 
   return (
     <FormProvider {...methods}>
--- a/src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
+++ b/src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
@@ -56,21 +56,7 @@ export const ResizeImagePopover = ({
       return
     }
 
-    const isomerImageAttrs = {
-      src: node.attrs.src,
-      alt: node.attrs.alt,
-      href: node.attrs.href,
-    }
-
-    editor
-      .chain()
-      .focus()
-      .updateAttributes("image", {
-        // use locally scoped style to override width in template scss
-        style: `width: ${data.value}%;`,
-        ...isomerImageAttrs,
-      })
-      .run()
+    editor.chain().focus().setImageStyle(`width: ${data.value}%;`).run()
 
     onClose()
   }
--- a/src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
+++ b/src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
@@ -69,7 +69,7 @@ export const ResizeImagePopover = ({
         // use locally scoped style to override width in template scss
         style: `width: ${data.value}%;`,
         // To display the resized width value in the form
-        width: `${data.value}%`,
+        // width: `${data.value}%`,
         ...isomerImageAttrs,
       })
       .run()
@@ -80,12 +80,16 @@ export const ResizeImagePopover = ({
   useEffect(() => {
     if (isOpen) {
       const initialValue: string | number =
-        editor.state.selection.content().content.firstChild?.attrs.width ?? 100
+        editor.state.selection
+          .content()
+          .content.firstChild?.attrs.style?.match(/width: (\d+)%/)?.[1] ?? 100
 
       if (typeof initialValue === "string") {
         methods.setValue(
           "value",
-          initialValue.endsWith("%") ? Number(initialValue.slice(0, -1)) : 100
+          initialValue.endsWith("%")
+            ? Number(initialValue.slice(0, -1))
+            : Number(initialValue)
         )
       } else {
         methods.setValue("value", initialValue)
--- a/src/layouts/components/Editor/extensions/Image/Image.ts
+++ b/src/layouts/components/Editor/extensions/Image/Image.ts
@@ -16,7 +16,6 @@ interface IsomerImageOptions {
 
 interface IsomerImageStyle {
   width?: number
-  backgroundColor?: string
 }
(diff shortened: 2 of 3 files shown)


src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
```diff
@@ -65,7 +65,10 @@ export const ResizeImagePopover = ({
     editor
       .chain()
       .focus()
-      .setImage({
+      .updateAttributes("image", {
+        // use locally scoped style to override width in template scss
+        style: `width: ${data.value}%;`,
+        // To display the resized width value in the form
         width: `${data.value}%`,
```
dcshzj, line 72: nit: I think this width should be removed since it is no longer needed, and we should adjust ResizeImagePopover to parse the `style` attribute.

src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
```diff
@@ -65,7 +65,10 @@ export const ResizeImagePopover = ({
     editor
       .chain()
       .focus()
-      .setImage({
+      .updateAttributes("image", {
```
dcshzj, line 68: question: I had the impression that this wouldn't work if there are more than one image in the editor?

seaerchin: could we hide this `updateAttributes` method and expose a `setWidth` method on the `Image` extension instead? 

more inline with the principle of least privilege + it's obvious to devs when reading the code for `Image` how the attributes are set

src/layouts/components/Editor/extensions/Image/Image.ts
```diff
@@ -40,6 +40,9 @@ export const IsomerImage = Image.extend({
       height: {
         default: "auto",
       },
+      style: {
+        default: null,
```
harishv7, line 44: why do we need to specify the default as null?

QiluXie: Thanks for the good catch @harishv7! Yeah, I think we should set it to `100%` or `auto`, I will go with `100%` if you guys are okay with it!

src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
QiluXie: Thanks for the suggestion! Fixed in 88c13554923d35df10f6c39727505f547c759108 and be4b5a51a5447f92db613b49f92d5bbc5bd26621

src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
QiluXie: > could we hide this `updateAttributes` method and expose a `setWidth` method on the `Image` extension instead?
> 
> more inline with the principle of least privilege + it's obvious to devs when reading the code for `Image` how the attributes are set

Thanks for the suggestion! Did the refactoring in 688d5a3e7bf09f0d97e8a4b8402aef05c68156d3

Instead of exposing `setWidth` method, I added a method to `setImageStyle` on the `image` extension which can be used to add a locally scoped `style` tag to the image node. It accepts a string style and added to the node. 

Thought this would be a more suitable method since 1) there could be future cases to add other attribute in `style` tag than just `width`, 2) `width` is a string attribute inside `style` tag, feels it's more appropriate to expose method to manipulate on tag level rather than a specific string value. 

@seaerchin wdyt?

seaerchin: sure, sounds good! do take note that currently we give the type as `style: string`, which is very risky - we might wanna consider giving an explicit type like:
```ts
style: { width: number }
```
and the method itself does the conversion to string.

the benefits is that it stops us from making careless mistakes (is width specified as px? %? if we wanted to add colours, we can typecheck to a fixed universe of colours maybe blue/green only) and gives feedback to the devs too.

src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
```diff
@@ -56,39 +56,35 @@ export const ResizeImagePopover = ({
...
   }
 
   useEffect(() => {
     if (isOpen) {
       const initialValue: string | number =
-        editor.state.selection.content().content.firstChild?.attrs.width ?? 100
+        editor.state.selection
+          .content()
+          // regex to parse the width value from style tag
+          .content.firstChild?.attrs.style?.match(/width: (\d+)%/)?.[1] ?? 100
```
seaerchin, line 74: sanity check - why are taking the first element here?

src/layouts/components/Editor/extensions/Image/Image.ts
```diff
@@ -11,13 +11,20 @@ interface IsomerImageOptions {
   width?: string
   height?: string
   href?: string
+  style?: string
+}
+
+interface IsomerImageStyle {
+  width?: number
+  backgroundColor?: string
```
seaerchin, line 19: can delete first! was just raising an example HAHA

src/layouts/components/Editor/extensions/Image/Image.ts
```diff
@@ -80,6 +90,40 @@ export const IsomerImage = Image.extend({
...
+              "Invalid width value. It should be between 1 and 100."
+            )
+            return false
+          }
+          style += `width: ${styleOptions.width}%;`
+        }
+
+        if (styleOptions.backgroundColor) {
+          style += `background-color: ${styleOptions.backgroundColor};`
+        }
```
seaerchin, line 111: can consider removing this first?

src/layouts/components/Editor/extensions/Image/Image.ts
```diff
@@ -80,6 +90,40 @@ export const IsomerImage = Image.extend({
...
+
+      setImageStyle: (styleOptions) => ({ tr, dispatch, editor }) => {
+        const { from, to } = tr.selection
+
+        let style = ""
+
+        if (styleOptions.width) {
+          if (styleOptions.width < 1 || styleOptions.width > 100) {
+            console.error(
+              "Invalid width value. It should be between 1 and 100."
```
seaerchin, line 102: just double checking here - do we have validation on the `form` for this range? 

we can sanitise + validate the input at point of entry and provide more feedback to user + safety to us

seaerchin (approved):
lgtm, we can do the enhancements (validation shifting) in a follow-up PR, functionality-wise looks good!

src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
seaerchin: sgtm! was only giving `backgroundColor` as an example, so we might wanna remove it @__@

src/layouts/components/Editor/extensions/Image/Image.ts
QiluXie: Yes! As I mentioned [in this comment](https://github.com/isomerpages/isomercms-frontend/pull/1736#discussion_r1422241724), we already have the validation on the `form` UI to disable the save button when the value is not within the range of 1-100. 

This is just an extra layer of protection for cases like users using inspection mode to call the method directly that bypasses the UI. I'm inclined to keep it.

seaerchin: i think if they inspect it's kinda in their own hands already but sure, no issues with keeping this!

src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
QiluXie: My understanding is at a given time, only one element can be selected by the user (at least that's the behaviour for our RTE currently). So the first `child` of the selected `content` is the one that user is trying to modify.

src/layouts/components/Editor/components/ImageBubbleMenu/ResizeImagePopover.tsx
QiluXie: Yeah no worries, just removed it in 103897178315870ae1628dd281b0ef10b6d7f671

src/layouts/components/Editor/extensions/Image/Image.ts
QiluXie: Sure removed in 103897178315870ae1628dd281b0ef10b6d7f671

src/layouts/components/Editor/extensions/Image/Image.ts
QiluXie: Sure removed in 103897178315870ae1628dd281b0ef10b6d7f671

PR #1738, opened 2023-12-11
IS-807 Fixed duplicating issue when setting hyperlink & alt text for image in RTE

QiluXie:
## Problem
Closes IS-807

The Hyperlink and AltText options of the image node in our new RTE has user-impacting bugs. Everytime users add or change the values through the handler, the RTE creates a duplicated image and applies the changes to the newly duplicated image, resulting in user confusion and very bad user experience. 

## Solution

The bug resides in the `setImage` method of the `image` extension. The logic of selecting the current node and updating its attributes is incorrect. This PR fixes the logic therefore fixes the bug for both options. 

**Breaking Changes**

  - Details ...

**Bug Fixes**:

- The image duplicating issue is fixed for both `Hyperlink` and `AltText` options of image node in the new RTE

## Before & After Screenshots

**BEFORE**:

https://github.com/isomerpages/isomercms-frontend/assets/14878879/4a7c8213-f157-46df-bb13-c93070bbc888

**AFTER**:

https://github.com/isomerpages/isomercms-frontend/assets/14878879/9cb08ff5-fd9d-4fd2-84c2-df1968782e84

## Tests

- UAT test in staging environment using the following steps:
- [ ] Go to staging CMS and click the repo `qilu-site-test`
- [ ] Go to rte test page 
- [ ] There are 3 existing images, try to modify the `hyperlink` and `altText` of them, notice that no duplicate image is created and the properties are modified successfully 
- [ ] Try to create a new image by clicking the icon in the toolbar 
- [ ] Observe the image creation is successful 
- [ ] Try to delete those images and observe the deletions are successful

### DIFF
--- a/src/layouts/components/Editor/extensions/Image/Image.ts
+++ b/src/layouts/components/Editor/extensions/Image/Image.ts
@@ -22,6 +22,7 @@ declare module "@tiptap/core" {
   interface Commands<ReturnType> {
     isomerImage: {
       setImage: (options: IsomerImageOptions) => ReturnType
+      setImageMeta: (options: IsomerImageOptions) => ReturnType
       deleteImage: () => ReturnType
       setImageStyle: (style: IsomerImageStyle) => ReturnType
     }
@@ -60,6 +61,32 @@ export const IsomerImage = Image.extend({
   addCommands() {
     return {
       setImage: (options) => ({ tr, dispatch, editor }) => {
+        if (!options.src) {
+          return false
+        }
+
+        if (dispatch) {
+          const node = editor.schema.nodes.image.create(options)
+
+          if (tr.selection.empty) {
+            const offset = tr.selection.to
+
+            tr.insert(offset, node)
+              .scrollIntoView()
+              .setSelection(Selection.near(tr.doc.resolve(offset)))
+          } else {
+            const offset = tr.selection.to + 1
+
+            tr.insert(offset, node)
+              .scrollIntoView()
+              .setSelection(Selection.near(tr.doc.resolve(offset)))
+          }
+        }
+
+        return true
+      },
+
+      setImageMeta: (options) => ({ tr, dispatch, editor }) => {
         if (!options.src) {
           return false
         }
--- a/src/layouts/components/Editor/extensions/Image/Image.ts
+++ b/src/layouts/components/Editor/extensions/Image/Image.ts
@@ -11,7 +11,6 @@ interface IsomerImageOptions {
   width?: string
   height?: string
   href?: string
-  style?: string
 }
 
 interface IsomerImageStyle {
@@ -22,7 +21,9 @@ declare module "@tiptap/core" {
   interface Commands<ReturnType> {
     isomerImage: {
       setImage: (options: IsomerImageOptions) => ReturnType
-      setImageMeta: (options: IsomerImageOptions) => ReturnType
+      setImageMeta: (
+        options: Pick<IsomerImageOptions, "alt" | "href">
+      ) => ReturnType
       deleteImage: () => ReturnType
       setImageStyle: (style: IsomerImageStyle) => ReturnType
     }
@@ -87,10 +88,6 @@ export const IsomerImage = Image.extend({
       },
 
       setImageMeta: (options) => ({ tr, dispatch, editor }) => {
-        if (!options.src) {
-          return false
-        }
-
         const { from, to } = tr.selection
 
         tr.doc.nodesBetween(from, to, (node, pos) => {
--- a/src/layouts/components/Editor/extensions/Image/Image.ts
+++ b/src/layouts/components/Editor/extensions/Image/Image.ts
@@ -64,22 +64,16 @@ export const IsomerImage = Image.extend({
           return false
         }
 
-        if (dispatch) {
-          const node = editor.schema.nodes.image.create(options)
+        const { from, to } = tr.selection
 
-          if (tr.selection.empty) {
-            const offset = tr.selection.to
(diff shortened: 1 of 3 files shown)


sehyunidaaa:
@QiluXie Just double-checking the behavior: if an image already has, say an alt text during insert (user puts alt text inside the modal), and tries to edit the alt text using the bubble menu, will they see the existing alt text? Or will it be an empty field?

sehyunidaaa:
(Nit) @QiluXie Not related to this PR, but the link icon is not vertically aligned to the rest of the icons (it's higher up). Any chance we could tweak it as well? Thank you

image

QiluXie:
> (Nit) @QiluXie Not related to this PR, but the link icon is not vertically aligned to the rest of the icons (it's higher up). Any chance we could tweak it as well? Thank you
> 
> image

Thanks @sehyunidaaa for catching this. I will take a look but inclined to not tackle it in this PR (also because it's not a high priority one atm). I will probably create a separate PR to fix similar small UI alignment issues so it's easier for other engs to review.

src/layouts/components/Editor/extensions/Image/Image.ts
```diff
@@ -55,22 +55,16 @@ export const IsomerImage = Image.extend({
...
-          } else {
-            const offset = tr.selection.to + 1
+        const { from, to } = tr.selection
 
-            tr.insert(offset, node)
-              .scrollIntoView()
-              .setSelection(Selection.near(tr.doc.resolve(offset)))
+        tr.doc.nodesBetween(from, to, (node, pos) => {
+          if (node.type === editor.schema.nodes.image) {
+            tr.setNodeMarkup(pos, null, { ...node.attrs, ...options })
```
seaerchin, line 62: fix here is fine, but it might remove image _creation_ functionality as we assume that the node between `from, to` is already an `image` (we do an `if` check) 

we should rebase this off your branch and rename this `createImage` so intended usage is clear + shift all usages _outside_ of `EditPage` (alt text, hyper link, resize) to new commands.

i think the new command for `alt text/hyperlink` can be called `setImageMeta` (or just create two small commands that do the same to avoid devs lumping everything into those commands).

this solves the problem + sets a good precedent for code organisation to avoid future bugs popping out.

seaerchin (commented):
could we double check image creation functionality + shift the methods for clearer organisation?

seaerchin:
can we also add a detailed list of steps for the tests? see other PRs for an example!

src/layouts/components/Editor/extensions/Image/Image.ts
QiluXie: Oh yeah! Thanks @seaerchin you are right. It affects the image creation, I should have tested it more thoroughly, my bad. 
I will fix it and also separate the functionality of editing image from creating image in the commands. 

Will update once the new commit is pushed.

QiluXie:
> can we also add a detailed list of steps for the tests? see other PRs for an example!

Sure, added them!

src/layouts/components/Editor/extensions/Image/Image.ts
QiluXie: Sure, refactored in 5be2b182bc3a214eeedc184e83bb73ea29bf32be

I added a method called `setImageMeta` to handle all the top level attributes change of image node, including `hylerLink` and `altText`. Decided to not create small commands for those top-level attributes as no clear benefits. The separation of `style` tag is useful because we can enforce the validity of `width` value inside the `style` string. For top-level attributes, we already have types defined in `IsomerImageOptions`.

src/layouts/components/Editor/extensions/Image/Image.ts
```diff
@@ -85,6 +86,26 @@ export const IsomerImage = Image.extend({
         return true
       },
 
+      setImageMeta: (options) => ({ tr, dispatch, editor }) => {
```
seaerchin, line 89: this method has the same signature as `setImage` - we probably _don't_ want this as this method should only allow devs to set certain meta level attributes post image creation (so things like `src` cannot be set). 

given this, can we tighten options here? perhaps a quick fix would be to just use the type `Pick<IsomerImageOptions, 'alt' | 'href'>`. this makes it obvious that this method cannot be used to set other attributes (atm it can) and make it distinct from `setImage`

seaerchin (approved):
given that this is a high priority bug-fix, i'm approving this first. 

however, we should aim to make the changes outlined here to allow for maintainability + less overhead required in the future.

src/layouts/components/Editor/extensions/Image/Image.ts
QiluXie: @seaerchin Sure, this makes sense and agree that it's better for maintainability. Other than just `alt` and `href`, I want to allow this method to set more attributes so it can fulfil future use cases. How about the following list:
```
  alt?: string
  title?: string
  width?: string
  height?: string
  href?: string
```
`style` shouldn't be allowed otherwise it's risky cause the string can have wrong value. They should only be able to set style by using `setImageStyle`. Similarly, `src` is not allowed. 
wdyt?

seaerchin: we can leave it as just `href/alt` first and add in when there's a need. we should avoid having [god methods](https://wiki.c2.com/?GodMethod) that do everything and add on a as-required basis.

QiluXie: @seaerchin Sure, done it in ff27724e206dd2a25b3d3eed7088be9c6e496bc8

PR #1758, opened 2023-12-21
fix: limit for get media

harishv7:
current limit value of 1000 affects ggs repos as all files are being fetched. changed it to be 15. For github repos, they are not affected as we do not pass the limit value to the read call

src/services/db/RepoService.ts
```
const contents = await super.readDirectory(sessionData, {
        directoryName,
      })
```


seaerchin (approved):
@seaerchin approved this PR from Slack with [Graphite](https://app.graphite.dev/github/pr/isomerpages/isomercms-frontend/1758)

PR #1762, opened 2023-12-28
fix(styles): fix overlapping page title issue

sehyunidaaa:
## Problem

On normal pages, the staging build indicator overlaps the page title if the title is too long.

image

## Solution

Done as a part of Design team learning thanks @seaerchin for demoing the solution and explaining concepts! 🙇 

## Before & After Screenshots

**BEFORE**:

**AFTER**:

## Tests

1. Open / create a normal page with a long title (100 chars)
2. Check that status indicator doesn't overlap page title (title should truncate)
3. Change folder name to something long (30 chars)
4. Check that no elements on header overlap each other

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/src/components/Header.jsx
+++ b/src/components/Header.jsx
@@ -129,13 +129,15 @@ const Header = ({
         </HStack>
         {/* <Spacer /> */}
         {title ? (
-          <Flex 
-            minWidth="-webkit-min-content" 
+          <Flex
+            minWidth="-webkit-min-content"
             alignItems="center"
             flex={1}
             justifyContent="center"
-            >
-            <Text textStyle="h5" noOfLines={1}>{title}</Text>
+          >
+            <Text textStyle="h5" noOfLines={1}>
+              {title}
+            </Text>
           </Flex>
         ) : null}
         <HStack flex="0 1 content" justifyContent="flex-end" w="fit-content">
--- a/src/components/Header.jsx
+++ b/src/components/Header.jsx
@@ -97,8 +97,9 @@ const Header = ({
         bg="white"
         h="4rem"
         w="100%"
+        spacing="0.5rem"
       >
-        <HStack spacing="1.25rem" flex={1}>
+        <HStack spacing="1.25rem" flex="1 1 content">
           {!showButton ? (
             <Box w="180px">
               <img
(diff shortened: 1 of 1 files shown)


src/components/Header.jsx
```diff
@@ -97,8 +97,9 @@ const Header = ({
         bg="white"
         h="4rem"
         w="100%"
+        spacing="0.5rem"
       >
-        <HStack spacing="1.25rem" flex={1}>
+        <HStack spacing="1.25rem" flex="1 1 content">
```
seaerchin, line 102: this means that the `HStack` will have space to grow! we might not want this - we want all the available space to go to the title instead. should this be `0 0 content` instead? 

this means that it has enough space to fit the content and won't grow/shink. (if you feel that it should shrink then feel free to change)

src/components/Header.jsx
```diff
@@ -128,11 +129,18 @@ const Header = ({
...
+            flex={1}
+            justifyContent="center"
+          >
+            <Text textStyle="h5" noOfLines={1}>
+              {title}
+            </Text>
           </Flex>
         ) : null}
-        <HStack flex={1} justifyContent="flex-end">
+        <HStack flex="0 1 content" justifyContent="flex-end" w="fit-content">
```
seaerchin, line 143: see above comment

seaerchin (commented):
btw, i also think we set `noOfLines = {1}` on the back button text so it truncates nicely - should be also be doing it here (eg: we access a page on a subfolder iwth a very long name)

src/components/Header.jsx
sehyunidaaa: Ah ok! Removing the grow, tysm

If we don't set `flex-shrink` for all of the contents in the stack tho, how do we prevent the header contents from exceeding the width like below?

image

src/components/Header.jsx
sehyunidaaa: changing this to `0 0 content`!

src/components/Header.jsx
sehyunidaaa: Will remove the growth factor! I think it should shrink just in case the text is rlly long!

> btw, i also think we set noOfLines = {1} on the back button text so it truncates nicely - should be also be doing it here (eg: we access a page on a subfolder iwth a very long name)

Also added this

seaerchin (approved):
remember 2 squash n merge. 

could i also trouble uyou to add testing steps?

sehyunidaaa:
@seaerchin Thanks! added testing steps below `Tests`

PR #1768, opened 2024-01-09
fix(editpage): remove all trailing empty p tags if present

dcshzj:
## Problem

In the Tiptap editor, trailing empty &lt;p> tags are being rendered as text, which negatively affects the rendering of the page when the last node by the user is something other than a paragraph.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Remove any trailing empty &lt;p> tags just before sending to the backend to save.
    - Note: This does not affect the trailing node added by Tiptap, as it is always added automatically when the user loads the page.

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-01-09 at 17 44 48

**AFTER**:

Screenshot 2024-01-09 at 17 45 16

## Tests

- [x] Smoke tests
    - [ ] Navigate to any site and edit using the Tiptap editor.
    - [ ] Add an image at the end of the page and save.
    - [ ] Verify that on the staging site, there is no text that appears after the image.
    - [ ] After the image, add multiple empty paragraphs and save.
    - [ ] Verify that on the staging site, there is no text that appears after the image.
    - [ ] Edit a page using the old Markdown editor and add `Blahblah<p></p>` to the end of the page.
    - [ ] Verify that on the staging site, only "Blahblah" appears without `<p></p>` being rendered.

## Deploy Notes

*None*

### DIFF
--- a/src/hooks/pageHooks/useUpdatePageHook.jsx
+++ b/src/hooks/pageHooks/useUpdatePageHook.jsx
@@ -26,11 +26,17 @@ export function useUpdatePageHook(params, queryParams) {
     (body) => {
       const { newFileName, sha, frontMatter, pageBody } = extractPageInfo(body)
 
+      let updatedPageBody = pageBody
+
+      while (updatedPageBody.endsWith("<p></p>")) {
+        updatedPageBody = updatedPageBody.slice(0, -7)
+      }
+
       return pageService.update(params, {
         newFileName,
         sha,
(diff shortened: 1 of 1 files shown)


dcshzj:
Declining, solution doesn't fix the root cause of the problem. We will need to explore how to fix the HTML output from Tiptap, or force Jekyll to render as HTML.

PR #1783, opened 2024-01-23
fix(modals): set motion preset to none

dcshzj:
> [!IMPORTANT]
> Requires testing on multiple devices + GSIB

## Problem

The opening and closing of modals in a GSIB is extremely slow in the default rendering mode.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Set the `motionPreset` for all modals to be `none`, so the modal will appear immediately instead of having an animation in which the modal slowly appears.

## Tests

- [x] Smoke tests
    - [ ] Go to any site.
    - [ ] Click on every possible action, verify that modals that are opened and closed appear and disappear immediately without any animation.

## Deploy Notes

*None*

### DIFF
--- a/src/components/CollaboratorModal/components/MainSubmodal.tsx
+++ b/src/components/CollaboratorModal/components/MainSubmodal.tsx
@@ -8,7 +8,6 @@ import {
   Divider,
   FormControl,
   Input,
-  Modal,
   ModalOverlay,
   ModalContent,
   ModalProps,
@@ -28,6 +27,8 @@ import { useEffect } from "react"
 import { FormProvider, useForm } from "react-hook-form"
 import { BiTrash } from "react-icons/bi"
 import { useParams } from "react-router-dom"
+
+import { Modal } from "components/Modal"
 
 import { useLoginContext } from "contexts/LoginContext"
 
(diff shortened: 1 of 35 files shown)


alexanderleegs (dismissed):
code wise lgtm, but i'm wondering if we should create a custom component (in a separate pr) which extends the design-system modal so that we don't have to insert this field in every time, since we want all modals transitionless to better support gsib?

kishore03109:
+1 on @alexanderleegs, this feels dev heavy to remember. Would much rather use IsomerModal + do a string search for Modal to make sure we arent using the chakra's/design system (forgot which one) to ensure we didnt miss out anything!

dcshzj:
@isomerpages/iso-engineers re-requesting review to use the approach suggested above (i.e. to have our own Modal component that sets the `motionPreset` to `none`).

seaerchin (approved):
Code lgtm. Just wondering if we should remove the parameter totally? Don't really have a strong opinion here

dcshzj:
@seaerchin I'm keeping for now to maintain maximum compatibility, in case there are some use cases which still needs a custom `motionPreset`.

PR #1786, opened 2024-01-24
fix(infocols): should not be allowed to rm last

kishore03109:
## Problem

Bug fix, disable deleting last infobox

https://github.com/isomerpages/isomercms-frontend/assets/42832651/673c0aaf-5ee4-40f2-81e4-af89deaacb20


alexanderleegs (approved):
lgtm, are there any more sites we need to fix this for?

kishore03109:
@alexanderleegs actually have good catch!

PR #1790, opened 2024-01-25
feat(e2e): migrate unlinked page tests to playwright

seaerchin:
## Problem
Previously, the tests for editing of unlinked pages were done in cypress. This PR migrates them to playwright. importantly, do take note that there's a shift to testing them on **email based login** - easier to setup (done alr in prev PR) + majority of our sites are on email based login. (also more complex)

## Solution
1. Add a `getApi` method - we need this because we set `baseURL` to be that for our frontend (so that navigation works fine) 
2. migrate from using `cypress` selectors to `playwright` locators.


e2e/editPage.spec.ts
```diff
@@ -0,0 +1,374 @@
...
+      context,
+    }) => {
+      // Arrange
+      const api = await getApi(await context.storageState())
+      const header = page.locator("#display-header")
+
+      // Act
+      await api.post("v2/sites/e2e-email-test-repo/settings", {
+        data: BASE_SETTINGS,
+      })
```
dcshzj, line 118: question: does the colour get updated immediately without refreshing the page?

Actually I'm a bit confused here. We seem to be setting the colour then testing if the page uses the same colour (testing 2 things), rather than testing if the page uses the colour that was originally set on the settings page (testing 1 thing).

dcshzj (commented):
I think edit page spec might get quite big + encompass a lot of different types/permutations of editing. Should we consider splitting into markdown editing and tiptap editing?

Also, I think we can separate out the collection + resource room editing into a separate spec that focuses more on testing aspects that are more specific to those types of editing.

e2e/editPage.spec.ts
```diff
@@ -0,0 +1,374 @@
...
+      await expect(page.getByText(SUCCESSFUL_EDIT_PAGE_TOAST)).toBeVisible()
+
+      // 2. Content is there even after refreshing
+      await page.reload()
+      await expect(
+        page.getByText(TEST_SANITIZED_INSTAGRAM_EMBED_SCRIPT)
+      ).toBeVisible()
+    })
+
+    test.skip("Edit page (unlinked) should not allow users to add untrusted external scripts", async ({
```
kishore03109, line 328: can we add a todo here to be fixed later?

kishore03109 (approved):
i dont have blocking comments for this specific pr, but i do air similar concerns to zj + think we need to cover the edge cases for the two different editors

e2e/editPage.spec.ts
alexanderleegs: Hmm i feel that this is actually testing a single thing, that setting a colour causes the page colour to change as well.

I think we can use your suggestion (check colour of edit page vs colour specified on settings) if we change the default colour of the repo though! Otherwise we have a fallback colour - we need to be able to distinguish between using the fallback colour and the actual colour

e2e/editPage.spec.ts
```diff
@@ -0,0 +1,374 @@
...
+      await expect(await page.getByText(TEST_UNTRUSTED_SCRIPT).count()).toEqual(
+        0
+      )
+    })
+
+    test.skip("Edit page (unlinked) should not allow users to add inline scripts", async ({
+      page,
+    }) => {
+      // NOTE: THis test fails - we can still add inline scripts.
+      // Arrange
```
alexanderleegs, line 355: Would be good to add a todo here!

e2e/editPage.spec.ts
seaerchin: i actually didn't realise this - the test still passes but i agree w/ the reload

seaerchin:
> I think edit page spec might get quite big + encompass a lot of different types/permutations of editing. Should we consider splitting into markdown editing and tiptap editing?
> 
> Also, I think we can separate out the collection + resource room editing into a separate spec that focuses more on testing aspects that are more specific to those types of editing.

agreed for `tiptap + markdown` - right now, goal is **not** to add more test specs but rather, to get to parity. lmk if you're ok w/ just adding md specs here and tiptap specs further down the line

dcshzj:
> agreed for `tiptap + markdown` - right now, goal is **not** to add more test specs but rather, to get to parity. lmk if you're ok w/ just adding md specs here and tiptap specs further down the line

Yup I think we can add the tiptap ones later. I think we are still in the midst of stablising tiptap so it would be a bit troublesome to create them right now.

e2e/editPage.spec.ts
seaerchin: added a `page.reload`

PR #1801, opened 2024-02-05
feat(cards): make the card link optional

dcshzj:
## Problem

Users have been requesting that the card block do not need to contain a link. Let's make their wishes come true.

## Solution

**Breaking Changes**

**Features**:

- Updated the validation schema for the card grid block to make links optional, unless a link URL has been provided.
    - If the link URL is not present, but the link text is provided, it is still treated as a card without a link and the link text does not appear on the card
- Made the validation happen on blur, so that the error message can appear to prompt the user to add the link text as soon as they have added the link URL.

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-02-05 at 13 32 41

**AFTER**:

Screenshot 2024-02-05 at 13 33 29
Screenshot 2024-02-05 at 13 33 47

## Tests

- [x] Smoke tests
    - [ ] Navigate to any site and edit a page using the new Tiptap editor
    - [ ] Add a card grid block
    - [ ] Create a card without putting any link URL or link text, then save the changes to the card grid. Verify that you should be able to save.
    - [ ] Create another card in the same card grid block and put in only the link URL. Verify that you should not be able to save the changes to the card grid.
    - [ ] Use the same card and add in the link text, then save the changes to the card grid. Verify that you should be able to save.
    - [ ] Create another card in the same card grid block and put in only the link text, then save the changes to the card grid. Verify that you should be able to save, but the link text does not appear in the editor preview.

## Deploy Notes

*None*

### DIFF
--- a/src/components/EditorCardsDrawer/EditorCardItem.tsx
+++ b/src/components/EditorCardsDrawer/EditorCardItem.tsx
@@ -196,12 +196,15 @@ export const EditorCardItem = ({
           </FormControl>
 
           {/* Card link URL */}
-          <FormControl isRequired isInvalid={!!errors.cards?.[index]?.linkUrl}>
+          <FormControl isInvalid={!!errors.cards?.[index]?.linkUrl}>
             <FormLabel mb="0.75rem">Link URL</FormLabel>
             <Input
               type="text"
               placeholder="Insert /page-url or https://"
-              {...methods.register(`cards.${index}.linkUrl`, { onChange })}
+              {...methods.register(`cards.${index}.linkUrl`, {
+                onChange,
+                onBlur: () => methods.trigger(),
+              })}
             />
             <FormErrorMessage>
               {errors.cards?.[index]?.linkUrl?.message}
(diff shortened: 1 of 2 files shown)


src/components/EditorCardsDrawer/EditorCardItem.tsx
```diff
@@ -196,25 +196,34 @@ export const EditorCardItem = ({
...
-          <FormControl isRequired isInvalid={!!errors.cards?.[index]?.linkUrl}>
+          <FormControl isInvalid={!!errors.cards?.[index]?.linkUrl}>
             <FormLabel mb="0.75rem">Link URL</FormLabel>
             <Input
               type="text"
               placeholder="Insert /page-url or https://"
-              {...methods.register(`cards.${index}.linkUrl`, { onChange })}
+              {...methods.register(`cards.${index}.linkUrl`, {
+                onChange,
+                onBlur: () => methods.trigger(),
```
seaerchin, line 206: clarification - this is to trigger validation on blur right?

seaerchin (approved):
lgtm; sanity check - any places that expect a link text to always be present (templat repo maybe)? otherwise feel free to merge

src/components/EditorCardsDrawer/EditorCardItem.tsx
dcshzj: Yup, otherwise the validation does not really trigger for some weird reason.

dcshzj:
> lgtm; sanity check - any places that expect a link text to always be present (templat repo maybe)? otherwise feel free to merge

This one is handled entirely within Tiptap:

https://github.com/isomerpages/isomercms-frontend/blob/d51baf95c2ee54904ce194a24d1703c169a7a4e3/src/layouts/components/Editor/extensions/Cards/Cards.ts#L64-L66

So both the link and the link text (i.e. footer) needs to be present before it is displayed on the card.

PR #1807, opened 2024-02-13
fix: allow adding multiple collaborators consecutively

dcshzj:
## Problem

It is not possible to add multiple collaborators consecutively, due to an issue with the way the `useEffect` hook is being called.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Removed the need for a `useEffect` hook by handling the finished states via react-query directly.

## Tests

- [x] Smoke tests
    - [ ] Navigate to any site using email login.
    - [ ] Open the collaborators modal and add multiple collaborators consecutively (without closing the modal or doing anything else). Verify that you can add multiple collaborators consecutively.
    - [ ] Add an invalid email address, verify that the error message says "That doesn't look like a valid email. Try a gov.sg or other whitelisted email."
    - [ ] Add a valid email address that is unlikely to be whitelisted (i.e. `test@example.com`), verify that the error message says "This collaborator couldn't be added. Visit our guide for more assistance."

## Deploy Notes

*None*

### DIFF
--- a/src/components/CollaboratorModal/components/MainSubmodal.tsx
+++ b/src/components/CollaboratorModal/components/MainSubmodal.tsx
@@ -23,7 +23,7 @@ import {
   Button,
 } from "@opengovsg/design-system-react"
 import _ from "lodash"
-import { useEffect } from "react"
+import { useEffect, useState } from "react"
 import { FormProvider, useForm } from "react-hook-form"
 import { BiTrash } from "react-icons/bi"
 import { useParams } from "react-router-dom"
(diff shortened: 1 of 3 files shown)


src/components/CollaboratorModal/components/MainSubmodal.tsx
```diff
@@ -148,20 +148,10 @@ export const MainSubmodal = ({
   ...props
 }: MainSubmodalProps): JSX.Element => {
   const { siteName } = useParams<{ siteName: string }>()
+  const [errorMessage, setErrorMessage] = useState<string>("")
```
seaerchin, line 151: i can't seem to find where this is used?

src/components/CollaboratorModal/components/MainSubmodal.tsx
```diff
@@ -176,16 +166,23 @@ export const MainSubmodal = ({
...
         id: "add-collaborator",
         description: "Collaborator added successfully",
       })
       collaboratorFormMethods.reset()
-    }
-  }, [addCollaboratorSuccess, collaboratorFormMethods, successToast])
+    },
+    onError: (err) => {
+      setErrorMessage(extractErrorMessage(err.response?.data.error))
+    },
```
seaerchin, line 184: this honestly feels like a code smell ._. it's not immediately obvious where this error message is used and there doesn't seem to be feedback?

seaerchin (commented):
i get that this is required for a bug fix, but the error doesn't seem to be handled and user doesn't seem to get feedback? maybe we could pair and you could explain this in abit more detail to me? 

as is, this seems ok for bugfix but might lead to worse maintainability of our codebase (fix now, worsen our lives by abit)

src/components/CollaboratorModal/components/MainSubmodal.tsx
dcshzj: Hmm the `errorMessage` is used here:

https://github.com/isomerpages/isomercms-frontend/blob/0a7c10964894e9cfa85cae8b8d9d44d01b2d0b13/src/components/CollaboratorModal/components/MainSubmodal.tsx#L233

This has always existed. The main difference is how this `errorMessage` is created. Previously it was a `useEffect` hook and now it has shifted to a callback function within the `react-query` hook.

seaerchin: oh thanks! i saw that this was a new change so i was wondering where it's used (didn't review in vscode)

seaerchin (approved):

PR #1809, opened 2024-02-21
fix(dompurify): further limit src

kishore03109:
## Problem

our csp headers are too permissive for a wholistic block of allowed lists. 

## Solution

while we should be limiting csp headers, a quick fix here is to limit the allowed src. 

This is a breaking change for our users who have been using the `script` tag, to ease in this change, have looked at all the script tags that our agency users have currently been using in their md files, and whitelisted them to achieve parity 

**Breaking Changes**

  - Moving forward, ops would need to inform eng if any agency would like to use a script tag with a different source. this change already breaks some agencies code, will be following up with ops to see if they are willing to change this on their end. 
- [ ] No - this PR is backwards compatible with ALL of the following feature flags in this [doc](https://www.notion.so/opengov/Existing-feature-flags-518ad2cdc325420893a105e88c432be5)

## Tests

https://github.com/isomerpages/isomercms-frontend/assets/42832651/0400a94f-8439-40ef-81dc-262f19418144

- [X] Smoke tests
- [ ] put in the following content into the markdown files
```
Allowed script 

<script src="//www.instagram.com/embed.js" async="true"></script>

Disallowed in csp

<script src="//www.evil.com/embed.js" async="true"></script>

Allowed in csp but should also be controlled

<script src="//www.blah.cloudfront.net/embed.js" async="true"></script>


Attributes not allowed 

<svg><script src="/jquery/jquery.min.js" href="THISFILEWILLBELOADED"></script></svg>
<svg><script src="/jquery/jquery.min.js" xlink:href="THISFILEWILLBELOADED"></script></svg>

```
- [ ] press save 
- [ ] accept the disclaimer if it pops up
- [ ] post save, only one of the 3 script tags should persist 

note, this same code will be used in the backend as well

### DIFF
--- a/src/layouts/EditPage/utils.ts
+++ b/src/layouts/EditPage/utils.ts
@@ -38,15 +38,19 @@ DOMPurify.setConfig({
 })
 DOMPurify.addHook("uponSanitizeElement", (node, data) => {
   // Allow script tags if it has a src attribute
-  // Script sources are handled by our CSP sanitiser
-  if (
+  const hasUnallowedSrcValue =
     data.tagName === "script" &&
     !(
       node.hasAttribute("src") &&
       node.innerHTML === "" &&
       ALLOWED_SRC.includes(node.getAttribute("src") ?? "")
     )
-  ) {
+
+  const hasUnallowedScriptAttribute =
+    data.tagName === "script" &&
+    (node.hasAttribute("href") || node.hasAttribute("xlink:href"))
+
+  if (hasUnallowedSrcValue || hasUnallowedScriptAttribute) {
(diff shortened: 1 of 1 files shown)


src/layouts/EditPage/utils.ts
```diff
@@ -29,7 +41,11 @@ DOMPurify.addHook("uponSanitizeElement", (node, data) => {
   // Script sources are handled by our CSP sanitiser
```
dcshzj, line 41: nit: same comment as from backend, we don't actually use our CSP now.

dcshzj (approved):

PR #1819, opened 2024-03-01
fix(media): should not have files that begin with underscore

kishore03109:
## Problem

currently files that start with _ are ignored. this means that in the final output, the site does not have the broken link.

this issue is quite existent in our sites (eg. https://www.cdc.gov.sg/our-programmes/gallery/2021/) 

when i did a string search, this lead to quite a number of sites with these types of images. this is not ideal, and as such this edge case will be coded out as part of the link checker for user to fix

Tests 
- [ ] enter into a repo and try to rename an image into something with an leading underscore.
- [ ] asset that you get the error message as shown below
Screenshot 2024-03-05 at 2 03 20 PM
- [ ] upload an image named '_name.png' into the images folder. assert that the leading underscore gets stripped off

https://github.com/isomerpages/isomercms-frontend/assets/42832651/1ce91503-04b0-442e-8fef-10ae99e3129c

### DIFF
--- a/src/components/MediaSettingsModal/MediaSettingsSchema.jsx
+++ b/src/components/MediaSettingsModal/MediaSettingsSchema.jsx
@@ -29,9 +29,9 @@ export const MediaSettingsSchema = (existingTitlesArray = []) =>
       )
       .test(
         "File not supported",
-        "File names cannot begin with an underscore",
+        "File names must begin with a letter or number",
         (value) => {
-          return !value.startsWith("_")
+          return /^[a-zA-Z0-9]/.test(value)
         }
       )
       .min(
--- a/src/components/MediaSettingsModal/MediaSettingsSchema.jsx
+++ b/src/components/MediaSettingsModal/MediaSettingsSchema.jsx
@@ -27,6 +27,13 @@ export const MediaSettingsSchema = (existingTitlesArray = []) =>
           return (value.match(/\./g) || []).length <= 1
         }
       )
+      .test(
+        "File not supported",
+        "File names cannot begin with an underscore",
+        (value) => {
+          return !value.startsWith("_")
+        }
+      )
       .min(
         MEDIA_SETTINGS_TITLE_MIN_LENGTH,
         `Title must be longer than ${MEDIA_SETTINGS_TITLE_MIN_LENGTH} characters`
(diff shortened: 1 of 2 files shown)


kishore03109:
Current dependencies on/for this PR:
* **#1819** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-frontend/1819?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/isomerpages/isomercms-frontend/1819?utm_source=stack-comment).

src/components/MediaSettingsModal/MediaSettingsSchema.jsx
```diff
@@ -27,6 +27,13 @@ export const MediaSettingsSchema = (existingTitlesArray = []) =>
           return (value.match(/\./g) || []).length <= 1
         }
       )
+      .test(
+        "File not supported",
+        "File names cannot begin with an underscore",
+        (value) => {
+          return !value.startsWith("_")
+        }
+      )
```
alexanderleegs, line 36: Hmm this is only a check for the user-provided name yeah? On the backend I believe we convert special characters into `_`, so something like `!!image.jpg` will be converted to `__image.jpg` and bypass this check. I think we might either need to tighten up this check to block starting with special characters entirely, or move this check to the backend?

alexanderleegs (changes requested):

src/components/MediaSettingsModal/MediaSettingsSchema.jsx
kishore03109: this is a frontend thing. keeping the rules consistent for both image upload and renames

src/components/MediaSettingsModal/MediaSettingsSchema.jsx
```diff
@@ -29,9 +29,9 @@ export const MediaSettingsSchema = (existingTitlesArray = []) =>
       )
       .test(
         "File not supported",
-        "File names cannot begin with an underscore",
+        "File names must begin with a letter",
         (value) => {
-          return !value.startsWith("_")
+          return /^[a-zA-Z]/.test(value)
```
alexanderleegs, line 34: I think we can also include numbers!

alexanderleegs (approved):
lgtm, small comment about allowing numbers

src/components/MediaSettingsModal/MediaSettingsSchema.jsx
kishore03109: done

PR #1828, opened 2024-03-08
fix(path-traversal): add regex to prevent path traversal attack

seaerchin:
## Problem
Currently, our frontend is vulnerable to path traversal attacks due to non-sanitisation of path parameters. 

## Solution
Validate site name according to teh regex in the picture below. sanitization is **not** done for other stuff like resource room name because **these allow spaces in the name.**

Screenshot 2024-03-08 at 1 37 08 PM

Attempting to use `%20` in the regex will lead to erroneous validation (as seen in the ss). We also cannot allow `%` because this will lead to allowing url encoded characters (`%5c`).
Screenshot 2024-03-08 at 1 38 41 PM

### DIFF
--- a/src/routing/RouteSelector.jsx
+++ b/src/routing/RouteSelector.jsx
@@ -66,18 +66,18 @@ export const RouteSelector = () => {
         <ProtectedRouteWithProps
           exact
           path={[
-            "/sites/:siteName/resourceRoom/:resourceRoomName/resourceCategory/:resourceCategoryName/editPage/:fileName",
-            "/sites/:siteName/folders/:collectionName/subfolders/:subCollectionName/editPage/:fileName",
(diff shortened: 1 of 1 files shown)


alexanderleegs (approved):

timotheeg:
It's strange to have a path-traversal fix done on frontend 🤔 . Typically path traversal is something that gets exercised (and fixed) on the backend (cause frontend can always be bypassed).

Can you describe the attack behaviour that was active, before that is this PR is meant to disable?

timotheeg (approved):

seaerchin:
from teh VAPT report, what happened was that this was a frontend oriented attack, because the cure53 user had access to both repositories, so they were able to escape `sites/a` and go to `sites/b` and have the api call out.

normally on our backend, we have `verifySiteMember` which prevents this - hence thise is a FE oriented fix

PR #1843, opened 2024-03-14
fix(permalinks): default permalink for create

kishore03109:
## Problem

We have quite a number of duplicate permalinks
this does not fix the problem completely, but it does allow for some sort of preventive measure to prevents accidental dups. 

we are not sure if of the number of agencies that use dup permalinks for the sake of playground, and it would be hard to enforce this until we have the functionality for playground out. more context [here](https://opengovproducts.slack.com/archives/CK68JNFHR/p1710317487152189)

## Solution

sane default when creating a page.

**Breaking Changes**

  - Details ...

## Tests

follow video

https://github.com/isomerpages/isomercms-frontend/assets/42832651/3c3657dc-b9a1-4a40-9e49-07d71c32fd0a

- [ ] create a page, note that you should be unable to modify the permalink 
- [ ] edit page modal's functionality remains the same, we dont still allow them to change to whatever.

### DIFF
--- a/src/components/PageSettingsModal/PageSettingsModal.tsx
+++ b/src/components/PageSettingsModal/PageSettingsModal.tsx
@@ -96,6 +96,7 @@ export const PageSettingsModal = ({
 }: PageSettingsModalParams) => {
   const { siteName, fileName } = params
   const isTiptapEnabled = useFeatureIsOn("is-tiptap-enabled")
+  const isPageCreated = !fileName
 
   const existingTitlesArray = pagesData
     .filter((page) => page.name !== fileName)
@@ -146,6 +147,23 @@ export const PageSettingsModal = ({
     })
   }
 
+  const getDefaultPermalink = (title: string) => {
+    if (!title) return ""
+    const titleSlug = title
+      .toLowerCase()
+      .replace(/[^a-z0-9]+/g, "-")
+      .replace(/^-+|-+$/g, "")
+    return `/${titleSlug}/`
+  }
+
+  const currTitle = watch("title")
+  const currPermalink = watch("permalink")
+  useEffect(() => {
+    if (isPageCreated && currPermalink !== getDefaultPermalink(currTitle)) {
+      setValue("permalink", getDefaultPermalink(currTitle))
+    }
+  }, [isPageCreated, setValue, currTitle, currPermalink])
+
   return (
     <Modal isOpen onClose={onClose}>
       <ModalOverlay />
@@ -169,8 +187,17 @@ export const PageSettingsModal = ({
                 isLink={false}
               />
               {/* Title */}
-              <FormControl isRequired isInvalid={!!errors.title?.message}>
-                <FormLabel>Page title</FormLabel>
+              <FormControl
+                isRequired
+                isInvalid={!!errors.title?.message}
+                mb="1rem"
+              >
+                <Box mb="0.75rem">
+                  <FormLabel mb={0}>Page title</FormLabel>
+                  <FormLabel.Description color="text.description">
+                    This appears on the top of the browser window or tab
+                  </FormLabel.Description>
+                </Box>
                 <Input
                   placeholder="Page title"
                   id="title"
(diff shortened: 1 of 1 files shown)


kishore03109:
* **#1843** <a href="https://app.graphite.dev/github/pr/isomerpages/isomercms-frontend/1843?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `develop`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @kishore03109 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

src/components/PageSettingsModal/PageSettingsModal.tsx
```diff
@@ -146,6 +147,23 @@ export const PageSettingsModal = ({
     })
   }
 
+  const getDefaultPermalink = (title: string) => {
+    if (!title) return ""
+    const titleSlug = title
+      .toLowerCase()
+      .replace(/[^a-z0-9]+/g, "-")
```
seaerchin, line 154: should we replace `A-Z`? maybe the regex should include `A-Z` to prevent replacement

src/components/PageSettingsModal/PageSettingsModal.tsx
```diff
@@ -146,6 +147,23 @@ export const PageSettingsModal = ({
...
+      .replace(/^-+|-+$/g, "")
```
seaerchin, line 155: i think we only need `/^-+$/g`

separately, i'm not sure why we're replacing all instances of `-` with an empty string - that seems like a mistake? (eg: `a school bus` -> `a-school-bus` -> `aschoolbus`)

src/components/PageSettingsModal/PageSettingsModal.tsx
```diff
@@ -169,32 +187,54 @@ export const PageSettingsModal = ({
...
                   placeholder="Insert /page-url or https://"
+                  isDisabled={isPageCreated}
                 />
+                {isPageCreated && (
+                  <Box mt="0.5rem" mb="0.5rem">
+                    <Text textStyle="body-2">
+                      {`${siteUrl}${watch("permalink")}`}
+                    </Text>
+                  </Box>
+                )}
```
seaerchin, line 234: seems to be repeating the form description? not sure if intended (can clsoe if so)

src/components/PageSettingsModal/PageSettingsModal.tsx
```diff
@@ -169,32 +187,54 @@ export const PageSettingsModal = ({
...
                   </FormLabel.Description>
                 </Box>
                 <Input
                   {...register("permalink", { required: true })}
                   id="permalink"
                   placeholder="Insert /page-url or https://"
+                  isDisabled={isPageCreated}
                 />
+                {isPageCreated && (
+                  <Box mt="0.5rem" mb="0.5rem">
```
seaerchin, line 229: not needed to fix but this `mt, mb` can be combined to `my` fyi

seaerchin (commented):
blocking on regex

src/components/PageSettingsModal/PageSettingsModal.tsx
kishore03109: hey @seaerchin, nope we are only replacing instances - at the start or at the end of the string. 

```
> "a school bus".toLowerCase().replace(/[^a-z0-9]+/g, "-").replace(/^-+|-+$/g, "")
'a-school-bus'
```

src/components/PageSettingsModal/PageSettingsModal.tsx
kishore03109: hey, i intentionally wanted to replace. we did a similar thing regarding the amplify migration, where we changed all the files to lowercase in an attempt to standardise things. my hunch here is that making all permalinks lowercase helps us standardise things, now that we are using a infra(amplify) that is case sensitive

src/components/PageSettingsModal/PageSettingsModal.tsx
kishore03109: hmm to be clear no ya, when `isPageCreated` is true, the form description is `You can change this later in Page Settings.`

we only show the text at the bottom iff the user is currently editing the page.

src/components/PageSettingsModal/PageSettingsModal.tsx
seaerchin: ah ok, my bad. so this first does a replacement of spaces to `-` then removes leading/trailing `-`.

src/components/PageSettingsModal/PageSettingsModal.tsx
seaerchin: oops, i didn't see the `toLowerCase`

src/components/PageSettingsModal/PageSettingsModal.tsx
seaerchin: yeap, i skimmed so i didn't notice but yea they're mutually exhaustive so lgtm!

seaerchin (approved):
for resource pages, do they also use this `pageSettingsModal`? otherwise, we might also need to duplicate the changes here over.

PR #1854, opened 2024-03-18
chore(validation): add validation for all routes

seaerchin:
## Problem
Some of our routes don't have validation - this leads to security vulnerabilities where path traversal attacks can be done. 

## Solution
Validate routes based on teh following criteria (observed behaviour): 

1. top level sites/resource categories/resource rooms will **not** have special characters; spaces will get converted to `-` in the url so it's ok if we validate alphanumeric + `-`. do note that **non-ascii** characters (chi/tamil) might have issues...
2. for filenames, we have the extension + we allow non-ascii characters (chi/tamil) so we will only check for special characters (inclusive of `.`) 
3. for media folders, we allow `/` (encoded) so we have to split and validate each one separately .___.

## Manual tests 
**NOTE:** refer to this [doc](https://docs.google.com/document/d/1_ouaNW3S3grk_ROjU7pHiskqzgULMEsxoUXV1yNCY7U/edit#heading=h.vnvoaoe7cnsx) for special chars

**Sitename/resource name/collection**
- [ ] ensure that you cannot create any of these with a special character
- [ ] create one of these (less sitename) with a space + a dash 
- [ ] ensure that you can access them (note that hte url converts space to a dash) 

**filename**
- [ ] create a filename with the **allowed** special characters (given in the doc) 
- [ ] create a filename w/ chi/tamil characters (non-ascii) 
- [ ] ensure that you can access those pages

**sub-collection + resource category**
- [ ] ensure that you cannot create a sub-collection with teh disallowed special characters 
- [ ] create a sub-collection with the **allowed** special characters (see doc) + ensure that the sub-collection has a dash + a space
- [ ] ensure that you can access said sub-collection

**media**
note that for media stuff, we can create nested folders and these folders are encoded using `%2f` 
- [ ] ensure that you cannot create a media folder with teh disallowed special characters 
- [ ] ensure that you can create a media folder with the **allowed** special characters
- [ ] ensure that you can upload images/files with special characters and that the special characters are renamed to `_` 
- [ ] ensure that you can rename the images/files to allowed special characters 
- [ ] ensure that you **cannot** rename the images/files to disallowed special characters

### DIFF
--- a/src/routing/RouteSelector.jsx
+++ b/src/routing/RouteSelector.jsx
@@ -66,9 +66,9 @@ export const RouteSelector = () => {
         <ProtectedRouteWithProps
           exact
           path={[
-            "/sites/:siteName([a-zA-Z0-9-]+)/resourceRoom/:resourceRoomName/resourceCategory/:resourceCategoryName/editPage/:fileName",
-            "/sites/:siteName([a-zA-Z0-9-]+)/folders/:collectionName/subfolders/:subCollectionName/editPage/:fileName",
-            "/sites/:siteName([a-zA-Z0-9-]+)/folders/:collectionName/editPage/:fileName",
+            "/sites/:siteName([a-zA-Z0-9-]+)/resourceRoom/:resourceRoomName([a-zA-Z0-9-]+)/resourceCategory/:resourceCategoryName/editPage/:fileName",
+            "/sites/:siteName([a-zA-Z0-9-]+)/folders/:collectionName([a-zA-Z0-9-]+)/subfolders/:subCollectionName/editPage/:fileName",
+            "/sites/:siteName([a-zA-Z0-9-]+)/folders/:collectionName([a-zA-Z0-9-]+)/editPage/:fileName",
             "/sites/:siteName([a-zA-Z0-9-]+)/editPage/:fileName",
           ]}
           component={injectApprovalRedirect(EditPage)}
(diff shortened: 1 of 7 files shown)


src/routing/RouteSelector.jsx
```diff
@@ -66,19 +68,49 @@ export const RouteSelector = () => {
         <ProtectedRouteWithProps
           exact
           path={[
-            "/sites/:siteName([a-zA-Z0-9-]+)/resourceRoom/:resourceRoomName/resourceCategory/:resourceCategoryName/editPage/:fileName",
-            "/sites/:siteName([a-zA-Z0-9-]+)/folders/:collectionName/subfolders/:subCollectionName/editPage/:fileName",
-            "/sites/:siteName([a-zA-Z0-9-]+)/folders/:collectionName/editPage/:fileName",
+            // DONE
```
kishore03109, line 71: sir

kishore03109 (approved):
hmm my only concern here is testing.
this pr could potentially do a 404 in a vaild path
could you add some basic testing to go into all the affected path just as a sanity check that this is working as intended?

src/routing/RouteSelector.jsx
seaerchin: oops

seaerchin:
> hmm my only concern here is testing. this pr could potentially do a 404 in a vaild path could you add some basic testing to go into all the affected path just as a sanity check that this is working as intended?

yea, i checked on local and there's unit tests also :pog: (`RouteSelector.spec.jsx`). 

i'll add a set of manual tests to run through during releaes so it's not just me

PR #1865, opened 2024-03-27
chore(mergify): enhance mergify

seaerchin:
## Problem
package updates are not automated and we have to look through even minor ones. minor updates should be approved + merged automatically

## Solution
update dependabot config to merge minor packages automatically. we auto-merge snyk cos is security issue (open to also changing this to minor)

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -3,9 +3,8 @@ pull_request_rules:
     conditions:
       - author~=^dependabot\[bot\]$
       - check-success~=lint
-      - check-success~=build
       - check-success~=test
-      - title~=Bump [^\s]+ from ([\d]+)\..+ to \1\.
+      - dependabot-update-type = version-update:semver-minor
     actions:
       review:
         type: APPROVE
@@ -16,9 +15,9 @@ pull_request_rules:
     conditions:
       - author=snyk-bot
       - check-success~=lint
-      - check-success~=build
       - check-success~=test
-      - title~=^\[Snyk\]
+      - title~=^\[Snyk\]]
+      - base = develop
     actions:
       review:
         type: APPROVE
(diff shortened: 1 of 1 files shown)


.github/mergify.yml
```diff
@@ -3,9 +3,8 @@ pull_request_rules:
     conditions:
       - author~=^dependabot\[bot\]$
       - check-success~=lint
-      - check-success~=build
       - check-success~=test
-      - title~=Bump [^\s]+ from ([\d]+)\..+ to \1\.
+      - dependabot-update-type = version-update:semver-minor
```
kishore03109, line 7: what about patches?

.github/mergify.yml
```diff
@@ -16,9 +15,9 @@ pull_request_rules:
     conditions:
       - author=snyk-bot
       - check-success~=lint
-      - check-success~=build
```
kishore03109, line 19: may i get context for this regarding why this line was removed?

.github/mergify.yml
seaerchin: updated to auto-update for all non-major

.github/mergify.yml
seaerchin: added back in, removed by accident.

.github/mergify.yml
```diff
@@ -3,8 +3,9 @@ pull_request_rules:
     conditions:
       - author~=^dependabot\[bot\]$
       - check-success~=lint
+      - check-success~=build
       - check-success~=test
-      - dependabot-update-type = version-update:semver-minor
+      - -dependabot-update-type = version-update:semver-major
```
kishore03109, line 8: wait am a bit confused, should this not be minor then?

seaerchin: Screenshot 2024-04-09 at 1 35 38 PM

didn't duplicate the comment here but this negates the condition so this is essentially everything but major (minor + patches)

kishore03109 (approved):

PR #1887, opened 2024-05-20
chore: increase contact us field limits

alexanderleegs:


src/utils/validators.js
```diff
@@ -166,17 +166,17 @@ const INFOCOLS_INFOBOX_DESCRIPTION_MAX_LENGTH = 120
...
-const LOCATION_TITLE_MAX_LENGTH = 30
-const LOCATION_ADDRESS_MAX_LENGTH = 30
+const LOCATION_TITLE_MAX_LENGTH = 50
+const LOCATION_ADDRESS_MAX_LENGTH = 50
 const LOCATION_OPERATING_DAYS_MIN_LENGTH = 2
-const LOCATION_OPERATING_DAYS_MAX_LENGTH = 30
+const LOCATION_OPERATING_DAYS_MAX_LENGTH = 50
 const LOCATION_OPERATING_HOURS_MIN_LENGTH = 2
-const LOCATION_OPERATING_HOURS_MAX_LENGTH = 30
+const LOCATION_OPERATING_HOURS_MAX_LENGTH = 50
```
seaerchin, line 179: ok on the frontend - but are these also coded into our backend?

seaerchin (approved):
lgtm but please check backend to see if it also enforces these limits

src/utils/validators.js
alexanderleegs: No changes required for backend! Backend checks for type of input, but not the length - I think we're okay with this because it's more of a restriction to help users not misuse the fields - but Shaz was showing use cases where the current limit is insufficient

PR #1888, opened 2024-05-21
chore: add frustration signal tracking

seaerchin:
## Problem
No frustration signal tracking in dd meant that we weren't getting important dd points

Closes [ISOM-1309]

## Solution
Add frustration signal trackign in + remove `as const` cast. the cast had to be removed as the library type doesn't allow readonly assignment to a mutable property...


kishore03109 (approved):
may i get context to this ah? 
if the library exports a type def, y not just use

PR #1908, opened 2024-07-05
feat: enhance Telegram URL regex

dcshzj:
## Problem

It appears that Telegram invite links to private groups are created in the format `t.me/+...`, which does not match the existing regex we have.

## Solution

**Breaking Changes**

**Features**:

- Enhance the Telegram URL regex to allow for an optional `+` at the start of the path.

## Before & After Screenshots

**BEFORE**:

Screenshot 2024-07-05 at 15 09 30

**AFTER**:

Screenshot 2024-07-05 at 15 09 41

## Tests

- [x] Smoke tests
    - [ ] Go to any site and visit the site settings page.
    - [ ] Attempt to add a Telegram invite link under the social media settings: `https://t.me/+Ird7tka7i7U3N2l1/`. No validation errors should appear.
    - [ ] Save the page, the changes should persist.
    - [ ] Attempt to add a normal Telegram group link: `https://t.me/group_name`. No validation errors should appear.
    - [ ] Attempt to add an invalid link: `https://t.me/++Ird7tka7i7U3N2l1/`. An error message should appear on blur.

## Deploy Notes

_None_

### DIFF
--- a/src/utils/validators.js
+++ b/src/utils/validators.js
@@ -16,7 +16,7 @@ const PERMALINK_REGEX = "^((/([a-zA-Z0-9]+-)*[a-zA-Z0-9]+)+)/?$"
 export const URL_REGEX_PREFIX = "^(https://)?(www.)?("
 export const URL_REGEX_SUFFIX = ".com/)([a-zA-Z0-9_-]+([/.])?)+$"
 export const YOUTUBE_REGEX = ".com/(@)?)([a-zA-Z0-9_-]+([/.])?)+$"
-export const TELEGRAM_REGEX = "telegram|t).me/([a-zA-Z0-9_-]+([/.])?)+$"
+export const TELEGRAM_REGEX = "telegram|t).me/\\+?([a-zA-Z0-9_-]+([/.])?)+$"
 export const TIKTOK_REGEX = ".com/@)([a-zA-Z0-9_-]+([/.])?)+$"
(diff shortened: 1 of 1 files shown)


src/utils/validators.js
```diff
@@ -16,7 +16,7 @@ const PERMALINK_REGEX = "^((/([a-zA-Z0-9]+-)*[a-zA-Z0-9]+)+)/?$"
 export const URL_REGEX_PREFIX = "^(https://)?(www.)?("
 export const URL_REGEX_SUFFIX = ".com/)([a-zA-Z0-9_-]+([/.])?)+$"
 export const YOUTUBE_REGEX = ".com/(@)?)([a-zA-Z0-9_-]+([/.])?)+$"
-export const TELEGRAM_REGEX = "telegram|t).me/([a-zA-Z0-9_-]+([/.])?)+$"
+export const TELEGRAM_REGEX = "telegram|t).me/\\+?([a-zA-Z0-9_-]+([/.])?)+$"
```
seaerchin, line 19: shouldn't it be `\/`?

dcshzj: Hmm no, we want to add the optional `+` at the start after the `/`, the `/` doesn't need to be escaped here but the `+` need.

seaerchin: ok clarified offline - this is a string, so we need to escape the escape character

seaerchin (approved):

PR #1937, opened 2024-10-18
fix: use menlo security for gsib

seaerchin:
# problem
1. gsib access to staging site fails as gsib uses a security scan which. leasd to too many redirects

# solution
1. use `menlosecurity` and prepend if gsib is detected


adriangohjw (approved):
LGTM

PR #1950, opened 2025-11-12
fix: prevent path traversal for API endpoint URL

dcshzj:
## Problem

Closes ISOM-2132.

## Solution

**Breaking Changes**

**Bug Fixes**:

- Adds a check on the input parameters to prevent any potential path traversals before crafting the final API endpoint.

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

- [x] Smoke tests
    - [ ] Log in to the CMS, then try this URL: `https://staging-cms.isomer.gov.sg/sites/isomer-vapt2025one/folders/example-folder/subfolders/example-subfolder/editPageSettings/x%5C..%5C..%5C..%5C..%5C..%5C..%5C..%5C..%5Cfoo`
    - [ ] Verify in the Console tab that there is an error message that starts with "Error: Unsafe path detected in parameter"
    - [ ] Verify in the Network tab that there is no requests with a 403 response to a URL that ends with `foo`.

### DIFF
--- a/src/services/PageService.jsx
+++ b/src/services/PageService.jsx
@@ -1,3 +1,5 @@
+const { isSafePath } = require("../utils/misc")
+
 export class PageService {
   constructor({ apiClient }) {
     this.apiClient = apiClient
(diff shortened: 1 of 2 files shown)


seaerchin (commented):
shuold we use `starter-kitty`?

dcshzj:
> shuold we use `starter-kitty`?

I tried previously but got some issues I think was the ESM/CJS thingy, but that was for the backend so probably frontend is quite ok.

But for this I think it's slightly different cos starter-kitty assumes backend but this is running on the frontend. It has a dependency on `fs` which is only available in nodejs.

seaerchin (approved):

